PR #1, opened 2016-09-26
Adrian: Updates to German translation

AdrianLxM:


app/src/main/res/values-de/strings-de.xml
```diff
@@ -289,14 +289,14 @@
...
-	<string name="enter_second_bg_value">Zweiter BG Wert</string>
+	<string name="enter_first_bg_value">Erster BG-Wert</string>
+	<string name="enter_second_bg_value">Zweiter BG-Wert</string>
 	<string name="language_editor">Angezeigte Sprache</string>
 	<string name="configure_missed_readings">Einstellungen für verpasste Werte</string>
 	<string name="install_pebble_classic_trend_watchface">Pebble Classic Trend Watchface installieren</string>
 	<string name="install_pebble_trend_watchface">Pebble Trend Watchface installieren</string>
 	<string name="install_pebble_watchface">Pebble Watchface installieren</string>
-	<string name="glucose_history">Glucose Geschichte</string>
+	<string name="glucose_history">BG-Historie</string>
```
jamorham: The english has "Glucose" not "BG" which I understand would be abbreviation for Blood Glucose which is not the same.

AdrianLxM:
Hi @jamorham, I had already changed that to just "Historie" ( it also contains treatments,...). I guess you are reading just one commit and not the overall diff ("Changed Files")?

app/src/main/res/values-de/strings-de.xml
```diff
@@ -17,10 +17,10 @@
 	<string name="title_activity_start_new_sensor">Neuen Sensor starten</string>
 	<string name="title_activity_new_sensor_location">Position des neuen Sensors</string>
 	<string name="title_activity_stop_sensor">Sensor stoppen</string>
-	<string name="title_activity_nav_drawer">Nav Schublade</string>
+	<string name="title_activity_nav_drawer">Navigationsleiste</string>
 	<string name="navigation_drawer_open">Navigationsleiste öffnen</string>
 	<string name="navigation_drawer_close">Navigationsleiste schließen</string>
-	<string name="title_activity_calibration_check_in">Kalibrierung</string>
+	<string name="title_activity_calibration_check_in">Kalibrierungen einspielen (per USB)</string>
```
jamorham: The original text doesn't mention USB

AdrianLxM:
As my email aparently didn't come through, again via the website:

@jamorham Did I change the wrong one? I thought I just added it to the one where you check them in via USB. Share will do it automatically. So it shouldn't have this activity in the first place.
Do you want me to remove the "(via USB)"? I think it should be there for the user to know that it is a wizard to enter fingersticks.

jamorham:
I appreciate you are trying to improve usability by adding that regarding USB but it just isn't a direct translation. The english just says "Calibration Activity" 

To change things like that (if needed) we also need to change the english version so that it can be consistent otherwise there are potential risks, such as reusing the english string for something else not knowing that a translated version has a different meaning etc. 

I have already merged ~61 of the 63 submitted translations here. Unfortunately github doesn't appear to show a true `diff` of the differences between the branches so it isn't obvious on the PR.

AdrianLxM:
@jamorham Did I change the wrong one? I thought I just added it to the 
one where you check them in via USB. Share will do it automatically. So 
it shouldn't have this activity in the first place.
Do you want me to remove the "(via USB)"? I think it should be there for 
the user to know that it is a wizard to enter fingersticks.

PR #2, opened 2016-09-26
Allow more notification sounds

AdrianLxM:
There are very little and easily overhearable notification sounds on some phones.

The parameter "all" gives also access to the ringtones and alarm sounds (better for the low predictive alert).

### DIFF
--- a/app/src/main/res/xml/pref_notifications.xml
+++ b/app/src/main/res/xml/pref_notifications.xml
@@ -85,7 +85,7 @@
 
                     android:defaultValue="content://settings/system/notification_sound"
                     android:key="calibration_notification_sound"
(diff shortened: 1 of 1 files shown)


jamorham:
Do ringtone sounds automatically loop? For some reason I thought they did. I'm wondering what that will do if so.

AdrianLxM:
> Do ringtone sounds automatically loop? 

On my phone they don't. With the standard alarm system the automatic loop had to be triggered as well.

Looping the low predict alarm until the notification is acknowledged would be a great extra feature but at the moment I'm happy to have something that sounds a bit longer (but stops as soon as I dismiss the Notification) :)

PR #3, opened 2016-09-26
Tooltips for sensor values and treatments in history again

AdrianLxM:
When it was changed that tapping on a treatment or value opened an interactive snackbar instead of a tooltip, the history could no longer show it.

Now you can see the notes form the past as tooltips again.


jamorham:
Can you tell me if my understanding of the changes here are correct:

The snackbar feature when clicking on a graph point in the Home page broke the clickable points within the History page.

The changes in this PR preserve the existing behavior on the Home activity but re-introduce Toast pop-ups in the old style when clicking on items in the History activity?

Please also confirm that you have tested this and it works and that the add note etc functionality on Home also still works. I assume you have but I would just like to have that confirmation in the PR. Thanks

AdrianLxM:
Yes, I have tested it.
There is no change on Home - still snackbar appears and adding notes (and undo)  works.

PR #4, opened 2016-09-26
readings excludeable from statistics

AdrianLxM:
This is apart from my calibration parameters one of the last things from xDrip missing:

On a long-press on an element in the BG-Datatable it will show a dialog to flag a reading.
A reading that is flagged will have no influence on the statistics.

I use this when the values falsely spike after a transmitter swap or on clear compression lows to not ruin my statistics.

It also corrects the cutoff for statistics to 39. Values below that are used as special values and have another meaning.


AdrianLxM:
It also adds the bugfix from January that 'avg' is a double.

jamorham:
Please confirm that you have tested this also.

I am testing to ensure that json processing of `BgReading` is not impacted, for example with followers.

AdrianLxM:
Yes, tested for normal use.
As I don't use follower, I can't test that part. It would be good, if you could.

PR #5, opened 2016-09-27
fix for entries between 12.00 and 12.59

AdrianLxM:
Treatments with time between 12.00 and 12.59 were mapped to the hour after midnight - as well as 00.01 to 00.59.

According to http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html
`hh` represents the hours in 1-12 hour format and `HH` in 0-23 hour format.


AdrianLxM:
I added another thing: while it make no sense to enter "0 carbs", a time that equals 0 ("0:00" = midnight) is a valid time.
This is what happened before:
screenshot_2016-09-27-03-17-43

Now:
screenshot_2016-09-27-03-19-01

PR #6, opened 2016-09-27
Disable alerts

tzachi-dar:
Import of the code that disables alerts.

Since it can be deleted, it brings back the 55 undeletable alert. We might consider allowing to delete it.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/AlertList.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/AlertList.java
@@ -15,6 +15,12 @@ import android.widget.AdapterView;
 import android.widget.Button;
 import android.widget.ListView;
 import android.widget.SimpleAdapter;
+import android.widget.TextView;
+import android.widget.SimpleCursorAdapter.ViewBinder;
+
+//#import android.widget.SimpleCursorAdapter;
+//#import android.database.Cursor;
+import android.graphics.Paint;
 
 import com.eveningoutpost.dexdrip.Models.AlertType;
 import com.eveningoutpost.dexdrip.Models.UserError.Log;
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/AlertList.java
```diff
@@ -159,9 +173,36 @@ public void onClick(View v) {
...
     }
 
+    void displayWarning() {
+        runOnUiThread(new Runnable() {
+            @Override
+            public void run() {
+
+                if (!isFinishing()){
+                    new AlertDialog.Builder(AlertList.this)
+                      .setTitle("Warnining !")
```
jamorham: Can we change the title to: "Warning !"

app/src/main/java/com/eveningoutpost/dexdrip/AlertList.java
```diff
@@ -159,9 +173,36 @@ public void onClick(View v) {
...
+                      .setMessage("No active low alert exists, it is highly recomended that you will add a low alert.")
```
jamorham: Can we change the message to: "No active Low Alert exists, without this there will be no alert on low glucose! Please add or enable a low alert."

app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
```diff
@@ -361,10 +358,10 @@ public static void update_alert(
         at.save();
     }
     public static void remove_alert(String uuid) {
-  /*      if(uuid.equals(LOW_ALERT_55)) {
+        if(uuid.equals(LOW_ALERT_55)) {
```
jamorham: There should not be any alerts which cannot be removed/disabled if the user wishes to do so. This restriction should be reverted.

jamorham:
I've added some line comments to hopefully make it easier with my suggested changes. 

These are for wording on the warning and that we should not have an unchangeable alert because it removes flexibility and user choice and would change how the code currently works which I think is the most agreeable.

tzachi-dar:
I have fixed the issues that you have raised.
As for testing, I have checked that the UI works.
The code is just a clone of what we had in xDrip.

I will test actual alerts tonight.

tzachi-dar:
Verified alerts are working/not working as needed.

PR #7, opened 2016-09-28
Adrian widgetstatus

AdrianLxM:
Adds the extra status line to the widget.
The settings were already present but had no functionality.


jamorham:
Can you confirm that this is tested both enabled and disabled and it works okay? Thanks.

AdrianLxM:
Works ok on my phone. Tested enabling/disabling it just for the widget and in general. Will update immediately if the widget line settings are changed (add/delete on statistics element e.g.).

Tested that it works on the widget after a restart of the phone even if `Home` was never opened before.

PR #8, opened 2016-09-28
manual calibration overwrite in engineering mode

AdrianLxM:
This will allow to manually set the slope and intercept from a menu in the calibration graph.

screenshot_2016-09-28-16-58-58

screenshot_2016-09-28-16-58-51

The text entry is limited to decimals and for slopes also to positive values:
screenshot_2016-09-28-17-42-00

It will modify the last valid calibration; the calibration graph is updated:
screenshot_2016-09-28-17-06-42

When engineering mode is disabled, the menu won't show:
screenshot_2016-09-28-17-00-40

No History rewrite happens. Just the values are calculated with that slope from now on.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
@@ -1,12 +1,24 @@
 package com.eveningoutpost.dexdrip;
 
+import android.content.DialogInterface;
+import android.content.Intent;
+import android.content.SharedPreferences;
 import android.graphics.Color;
 import android.os.Bundle;
+import android.preference.PreferenceManager;
 import android.support.annotation.NonNull;
+import android.support.v7.app.AlertDialog;
+import android.text.InputType;
+import android.text.TextUtils;
+import android.view.Menu;
+import android.view.MenuItem;
+import android.widget.EditText;
 import android.widget.TextView;
 
 import com.eveningoutpost.dexdrip.Models.Calibration;
+import com.eveningoutpost.dexdrip.Models.JoH;
 import com.eveningoutpost.dexdrip.utils.ActivityWithMenu;
+import com.eveningoutpost.dexdrip.wearintegration.WatchUpdaterService;
 
 import java.text.DateFormat;
 import java.text.DecimalFormat;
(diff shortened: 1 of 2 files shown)


AdrianLxM:
I also testet it with German langage settings where we usually use another decimal separator. It works - just like when entering fingersticks for calibration.

app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
```diff
@@ -114,4 +126,111 @@ public Line getCalibrationsLine(List<Calibration> calibrations, int color) {
...
+    public boolean onCreateOptionsMenu(Menu menu) {
+
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(xdrip.getAppContext());
+
+
+        //Just generate the menu in engineering mode
+        if (!prefs.getBoolean("engineering_mode", false)){
+            return false;
+        }
+
```
tzachi-dar: There is a function for it in xdrip-plus: home.getPreferencesBooleanDefaultFalse()

jamorham:
This looks interesting and probably worthwhile being hidden behind engineering mode. I don't think non-developers should ever be entering numbers manually. I feel that a calibration algorithm has failed if that is happening. I know that isn't the intention behind this PR I'm just stating what I think the position should be regarding this sort of things.

app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
jamorham: yes I would recommend that too

app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
```diff
@@ -114,4 +126,111 @@ public Line getCalibrationsLine(List<Calibration> calibrations, int color) {
...
+        new AlertDialog.Builder(this)
+                .setTitle("Ovewrite Intercept")
+                .setMessage("Overwrite Intercept")
+                .setView(editText)
+                .setPositiveButton("OK", new DialogInterface.OnClickListener() {
+                    public void onClick(DialogInterface dialog, int which) {
+                        String text = editText.getText().toString();
+                        if (!TextUtils.isEmpty(text)) {
+                            double doubleValue = Double.parseDouble(text);
```
jamorham: For some reason I didn't think `Double.parseDouble(String str)` was locale aware? I made: `JoH.tolerantParseDouble(String str)` to handle issues with different decimal separators. Java documentation seems to say it does support localized. Possibly it becomes an issue if the string is stored and later interpreted..

tzachi-dar:
This is indeed a needed feature. It will allow us to test more calibration algorithms.

One change that I would do to it is use slide bars instead of entering data directly.
The advantages of slide bars are: 1) we can limit the input of users, for example intercept between 20-60.
2) The user can play with the values, see the influence on the screen and find the optimum easily.
There is an example for this (I started working on this but never finished) on xdrip experimental calibration_slide.

As for which mode to put it. I'm fine with having it behind a key that is not exposed to users that don't compile by themselves.

I guess that we will need to do much more to this mode until we can push it to users.

app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
AdrianLxM: I did test that. The input field wouldn't allow to enter a comma anyways. But I can cahnge it.

app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
AdrianLxM: I will change that.
Any chance that we could later move these functions into a separate utility class?

tzachi-dar:
By the way, we might enable 2 modes. One that adrian has created for small screens, and one for bigger screens that will have much more options.

AdrianLxM:
Did the requested changes and tested them.

PR #9, opened 2016-09-28
restrict timeframe when updating persistant notification

AdrianLxM:
Not yet tested


AdrianLxM:
@jamorham, can I delete this now?

PR #10, opened 2016-09-29
hidden calibration params for adrian - only in engineering mode

AdrianLxM:
This just works in engineering mode. I used to patch in these parameters manually to get the calibration algorithm working for me.

Tested that even when the new setting was on, it will NOT be used when engineering mode is switched off.

Tested that slopes now can really be lower if the regression algorithm has that result. -> working.


AdrianLxM:
Thanks Tzachi,

about the naming: It is not a feature that a user who is not in "engineering mode" can see.
When, according to plan, have the calibration algoriths pluggable and probably have multiple ones to choose from (in engineering-mode), I'd really like to know under whose scope they are.

PR #11, opened 2016-09-29
Allow the user to configure more logging to be stored by xDrip.

tzachi-dar:
This is needed if the user has a problem in some component and we want to allow him to send more logs without creating a private build for him.

Once approved I'll remove the logs with tag "dddd" that were used to debug this code.

Tested by logging different log components.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/UserError.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/UserError.java
@@ -7,8 +7,10 @@ import com.activeandroid.Model;
 import com.activeandroid.annotation.Column;
 import com.activeandroid.annotation.Table;
 import com.activeandroid.query.Select;
+import com.eveningoutpost.dexdrip.Home;
 
 import java.util.Date;
+import java.util.Hashtable;
 import java.util.List;
 
 /**
@@ -18,6 +20,8 @@ import java.util.List;
 @Table(name = "UserErrors", id = BaseColumns._ID)
 public class UserError extends Model {
 
+	private final static String TAG = UserError.class.getSimpleName();
+	
     @Column(name = "shortError")
     public String shortError; // Short error message to be displayed on table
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Models/UserError.java
```diff
@@ -151,52 +155,139 @@ public static void e(String a, String b){
             new UserError(a, b);
         }
 
-        public static void e(String a, String b, Exception e){
-            android.util.Log.e(a, b, e);
-            new UserError(a, b + "\n" + e.toString());
+        public static void e(String tag, String b, Exception e){
+            android.util.Log.e(tag, b, e);
+            new UserError(tag, b + "\n" + e.toString());
```
jamorham: This and similar lines, is this change from `a` to `tag` just a refactor to make it more readable?

app/src/main/java/com/eveningoutpost/dexdrip/Models/UserError.java
```diff
@@ -151,52 +155,139 @@ public static void e(String a, String b){
...
+    	}
+    	
+    	/*
+    	 * This function reads a string representing tags that the user wants to log
+    	 * Format of string is tag1:level1,tag2,level2
+    	 * Example of string is Alerts:i,BG:W
+    	 * 
+    	 */
+    	public static void readPreference(String extraLogs) {
+    		Log.e("dddd", "called " +  extraLogs);
```
jamorham: This and similar lines, is `dddd` just an internal tag for your use when debugging this code or does it have any special significance?

tzachi-dar:
OK,
I have moved things down in the menu to below the "Store Logs for Troubleshooting.
So I have removed all the dddd logging tags.

I have fixed the preferences conflicts.
I'm opening this as a new PR in order to make it more clear of what I have made.

jamorham:
I'm pretty sure this was already merged from another PR so I am closing those one as I don't think it is required any more.

PR #12, opened 2016-09-30
Race condition workaround for predictive alert

AdrianLxM:
This should fix the issue I have that a low prediction alarm is risen even though the trend starts to bend and with the new value in, it should not fire... but still does as it is looking at the old trend/prediction withouth regarding the new value.

NOT TESTED YET!


jamorham:
I merged this along with changes to only regenerate the data if needed. Testing locally the last few days it appears to work. Please let me know if you see the same good results with this at your end.

AdrianLxM:
Thanks.

I've been running this since a bit before the pull request and I didn't experience this phenomenon since. Happened to me more than once a day before that.

AdrianLxM:
I will intall your version tomorrow and test it.

PR #13, opened 2016-10-01
adrian calibration: default slope correction

AdrianLxM:


AdrianLxM:
There is a slight difference between 70 and .70 :-/

PR #15, opened 2016-10-03
Statistics in History

AdrianLxM:
I've added a small statistics overlay in the History:
screenshot_2016-10-03-13-57-26
It can be enabled/disabled from the menu (for small screen phones e.g.):

screenshot_2016-10-03-13-57-32
screenshot_2016-10-03-13-57-38

I checked that the values shown for "Yesterday" are the same as in the statistics activity.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/BGHistory.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/BGHistory.java
@@ -5,6 +5,8 @@ import android.app.Dialog;
 import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
+import android.view.Menu;
+import android.view.MenuItem;
 import android.view.View;
 import android.widget.AdapterView;
 import android.widget.ArrayAdapter;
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java
```diff
@@ -110,4 +128,14 @@ public String getAverageUnitised(){
         return "Avg:" + (new DecimalFormat("#.0")).format(avg*Constants.MGDL_TO_MMOLL);
     }
 
+    public String getCapturePercentage(boolean extended){
+        String result =  "Cap:" + ((possibleCaptures>0)?Math.round(getTotalReadings()*100d/possibleCaptures) + "%":"-%");
```
jamorham: Where I added Cap: for the extra status line I don't think it is possible for it to ever get to 100% the use of Math.round in your version here probably will work better. Can we get both in sync in due course?

app/src/main/java/com/eveningoutpost/dexdrip/BGHistory.java
```diff
@@ -144,8 +154,40 @@ private void setupCharts() {
...
         previewChart.setViewportChangeListener(new ViewportListener());
         chart.setViewportChangeListener(new ChartViewPortListener());
+
+        setupStatistics(date1.getTimeInMillis(), endDate.getTimeInMillis());
     }
-    
+
+    private void setupStatistics(long from, long to) {
+
+        if (Home.getPreferencesBoolean("show_history_stats", true)) {
```
jamorham: Just to confirm that defaulting to stats enabled in intentional?

AdrianLxM:
@jamorham, I added the following:
* History should now work with 2min intervals from LimiTTer
* Extra status line capture percentage is now also calculated with rounding.
* Setting for extra status line to switch to a 24h window.
* Widget will now only update once on widget line preference change and it will update on all setting changes.

Yes, it is intentional to show the stats on default. I guess this is preferred.The calculation overhead is not too much as it is triggered by the user, small timeframe and it is simple queries (no stdDev e.g.)
 Just for small screens it might be good to be able to switch it off.

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -739,7 +739,7 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
             final PreferenceCategory alertsCategory = (PreferenceCategory) findPreference("alerts_category");
             final Preference disableAlertsStaleDataMinutes = findPreference("disable_alerts_stale_data_minutes");
             final Preference widgetRangeLines = findPreference("widget_range_lines");
-            final Preference adrian_calibration_mode = (Preference) findPreference("xdrip_plus_motion_settings");
+            final Preference adrian_calibration_mode = findPreference("xdrip_plus_motion_settings");
```
jamorham: Shouldn't the preference being found be: `adrian_calibration_mode` ?

AdrianLxM:
@jamorham fixed the settings. There was a whole commit missing. Good catch!

PR #23, opened 2016-10-07
Send sensor data to followers.

tzachi-dar:
This is a first commit in updating calibration and sensor data to followers.

Still need to do the same for calibrations, and also write the code that will recognize that a sensor/calibration is missing. In such a case, will send sensor/calibration request to the remote side.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
@@ -13,20 +13,27 @@ import android.os.Bundle;
 import android.os.PowerManager;
 import android.preference.PreferenceManager;
 import android.support.v4.content.LocalBroadcastManager;
-import android.util.Log;
+import android.util.Base64;
 import android.widget.Toast;
 
 import com.eveningoutpost.dexdrip.Models.BgReading;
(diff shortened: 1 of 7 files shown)


tzachi-dar:
Added code to send calibrations with sensor.
Currently, only 15 last calibrations will be sent. I still don't understand why the number is so small, since I don't get close to the 4k limit.

app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
```diff
@@ -369,12 +431,18 @@ private static synchronized String sendMessageNow(String identity, String action
             data.putString("action", action);
             data.putString("identity", identity);
 
-            if (payload.length() > 0) {
-                data.putString("payload", CipherUtils.encryptString(payload));
+            if(action.equals("sensorupdate") ) {
```
jamorham: All payload data must be encrypted for privacy and security.

tzachi-dar:
Make sure that sensor updates are being sent on more cases:

1) When the follower sees, an unrecognized calibration.
2) When doing a bg sync.

Don't send a separate calibration update if this is a master, since we already send the sensor and all last calibrations.

PR #25, opened 2016-10-19
Add back-filled Libre readings to SendQueue.

jotomo:
Add historic/missed readings to BgSendQueue so they're pushed out to MongoDB. 
Just tested it, works for me :tm:  :-)


jamorham:
Can you explain this patch? isn't ` BgSendQueue.handleNewBgReading` already called here:

https://github.com/NightscoutFoundation/xDrip/blob/fa2cf2c4ce567e761c4f1632e602dbda6908b2fc/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java#L429

jotomo:
I looked for that call but missed it. Yet I'm only seeing those values being pushed _with_ the patch. I'll fire up the debugger when I get missed readings again to see which code paths are being run.

jotomo:
Hm, whatever it was that I saw, it wasn't down to xDrip not passing those values along, which is another issues that seems to have resolved itself ...
Sorry for the noise!

PR #27, opened 2016-10-19
Tzachi calibration sensor follower2

tzachi-dar:
This is a rebase of the previous pull request over new master.

It also adds encryption to packet.
Number of calibrations passed was decreased to 9, to make sure that it will always pass.


tzachi-dar:
Added more code:
1) Pushing of calibration data in more casses.
2) Start of implementation of pushCalibrations2(). (still no code that listens to them).

I also need to have code for override calibrations (seems simple).

I'm not 100% sure what I should do in the case of automatedCalibration(). Should they also be sent to master like the other calibrations? can I assume that only the master processes them?

PR #28, opened 2016-10-20
Fix for cascading alerts.

tzachi-dar:
See https://github.com/StephenBlackWasAlreadyTaken/xDrip-Experimental/pull/268 for more info.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -238,15 +238,14 @@ public class Notifications extends IntentService {
                 // We should not do anything if we are snoozed for the 80...
                 // If one allert was high and the second one is low however, we alarm in any case (snoozing ignored).
                 boolean opositeDirection = AlertType.OpositeDirection(activeBgAlert, newAlert);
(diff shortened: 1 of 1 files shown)


jamorham:
I read through that PR link but I just want to clarify.

Is this so that if we had for example an alert at 400 which was triggered and then snoozed and an alert at 300 it would also snooze the 300 one because a higher alert was already snoozed? Is that the new logic here?

AdrianLxM:
> Is this so that if we had for example an alert at 400 which was 
triggered and then snoozed and an alert at 300 it would also snooze the 
300 one because a higher alert was already snoozed? Is that the new 
logic here?

I didn't look at the code of the PR yet, but this was the logic we had 
in xDrip. I really grew fond of it.
You can do alarm escalation cascades like at 75mg/dl, 70mg/dl, ... but 
the 75mg/dl alert will not go off when the BG rises above 70mg/dl again 
(unless the snooze time is over of course).

Falling below 70mg/dl deletes the snooze period for the 75mg/dl alert, 
so when the BG was heading into the right direction it basically meant 
running through all alerts again.

Thanks @tzachi-dar

tzachi-dar:
Adrian explanation was correct. (Well, he has found this bug in the first place).

PR #29, opened 2016-10-20
Increase fonts for big tablets.

tzachi-dar:

Obviously, there will be more changes needed as I start playing with treatments, predictions and so on, but this gives the big tablets a much better look.


tzachi-dar:
Thanks.

PR #30, opened 2016-10-20
Fix BgReading exception jotomo found upon first access to syncBgData

kskandispersonal:
This is to resolve the exception jotomo found in the original PR for PersistentStore relating to syncBgData calling BgReading() constructure upon first access to the database.  

He also found unused methods and variables and recommended doing an Inspect code under the Analyze menu which this PR does NOT fully include, but I did remove some unused methods and variables and imports.

### DIFF
--- a/wear/src/main/java/com/eveningoutpost/dexdrip/ListenerService.java
+++ b/wear/src/main/java/com/eveningoutpost/dexdrip/ListenerService.java
@@ -4,7 +4,6 @@ import com.eveningoutpost.dexdrip.Models.BgReading;//KS
 import com.eveningoutpost.dexdrip.Models.Calibration;//KS
 import com.eveningoutpost.dexdrip.Models.Sensor;//KS
 import com.eveningoutpost.dexdrip.Models.TransmitterData;
-import com.eveningoutpost.dexdrip.Models.UserError;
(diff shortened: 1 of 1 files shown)


jotomo:
The issue causing the exception - being that the DB isn't ready at that point - is only deferred a few lines to https://github.com/NightscoutFoundation/xDrip/pull/30/files#diff-2fe96e74ad12748a3a7c51ce3bf38ccfR634, which also causes a BgReading to be constructed.
I haven't further analyzed this as I have no clue yet how Android inits anything, but @kskandispersonal said _Sensor.InitDB_ has likely not been called at that point, which initializes the DB.

kskandispersonal:
@jotomo, is the following now causing an exception?
BgReading bgData = gson.fromJson(bgrecord, BgReading.class);

I can't reproduce this.  After reviewing, I would have thought that if Sensor.InitDb was the cause then Sensor.currentSensor(), called a few lines earlier, would have also caused an exception, which it apparently does not, right?  Currently Sensor.InitDb is called upon syncSensorData() which should be the first data initialized by the phone, but it could be a timing issue and the syncBgData() request comes in first... I will re-evaluate and post a new fix.

PR #31, opened 2016-10-20
Kskandis update1

kskandispersonal:
This is to resolve the exception jotomo found in the original PR for PersistentStore relating to syncBgData calling BgReading() constructure upon first access to the database.


jamorham:
Is the changeset in this PR correct? It seems to delete quite a lot of lines, is that some other general cleanup occurring at the same time?

jotomo:
Running the code I get:
```
E/AndroidRuntime: FATAL EXCEPTION: WearableListenerService
                  Process: com.eveningoutpost.dexdrip, PID: 6850
                  java.lang.RuntimeException: Failed to invoke public com.eveningoutpost.dexdrip.Models.BgReading() with no args
                      at com.google.gson.internal.ConstructorConstructor$3.construct(ConstructorConstructor.java:111)
                      at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:207)
                      at com.google.gson.Gson.fromJson(Gson.java:814)
                      at com.google.gson.Gson.fromJson(Gson.java:779)
                      at com.google.gson.Gson.fromJson(Gson.java:728)
                      at com.google.gson.Gson.fromJson(Gson.java:700)
                      at com.eveningoutpost.dexdrip.ListenerService.syncBgData(ListenerService.java:626)
                      at com.eveningoutpost.dexdrip.ListenerService.onDataChanged(ListenerService.java:419)
                      at com.google.android.gms.wearable.WearableListenerService$zzc$1.run(Unknown Source)
                      at android.os.Handler.handleCallback(Handler.java:739)
                      at android.os.Handler.dispatchMessage(Handler.java:95)
                      at com.google.android.gms.wearable.WearableListenerService$zzb.dispatchMessage(Unknown Source)
                      at android.os.Looper.loop(Looper.java:148)
                      at android.os.HandlerThread.run(HandlerThread.java:61)
                   Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String com.activeandroid.TableInfo.getIdName()' on a null object reference
                      at com.activeandroid.Model.<init>(Model.java:55)
                      at com.eveningoutpost.dexdrip.Models.BgReading.<init>(BgReading.java:0)
                      at java.lang.reflect.Constructor.newInstance(Native Method)
                      at com.google.gson.internal.ConstructorConstructor$3.construct(ConstructorConstructor.java:104)
                      at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:207) 
                      at com.google.gson.Gson.fromJson(Gson.java:814) 
                      at com.google.gson.Gson.fromJson(Gson.java:779) 
                      at com.google.gson.Gson.fromJson(Gson.java:728) 
                      at com.google.gson.Gson.fromJson(Gson.java:700) 
                      at com.eveningoutpost.dexdrip.ListenerService.syncBgData(ListenerService.java:626) 
                      at com.eveningoutpost.dexdrip.ListenerService.onDataChanged(ListenerService.java:419) 
                      at com.google.android.gms.wearable.WearableListenerService$zzc$1.run(Unknown Source) 
                      at android.os.Handler.handleCallback(Handler.java:739) 
                      at android.os.Handler.dispatchMessage(Handler.java:95) 
                      at com.google.android.gms.wearable.WearableListenerService$zzb.dispatchMessage(Unknown Source) 
                      at android.os.Looper.loop(Looper.java:148) 
                      at android.os.HandlerThread.run(HandlerThread.java:61)
```
...

kskandispersonal:
Yes, @jamorham, I did some partial cleanup - removed some comments, removed a some unused variables and unused methods.  Pertinent code changes are mainly in syncBgData().

Would you rather I didn't do any cleanup while making a code fix?  I think I might be making it more confusing:(

kskandispersonal:
@jotomo Is this the same result as before?  So the InitDb() did not help at all.  Is this after a fresh install?  Could you post the app and wear log of the debug statements around the syncDbData call?

jamorham:
@kskandispersonal  Its quite okay to make cleanups, I just was drawing it to your attention after we had previous issues with changes reverting, I wanted to ensure you were not getting in to a situation like that again. Also it is quite easy to be working on multiple files, fix a bug in one and commit that fix and other changes which have dependencies elsewhere which are not committed and end up breaking things. I just wanted to make sure you had checked the other changes because from reading the discussion I couldn't see how they were related.

tzachi-dar:
I'm not familiar with wear, so I don't have a real input about this pr.
But I must say that all the cleanups should come in a separate file other wise it will confuse anyone looking at this commit.

jotomo:
@kskandispersonal Clean reinstall, issue persists. 
Logs: [app](https://gist.github.com/jotomo/683d667d37ec3200e2014dbf542368a2), [wear](https://gist.github.com/jotomo/d53e71777d0b1da0a3121f33214cc987)

jotomo:
@kskandispersonal No, the added _InitDB_ call doesn't make a difference, it's the same as before.

kskandispersonal:
Thanks, @jotomo for the logs and testing.  I made some changes to syncBgData and hope this will resolve the issue.  This commit fixes a few issues.  
1, BG and Calibration wear data were only getting created as new records, not updated with the phone's data.  So calibration overrides, and changes in BgReading data were not getting reflected on the wear device.  This fix now saves changes to the wear device db.
2. Several issues with preferences not being synced properly.
More thorough testing needs to be done with the preferences but this update provides additional G5 preferences and Collection method is being processed now on the wear side.

jotomo:
@kskandispersonal Thanks! I need to resolder a broken joint on my Limitter tomorrow, so I can't test right now. But I'll test it once my LimiTTer is working or again - or when you say it's a good time to test this.

kskandispersonal:
@jotomo Thanks for testing.  Maybe hold off though as I see a new issue now:(  Since this change (saving modified phone Bg and Cal records), wear is now calculating a different BG from phone, so I am doing something wrong.  I'm getting two parallel lines displayed on the watch, one for the phone and one for wear:(  I can't test now since my sensor failed but will try to do so later today.

AdrianLxM:
@kskandispersonal are you in the first 2 days of sensor start?

xDrip does use an age-adjustment on the raw value on the first two days of sersor insertion to adjust for wound healing. I'd check if you use `BgReading.age_adjusted_raw_value` and not `raw_data` if this happens in the first two days.

kskandispersonal:
@AdrianLxM Thanks, yes, wear is basically using your same app.BgReading code and does perform calculateAgeAdjustedRawValue().
However, I failed to merge in recent changes to app.Calibration and see that lastValid and latestValid methods have changed.  App.Calibration now uses criteria .where("slope!=0") whereas wear was still using .where("(slope!=0 and intercept!=0)).  I'm not sure if this is the cause for the BG parallel discrepancy between phone and wear, but lastValid is used by TransmitterData as well as BgReading among other classes.  I made the change and will retest once my sensor comes on line.  My sensor was only 11 days old but it suddenly registered a value of 1200+ and then continued at HIGH for several readings!:(

kskandispersonal:
@jamorham I've made a few more changes in the logic dealing with syncing BGs, mainly related to new calibrations performing adjustRecentBgReadings() which was causing phone and wear to become out-of-sync.  Still testing, but it is looking more stable.  
Still TODO: 
1) Add Adrian's recent, new Calibration Plugin code.  In general, app and wear "shared" code needs to be kept in-sync, otherwise, the BG graphs of the devices will become out-of-sync!  I've been doing a diff on app/wear folders and bringing in the phone code as required, but have not gotten to the Cal Plugin.  There may be other out-of-sync files, too, as I haven't done the diff for a few days now.
2) Add logic to support multiple wear devices, as requested by @taylorfowler, to track connectG5 usage.
3) Notifications, but I think this is a low priority, at least for me.
4) Investigate G5CollectionService to reduce signal drops.  I am still getting signal drops on the wear device (but better than my Samsung Galaxy Note 4 phone though!).  Wear is using the restartWatchDog() which restarts the G5 service on missedReadingAlert().  Dropped signals mainly occur when 1) glucose is changing rapidly therefore is probably a dexcom issue, and 2) when wear goes into deep sleep mode.  G5CollectionService uses setExactAndAllowWhileIdle() in keepAlive() which should allow the Alarm to wakeup the wear device when it is asleep.  However, I've seen references online where wear developers have had issues with this.  Some recommend using WakefulBroadcastReceiver() but this would change the G5 Service to an IntentService.  In the meantime, I've had great success in preventing the dropped signals (at night mainly due to deep sleep mode) by setting my App Wear setting on my phone to "Always-on screen" under the Settings - Smartwatch 3 configuration.  For the past several nights I've gotten no or limited dropped signals.  I don't know if this setting causes an increase in battery usage, but with it on, it is about 20% overnight (8 hours or so) so not huge.

kskandispersonal:
I made some fixes to reflect new calibrations which were out-of-sync w/ the phone due to not having ALL calibration records for the sensor and the phone's adjusted BGs (count rewrite_history) were not getting saved properly on the watch.  Also, updated the bgDataList collection to replace existing records, rather than just add all entries in the DataMap.  This was causing duplicate entries on the watch graph.  Another fix was related to add super() to class model constructors in attempt to resolve System: ClassLoader referenced unknown path error during debug mode.  I still need to do the TODOs listed above, as well as process other Calibration messages such as the UndoRedo.

kskandispersonal:
This last commit added the CapabilityApi to the wear app, see wear.xml under app and wear\...\res\values.  This is needed to restrict devices to only those that process the xDrip capabilities.  This is only needed if users have other wear devices paired but are not related to xDrip+ ecosystem.  I also added support for multiple wear watches.  To activate a watch as the phone's "use_connectG5" watch preference, just select that option on the target watch.  That preference along with its device name and id will be sent to the phone and will trigger the G5CollectionService to run on the identified watch device.  To switch to another watch, select the "use_connectG5" option on the new, target watch.  I have not tested this will multiple watches, only on a single watch, Sony Smartwatch 3, but it should work with multiple watches in the ecosystem.

jamorham:
All these changes sound really good. How close is this code to being ready for merge do you think?

kskandispersonal:
I think I should test a few more days.  I can't test the multi-watch scenario.  Btw, I got a 99% cap rate on the watch yesterday:)  Unfortunately, today there was a 1+hour signal loss while sleeping.  I'm not sure why because that has not happened since I started using the "Always-on screen" in the Phone's Wear App setting. It started working fine as soon as I connected the watch to the USB port and nothing showed up in the log.  Maybe it is related to the new CapabilityApi code but I don't think so...

Btw, I am getting an error in Simulation.java when treatments are passed to the watch - caught by the two catch statements in prcessingIncomingIntent.  Has this been working?  I don't know if something I did caused this to break.  I will see if I can track down the issue.

kskandispersonal:
I made a change to send / sync calibration data.  Calibration records were getting duplicated on wear when their uuid already existed in the wear db.  Now, both phone and wear are calculating the same values for BGs after a calibration, but I still find it necessary to send Calibrations.allForSensor() and rewrite_history BGs to the phone to get the correct results.   If we don't care about the rewrite_history BGs which are calculated after a calibration for the past 2-30 BGs in Calibration.adjustRecentBgReadings() then we don't have to send the rewrite_history BGs.  I find it useful to see these adjustments, and it keeps the watch in-sync with the phone display.  

I still have quite a bit of logging still in the code which I can remove but for now I've left it in.  Other than that I think it is ready to be merged but please let me know if there is something I've got wrong!

Also someone mentioned today on gitter that they would like to use the watch for G4+xbridge.  I was unaware that xbridge made it possible to use Bluetooth in place of the receiver.  Now I understand what Adrian was referring to at the beginning of this project when he mentioned he'd like to see a common collector!  I thought only G5 was able to use Bluetooth.  Anyway, the G4+xbridge collector could be added to the wear solution as well.

app/src/main/java/com/eveningoutpost/dexdrip/Services/G5CollectionService.java
```diff
@@ -266,13 +266,11 @@ public void onDestroy() {
 
         Log.d(TAG, "onDestroy");
         SharedPreferences sharedPrefs = PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
-        if (sharedPrefs.getBoolean("use_wear_connectG5", false)) {//KS
```
jamorham: Is this a fix to something which previously got broken? Can you explain the significance of this change just so I understand how it relates?

kskandispersonal: Yes, this was a fix.  I forget now the exact problem but it was causing the Wear G5CollectionService to not shutdown properly.  Instead, I moved the shutdown criteria into ListenerService which has more control and logic to determine whether G5 should be shutdown.
A related fix in the app/CollectionServiceStarter.java line 182.

PR #32, opened 2016-10-22
Import re-raise alerts from xdrip.

tzachi-dar:
Signed-off-by: Tzachi Dar tzachi.dar@gmail.com

This adds the re-raise alerts feature from xdrip.
This is needed since some phones, only play the alert once. 
So, when one is sleeping there is a chance that a single playing of the alert will not wake him.
This feature is off by default (so if the user does not change anything, nothing will happen).

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
@@ -22,6 +22,7 @@ import java.util.Calendar;
 import java.util.Date;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
@@ -152,7 +153,7 @@ public class AlertType extends Model {
      * In the case of "unclear state" for more than predefined time, return the "55" alert
      * In case that alerts are turned off, only return the 55.
      */
-    public static AlertType get_highest_active_alert(Context context, double bg) {
+    public static AlertType get_highest_active_alert(Context context, double bg, AtomicBoolean unclearReading) {
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
         if(prefs.getLong("alerts_disabled_until", 0) > new Date().getTime()){
             Log.d("NOTIFICATIONS", "Notifications are currently disabled!!");
@@ -171,6 +172,7 @@ public class AlertType extends Model {
         AlertType at;
         if (UnclearTime >= UnclearTimeSetting && bg_unclear_readings_alerts ) {
             Log.d("NOTIFICATIONS", "Readings have been unclear for too long!!");
+            unclearReading.set(true);
             Notifications.bgUnclearAlert(context);
         }
         if ((UnclearTime > 0 ) && bg_unclear_readings_alerts) {
@@ -427,15 +429,15 @@ public class AlertType extends Model {
 
 
     public static void testAll(Context context) {
-
+        AtomicBoolean unclearReading = new AtomicBoolean(false);
         remove_all();
         add_alert(null, "high alert 1", true, 180, true, 10, null, 0, 0, true, 20, true, true);
         add_alert(null, "high alert 2", true, 200, true, 10, null, 0, 0, true, 20, true, true);
         add_alert(null, "high alert 3", true, 220, true, 10, null, 0, 0, true, 20, true, true);
         print_all();
-        AlertType a1 = get_highest_active_alert(context, 190);
+        AlertType a1 = get_highest_active_alert(context, 190, unclearReading);
         Log.d(TAG, "a1 = " + a1.toString());
-        AlertType a2 = get_highest_active_alert(context, 210);
+        AlertType a2 = get_highest_active_alert(context, 210, unclearReading);
         Log.d(TAG, "a2 = " + a2.toString());
 
 
@@ -445,11 +447,11 @@ public class AlertType extends Model {
         add_alert(null, "low alert 1", false, 80, true, 10, null, 0, 0, true, 20, true, true);
         add_alert(null, "low alert 2", false, 60, true, 10, null, 0, 0, true, 20, true, true);
 
-        AlertType al1 = get_highest_active_alert(context, 90);
+        AlertType al1 = get_highest_active_alert(context, 90, unclearReading);
         Log.d(TAG, "al1 should be null  " + al1);
-        al1 = get_highest_active_alert(context, 80);
+        al1 = get_highest_active_alert(context, 80, unclearReading);
         Log.d(TAG, "al1 = " + al1.toString());
-        AlertType al2 = get_highest_active_alert(context, 50);
+        AlertType al2 = get_highest_active_alert(context, 50, unclearReading);
         Log.d(TAG, "al2 = " + al2.toString());
 
         Log.d(TAG, "HigherAlert(a1, a2) = a1?" +  (HigherAlert(a1,a2) == a2));
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
@@ -153,7 +153,7 @@ public class AlertType extends Model {
      * In the case of "unclear state" for more than predefined time, return the "55" alert
      * In case that alerts are turned off, only return the 55.
      */
-    public static AlertType get_highest_active_alert(Context context, double bg, AtomicBoolean unclearReading) {
+    public static AlertType get_highest_active_alert(Context context, double bg) {
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
         if(prefs.getLong("alerts_disabled_until", 0) > new Date().getTime()){
             Log.d("NOTIFICATIONS", "Notifications are currently disabled!!");
@@ -164,21 +164,7 @@ public class AlertType extends Model {
             return null;
         }
 
-        Boolean bg_unclear_readings_alerts = prefs.getBoolean("bg_unclear_readings_alerts", false);
-        Long UnclearTimeSetting = Long.parseLong(prefs.getString("bg_unclear_readings_minutes", "90")) * 60000;
-
-        Long UnclearTime = BgReading.getUnclearTime(UnclearTimeSetting);
-
         AlertType at;
-        if (UnclearTime >= UnclearTimeSetting && bg_unclear_readings_alerts ) {
-            Log.d("NOTIFICATIONS", "Readings have been unclear for too long!!");
-            unclearReading.set(true);
-            Notifications.bgUnclearAlert(context);
-        }
-        if ((UnclearTime > 0 ) && bg_unclear_readings_alerts) {
-            Log.d(TAG_ALERT, "We are in an clear state, but not for too long. Alerts are disabled");
-            return null;
-        }
         at = get_highest_active_alert_helper(bg, prefs);
         if (at != null) {
             Log.d(TAG_ALERT, "get_highest_active_alert_helper returned alert uuid = " + at.uuid + " alert name = " + at.name);
@@ -429,15 +415,14 @@ public class AlertType extends Model {
 
 
     public static void testAll(Context context) {
-        AtomicBoolean unclearReading = new AtomicBoolean(false);
         remove_all();
         add_alert(null, "high alert 1", true, 180, true, 10, null, 0, 0, true, 20, true, true);
         add_alert(null, "high alert 2", true, 200, true, 10, null, 0, 0, true, 20, true, true);
         add_alert(null, "high alert 3", true, 220, true, 10, null, 0, 0, true, 20, true, true);
         print_all();
-        AlertType a1 = get_highest_active_alert(context, 190, unclearReading);
+        AlertType a1 = get_highest_active_alert(context, 190);
         Log.d(TAG, "a1 = " + a1.toString());
-        AlertType a2 = get_highest_active_alert(context, 210, unclearReading);
+        AlertType a2 = get_highest_active_alert(context, 210);
         Log.d(TAG, "a2 = " + a2.toString());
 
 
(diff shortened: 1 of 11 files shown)


jamorham:
Hi, thanks for submitting this, I have a few questions about it.

Does this only affect "other alerts" ? Would it affect low prediction and persistent high alerts which I think also use some of the same alerting framework.

What is the reason for passing the `unclearReading` boolean and does this need to be an `AtomicBoolean` ?

app/src/main/java/com/eveningoutpost/dexdrip/Services/SnoozeOnNotificationDismissService.java
```diff
@@ -42,4 +60,10 @@ protected void onHandleIntent(Intent intent) {
 
         AlertPlayer.getPlayer().Snooze(getApplicationContext(), snooze);
     }
+    
+    private void snoozeOtherAlert(String alertType) {
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
+        int snoozeMinutes = MissedReadingService.readPerfsInt(prefs, "other_alerts_snooze", 20);
+        UserNotification.snoozeAlert(alertType, snoozeMinutes);
```
AdrianLxM: Shouldn't this be only called when the reraise feature is activated?

Scenario: Reraise feature is deactivated as the user does just want an unintrusive Notification sound; Snooze is set to 30 minutes.
At time 0 a predictive alert happens. At time 25 the user looks at the phone and swipes away the notification.
What happens at time 30?
I believe with the current logic a new predictive alert would rise. After the PR it would still be snoozed until time 55.

app/src/main/res/xml/pref_notifications.xml
```diff
@@ -171,8 +171,20 @@
                         android:defaultValue="20"
                         android:key="other_alerts_snooze"
                         android:numeric="integer"
-                        android:summary="Minimum number of minutes to pass before raising the same alert."
+                        android:summary="Number of minutes before raising the same alert after snooze."
                         android:title="Alert Snooze" />
+                    <CheckBoxPreference
+                        android:key="disable_alerts_reraise"
```
AdrianLxM: Isn't this more an "enable" than a "disable"?

app/src/main/java/com/eveningoutpost/dexdrip/Services/SnoozeOnNotificationDismissService.java
tzachi-dar: This is correct that there is a functional behavior change here. In the case that you mention, the next alert will be after 55 minutes.
I think that this is ok, since usually, when alert is snoozed the snooze time will be calculated from current time.
Until my change, there was no way to snooze the other alerts, and I think that this is something that is needed.
Still, if we fill that we don't want any behavior change, I can easily change that.

app/src/main/res/xml/pref_notifications.xml
tzachi-dar: Yes, you are right, I'll change that soon.

tzachi-dar:
This will affect also low prediction and persistent high alerts (if the user chooses to use them. See also answer to adrians question).
I think that since this alerts happen even more frequently, they can enjoy from a snooze feature even more. Obviously, we can change that not to affect them, but I believe that the main reason that this alerts did not receive a snooze feature was since we didn't have the resources to add it.
In any case, it might also be a good time to add a snooze time for each of them...

As for the atomic variable. 
We need to pass a return value from a few functions. 
onHandleIntent() ->notificationSetter()->FileBasedNotifications()->get_highest_active_alert().
Some of them already have a return value, others don't but I don't think that using the unclearReading as their return value is reasonable.
There are a few ways that are being discussed on how to return more than one value in java (see http://stackoverflow.com/questions/457629/how-to-return-multiple-objects-from-a-java-method) but all have a disadvantege : the parameters don't have names. For example an array of objects, where we have the first parameter and the second but with no names. We can have a pair, but again, different variables will not have a name. We can also have a class defined for each function, but this looks too complicated.
So, the way that I chose was to pass the variable by reference. This is very common in C++ (and less in java), but I think that this makes sense in this example.
Following http://stackoverflow.com/questions/4319537/how-do-i-pass-a-primitive-data-type-by-reference brought me to the atomic types, which are already built in.

jamorham:
Persistent High and Predicted Low alerts already can be snoozed.

With the AtomicBoolean is it only `set()` in one place and `get()` in another?

tzachi-dar:
Yes, for the AtomicBoolean.
How can this alerts be snoozed?

jamorham:
With the AtomicBoolean, so is this created in `Notifications.java` and only ever set after it is calculated in `AlertType.java` and then passed onward through the methods until it is used in `calcuatleArmTimeUnclearalert` ? Is it used/modified anywhere else?

tzachi-dar:
Just like you said, starts as false, get set in get_highest_active_alert and being used in calcuatleArmTimeUnclearalert.

jamorham:
Okay, I think there is a simpler and more elegant way of doing this which also allows for future expansion without requiring lots of changes to methods. The unclear readings boolean is a representation of the database involving processing which is also only applicable with a data source which supports filtered.

I will annotate the listing where I think you could pass this parameter in a different way and there are also some optimizations which can be applied but we can deal with that in a later PR.

app/src/main/java/com/eveningoutpost/dexdrip/Models/AlertType.java
```diff
@@ -171,6 +172,7 @@ public static AlertType get_highest_active_alert(Context context, double bg) {
         AlertType at;
         if (UnclearTime >= UnclearTimeSetting && bg_unclear_readings_alerts ) {
             Log.d("NOTIFICATIONS", "Readings have been unclear for too long!!");
+            unclearReading.set(true);
```
jamorham: How about instead:
`PersistentStore.setLong("unclearReading",2); // 0=unset  1=false  2=true`

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
```diff
@@ -111,6 +112,7 @@ protected void onHandleIntent(Intent intent) {
         PowerManager pm = (PowerManager) getApplicationContext().getSystemService(Context.POWER_SERVICE);
         PowerManager.WakeLock wl = pm.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, "NotificationsIntent");
         wl.acquire(60000);
+        AtomicBoolean unclearReading = new AtomicBoolean(false);
```
jamorham: How about instead:
`PersistentStore.setLong("unclearReading",1); // 0=unset  1=false  2=true`

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
```diff
@@ -338,26 +341,69 @@ private void notificationSetter(Context context) {
...
+        Log.d("Notifications" , "calcuatleArmTimeBg returning: "+ new Date(wakeTimeBg) +" in " +  (wakeTimeBg - now)/60000d + " minutes");
+        return wakeTimeBg;
+    }
+    
+    
+    
+ // This is the absolute time, not time from now.
+    private long calcuatleArmTime(Context ctx, long now, boolean unclearAlert) {
+        Long wakeTimeBg = calcuatleArmTimeBg(now);
+        Long wakeTimeUnclear = calcuatleArmTimeUnclearalert(ctx, now, unclearAlert);
```
jamorham: How about instead:
`Long wakeTimeUnclear = calcuatleArmTimeUnclearalert(ctx, now, (PersistentStore.getLong("unclearReading") == 2) ? true : false);`

Then none of the reference passing is needed which I think would simplify this process and the PR.

We can later also optimize the calculation segment with `DexCollectionType.hasFiltered()` because I don't think the database needs to be processed when using a different sensor type.

tzachi-dar: Using persistent storage is actually using a global variable but with another technique.
I think that having it a class member in notifications.java will make it more readable.

This leaves us with the question of how to return it from getHighesetActiveAlert(). I guess that we can live it as atomic boolean, or return a pair (this will only be one pair this time).

Does it sound reasonable?

tzachi-dar:
Here are some of the reasons to avoid global variables:

http://stackoverflow.com/questions/17874946/is-there-any-disadvantage-to-declare-a-variable-global
http://stackoverflow.com/questions/484635/are-global-variables-bad
http://arstechnica.com/tech-policy/2009/05/buggy-breathalyzer-code-reflects-importance-of-source-review/

AdrianLxM:
I agree with Tzachi, that persistant storage has the smell of global variables and will introduce states that all need to be covered.
It will mean that the order in which methods are called cannot be changed. Also such states are hard to manage as soon as concurrency is present. And this really can happen that the alarm system is called twice (once from a timer, then from a late packet received e.g.). IMHO this would mean a higher effort in synchronization and/or sanity checks.

At the moment it would look like this:
```
a(){
	getHighest()
	c()
}

 getHighest(){
	unclear= calculate_unclear();
	make_persistant(unclear);
	if (unclear) {
		// fire unclear alert
	}
	else do something else;
 }

 c(){
	read_unclear()
	do_sanity_check_for_unclear();
	if (unclear) do something
	else do something else;
 }

calculate_unclear() {
	return ...
}
```

Why not calculating `unclear` one step higher in the call hierarchy? This would also make `getHighesetActiveAlert()` really do what the name says and not also trigger "unclear alerts".

```
a(){
	unclear = calculate_unclear_and_fire_alert_if_necessary()
	getHighest(unclear)
	c(unclear)
}

 getHighest(boolean unclear){
	if (unclear) do something or possibly nothing
	else do something else and return highest;
 }

 c(boolean unclear){
	if (unclear) do something
	else do something else;
 }

calculate_unclear_and_fire_alert_if_necessary() {
	// check if unclear and fire
	// return boolean if "consumed" the alarm (by firing unclear alert)
	return ...
}
```

cheers, Adrian

tzachi-dar:
I tend to agree with adrian. Seems like the mess has started since calculation of unclear reading is done in get_highest_active alert().
I'll take it to it's own function and things will be much clearer.

jamorham:
Okay this sounds like a good plan. My suggestion for using persistent storage was just because as far as I could see the boolean was only a product of the contents of the database and just dependent on the state of that.

With moving the calculation of unclear reading to a higher function I think it could also maybe be optimized in two ways:

1) Does it need to be calculated at all if the preference option for unclear alerts is disabled? 
2) Does it need to be calculated at all if the data source does not support "filtered" data? There is `DexCollectionType.hasFiltered()` which can check for this.

I may not fully understand the usage of unclear readings so please consider the implications of those optimizations.

tzachi-dar:
Here is a small explanation of the feature:

Sensors might get noisy sometimes. In such cases especially at night, there is on point in alerting, there is a good chance that it will be a false alarm.
But how is noisy system defined?
Well the definition that I was using was simple: for each bg measurement compare (bg_raw / bg_filtered). If 0.7< result < 1.3 than this is not defined as noise. If there is noise in one point we do not alert for the next 30 minutes.

This is the highlight of the feature. Will probably need to improve it in future, although I must say from my experience this gives very good results in detecting a failing sensor or a night comprasion.

All that said:
1) Does it need to be calculated at all if the preference option for unclear alerts is disabled?  - Well no, currently we calculate it but ignore the results... I'll fix it.
2) Does it need to be calculated at all if the data source does not support "filtered" data? No - will fix it to use  DexCollectionType.hasFiltered().

And another question going back to the beginning of this mail thread: 
How are low prediction and persistent high alerts snoozed? can you give me a pointer to the code.

AdrianLxM:
@tzachi-dar, not sure if I mentioned it before: It would be good to not have a behavioural change if the new setting is not ticked off.

Did you compare this to the noise model that is used for the predictive alerts?

jamorham:
Thanks for the explanation of the unclear readings.

The code relating to having snoozed low prediction is here: https://github.com/NightscoutFoundation/xDrip/blob/ce127382227f66e87737500fc3f11b04f9684a3f/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java#L692L696 there is very similar code for the persistent high alert.

tzachi-dar:
Ok, this snoozes these alerts as it snoozes all low/high alerts.
I think that this alerts deserve a snooze of their own as well.
My code will not change that, but it will allow one to do the following:
1) snooze only this alert for a given time. (not all low alerts)
2) allow to control the time of re-raise for this alert. (This is important if you want to make sure you hear this alert).

Do you want me to add this two alerts to this pr, or do you want to live this for some other time?

jamorham:
I think we should try to keep this PR as simple as possible for now especially as we need to ensure that changes don't have any unintended consequences.  If that means not having re-raise for the persistent high and low predict then that might be the ideal as they do already have some repeating. We can look at that later I think? 

I agree with @AdrianLxM that it should be possible to have the behavior exactly as it is at the moment with the option in its default state but allow re-raise to be enabled for those who want it. Everyone has different needs.

AdrianLxM:
@jamorham, at the moment the alarm will NOT repeat. It will just sound once and then will be automatically snoozed.

I would like the feature to have the predictive low alert being called every 30 seconds or minute until I dismiss the notification and only then enter the snooze period.

@tzachi-dar keeping the old behaviour would be just one line of code.

tzachi-dar:
I think that having this feature per alert (with it's own ui) should not be
too defecault, So I can do that and make anyone happy.

בתאריך 25 באוק׳ 2016 6:40 אחה״צ,‏ "AdrianLxM" <notifications@github.com>
כתב:

tzachi-dar:
Just added 3 commits to this pr:
1) Move the calculating of unclear alerts to a higher level, so it can be used by all places in stack.
    Also calculate it only when filters data exists, and feature is turned on.
2) Make sure that behavior is preserved unless the user has specifically decided to use shorter reraise.
3) Rename the disable alerts reraise (enable alerts reraise).

My next step, if there are no major changes is to add the reraise feature to each of this alerts specifically, so users can choose which reraise time they want for each alert. 

Please note that I haven't yet tested any of this changes.

tzachi-dar:
Updated all alerts to have their own snooze and reraise options.
Still needs testing.

tzachi-dar:
Here is some short status.

Missed reading alerts seems to be working well.

persistentHighAlert - seems not to work that well. (some thought of alert immidiatly that needs to be debugged.)
predictive low - does not seem to work well.

Looking at the code it seems that the two alerts above don't have a mechanism that will call them according to the time of reraise. (In other words, they are only called again when there is a new bg, or the timer of other alerts has expired, but they don't have a timer of their own).

It seems to me that this pr is getting too big, so I wander what is the right step right now.
I think that not enabling re-raise for them at the current step is the right thing to do.

Any thoughts?
Tzachi

jamorham:
If there are problems with the persistent high and predictive low then I reckon remove the extra re-raise functionality from those so we can get this merged more easily and re-visit those later?

tzachi-dar:
I have returned persistent high and predictive low to their original forum.

That said, it seems that moving unclear reading a place has caused a few side effects. I hope I have handled them all, but I continue in my testing.
Anyone who has BW to test is welcomed to do it.

Tzachi

jamorham:
What sort of side effects might be seen? Do you think this would be safe to merge yet?

tzachi-dar:
The main side effects that I have seen were caused by moving the place where unclerreading was calculated.
In the original code, there have been the following changes (compared with moving the check higher):
1) unclerreading was checked after all alerts disabled was checked. (so there was no such alert, if alerts are disabled.)
2) unclerreading was checked a few times, but the boolean itself was only set if it was unclear for a long time (long enough to trigger the alert ).
3) As a matter of fact, the alert was never removed from the db tracking the alerts. This had influence on the time calculations later.

I hope I have fixed all this issues, by yesterdays checkin.

I have by now, completed the tests of missed reading and persistent high, and still need to test predicted low and unclear readings, so I probably need 1-2 more days to be sure that things are working.

tzachi-dar:
OK, all alerts were checked, can be merged now.

PR #34, opened 2016-11-04
Take snooze time from the correct place, for missed-readings and noisy sensor alerts.

tzachi-dar:
Still need some more testing.


tzachi-dar:
Tested, working correctly.

PR #36, opened 2016-11-09
Return to the previous volume in the case that an alert has been snoozed.

tzachi-dar:
Also fix error printing.
Still not tested through.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
@@ -90,8 +90,8 @@ public class AlertPlayer {
 
     private final static String TAG = AlertPlayer.class.getSimpleName();
     private MediaPlayer mediaPlayer;
-    int volumeBeforeAlert;
-    int volumeForThisAlert;
+    int volumeBeforeAlert = -1;
+    int volumeForThisAlert = -1;
 
     final static int ALERT_PROFILE_HIGH = 1;
     final static int ALERT_PROFILE_ASCENDING = 2;
@@ -147,6 +147,7 @@ public class AlertPlayer {
             mediaPlayer.release();
             mediaPlayer = null;
         }
+        revertCurrentVolume(ctx);
     }
 
     //  default signature for user initiated interactive snoozes only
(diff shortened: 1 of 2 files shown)


TecMunky:
I tested this, and it "almost" works.  "Ascending" volume selected for this test.

When the audio alarm kicks in, it drops the volume to zero.  When the alarm is dismissed, the volume returns to previous setting.  But, until the alarm is dismissed, the volume is zero and the program playing in the background continues playing, at zero volume.  The volume can still be adjusted.  

What I expect is that the program will pause until the alarm is dismissed or finishes playing, but that does not happen.

What actually happens:
first alert - vibrate, volume does not change
second alert - vibrate, volume drops to zero - after alarm stops volume returns to previous setting
third alert - vibrate and sound - volume drops to just above zero during sound, at completion the volume returns to previous setting
following alerts - vibrate and sound - volume successively raises for each alert, and at completion of sound volume returns to previous setting.

During alerts, the program continues playing and does not pause - this is an issue.  For every other app, when sounds occur, my podcast player pauses during the sound - if this happened I would be okay with the current behavior - but as it is I miss some of the program if I don't take action by rewinding, or dismissing the alarm immediately.  If I happen to be driving, this is extra workload - which is potentially dangerous.

This is not a complete solution.

TecMunky:
I tried again using the "medium" volume setting for alerts, and playing sound from a different app - the volume changed to a medium level during the alarm, then returned to the previous setting when the alarm was finished - but the program playing still did not pause during the alarm.

tzachi-dar:
Hi,

Thanks for testing this.

This PR is mainly to see what happens when one is actively dismissing an alert, not waiting for it to finish.

The other feature you have asked for is audio focus. http://stackoverflow.com/questions/15390126/how-to-stop-other-apps-playing-music-from-my-current-activity 

I'll look at it, but it is not 100% related with this specific pr.

TecMunky:
UPDATE:  I just experienced a notification from another app while I was listening to a podcast.  The volume decreased during the notification sound and vibration, and the program continued playing during the notification.  

I therefore conclude that the problem is not with this code fix, but something to do with my phone setup.

That means that this fix works, and can be merged into the code base.

jamorham:
@tzachi-dar  Is this sufficiently tested for merging do you think?

tzachi-dar:
I have been running with this for a few days so I think that it is ok.
After the merge I'll try to also take the audio focus and we will see how it will work.

jamorham:
I see some changes to the alarm time as well. Are those related?

tzachi-dar:
There are 3 changes in this pr:
1) Change for the volume itself after snooze.
2) Change for printing itself (should not change anything in behavior, but the printing about negative time should not happen anymore).
3) Changes of what happens when we are asked to do the arming "now". This is a case that should not happen, but if it will happen, we will be in an infinite loop, where the timer is fired all the time. So my mitigation to this was to sleep for a second. In the code I had a bug that we slept for 1ms, and I have now fixed that to be a second.

PR #38, opened 2016-11-19
Tzachi calibration sensor follower5

tzachi-dar:
Last code for this pr, with a few changes to logging code.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
@@ -227,12 +227,17 @@ public class GcmActivity extends Activity {
         }
     }
 
-    public static void syncSensor(Sensor sensor) {
+    public static void syncSensor(Sensor sensor, boolean forceSend) {
         Log.i(TAG, "syncSensor called");
         if(sensor == null) {
             Log.e(TAG, "syncSensor sensor is null");
             return;
         }
+        if((!forceSend) && JoH.pratelimit("GcmSensorCalibrationsUpdate", 300) == false) {
+            Log.i(TAG, "syncSensor not sending data, because of rate limiter");
+            return;
+        }
+
         String json = sensorAndCalibrationsToJson(sensor);
         GcmActivity.sendMessage(GcmActivity.myIdentity(), "sensorupdate", json);
     }
@@ -342,6 +347,8 @@ public class GcmActivity extends Activity {
     }
 
     public static void syncBGTable2() {
+        // Since this is a big update, also update sensor and calibrations
+        syncSensor(Sensor.currentSensor(), true);
         new Thread() {
             @Override
             public void run() {
@@ -399,6 +406,13 @@ public class GcmActivity extends Activity {
             GcmActivity.sendMessage("sbr", ""); // request sensor battery update
         }
     }
+    
+    public static void requestSensorCalibrationsUpdate() {
+        if (Home.get_follower() && JoH.pratelimit("SensorCalibrationsUpdateRequest", 300)) {
+            Log.d(TAG, "Requesting Sensor and calibrations Update");
+            GcmActivity.sendMessage("sensor_calibrations_update", "");
+        }
+    }
 
     public static void pushTreatmentAsync(final Treatments thistreatment) {
         new Thread() {
@@ -442,6 +456,10 @@ public class GcmActivity extends Activity {
 
     public static void pushCalibration(String bg_value, String seconds_ago) {
         if ((bg_value.length() == 0) || (seconds_ago.length() == 0)) return;
+        if (Home.get_master()) {
+            // For master, we now send the entire table, no need to send this specific table each time
+            return;
+        }
         String currenttime = Double.toString(new Date().getTime());
         String tosend = currenttime + " " + bg_value + " " + seconds_ago;
         sendMessage(myIdentity(), "cal", tosend);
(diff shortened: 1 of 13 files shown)


tzachi-dar:
Added a few additions to this PR, after sending data to xdrip using curl.
1) Protect against IlligalArgumentException.
2) Protect against null calibration.
3) Send an array of calibrations instead of only a single calibration (only change of protocol right now, implementation can be changed later).

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -928,6 +940,7 @@ public String toJSON() {
             jsonObject.put("raw_calculated", raw_calculated);
             jsonObject.put("raw_data", raw_data);
             jsonObject.put("calculated_value_slope", calculated_value_slope);
+            jsonObject.put("calibration_uuid", calibration_uuid);
```
tzachi-dar: Looking at a,b,c I think that they are used on the fllower. For example on the function:
displayCurrentInfo()
that calls displayCurrentInfoFromReading().
That calls estimated_bg().

There are many other calls that seems to be using a,b,c.
I'm not sure that the extra data is a real issue except for when sending data for the entire day, which should not really happen (but I guess that data costs differently in other places).

If we do want to save this data, than my suggestion is this:
Since a,b,c are double, they typically looks like:
a = 2.67846736149458e-10, b = -772.056833308984, c = 556355252287477.0
I believe that keeping more than 5 digits has no real influence on our accuracy. 
So we can probably send:
a = 2.6784, b = -772.05, c=55635e10
This will have almost the same values, but will use a much smaller space size.
I haven't still looked at how this look at json, but I wanted to know what do you think.

Thanks
Tzachi

tzachi-dar: By the way, another solution, is to only add the calibration string, when sending a single packet, but not when sending more than one packet.

jamorham: The `estimated_bg` method is only used by the depreciated and broken "predictive" mode. This setting is only accessible via engineering mode and is checked and warned for by `checkBadSettings()` if anyone were to enable it. Eventually I think it should be removed completely. I don't think that is needed, especially not on the follower. 

The other usages of a,b,c are just when calculating curves for calibration which should only happen on the master. What do you think? I reckon it should be possible to drop them unless I have missed something but I did look at this before when I was trying to work out exactly what a,b,c even were for.  I can test removing these tonight to see whether there is any visible impact on the follower.

Reducing the size of them might be an idea but I am not certain what the implications of that might be. 

Your other suggestion of only adding the calibration uuid on single entries might also work as a good alternative. I would have to look at the code flow as to whether we know this ahead of time. For example when batching the entries due to rate limit (which would normally only happen with an nfc based sensor)

jamorham: To answer my own question, yes we do appear to know ahead of time when a json entry is going to be part of a batch so we could exclude the calibration uuid in that case? 

Seems like the least impact all round solution with only a slight increase in complexity, eg adding a boolean to indicate reduced size desired of the json record and add a wrapper for the default signature when a boolean is not present for the original behavior?

jamorham:
An impact of this PR is that if a follower updates to a new version but the master doesn't then the `cal2` message wont be understood but I think that is a small price to pay for the additional functionality and I think somewhat unavoidable given the design of the original `cal` message. It will be fine the other way around though I believe with the master understanding both message types so upgrade of the master first shouldn't cause an issue.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
tzachi-dar: OK, so I'll fix this according to the last suggestion.

tzachi-dar:
I believe that if new messages will be added using json, it will be easier to extend them next time that there is a need to do that.

tzachi-dar:
I have exclude the calibration uuid when sending syncBGTable2.

As a side note, I have seen that bgs message reaches a few times to the master and the follower.
I believe that it should be easy to make sure that the master does not get it's own messages.
This will probably save a lot of data. What do you think ? (not as part of this PR).

app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
```diff
@@ -242,10 +242,11 @@ public void run() {
     }
 
     public synchronized static void syncBGReading(BgReading bgReading) {
+        Log.d(TAG, "syncBGReading called");
         if (JoH.ratelimit("gcm-bgs-batch", 15)) {
-            GcmActivity.sendMessage("bgs", bgReading.toJSON());
+            GcmActivity.sendMessage("bgs", bgReading.toJSON(true));
         } else {
-            PersistentStore.appendString("gcm-bgs-batch-queue", bgReading.toJSON(), "^");
+            PersistentStore.appendString("gcm-bgs-batch-queue", bgReading.toJSON(true), "^");
```
jamorham: I think batch queue entries also should be `false` as they are space sensitive and will come in a block preceded by a standard solo reading.

tzachi-dar:
Change done.

jamorham:
Testing this locally now..

jamorham:
Merged! Thanks.

tzachi-dar:
Thanks.

PR #43, opened 2016-12-07
If the snooze dialog was not pressed after 120 seconds it will be rem…

tzachi-dar:
…oved from the screen.

I have noticed that many times, the confirm remote snooze dialog is not pressed.
This happens when the user dismisses the alerts from the notification dialog, or simply when he just snoozes the alerts and does not look on the tablet any more.

This will close the dialog.

Verified to work, but not really tested through.


tzachi-dar:
Ok, tested it for a few days on 4 machines. Worked well.
I think it can be merged.

tzachi-dar:
Thanks.

PR #47, opened 2016-12-22
Split the isBTWixel to two functions:

tzachi-dar:
isBTWixel, isLimiter

This will make code more readable.
This is very simple code when looking in the eye, but more complicated to test considering all the needed devices.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/BluetoothScan.java
@@ -350,7 +350,8 @@ public class BluetoothScan extends ListActivityWithMenu {
 
         } else if (device.getName().toLowerCase().contains("drip")) {
             if (!
-                    (CollectionServiceStarter.isBTWixel(getApplicationContext())
+                    (CollectionServiceStarter.isBTWixel(getApplicationContext()) ||
+                            CollectionServiceStarter.isLimitter(getApplicationContext())
                             || CollectionServiceStarter.isWifiandBTWixel(getApplicationContext())
                     ) || CollectionServiceStarter.isLimitter()) {
                 prefs.edit().putString("dex_collection_method", "BluetoothWixel").apply();
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -1749,6 +1749,7 @@ public class Home extends ActivityWithMenu {
             btnRedo.setVisibility(View.INVISIBLE);
         }
         boolean isBTWixel = CollectionServiceStarter.isBTWixel(getApplicationContext());
+        boolean isLimter = CollectionServiceStarter.isLimitter(getApplicationContext());
         // port this lot to DexCollectionType to avoid multiple lookups of the same preference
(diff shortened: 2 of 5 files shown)


tzachi-dar:
By the way, another option for this pr, is simply rename the function isBTWixel isBTWixelOrLimitter.

jamorham:
All new code relating to this sort of thing has instead been using the helper class `DexCollectionType` which is a function set and enum orientated way of making the same determinations avoiding much of the convoluted boolean logic seen with the use of the functions from `CollectionServiceStarter`. 

This allows for more efficient lookup, easier expansion for new collection systems, reduces the risk of typos and focuses on the function required rather than the specifically selected collection type. 

For example in `DexCollectionService` which supports both the classic bluetooth xdrip protocol and xBridge protocol and needs to know whether to set its failover timers, the logic can just be simplified to:  `if (DexCollectionType.hasBtWixel() || DexCollectionType.hasXbridgeWixel()) {` 

If we want to later add specific support for example for BlueReader or Transmiter devices then if these are supported by the same service then it is as simple as adding `BlueReader` to the `usesBtWixel` collection. The collections also allow for combinations like WiFi and Bluetooth without it requiring complex logic, that combo collection type is simply in both sets.  Imagine if we want to add LimiTTer & WiFi dual collection method in the future, it is really easy to do with `DexCollectionType`

Likewise anything which uses bluetooth at all can be in `usesBluetooth` and checked with `DexCollectionType.hasBluetooth()` for more generic bluetooth functions such as automatically switching bluetooth etc. Wherever there is a function which might start to involve excessive boolean logic for differentiation between different collection types we can add a helper set and/or method to simplify this.

In other places like `TransmitterData` for differentiating LimiTTer specific protocol quirks we use like: `DexCollectionType.getDexCollectionType() == DexCollectionType.LimiTTer`

I don't think it is really worthwhile to add extra helper methods to `CollectionServiceStarter` like this because, while being readable, they make the logic more complex and boolean conditionals using them would have to be reproduced anywhere else they were used. `DexCollectionType` is a more extensible and simpler way of doing the same thing. 
 
At the same time I haven't gone over all the functions and eliminated the use of the old methods due in part to the rigorous testing needed to evaluate that the logic remains correct for every combination which might occur. There are some significant complexities, such as what happens when switching from one collection type to another. The current code works, eventually the old code can be replaced but I prefer a gradual approach as new code or changes to the logic require specific work and testing on one part where any breakage will become obvious as part of the testing of that specific feature.

kskandispersonal:
Thanks, @jamorham, this helps me to understand the services a bit more as it relates to G4 on the watch.  I am thinking that only the DexCollectionService needs to be supported on the Watch which is currently started by CollectionServiceStarter.startBtWixelService().  Or would there be a requirement for DexShareCollectionService (ie., the dexcom G4 Share Receiver) on the watch, as well?  I assume all of the wifi collection service code is out-of-scope for the watch as well.  Right now wear ListenerService and the app WatchUpdaterService handles most of the collection service start/stop via processConnectG5().  This may need to be replaced by including something like CollectionServiceStarter code.

kskandispersonal:
Yes, that is what I thought, any `usesBluetooth` collection should be allowed on the watch, and then look to `CollectionServiceStarter` for the actual required service `start/stop` processes.

tzachi-dar:
OK,
Thanks for the detailed explanation.

PR #49, opened 2016-12-30
Add Dexcom BT collection services to wear

kskandispersonal:
Add Dexcom BT collection services to wear:
1. Added BT Collection Services (DexCollectionType.usesBluetooth) to wear.  One of these services will be enabled if enable_wearG5 is enabled when phone is not-in-range.  The DexCollectionType.DEX_COLLECTION_METHOD is determined by the Phone prefs and this is sent to the wear device.  A pref listener will stop/start the service depending on its value.  If force_wearG5 is enabled, the wear collection service will not be stopped.  It will continue to run on the wear device and will not be initiated on the phone.  This process is the same as the existing G5CollectionService process.
2. To assist in debugging, a wear pref, dex_collection_method, has been added to XDrip Prefs for display purpose only.  It will be updated whenever the method changes on the Phone.
3. Currently, G5CollectionService, DexShareCollectionService and DexCollectionService (used by Wixel, LimiTTer, etc.) are supported.  These wear versions are largely unchanged from their corresponding app versions.
4. xDrip+ app maintains several BT Advanced Settings.  Many of these are used by DexCollectionService, and a few by DexShareCollectionService.  Rather than re-create each of these on the wear device, they are sent via SendPrefData and watch will use the phone's settings.  If this proves to be problematic, they can be added separately on the watch.  In particular, DexCollectionService sets the bridge_battery preference.  Therefore, this preference is sent back to the phone to be synchronized with its preferences.  This process should be confirmed for its validity.
4. Note, the preference title/summary have been updated for the English strings, but their IDs have yet to be updated. So for example, enable_wearG5 is still be used in the code. These can be updated once testing is successful.
5. Testing was performed only G5 since hardware for the other data sources is not available.  Simple selection of the various data source shows they appear to work - handoff of services start / stop appear to by synchronized correctly but of course, no bgReadings could be confirmed.


jamorham:
Closing to recreate..

PR #52, opened 2017-01-05
Fixes for BT collection services on wear 

kskandispersonal:
Attempt to fix issue gitter user has with syncing data between phone and watch upon reconnection after watch was in standalone mode.  User says old phone data is getting sent to watch upon connection and being displayed on the watchface main screen (not chart).  
Also, synced app Calibration with wear Calibration for standalone pertinent methods (excludes calibration plugin and alarm/notifications code).


jamorham:
Do you think this is ready to be merged? I assume you have tested it locally even though you were unable to reproduce the original issue?

kskandispersonal:
Yes, it is ready to merge.  I did test locally w/ G5 and was never able to reproduce the issue.

PR #56, opened 2017-01-10
Fix wear issues including Battery Usage

kskandispersonal:
Users seemed confused with the original Battery Usage display on the watch, so the following fixes were made which also help identify which device is running the collection service.	
- Ensure Uploader battery displays phone battery when using the phone collector by calling sendData() which contains the phone's battery usage to after calls to sendWearBgData().
- Display "Uploader" or "Wear" battery usage & label to correspond to the device the collector is running on.
- Enable bridge_battery display only for DexCollectionType.hasBattery().
- Sync bridge_battery pref between watch and phone.  Send bridge_battery as dataMap element to allow the watch to display bridge_battery even when wear_integration is not enabled.

Unrelated to Battery Usage, but null pointer exception sometimes occurs in G5CollectionService when accessing UserError:
- Ensure database is initialized on entry into G5CollectionService which may use the UserError table.
- Remove duplicate BgReading and Calibration tables from manifest.

This code has been tested on G5 only.


kskandispersonal:
Fixed and tested code to prevent onConnected onCapabilityChanged being triggered multiple times.  I don't think this is related to the issue Mondelman is experiencing though.  It is ready to be merged.

PR #60, opened 2017-01-18
Add the data that is received from BT to the printing (in the case th…

tzachi-dar:
…at we are already printing).

This makes debugging the data that is received from Limitter for example simpler.
Only works when one has already enabled printing on this tag.
Was checked against all bytes. this seems to give good results when ascii text is sent (prints well to our log mechanism).
Negative bytes, are not printed well, but are not part of limiter/ xdrip data.
For some reason it does not print nice to my pc. Probably wrong character set or something similar.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
@@ -377,7 +377,7 @@ public class DexCollectionService extends Service {
             wakeLock1.acquire();
             try {
                 final byte[] data = characteristic.getValue();
-                Log.i(TAG, "onCharacteristicChanged entered " + HexDump.dumpHexString(data, 0, data.length));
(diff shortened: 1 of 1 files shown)


jamorham:
Elsewhere for bluetooth debugging we would normally print the non-ascii data with for example, `JoH.bytesToHex(data)`  Would it be worth including a hex dump for non-ascii protocols like xBridge, perhaps by using the `HexDump.dumpHexString(data)` from the usbserial package?  I believe it will print both Hex and ascii printable at the same time and is designed for this sort of thing.

Similar question, can invoking String on a byte array ever result in an encoding exception?

tzachi-dar:
Used HexDump.dumpHexString for printing.
I have checked a print of all bytes (from -128 to 127) and no error was detected.
Actually this also prints to to my pc. I did not see it yesterday, because ascii chars have moved to the next line.

jamorham:
If you use just `HexDump.dumpHexString(data)` then there isn't a risk of `data` being `null` causing an exception during reference of `data.length`

tzachi-dar:
Missed the last comment, fixed it now.

PR #63, opened 2017-01-27
Test fix for Always-on Screen using wakelock PowerManager.SCREEN_BRIGHT_WAKE_LOCK

kskandispersonal:
This is just test code to see if it helps Miguel's issue with G5 Collector failing to successfully scan the transmitter when watch is in ambient mode.  The test code runs only when tryForceScreenOn=true which is currently its default.  If this test is successful the boolean should be moved to a pref.


kskandispersonal:
I change the G5 Collector so that force screen to always-on retrieves the preference, so no more Boolean defaults.  A TODO is to implement this as default for certain watches, but only if this code helps!  Also, a TODO: send bridge_battery and certain other changeable prefs only with new TransmitterData, rather when the listener triggers onchange.

kskandispersonal:
Yes, I kept the `WatchUpdaterService.ACTION_STATUS_COLLECTOR` in SystemStatusFragment (as well as in the MegaStatus class) because of DexShareCollector does have a lastState status string and the System Status page displays the connect / not connected string which should come from the appropriate collector device.

PR #77, opened 2017-03-02
Fixes to sony.com.stepservice; Add Steps, Carbs, Insulin, StdDev to History page.

kskandispersonal:
I still have Log.e statements for the step code in ListenerService.  These can be removed once we confirm that the sony.com.stepservice is no longer an issue - i.e., if it stops, we force a re-reregister.  So far this has been working.


app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java
```diff
@@ -75,14 +77,22 @@ public StatsResult(SharedPreferences settings, long from, long to){
             avg = 0;
         }
 
+        if(getTotalReadings() > 0){
```
jamorham: Would it be better to move this inside `getStdevUnitised()` so that the sql query is only run if that figure is requested? In `getTotal_carbs()` we calculate it the first time if requested and then store the result for efficiency if it is queried more than once.

app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java
```diff
@@ -99,11 +109,11 @@ public int getIn() {
         return in;
     }
 
-    public int getTotal_carbs() {
+    public double getTotal_carbs() {
         if (total_carbs < 0) {
             Cursor cursor = Cache.openDatabase().rawQuery("select sum(carbs) from treatments  where timestamp >= " + from + " AND timestamp <= " + to, null);
             cursor.moveToFirst();
-            total_carbs = cursor.getInt(0);
+            total_carbs = cursor.getDouble(0);
```
jamorham: This is now a double but rounded later when displayed? Can you explain the difference here, I'm not sure if I'm missing something.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CollectionServiceStarter.java
```diff
@@ -2,6 +2,9 @@
 
 import android.app.AlarmManager;
```
jamorham: Imports but no code changes. Was this file added accidentally?

app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java
```diff
@@ -75,14 +77,22 @@ public StatsResult(SharedPreferences settings, long from, long to){
             avg = 0;
         }
 
+        if(getTotalReadings() > 0){
+            cursor= db.rawQuery("select ((count(*)*(sum(calculated_value * calculated_value)) - (sum(calculated_value)*sum(calculated_value)) )/((count(*)-1)*(count(*))) ) from bgreadings  where timestamp >= " + from + " AND timestamp <= " + to + " AND calculated_value > " + DBSearchUtil.CUTOFF + " AND snyced == 0", null);
+            cursor.moveToFirst();
+            stdev = cursor.getDouble(0);
+            stdev = Math.sqrt(stdev);
```
AdrianLxM: StatsResult is used in Home.java and the widget that is recalculated quite often where we decided to not show the stdDev as it needs quite some time to calculate.

I did't see where you use it, but I think it would be better to move this to  a different utility class than the one that is used for Home and the Widget.

app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java
kskandispersonal: Yes, I will do that.

app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java
kskandispersonal: I only added the rounding when displayed because that is how you had it formatted in Home.  I assumed you want it to take up less space on the home screen.  But on History pages, I thought it would be better to display it as a double.

app/src/main/java/com/eveningoutpost/dexdrip/Services/MissedReadingService.java
```diff
@@ -1,63 +1,91 @@
...
 
 import java.util.Calendar;
 import java.util.Date;
 
-public class MissedReadingService extends IntentService {
+import static com.eveningoutpost.dexdrip.Home.startWatchUpdaterService;
+import static com.eveningoutpost.dexdrip.Models.JoH.showNotification;
+
+public class MissedReadingService extends IntentService  implements
+        GoogleApiClient.ConnectionCallbacks,
```
kskandispersonal: Yes, you are right, I will add a request to WatchUpdaterService to implement this action.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/CollectionServiceStarter.java
kskandispersonal: Yes, probably:(  I'll remove it.  I guess this PR is not ready, obviously, to be merged.  In fact, I am still working on it and have found another issue in G5CollectionService re: Scan Constantly while testing.  So I'll pushing some changes to G5CollectionService as well that I hope you will review.

app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java
kskandispersonal: Yes, as @jamorham suggested, I'll move it into its own method so it is only calculated when required.  I added it for the History pages.  And actually, I would like to see it on the Home screen as well, so I've enabled it via a preference.

AdrianLxM: Sorry @kskandispersonal, I'm on a train and the notification about Jon's response arrived late.
Either lazy-loading or just calculating it when the setting is activated sounds good.
Nice doing this in SQL btw. (as SQLite does not offer this) :)

app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java
kskandispersonal: Please let me know if you want me to change this though.

kskandispersonal:
This PR is complete.  I am still TESTING it but the more testing the better.  

I am still unable to isolate why switching between devices can sometimes require 14 minutes before the G5CollectionService successfully gets a reading.  The 133 can occur repeatedly, sometimes it even takes a device restart.  The most common experience though is to get a reading at 14 minutes of missed readings.  Oddly, when the phone triggers the Disable Force on Missed Readings, the phone usually gets the very next reading when disable_wearG5_on_missedreadings_level is set to 0.  So I am thinking the missed readings when randomly switching devices may be related to timing.  The G5CollectionService computes getNextAdvertiseTime() which is used for wake up in keep-alive.  So maybe this value should be shared between devices??

Another related observation, sometime ago, perhaps before we made big changes in the G5CollectionService, the watch itself, almost always got the very next reading after force_wearG5 was enabled; the reverse was not true though.  Phone almost always required the minimum 14 minutes.  I'm not sure what changed in the watch/G5CollectionService to cause it to require the long delay in readings though.

Can you think of any reason why these long delays should be occurring?  I feel it is a bug in the code since at one point the watch did NOT have these delays, and it seems phone does NOT (but not always)have them when switched immediately after a watch reading.  Any correlations?

I also made some fixes to both G5CollectionService code around the Scan Constantly G5 Setting, scanConstantly().  When this setting is enabled, the collector will perform a delayed cycleScan(15000).  I believe this prevents the service from completely shutting down.  The collector wakes back up when a onScanResult arrives, and then continues to process transmitter readings.  So System Status would show that the phone G5CollectionsService was running, but actually, it was the watch collector that was processing the transmitter readings.  My testing of this code shows that the watch G5 collector does indeed get destroyed now, but please have a look to confirm what I did is correct!

app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java
jamorham: I don't think it matters hugely but I didn't expect input data for carbs to be in fractions of a gram.

doc77:
Thank you kskandispersonal for this most valuable addition. I am especially pleased with the extra status line information now in the history pages.

kskandispersonal:
Perhaps the only really required settings that should be displayed are the top 3 which are enable, force and the device running the collector.  The rest could be put onto their own  page?

I'm not sure what you mean by the force wear gets enabled?  How does that happen?  Wear will take over when enable_wearG5 is enabled only when the phone is out-of-range.  But I don't think we ever enable force_wearG5 programmatically.  I could be mistaken though.

jamorham:
I think the sequence goes like this.. 

1) enable `force wear` on phone, use watch as collector
2) watch runs out of battery
3) disable `force wear` on phone
4) some time later charge up watch, watch switches on
5) force wear gets enabled on the phone when the phone syncs its settings which still contain force wear
6) Take phone out and get surprised/confused when no readings occur.. turn `force wear off` on phone.. etc

Basically I think the solution is to raise a notification on the phone when the watches settings sync causes force wear to become enabled on the phone so that there is no surprise to the user.

kskandispersonal:
Oh, I see.  The battery runs completely out on the watch so that when phone programmatically disables its force_wearG5 it does not get sent to the watch for syncing.  Then when watch re-connects onPeerConnect, it sends its settings to the phone for syncing.  Hence, force_wearG5 gets enabled.

Yes, a notification would be good to see.  We already send a toast when it is disabled.  Do you think the toast messages should be notifications instead?

PR #86, opened 2017-03-21
Royce ratio

gitanO0:
Added Option for carbs/insulin ratio for extra line status.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java
@@ -21,7 +21,6 @@ public class StatsResult {
     private final int above;
     private double total_carbs = -1;
     private double total_insulin = -1;
-    private double ratio = -1;
     private double stdev = -1;
     private int total_steps = -1;
     private final double avg;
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java
```diff
@@ -133,6 +134,27 @@ public double getTotal_insulin() {
         return total_insulin;
     }
 
+    public double getRatio() {
```
jamorham: Could this method be simplified to:

```
 public double getRatio() {
        return getTotal_carbs() / getTotal_insulin();
    }
```

That way if someone's status line contains total carbs or total insulin as well as the ratio then it only has to go to the database once for each component as each method already caches the result.

jamorham:
Hi there,

I've added the "Royce Ratio" in this commit: https://github.com/NightscoutFoundation/xDrip/commit/3f43ceaa2af0f01a64b5cf202e8db6e27fcfb77c and credited you for it in the commit log.

This is an optimized version as per my comments on your PR last week. I hadn't seen any response to that so I thought rather than delay further the best thing to do was to get the feature added as it is only about 15 lines.

I've also added `.iml` to the `.gitignore` file to hopefully avoid anyone else having the same trouble you did with the iml files polluting their PR.

jamorham:
A version of this has been integrated. Thanks for the PR.

PR #88, opened 2017-03-21
Cancel static Alarms in DexCollectionService .onDestroy to prevent service from restarting

kskandispersonal:


app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -173,14 +168,35 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
         super.onDestroy();
         Log.d(TAG, "onDestroy entered");
         close();
         foregroundServiceStarter.stop();
-        setRetryTimer();
+        if (shouldServiceRun()) {//Android killed service
+            setRetryTimer();
+        }
+        else {//onDestroy triggered by CollectionServiceStart.stopBtService
+            if (serviceIntent != null) {
```
jamorham: I think we should wrap this with a try/catch block because what if `alarm.cancel()` throws an exception for example. I don't know under what circumstances it could but perhaps with a race condition between the alarm firing and us trying to cancel it.

PR #110, opened 2017-04-26
Added MissedReadingService RESTART COLLECTOR

kskandispersonal:
I disabled Wear's MissedReadingService alarms for missed readings.  I didn't think it would be a priority at this time.
Also, I noticed that restart always missed the next reading on RESTART and often additional readings.  I think this was because scanning was disabled due to the quick restart intervals (normally scheduled wake up times, then the ArmTimer triggered restart).  So I reduced the timing in scanLogic G5-scanlogic downto 1 from 2, and that allowed the scanning to occur on RESTART.  I also changed Home.stale_data_millis down to 5 minutes (for engineering).  This allows the service to restart right away and not miss the opportunity to get the next reading.  I left non-engineering mode at 11 minutes though.  Perhaps we should reduce it as well?  I left ArmTimer at 6 minutes though.  So now RESTART does not appears to consistently read the next reading upon RESTART.  Also, maybe we want to make these changes on the phone after the watch is more thoroughly tested by others.  Otherwise, the code is ready for your review and merge.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/wearintegration/WatchUpdaterService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/wearintegration/WatchUpdaterService.java
@@ -1289,6 +1289,7 @@ public class WatchUpdaterService extends WearableListenerService implements
             dataMap.putString("falling_bg_val", mPrefs.getString("falling_bg_val", "2"));
             dataMap.putBoolean("rising_alert", mPrefs.getBoolean("rising_alert", false));
             dataMap.putString("rising_bg_val", mPrefs.getString("rising_bg_val", "2"));
+            dataMap.putBoolean("aggressive_service_restart", mPrefs.getBoolean("aggressive_service_restart", false));
 
             //Extra Status Line
             dataMap.putBoolean("extra_status_line", mPrefs.getBoolean("extra_status_line", false));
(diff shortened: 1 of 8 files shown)


AdrianLxM:
You are fast. Wow :)

I've noticed that with the G5, there is some jitter in the times the reading come in. Not too much but it may get mapped up to two seconds (modulo 5 minutes) apart from each other.

Here:
https://github.com/NightscoutFoundation/xDrip/pull/110/files#diff-351ff2a4e5b520942cfbaf69b6f9f0b4R137
and here
https://github.com/NightscoutFoundation/xDrip/pull/110/files#diff-4c417e68ae702a6a6c03d76c64418c6cR437

it looks like it would check back at exactly 5 minutes. I wonder if it would be better to add a few grace seconds to not check back just in the moment where the new value is about to come in.

cheers, Adrian

kskandispersonal:
@AdrianLxM Yes, I noticed that, too, a few seconds off sometimes.  The wear code is not using the setAlarm() though.  That code is only for the Missed Alarms settings which I'm not sending over at this point, maybe if someone really wants it we can sync that pref, bg_missed_alerts.  Home.stale_data_millis() is used by Notifications.ArtTimer which is currently triggered at 6 minutes so setting Home.stale_data_millis()  to 5 seems to work.  It didn't work consistently when Home.stale_data_millis() was returning 6 minutes though.

AdrianLxM:
@kskandispersonal that makes sense. :+1: 
The case where it get's triggered earlier due to a new reading coming in definitely is not a missed reading case :)

jamorham:
Is this ready for merging? Does it test okay?

Thanks.

kskandispersonal:
Yes, it's ready for merging.  I've been running with it for 9 days.  Restarts are triggered as expected.

PR #117, opened 2017-05-11
installs differently on Wear 2.0

AdrianLxM:


jamorham:
So does this mean that the current version is actually compatible with Wear 2.0 just that the android wear interface doesn't make it obvious how to install it?

AdrianLxM:
For the watchfaces: Yes.
I can't say anything about wear collection though.

I don't have a wear 2.0 to test, just helped someone.

kskandispersonal:
Thanks, I also do not have a Wear 2.0 watch so cannot test.  I've seen a few queries on FB re: Wear 2.0 but I don't think anyone has gotten it to work, but maybe that is due to the odd way in which it must be installed (wishful thinking!).

PR #121, opened 2017-05-18
Replace old preferences with new IDs and text.  New text requires TRA…

kskandispersonal:
…NSLATION.

The 4 replacement preferences require translation to fi, fr, and nl.


kskandispersonal:
The Sync Watch Treatments is only synced FROM watch to phone; not from Phone to Watch as Watch does not display Treatments currently.  Sync Watch Treatments only support treatments on keypad.  Voice entered treatments remains the same for now (and will not be synced when out-of-range from the phone).  Sync occurs only approx. every 5 minutes typically when a new BG reading occurs.  Note, BG treatments are created automatically as a BloodTest, not as a calibration.  The user can then convert them to a Calibration on the xDrip+ graph by selecting the square box BG and tapping Yes to calibrate.  This is equivalent to the preference treatment_fingerstick_calibration_usage="auto" as processed in Home.processCalibrationNoUI, except there is no AlertDialog displayed and instead of AddCalibration, BloodTest.createFromCal is called.  Looking forward to user feedback to see if this is what the user expects.  This commit is based on the request from gitter user mark71.

kskandispersonal:
Please wait on this PR. Still need to make some code changes.

kskandispersonal:
Sorry, but I want to make a some additional changes to the treatment code so could you please hold off on your review?

kskandispersonal:
This PR is ready for review.  It includes a new, 3-dot menu on the Watchface to activate the Keyboard Input Activity to enter Treatments, and other menu options such as, activate XDrip Prefs, and view data table activity views for Treatments, BloodTest, Calibrations and BgReadings, and a command to Restart Collector on the watch, and refresh the data tables from Phone.

Treatments syncing was added as well esp to receive changes from NS.  Treatment deletion was added to MongoSendTask for the Watch, and I also activated it for the phone Treatments as the code identified that as a TODO.  MongoSendTask may need extra review since I changed the phone logic a bit.

Treatments are synced and displayed by enabling the new preference Show Treatments under Wear Integration on the phone, or in XDrip Prefs on the watch.

Note, this PR does NOT include any support for the Spoken Treatments Activity.  It behavior remains the same.  Thus, to use it, the user must be connected to the phone, otherwise the spoken treatments will not be processed.  If it is a high priority, we can add support, but from my experience, speech recognition is very poor on the SW3.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/UploaderQueue.java
```diff
@@ -146,7 +148,41 @@ public static UploaderQueue newEntry(String action, Model obj) {
         result.bitfield_wanted = DEFAULT_UPLOAD_CIRCUITS
                 | (Home.getPreferencesBooleanDefaultFalse("cloud_storage_mongodb_enable") ? MONGO_DIRECT : 0)
                 | (Home.getPreferencesBooleanDefaultFalse("cloud_storage_api_enable") ? NIGHTSCOUT_RESTAPI : 0)
-                | (Home.getPreferencesBooleanDefaultFalse("cloud_storage_influxdb_enable") ? INFLUXDB_RESTAPI : 0);
+                | (Home.getPreferencesBooleanDefaultFalse("cloud_storage_influxdb_enable") ? INFLUXDB_RESTAPI : 0)
```
jamorham: The way diff shows this really confused me as the changes are minimal in `newEntry()`

app/src/main/res/values-fi/strings-fi.xml
```diff
@@ -204,10 +204,10 @@
 	<string name="wear_integration">Wear Integration</string>
 	<string name="send_data_to_android_wear_watch">Lähetä tietoja Android Wear Watchface.</string>
 	<string name="android_wear_integration">Android Wear Integration</string>
-	<string name="pref_wear_connectG5">Yhdistä Wear G5</string>
-	<string name="pref_summary_wear_connectG5">Yhdistä Wear suoraan G5</string>
-	<string name="pref_use_wear_connectG5">Käytä Wear G5 Collection Service</string>
-	<string name="pref_summary_use_wear_connectG5">Force puhelin käyttää Wear G5 Collection Service</string>
+	<string name="pref_enable_wearG5">Enable Wear Collection Service</string>
```
jamorham: Where a translation hasn't been made yes the strings should be missing from the localized file so that they default to the english strings. So I would delete these additions

app/src/main/res/values-fr/strings-fr.xml
```diff
@@ -204,10 +204,10 @@
 	<string name="wear_integration">Intégration Wear</string>
 	<string name="send_data_to_android_wear_watch">Envoyer des données au Watchface Android Wear.</string>
 	<string name="android_wear_integration">Intégration Android Wear</string>
-	<string name="pref_wear_connectG5">Connectez Porter à G5</string>
-	<string name="pref_summary_wear_connectG5">Connectez Porter directement à G5</string>
-	<string name="pref_use_wear_connectG5">Utilisez Porter G5 Collection Service</string>
-	<string name="pref_summary_use_wear_connectG5">Téléphone Force pour utiliser Wear service G5 Collection</string>
+	<string name="pref_enable_wearG5">Enable Wear Collection Service</string>
```
jamorham: Same for french

app/src/main/res/values-nl/strings-nl.xml
```diff
@@ -204,10 +204,10 @@
 	<string name="wear_integration">Wear Integration</string>
 	<string name="send_data_to_android_wear_watch">Stuur gegevens naar Android Wear wijzerplaat.</string>
 	<string name="android_wear_integration">Android Wear Integration</string>
-	<string name="pref_wear_connectG5">Sluit Slijtage aan G5</string>
-	<string name="pref_summary_wear_connectG5">Maak rechtstreeks verbinding Slijtage aan G5</string>
-	<string name="pref_use_wear_connectG5">Gebruik Draag G5 Haalservice</string>
-	<string name="pref_summary_use_wear_connectG5">Force telefoon om Wear G5 Collection Service te gebruiken</string>
+	<string name="pref_enable_wearG5">Enable Wear Collection Service</string>
```
jamorham: same for dutch

app/src/main/res/values-fi/strings-fi.xml
kskandispersonal: Okay, then we should delete the new ones that I added.  I had also removed the old ones which were no longer being used, which was the original issue Jim E had mentioned - warnings about unused prefs, or something like that.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/UploaderQueue.java
kskandispersonal: That is the only change in NewEntry, but I added a new method just for watch to UploaderQueue.java rather than change NewEntry.

kskandispersonal:
I removed the non-localized entries from the <lang>/strings.xml files.

I also am looking into performance of treatments and would like to make a few changes before you merge this PR...

Enabling treatments on the phone does not seem to have any impact on battery usage. Phone is using about 5-6% / hour whether Show Treatments is enabled or not. On the other have, it does have an impact on the watch. When enabled, watch uses about 7%; not enabled, it uses about 3%. I'm not sure if this is due to the graph update, but probably. I am thinking it might be better to de-couple the "Show Treatments" preference from the Phone's "Show Treatments" pref. They really do not rely on each other and if the user is primarily entering treatments on the watch, then there is no reason to enable "Show Treatments" on the phone, except for integration with Nightscout. Also, the watch user can "sync" the Treatments data from the phone anytime using the new 3-Dot Menu command. This will reduce the overhead on the phone as well as watch.

app/src/main/res/xml/pref_advanced_settings.xml
```diff
@@ -70,6 +70,12 @@
                 android:summary="@string/pref_summary_wear_logs_prefix"
                 android:textIsSelectable="true"
                 android:title="@string/pref_wear_logs_prefix" />
+            <CheckBoxPreference
```
jamorham: After this merge I'm planning to fold Pebble and Wear settings in to their own screens as the preference page is really busy with options now.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BloodTest.java
```diff
@@ -124,6 +124,10 @@ private BloodTestMessage toMessageNative() {
     private static final long CLOSEST_READING_MS = 30000; // 30 seconds
```
jamorham: Can you explain what the changes in BloodTest do? It looks like the treatment de-duplication handling that I introduced for treatments but applied to bloodtests. This probably makes sense, but is this only applied to watch synced items?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoSendTask.java
```diff
@@ -94,10 +101,22 @@ public Void doInBackground(String... urls) {
                                         } else {
                                             Log.wtf(TAG, "Bloodtest with ID: " + up.reference_id + " appears to have been deleted");
                                         }
+                                    } else if (type.equals(Treatments.class.getSimpleName())) {
```
jamorham: This probably makes sense.. its a reminder that the treatment handling in the nighscout uploader was added independently before the uploader-queue framework existed and I haven't got round to moving it in to a proper place.

app/src/main/res/xml/pref_advanced_settings.xml
kskandispersonal: Okay, that makes sense.  Per my previous comment, I'd like to change the "Show Treatments" pref to "Sync Treatments" so that it is clearly distinct from the watch's "Show Treatments" as I don't think they should be unique.  Is it okay if I make changes in this PR still for this?

app/src/main/java/com/eveningoutpost/dexdrip/Models/BloodTest.java
kskandispersonal: Yes, exactly, it is only applied to watch right now - see Home.java line 815.  I didn't change code for the phone.  So, if the user enters a BloodTest on the watch, upon re-connect, app Home will process that data in processAndApproveTreatment() by calling BloodTest.createFromCal, with the new argument, thisuuid.  Watch passes thisuuid to Home in the receivedText and it is parsed in handleWordPair().  It is the same uuid used to create a Treatment, if the user has entered Treatment data as well on the Watch.  Watch also passes a "watchkeypad" to identify the receivedText is coming from the watch, and not phone.  I could have used the new "uuid" field in receivedText but thought it best to have a separate identifier in case in the future uuid is used by phone or other device. 

So the Phone will still  call processCalibrationNoUI() to process new BloodTests which currently displays the UI based on BloodTest prefs, auto, ask, etc.  The watch forgoes this UI and simply calls createFromCal() directly (Home, line 815).  Maybe we still need to display the UI for watch-entered BloodTests?  I thought it simpler to forgo the UI.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/MongoSendTask.java
kskandispersonal: Oh, I was wondering why there was a separate method postTreatments in NS.  I haven't changed the non Wear Queues, so NS is still using the postTreatments() method.

app/src/main/res/xml/pref_advanced_settings.xml
kskandispersonal: Sorry, I DO think they phone and watch treatment prefs should be unique.  I do not want to sync the Treatment preferences between the devices.

jamorham:
Reading through this I can't see any obvious problems and it doesn't look like it will interfere with any of the phone app functionality. With things like sync, the logic flow of what can happen can be complex so if there are any issues I'd expect to see them in wider testing.

Before I merge, can you confirm that you've tested this yourself and are confident that it passes your tests?

Thanks

kskandispersonal:
Yes, I've been testing it off and on for the past week w/ code tweaks here and there.  My only concern is the performance hit on the watch with Show Treatments enabled.  As I mentioned, with Show Treatments enabled, watch uses around 7% battery per hour versus around 3% with it disabled.  Because of that, I'd like to de-couple the Show Treatments preference which is currently synced across phone and watch in this PR.  I think phone should be able to enable "Sync Treatments" without the watch necessarily enabling "Show Treatments".  The phone battery usage doesn't seem to be effected whether treatments are synced or not.  But if you'd rather do the merge now that is okay, too.  I just think it would be easier to change the preference handling before anyone has a chance to do an install.

jamorham:
With show treatments on the watch, are you drawing the simulation graphs? That requires some fairly hefty computations.

kskandispersonal:
Yes, Show Treatments will cause the treatments to display on the graph. I thought it would be useful to see the treatment on the graph after the user enters it using the watch keypad, just to confirm the treatment was saved.  I guess Show Treatments can easily be disabled though via the 3-Dot Menu item XDrip Prefs.  That is what I am doing now as I want to conserve the battery in most instances.

jamorham:
I'm not sure if you misunderstood me. When treatments are entered, curves can display on the graph showing simulations of IoB carbs etc. Did you port that over to wear? The reason I ask is that it is pretty cpu intensive and that could explain the battery drain if so.

PR #135, opened 2017-06-25
DailyIntent Save Database before pruning with new option

TecMunky:
This change adds a new setting under "less common settings" to "Save Database Daily", modifies the DailyIntentService to execute saveSql when enabled, and overloads the saveSql function in DatabaseUtils.java to allow the name of the zip file to be changed to "daily..." vs "export...". Note that the name of the exported database (in the zip file) is not changed, just the name of the zip file. The strings.xml file is also modified with the strings for the new settings option, while the prefs_advanced_settings.xml file is modified to add the new setting.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/DailyIntentService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/DailyIntentService.java
@@ -15,6 +15,7 @@ import com.eveningoutpost.dexdrip.Models.UserError.Log;
 import com.eveningoutpost.dexdrip.UtilityModels.BgSendQueue;
 import com.eveningoutpost.dexdrip.UtilityModels.CalibrationSendQueue;
 import com.eveningoutpost.dexdrip.UtilityModels.UploaderQueue;
+import com.eveningoutpost.dexdrip.utils.DatabaseUtil;
 import com.eveningoutpost.dexdrip.utils.Telemetry;
 import com.eveningoutpost.dexdrip.wearintegration.WatchUpdaterService;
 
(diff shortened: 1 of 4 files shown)


TecMunky:
This change does NOT sync saved databases to online cloud storage.  It is my opinion that forcing users to use a specific cloud service is not appropriate.  There are many Android apps which can sync Android folders to the cloud.  One such App is "FolderSync" which is very flexible, powerful, and can sync to multiple cloud services.  When using such syncing apps, the folder to sync can be located at "/xDrip".  on my phone this folder is located at "/storage/emulated/0/xDrip".

Other options are DriveSync to sync to Google Drive or DropSync to sync to Dropbox.  Note that FolderSync can sync to both these cloud services, and many others.

TecMunky:
Also, while testing this change, I noted some interesting things about the Alarm Manager.  First, The setRepeating function does not guarantee when the alarm interval will actually be executed.  The DailyIntent is scheduled with the constant INTERVAL_DAY (24 hours).  In my testing, the actual interval ranged between 17 and 24 hours.  First, because since API 19 all repeating alarms are inexact.  And second, because inexact alarms occur when the system decides it is an opportune time - the system tries to group alarms to be executed together to save resources.

All this may not be news to everyone.  But the point of this comment is that during testing I tried to change the interval to be more often so I could test more quickly.  I tried INTERVAL_FIFTEEN_MINUTES, INTERVAL_HALF_HOUR, INTERVAL_HOUR, and I even tried a 2 hour interval by creating my own constant to pass to the alarm manager.  But with all these, the actual interval remained between 17 and 24 hours.  

The function I modified for these tests was CollectionServiceStarter.startDailyIntentService -- maybe there was something I missed.  If someone knows why changing the interval setting did not change the actual interval, I would like to know why.

jamorham:
This is a simple easy to review PR, just how I like them.

As with all PR's can you confirm that you have tested this code during normal use and that it works as desired?

Thanks.

TecMunky:
Yes, I can confirm I have tested this completely.  I had been using it for a couple months (without the option switch) before I decided it should be pushed.  I tested the option switch for about a week before I pushed it.  It took a week because the DailyIntentService runs every 17 to 24 hours.

PR #136, opened 2017-06-26
Fixes for watch Treatments; deprecate Manifest BIND_LISTENER.

kskandispersonal:
As well as some minor, performance improvements.  To conserve watch battery, watchface refresh defaults to every minute.


app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -2442,15 +2444,15 @@ public static String extraStatusLine() {
             if (prefs.getBoolean("status_line_carbs", false)) {
```
jamorham: Changes below improve the look of the extra status line or is there another reason?

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -3016,6 +3018,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 break;
             case R.id.action_open_watch_settings:
                 startService(new Intent(this, WatchUpdaterService.class).setAction(WatchUpdaterService.ACTION_OPEN_SETTINGS));
```
jamorham: This looks like a significant bug fix

app/src/main/java/com/eveningoutpost/dexdrip/Models/BloodTest.java
```diff
@@ -124,7 +124,7 @@ private BloodTestMessage toMessageNative() {
     private static final long CLOSEST_READING_MS = 30000; // 30 seconds
 
     public static BloodTest create(long timestamp_ms, double mgdl, String source) {
```
jamorham: I thought this was already fixed, is this github not showing the changes properly?

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
kskandispersonal: Just UI improvements to make it consistent with other status line elements where there is no space after the label.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
```diff
@@ -526,8 +526,9 @@ public static Calibration create(double bg, long timeoffset, Context context, bo
                     BgSendQueue.handleNewBgReading(bgReading, "update", context);
                     // TODO probably should add a more fine grained prefs option in future
                     calculate_w_l_s(prefs.getBoolean("infrequent_calibration", false));
-                    adjustRecentBgReadings(adjustPast ? 30 : 2);
                     CalibrationSendQueue.addToQueue(calibration, context);
+                    BgReading.pushBgReadingSyncToWatch(bgReading, false);
```
jamorham: Any reason for the order change of the adjusting recent readings?

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
kskandispersonal: Yes, I'm afraid I never tested this.  I think this was existing code when I joined.

app/src/main/java/com/eveningoutpost/dexdrip/Tables/CalibrationDataTable.java
```diff
@@ -118,7 +118,7 @@ public boolean onLongClick(View v) {
                         public void onClick(DialogInterface dialog, int which) {
                             switch (which){
                                 case DialogInterface.BUTTON_POSITIVE:
-                                    calibration.invalidate();
```
jamorham: this maybe has to be approached with caution, but tracing it through it should be an improvement. Possibly as it is from interactive it should have the from_interactive set to true but we can leave that for now.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
AdrianLxM: It never had mattered as there where only the two options in the menu and this was the second one.
Good example why to write it anyway - to not fall into the trap later :) 
Nice catch

app/src/main/java/com/eveningoutpost/dexdrip/Models/BloodTest.java
kskandispersonal: Probably you made the fix in master, and I didn't pull it down yet, but made it my local repo.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
kskandispersonal: Yes, the new calibration should be created on the watch before the BgReading gets associated to it.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
kskandispersonal: Oh, @AdrianLxM, you are right, I should have noticed it when I added additional cases:(

AdrianLxM: Sorry if that came out wrong. That was not what I've meant.

jamorham:
Okay, is this still ready for merge with the latest changes?

kskandispersonal:
Yes.  I made a few last minute changes realizing Treatments.systimestamp wasn't being set which effected population of the treatments new data table view.

PR #139, opened 2017-06-30
noise ultrasensitive mode

AdrianLxM:
Provide engineering mode feature to allow noise compensation to occur even with very low noise levels. (Ultra-sensitive threshold)

The purpose of this is to give the option to smooth glucose data effectively for occasional sensors which show noise during initial bedding in stage.


jamorham:
What is the status of this? Is it ready for merge? tested?

Thanks. Jon

AdrianLxM:
Yes, I tested it in all setting combinations and with two sensors in parallel (one rather good but fresh and one wonky one) think it's ready to merge.

PR #141, opened 2017-07-01
close confirmation on wear when confirmed on phone

AdrianLxM:
On the phone insulin and carbs can be confirmed individually by clicking on them instead of clicking on the green plus. If all are confirmed this way, the dialog closes on the phone but did not on the watch.

By sending this message to the watch it will close.

@kskandispersonal, could you please take a look at this. Thanks.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -989,7 +989,7 @@ public class Home extends ActivityWithMenu {
                 (btnInsulinDose.getVisibility() == View.INVISIBLE)) {
             hideAllTreatmentButtons(); // we clear values here also
             //send toast to wear - closes the confirmation activity on the watch
-            WatchUpdaterService.sendWearLocalToast("Treatment processed", Toast.LENGTH_LONG);
+            WatchUpdaterService.sendWearToast("Treatment processed", Toast.LENGTH_LONG);
             return true;
         } else {
             return false;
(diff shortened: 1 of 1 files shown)


kskandispersonal:
Thanks, @AdrianLxM.  Please replace the method sendWearLocalToast with sendWearToast:

WatchUpdaterService.sendWearToast("Treatment processed", Toast.LENGTH_LONG);

AdrianLxM:
Thanks @kskandispersonal.

I used the same as when all treatments are confirmed together. https://github.com/NightscoutFoundation/xDrip/blob/649a6e5f72ebcb554823cf647da81e5333dc771b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java#L796

So while I'm at it, should I replace it there too?

kskandispersonal:
That code snippet is used only if it was sent from Wear "if (watchkeypad)" , so only needs to display a local toast message.  The Approval Activity (Simulation) will already have been closed by tapping the + on Wear, which then processes the approval locally now, and then at the next connection will send the treatment data.

Line 809, is the method that should be used for phone treatments.

jamorham:
Should this PR be merged?

AdrianLxM:
I did do the changes that that @kskandispersonal suggested.
@kskandispersonal what do you say?

kskandispersonal:
Yes.  I've been using this code for a while now but it may conflict with the changes you made, @jamorham.

jamorham:
There don't appear to be any merge conflicts, but the code will fire even if wear is not in use, are we relying on in ` if ((googleApiClient != null) && (googleApiClient.isConnected())` in `sendWearToast` as the qualifier for detecting if wear is enabled? 

If the user enables and then disables wear in the preferences will the code still fire? In that case do we care or should there be some additional checks that wear is actually enabled before sending the toast?

jamorham:
Ok, so this should be merged? You have continued to test it with good results?

kskandispersonal:
Yes, I haven't seen any issues with it and I've been using Treatments with good results..

PR #146, opened 2017-07-11
Fix for Watch Uploader battery usage when Force Wear is disabled ; ad…

kskandispersonal:
Fix for Watch Uploader battery usage when Force Wear is disabled ; add refresh_on_change pref to force face refresh on data change rather than wait for to refresh each minute.

Watch Uploader battery bug was identified by @gooseodyssey.

Some deprecated code remains commented out 'til broader testing confirms fix is acceptable.


jamorham:
What is the current state of this PR? Should I be checking it for merge?

PR #154, opened 2017-08-08
Enhance the csv-Export for SiDiary

MasterPlexus:
Including also the treatment Data and BG-measurements.
Did not use the Table BloodTests at the moment, as for me is not clear what, and when there will be data stored. At the moment I see only data int he Calibration-Table, so I used that.

This Implementation was testet several times and work fine as a first shot. But expecting may to implement also some additional settings in future (for example, Export only data since the last export and chose what should be exported.


MasterPlexus:
Hi @jamorham, did not want to bother you, just want to know if the pull will be ok for you, or you may expect it in a other way? Would be nice to include it, because then I can switch back to nightlys ;)
Have a great Day ,

jamorham:
@AdrianLxM are the changes you requested now in this PR?

PR #187, opened 2017-09-28
Update Phones_For_G5.md

drnoname82:


jamorham:
Can you provide more information? Which collector was used, is this stock rom etc? What sort of capture rate was achieved? Thanks

drnoname82:
This is using the old collector (pre-OB1, "Authenticate G5 before each read" activated) on a Lenovo Moto G4 with Android 7.0 stock rom. Capture rates are around 93 %.

PR #188, opened 2017-09-29
Fix to display sensor and treatment data on phone on each sync.

kskandispersonal:
This issue was observed by wannago.  Moved Steps and Treatments data sync BEFORE BGs sync so the phone screen refresh, triggered by BG refresh, will show all new data.


jamorham:
Should I be doing anything with this PR? Is it ready for merge? Thanks.

kskandispersonal:
Yes, tested and ready for merge.  Thanks!

PR #190, opened 2017-10-02
Change visibility of all notifications to public

p5nbTgip0r:
This allows the notification content for xDrip's notifications to appear on the lock screen when 'Hide sensitive notification content' is chosen for the lock screen preferences of notifications.


jamorham:
Okay, what about users who *want* the information to remain hidden? 

I think this should be controlled by a preferences switch in the same way that `HIGH_PRIORITY` is currently handled and default to the current behavior as it is a privacy related issue. 

Search for references to `high_priority_notifications` to see how this is currently achieved in relation to that.

p5nbTgip0r:
> Okay, what about users who want the information to remain hidden
That's a scenario I didn't think about. In the latest commit I made an option to toggle public/private notifications.

Issue #126, opened 2017-05-29
BG broadcast issue with NFC and libre


gregorybel:
When I scan my libre over NFC with a few hours between each scan, then xdrip-plus broadcasts several times the same last value instead of broadcasting all new values since last scan.

See screenshot:

![2017-05-15 19 24 07](https://cloud.githubusercontent.com/assets/24234385/26554787/489b4828-4492-11e7-9f93-584e65a493b9.png)

jamorham:
Thanks for reporting this. I'm trying to determine where the problem is originating from.

As a test can you do this:

Go to `Settings` -> `Inter-app Settings` and un-check `Send Display Glucose` then repeat your test and let me know if you get the same problem?

gregorybel:
Hi John, 

I unchecked this and no more issue! 

Hope this help to fix issue. 
Thanks

jamorham:
Okay thanks, this means that the feature called `Display Glucose` which includes noise compensation and calibration plugins will need to be updated to support data coming in out of sequence which I think is what is causing your issue here.

gregorybel:
Jon, can you tell me where in the code you think the issue is then I will try to fix it and if I manage then will do a PR.

AdrianLxM:
@gregorybel, it must be in this call: https://github.com/NightscoutFoundation/xDrip/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgSendQueue.java#L238

gregorybel:
Thank you @AdrianLxM!

PR #196, opened 2017-10-06
Send Display Glucose (smoothed) via REST to NS

AdrianLxM:


jamorham:
Okay this PR looks pretty good. Can I confirm a few points:

1) Have you tested this to confirm it works?
2) I assume the switch to control the behavior defaults to off?
3) I assume those strings are already defined and make sense in this context?

AdrianLxM:
1. yes
2. yes
3. If you are referring to the "keys" in the json uploaded to NS.

jamorham:
I meant the preference description strings.

jamorham:
I'm concerned that this doesn't have time related safety checks like there is with `InjectNoise` for example:

```
 final BgReading bgReading = this;
        if (JoH.msSince(bgReading.timestamp) > Constants.MINUTE_IN_MS * 20) {
```

`getDisplayGlucose()` is going to work on the 2 most recent readings which might not be what is actually in `BgReading` `this` which would result in problematic behavior if that happened.

The display glucose also contains a noise value so we could also later merge the two methods once we are satisfied each is working correctly.

AdrianLxM:
Ad Question 3): Yes, those are the Strings used in the settings for sending Display Glucose in the inter-app settings.

Ad safety checks: I used the same logic as in BgSendQueue->local broadcast, where this isn't guarded either except for a simple check for `null` being returend from `getDisplayGlucose()`. I thought you did have a solution for out of sync/order data already in there?

jamorham:
Yes, I think it could run in to the same type of problem as cited in issue https://github.com/NightscoutFoundation/xDrip/issues/126 (which is going to require some restructuring to fix if it is appropriate)

As this new smoothing is off by default and a vast majority of people the data is going to be processed as it arrives in the correct sequence and with each reading spread out then I think this should be fairly safe to merge this and we can see what happens with it operationally.

PR #200, opened 2017-10-11
blukon Out of process algorithm 

tzachi-dar:
This is the first step in allowing out of process algorithms for blokon on xDrip.
A new ui checkbox was added and when set, the complete sensor data will be scanned by the blukon.
xDrip will gather this data and store it in one big buffer.
Currently this buffer is only written as a file.

All comments are welcomed.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NSEmulatorReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NSEmulatorReceiver.java
@@ -68,13 +68,14 @@ public class NSEmulatorReceiver extends BroadcastReceiver {
                             case Intents.XDRIP_PLUS_NS_EMULATOR:
 
                                 // in future this could have its own data source perhaps instead of follower
-                                if (!Home.get_follower() && DexCollectionType.getDexCollectionType() != DexCollectionType.NSEmulator) {
+                                if (!Home.get_follower() && DexCollectionType.getDexCollectionType() != DexCollectionType.NSEmulator && 
+                                    !Home.getPreferencesBooleanDefaultFalse("external_blukon_algorithm")) { //???DexCollectionType
                                     Log.e(TAG, "Received NSEmulator data but we are not a follower or emulator receiver");
                                     return;
                                 }
 
                                 if (!Home.get_follower()) {
-                                    // must be NSEmulator here
+                                    // must be NSEmulator here ???? Not true anymore.
                                     if (!Sensor.isActive()) {
                                         // warn about problems running without a sensor record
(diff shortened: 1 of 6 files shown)


jamorham:
@keencave @gregorybel I am not familiar with with the data flow in blukon, will this PR preserve the existing behavior when it is disabled (the default) ?

@tzachi-dar Is `JoH.bytesToHex()` or `HexDump.dumpHexString()` insufficient for the hex dumping?

@tzachi-dar I assume the plan will be to issue a broadcast intent rather than write to a file at some point in the future?

@tzachi-dar Have you tested this with the hardware device to determine if this sequence of commands gets the data required?

dabear:
@jamorham @tzachi-dar  this code also introduces a new way of communicating with the blucon (multiblock reads, these command has the 010d0f prefix)

You (or @keencave) should really test this code with an almost depleted battery. Error handling with multiblock reads is currently untested; It is currently unknown what kinds of Nacks we will receive when sending a multiblock request to the blucon when battery is low

keencave:
I can do the test with low battery but I need my laboratory current source for this. It is availbale to me on the weekend.

gregorybel:
One big issue I see is that code starts to be no longer readible. We need a big refactoring. Otherwise will comment code directly.

tzachi-dar:
I have added some more code to this PR
It can now send data as an intent to libreXposed and receive  the calculated data and show it. :-) 
Currently data is received as a nightscout emulator and I have made the minimal changes that are needed in order to receive data. I'm not sure that this is the correct way to go since I'm still missing the following functionality:
1) Allow the code to pass through the calibration mechanism.
2) Use the libre algorithm in order to close gaps.
So, we need to decide if I should change the ns- broadcast receiver and add the needed functionality to it, or create a new broadcast receiver. 
As for the previous remarks:
1) I have tested it with real hw, although there is obviously much more to test.
2) As for low battery testing, I guess that I'll have to wait for my battery to finish to see what really happen, but there are a few things that should help me fill good about my code:
a) I verify that all buffers are received from the blucon.
b) A nack anywhere will cause the intent not to be sent.
c) I hope that libre code check the checksum of the packet and if so, in any case of error they will detect it and return an error (I'll verify it). Also I understood that there is code that checks the checksum so I would like to use it and not sent intents with wrong data.
3) I have changed the code to use HexDump.dumpHexString

tzachi-dar:
One last comment: Writing the data to files help me collect data. I'll remove it before the code is merged.

keencave:
Just tested Tzachis code with low volatge (2.62. V). The code runs stable, I receive either having a battery low indication all historic data. Tzachis code can be swithced on/off with the new menue entry. I think it can be merged.

PR #204, opened 2017-10-16
Implement notification channels for Android O

jwoglom:
This change:
* creates a NotificationChannels class (that extends ContextWrapper) which creates notification channels for Android O+. Per Google documentation and examples, android ignores creation requests for channels if they already exist, so always attempting to create them (by call in the Home activity) is the right way to go.
* changes notification methods in the JoH helper class to use NotificationCompat.Builder, with an argument for the notification channel ID in these methods.
* updates most, but not all, notification-related calls that use JoH.showNotification to provide a NotificationChannels.XX_CHANNEL constant (created in NotificationChannels.java) as the channelId. There were a couple oddly-specific error cases that generate notifications which I did not create a notification channel for. 
* updates all other uses of Notification.Builder to use NotificationCompat.Builder, except for in GcmListenerSvc. It looked to me like updating this class to be channel-aware would require passing channelId's in bundled GCM objects which are picked up here, and I don't actually know how this class is used so I just left it alone.

The user-visible changes as a result of adding notification channels is that sound, vibration, do not disturb, and importance preferences can be set for different categories of notifications. Additionally, until at least one notification channel is created for an application running under Android O, vibration and importance preferences on all notifications for that app cannot be changed, which was functionality supported for all notifications by a single app in Android N.

This change has been tested on a Nexus 5X running Oreo, as well as on an emulated Nougat device to ensure notifications are still generated and the app doesn't crash when running on an earlier SDK version than 26, which introduces notification channels.

xdrip-channels-1
xdrip-channels-2
xdrip-channels-3
xdrip-channels-4

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -81,6 +81,7 @@ import com.eveningoutpost.dexdrip.UtilityModels.Experience;
 import com.eveningoutpost.dexdrip.UtilityModels.Intents;
 import com.eveningoutpost.dexdrip.UtilityModels.JamorhamShowcaseDrawer;
 import com.eveningoutpost.dexdrip.UtilityModels.NightscoutUploader;
+import com.eveningoutpost.dexdrip.UtilityModels.NotificationChannels;
 import com.eveningoutpost.dexdrip.UtilityModels.Notifications;
 import com.eveningoutpost.dexdrip.UtilityModels.PersistentStore;
 import com.eveningoutpost.dexdrip.UtilityModels.PrefsViewImpl;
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -255,10 +256,13 @@
 
     private static String statusIOB = "";
     private static String statusBWP = "";
+    private NotificationChannels notifChannels;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         mActivity = this;
+        // Builds notification channels if supported
```
jamorham: What happens if a notification is raised before `NotificationChannels` is instantiated? It is quite possible for `Home.onCreate()` to never be called. For example if the user reboots their phone, doesn't open the app and then receives a notification. If this is always needed then it should be appended to the initialization in the `xdrip` Application class I think.

jwoglom: Was wondering about this case; I was assuming that Home would be loaded early enough that this wouldn't matter, but that makes total sense. You're talking about onCreate() in xdrip.java, correct?

PR #213, opened 2017-11-09
Masterplexus blueReader adding better Battery (percent and rest days) and restart

MasterPlexus:
Adding Battery Stuff for blueReader:
- percentage of Battery
- Rest days of Battery
- hide Low/-Bat-message in Home with nrf-flag, as it was misinterpretated for blueReader (may rework later)

Further:
- add log output of errors specific for bluereader in DexCollector and TransmitterData
- add Restart/reset of blueReader when in ugly state (clearly to check in addition, may need also a hibernate-send, and Collectionservice needs then also may a restart)

Surely the coding is not as its best, so every comment is welcome ;)

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/SystemStatusFragment.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/SystemStatusFragment.java
@@ -263,10 +263,10 @@ public class SystemStatusFragment extends Fragment {
         } else {
             transmitter_status_view.setText("" + td.sensor_battery_level);
             GcmActivity.requestSensorBatteryUpdate(); // always ask
-            if (prefs.getString("btDevice","").equals("blueReader")) {
-                if (td.sensor_battery_level <= 3200) {
+            if (getBestLimitterHardwareName().equals("BlueReader")) {
+                if (td.sensor_battery_level <= 3300) {
                     transmitter_status_view.append(" - very low");
-                } else if (td.sensor_battery_level <= 3400) {
+                } else if (td.sensor_battery_level <= 3650) {
                     transmitter_status_view.append(" - low");
                 } else {
                     transmitter_status_view.append(" - ok");
--- a/app/src/main/java/com/eveningoutpost/dexdrip/SystemStatusFragment.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/SystemStatusFragment.java
@@ -430,6 +430,9 @@ public class SystemStatusFragment extends Fragment {
     private void setNotes() {
         try {
 
+            if(!prefs.getString("blueReader_Version","empty").startsWith("empty")) {
+                notes.append("\n- blueReader Firmware: " + prefs.getString("blueReader_Version","empty"));
+            }
             if ((mBluetoothManager == null) || ((android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.JELLY_BEAN_MR2) && (mBluetoothManager.getAdapter() == null))) {
                 notes.append("\n- This device does not seem to support bluetooth");
             } else {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/SystemStatusFragment.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/SystemStatusFragment.java
@@ -57,6 +57,7 @@ import java.util.List;
 import java.util.Set;
 
 import static com.eveningoutpost.dexdrip.Home.startWatchUpdaterService;
+import static com.eveningoutpost.dexdrip.Services.DexCollectionService.getBestLimitterHardwareName;
 import static com.eveningoutpost.dexdrip.utils.DexCollectionType.DexcomG5;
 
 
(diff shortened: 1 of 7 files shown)


jamorham:
@SandraK82 Can you comment on the bluereader protocol elements that this PR uses? I am not aware of them. Are they present on all firmwares?

MasterPlexus:
@jamorham , Just to precise: 
I did not add additional protocol elements except the restart-command, which is a basis command in the Firmware I guess (but clearly @SandraK82  should say something about it, may also if this is a good idea to implement it in that way).
The rest is more a recalculation of surely available values from the bluereader which are given in Limitter mode.
As far as I know Sandra may also works on better integration for the blueReader, but as I have my solution some weeks in place I think it was may a good point also for other users. But no problem if this will be rejected if Sandras solution will be better implemented.

jamorham:
I appreciate the PR, I just don't know enough about bluereader to be able to completely evaluate what these changes will do and I have to balance new features against whether existing users will experience problems due to compatibility etc.

SandraK82:
I will review the PR and check for problems regarding blueReader. But I think it will be good as masterplexus got the command documentation from me. 

Von meinem iPhone gesendet

MasterPlexus:
I've changed the handlin of messages from bluereader to a better place, and use now hibernate to bring the reader back to work if he is in a ugly state... @SandraK82 , now you can review the code, for me it works now fine...

jamorham:
Is this ready to be merged?

MasterPlexus:
On my side it works properly, and for a other tester also (which used my comits to build his own version). But I have no problem to wait for the review of @SandraK82 .

MasterPlexus:
Adding a safety Power off for the blueReader, as the Firmware seems to come to an problematic state if the battery was to low and try to switch it of itself. Changed the estimation of the ugly state, as TRANSMISSION_FAILED is not correct in addition.

app/src/main/java/com/eveningoutpost/dexdrip/SystemStatusFragment.java
```diff
@@ -262,13 +263,27 @@ private void setTransmitterStatus() {
         } else {
             transmitter_status_view.setText("" + td.sensor_battery_level);
             GcmActivity.requestSensorBatteryUpdate(); // always ask
-            if (td.sensor_battery_level <= Dex_Constants.TRANSMITTER_BATTERY_EMPTY) {
-                transmitter_status_view.append(" - very low");
-            } else if (td.sensor_battery_level <= Dex_Constants.TRANSMITTER_BATTERY_LOW) {
-                transmitter_status_view.append(" - low");
-                transmitter_status_view.append("\n(experimental interpretation)");
+            if (getBestLimitterHardwareName().equals("BlueReader")) {
```
jamorham: The classic system status tab is a legacy component which eventually will get removed once its features have been absorbed elsewhere. We shouldn't be adding new code here unless it is essential. I would suggest a better place for statistics and information display about bluereader is via the megastatus section in `DexCollectionService` somewhere like here: https://github.com/NightscoutFoundation/xDrip/blob/ab2a14277a98d171e0731db7e77901e02adb2a32/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java#L1275

Also it is usually only dexcom which has a transmitter battery as bluereader is a bridge, we would normally store that as a bridge battery. I haven't followed completely what you're doing here regarding the battery so perhaps I don't understand the purpose of this code. 

Data which you want to store persistently but which isn't actually a preference item. Eg we don't need to back it up from a configuration point of view and represents state etc should typically be placed in the `PersistentStore` There are of course places where xDrip historically uses preferences for this data which you may have copied the style from, but I would prefer if all new code uses the persistent store.

jamorham:
Thanks for the additional code, but unfortunately this PR is now suffering from what I consider a github bug where it isn't easy for me to see the change set because it is starting to show other merges to master as changes from your PR when they are not. 

What I would suggest to do is either:

1) Create a branch based on the previous set of changes before the most recent commits and resubmit the PR from that branch so that we can get those changes merged asap. I assume @SandraK82 expects them to work as you have the documentation required and have tested them and that they will work with bluereader firmwares out in the wild?

2) Continue with the development and resubmit the PR incorporating the changes I suggested and we can do the review process from there whenever you are ready.

I prefer option 1 as it is a lot easier for me to merge small working tested change sets in and the problem is that if it takes too long in the cycle to get a PR merged then there may be other code changes in xDrip which involve refactoring and that's going to involve more work for everyone and potentially make it harder to get features incorporated sucessfully.

I'd like to improve the bluereader support but I also have to be careful not to break things as well.

MasterPlexus:
Thanks for your comments, which are completely right. The changes I did was due to what I've seen in the code, so may I misinterpreted some points. I'll take a deeper look into it this evening, and come back to it. At the moment for me it would be good to leave this pull open, as I have a place to discuss may if I have further questions. At the end I think it would be the best to make a new pull request to see the real changes.

jamorham:
It is easily done, there is very little documentation and lots of legacy portions of code still in service. I really look forward to getting this PR integrated and I hope we can do so soon because I don't want the code base to move on and make merging impossible and create more work for you.

MasterPlexus:
I've put the most stuff now in an own class for the blueReader as it was done also for the blueCon. At the moment some small things needs tobe done, and I have to test it. may I'll commit a new PullRequest at Thursday or Friday. Unfortunately the next two days I've no big availabilitys, but we will see.

MasterPlexus:
So I'm near to commit the new changes and I'm on last running tests. I'll close this pull, and made a new one soon.

PR #217, opened 2017-11-19
Send data as intents to LibreXposed.

tzachi-dar:

This was tested in my private branch, but somehow slipped from the final merge.


jamorham:
Does this only send the intent when `external_blukon_algorithm` is enabled?

tzachi-dar:
Yes.

PR #218, opened 2017-11-21
Add uuid to BgReading being created from json.

tzachi-dar:
This should allow blukon to upload data to NS.

This change will influence all callers of this function.  If you fill it is unsafe, I'll protect it with a flag and so it will only affect the NSClientReceiver branch.


tzachi-dar:
Just to make my fix clearer. Without this UUID, the packet will be thrown in the function newEntry
With the error: "reference_uuid was null so refusing to create new entry"

tzachi-dar:
By the way, might be safer to pass the uuid inside the json, less side affect on other places.

tzachi-dar:
Closing it as https://github.com/NightscoutFoundation/xDrip/pull/219 is much safer.

PR #219, opened 2017-11-21
Add uuid to BgReading being created from json in NSEmulatorReceiver.

tzachi-dar:
This should allow blukon to upload data to NS.

This is a much safer version of the fix.


jamorham:
Has this change been tested and confirmed to work or has it been produced by following the code path to see what the problem with the nightscout upload queue is?

tzachi-dar:
My blukon is out of battery.
So, it was tested by sending an intent manually to the OOPAlgorithm.
The OOPAlgorithm sent an intent to xDrip, and xDrip showed the data on the screen and also sent it to NS.

Since xDrip already sends data to the OOPAlgorithm we should be fine.

PR #223, opened 2017-11-30
Remove the need to enter calibrations for the OOP algorithm.

tzachi-dar:

Background:
Currently it is not possible to pass the oop algorithm through the calibration mechanism.
In the future, I want to allow both the option to calibrate and the option not to calibrate.
In the near future I want to allow people to scan directly using the phone, and also use the OOP algorithm. 
(and again that might be with calibrations and without calibrations.)

In this PR I remove the need for calibration. This is based on running without calibrations such as the ns emulator (and a follower, that does have calibrations now).
This is one way to solve the problem. Another way is to create a "fake" calibration and let the code work with it. This option is closer to other options that xDrip has.

What do people think?

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -2186,7 +2186,7 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
         }
         if (isWifiWixel || isWifiBluetoothWixel || collector.equals(DexCollectionType.Mock)) {
             updateCurrentBgInfoForWifiWixel(notificationText);
-        } else if (is_follower || collector.equals(DexCollectionType.NSEmulator) ||DexCollectionType.isLibreOOPAlgorithm()) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -2186,7 +2186,7 @@ private void updateCurrentBgInfo(String source) {
         }
         if (isWifiWixel || isWifiBluetoothWixel || collector.equals(DexCollectionType.Mock)) {
             updateCurrentBgInfoForWifiWixel(notificationText);
-        } else if (is_follower || collector.equals(DexCollectionType.NSEmulator)) {
+        } else if (is_follower || collector.equals(DexCollectionType.NSEmulator) ||DexCollectionType.isLibreOOPAlgorithm()) {
```
jamorham: Can you instead pass `collector` as a parameter to `isLibreOOPAlgorithm` to avoid repeated look ups of the current collection type.

tzachi-dar: Passed it in both locations.

PR #224, opened 2017-12-03
If all alerts are disabled, do not alert for low bridge battery.

tzachi-dar:
Checked by running with low battery when all alerts are disabled, and only when low alerts are disabled.


tolot27:
Is there any issue associated with this PR?

Navid200:
@tzachi-dar Hi,
What are your thoughts about this pull request?
May I ask why it has failed one of the two tests?
Do you feel this pull request should be merged?

Navid200:
@tzachi-dar Would you please explain why this pull request is till open?
Do you want it to be merged?
Would you please explain why a user who has disabled all alerts cannot also disable the low bridge battery alert?  In other words, why do we need this capability?

Alternatively, can we close this pull request?

Thanks,

Navid

tolot27:
Currently, battery low is alerted, even if all alerts are disabled. This PR fix this inconsistency.

tolot27 (approved):
This PR improves the notification consistency by silencing the low battery alert if all alerts a disabled.

Navid200:
Thanks for all the work.

Is it ever possible that the low bridge battery could be considered an emergency?  In which case, we would be sorry that we added this capability.

tolot27:
> Thanks for all the work.
> 
> Is it ever possible that the low bridge battery could be considered an emergency? In which case, we would be sorry that we added this capability.

If a user decides to disable all alerts, all alerts should be disabled. Otherwise, a user should just disable certain alerts. That's a matter of consistency.

Navid200:
I hear you.
But, we have never been consistent.  We have a predicted low alert.  But, we don't have a predicted high alert.  The reason is that we are worried that the user may do something that could harm them.

Is there an answer for "whether low bridge battery is an emergency"?

tzachi-dar:
I think that low bridge battery can not be considered as an emergency, as there are still a few hours of work for that device.
Even when BG goes below 55 this does not override the "all alerts disabled". 
Furthermore, we have an alert for when the battery is realy dead (this is the no readings alert).

Please note that snoozing for this alert is very limited, and it is quite possible that it would cause a ring every 30 minutes, although there is no immediate problem.

tolot27:
> But, we have not always been consistent.

But we can start with.

> We have a predicted low alert. But, we don't have a predicted high alert.

Having an incomplete feature set is not the same of as having a "disable all alerts" feature, which does not disable all alerts.

It is quite annoying if you disable all alerts but during a long and important meeting, the battery low alert reoccurs. As @tzachi-dar pointed out, there is still a lot of remaining battery. And the notification itself is not deactivated, just the alarm.

jamorham (approved):
I think this should be okay

PR #233, opened 2017-12-06
Move the scan code to a function of it's own.

tzachi-dar:
This should make it simpler to have another function to read the tag.

This should make it simpler to have another function to read the tag.

Signed-off-by: Tzachi Dar tzachi.dar@gmail.com
This is the first step in creating another function that will read the libre code for the oop algorithm.
It should not change anything, and I did my best to make that very visible. (obviously I can be wrong).

This was tested with a dead sensor, and I'll test that with a real sensor once my libre sensor will get older.
JamOrHam, please read it and see if there is anything that needs to be changed here, or if you think that anything with program flow has changed with this PR.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -914,11 +914,7 @@ public class BgReading extends Model implements ShareUploadableBg {
     }
 
     public static BgReading readingNearTimeStamp(double startTime) {
-        double margin = (4 * 60*1000);
-        if(DexCollectionType.hasLibre()) {
-        	// Libre has data every 1 minute (and we force this in the scan functions).
-        	margin = 59*1000;
-        }
+        final double margin = (4 * 60*1000);
         final DecimalFormat df = new DecimalFormat("#");
         df.setMaximumFractionDigits(1);
         return new Select()
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -914,7 +914,11 @@ public class BgReading extends Model implements ShareUploadableBg {
     }
 
     public static BgReading readingNearTimeStamp(double startTime) {
-        final double margin = (4 * 60*1000);
+        double margin = (4 * 60*1000);
+        if(DexCollectionType.hasLibre()) {
+        	// Libre has data every 1 minute (and we force this in the scan functions).
+        	margin = 59*1000;
+        }
         final DecimalFormat df = new DecimalFormat("#");
         df.setMaximumFractionDigits(1);
         return new Select()
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
@@ -24,6 +24,7 @@ import com.eveningoutpost.dexdrip.ImportedLibraries.usbserial.util.HexDump;
 import com.eveningoutpost.dexdrip.Models.GlucoseData;
 import com.eveningoutpost.dexdrip.Models.JoH;
 import com.eveningoutpost.dexdrip.Models.LibreBlock;
+import com.eveningoutpost.dexdrip.Models.LibreOOPAlgorithm;
 import com.eveningoutpost.dexdrip.Models.PredictionData;
 import com.eveningoutpost.dexdrip.Models.ReadingData;
 import com.eveningoutpost.dexdrip.UtilityModels.PersistentStore;
@@ -165,6 +166,8 @@ public class NFCReaderX {
     }
 
     private static synchronized void doTheScan(final Activity context, Tag tag, boolean showui) {
+    	if (d)
+            Log.d(TAG, "doTheScan called");
         synchronized (tag_lock) {
             if (!tag_discovered) {
                 if (!useNFC()) return;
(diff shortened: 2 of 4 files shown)


tzachi-dar:
Added code that does the real scanning, and sending for the OOP algorithm.

Tested against dead sensors, with old versions I know that worked. Will be tested tomorrow against good sensor.

tzachi-dar:
Tested and works well with oop algorithm and without it.
(2 small fixes also applied).

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -277,125 +408,17 @@ protected Tag doInBackground(Tag... params) {
...
-                                    for (int j = 0; j < 3; j++) {
-                                        System.arraycopy(replyBlock, 2 + (j * 9), data, i * 8 + (j * 8), 8);
-                                    }
-                                } else {
-                                    System.arraycopy(replyBlock, 1, data, i * 8, replyBlock.length - 1);
-                                }
-                            }
+                        boolean readPatchSucceeded;
+                        if (Home.getPreferencesBooleanDefaultFalse("external_blukon_algorithm")) {
+                        	readPatchSucceeded = LibreOOPAlgorithm.readOOPData(nfcvTag);
```
jamorham: Instead why not change the reading block count from 40 to 42 if `external_blukon_algorithm` is set and use the local `readPatch` and then call `SendData` directly to eliminate duplicate code in the OOP class? Possibly do an array copy to get the buffer the exact size the interface is expecting. Maybe this is already what you are planning to do..

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -914,7 +914,11 @@ public static boolean isDataStale() {
     }
 
     public static BgReading readingNearTimeStamp(double startTime) {
-        final double margin = (4 * 60*1000);
+        double margin = (4 * 60*1000);
+        if(DexCollectionType.hasLibre()) {
```
jamorham: `hasLibre()` is too broad a condition at this point. There are many places in the code base where features are expecting the minimum period between readings to be 300 seconds. For example a limitter device might emit the same reading twice within a 5 minute window due to a re-transmission.  Like the rest of these changes, I think this should be tied to: `external_blukon_algorithm` preference item.  Possibly even then this could introduce unexpected results, for example with low prediction etc.

app/src/main/java/com/eveningoutpost/dexdrip/NSEmulatorReceiver.java
```diff
@@ -110,6 +110,7 @@ public void run() {
                                                         // sanity checking???
                                                         // fake up some extra data
                                                         faux_bgr.put("raw_data", json_object.getDouble("sgv"));
+                                                        faux_bgr.put("age_adjusted_raw_value", json_object.getDouble("sgv"));
```
jamorham: I'm not exactly sure what this will do but it is okay for completeness. I didn't think that field was used if `hasLibre()` is true.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
tzachi-dar: Well, read the code of nfcreaderx.readPatch again.
I guess that if all the conditions below will be true: 
1) nfc_test_diagnostic is off  
2) multiblock is on.
3) addressed is off
4) Increase the number of blocks that is read from 42 to 43 (that does not divide in 3)
Than the code should do the exact same thing.

So I'll change the function if that is what you think that I should do, but do you think that it will make the code more readable.

app/src/main/java/com/eveningoutpost/dexdrip/NSEmulatorReceiver.java
tzachi-dar: This was needed when I have started looking at applying calibrations on the oop algorithm as well.
But this will wait for the next pr.

tzachi-dar:
I have removed the code that allows more than one message in 5 minutes. There will probably be many things that will not work, since we have code that is saying something like:
Get latest X bg readings(), and that might break if there are readings every minute.
Furthermore, I'm not 100% sure it will work well, if we have a reading every 7 minutes as well.
This will probably work OK for blucon, as is.

But this will need some more thinking once we do fill gaps. In other words the question of what to do if one has scanned after 7 minutes. Should we show him the value of 2 minutes ago?
Well there is always the next pr to think about it.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
jamorham: As far as I can see the code in `LibreOOPAlgorithm.readPatch` duplicates the nfc transmission functionality of the existing `readPatch` method which has greater flexibility which was put in for speed and compatibility reasons. 

The only real difference I can see is that `LibreOOPAlgorithm.readPatch` reads slightly more data. My thinking is that this duplicate code should be removed and I assume that the existing `readPatch` would want to be modified slightly to read the additional data blocks as I assume your external interface needs *exactly* that many bytes of data, which is what I was getting at earlier.

If multiples of 3 is a problem then just read an extra block and truncate the resulting byte array before returning it or we can do something to read one or two singles blocks at the end if in multi-block mode if it is absolutely necessary.

If you *need* an initial preparation command like is in `getPatchInfo` then you can just inject the byte sequence in to the existing `diag` byte array and it will be transmitted before the main read occurs. As far as I am aware this is not required for successful read operations, but if you have tested using this already and feel it wont work without it then that is where I would suggest utilizing that functionality. 

Ideally have a method in `LibreOOPAlgorithm` which returns the byte sequence and populate `diag` from that if the preference item is set and then pass the diag `replyBlock` back along with the main data block or if you need to cancel before that then make a call to `resultOk` and skip/return early based on a failure there.

The idea is to keep the nfc reading in one place and any specific data processing in another class as much as possible.

tzachi-dar:
Closing PR since it was reimplemented on https://github.com/NightscoutFoundation/xDrip/pull/240

PR #237, opened 2017-12-11
Tzachi xdrip uploader

tzachi-dar:
With this change, xDrip will connect to the BT device and send the data that it collects to mongo DB. The data can be processed by another xDrip using wifi. This works for G4 and G5 devices.
Other data sources can be added later.

This should allow all house coverage for G5 or two parents to use one BT device transparently.
Basic testing on g4 worked correctly, and I'm starting a more serious testing phase. (I will not test G5 that I don't have, but change is trivial there).

Still need to implement rest uploads.

All comments are welcomed.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
@@ -13,6 +13,7 @@ import com.eveningoutpost.dexdrip.Services.Ob1G5CollectionService;
 import com.eveningoutpost.dexdrip.UtilityModels.Constants;
 import com.eveningoutpost.dexdrip.UtilityModels.NotificationChannels;
 import com.eveningoutpost.dexdrip.UtilityModels.PersistentStore;
+import com.eveningoutpost.dexdrip.UtilityModels.UploaderQueue;
 import com.eveningoutpost.dexdrip.utils.PowerStateReceiver;
 import com.eveningoutpost.dexdrip.xdrip;
 import com.polidea.rxandroidble.RxBleConnection;
@@ -467,6 +468,7 @@ public class Ob1G5StateMachine {
             return;
         } else {
             UserError.Log.d(TAG, "Created transmitter data " + transmitterData.uuid + " " + JoH.dateTimeText(transmitterData.timestamp));
+            UploaderQueue.newTransmitterDataEntry("create", transmitterData);
             // TODO timeInMillisecondsOfLastSuccessfulSensorRead = captureTime;
         }
         Sensor sensor = Sensor.currentSensor();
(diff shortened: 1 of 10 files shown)


jamorham:
Can you explain the use case in more detail? I would be skeptical that this would work for G5 because of the collisions which can occur on the bluetooth layer and the G5 requesting reauthentication when switching clients. 

With G4, do devices like the HM10 allow sending notifications to multiple clients? What will happen when one device is then out of range or partially in range will this jam communication to the other? The testing environment for this is complex. I assume you are going to be running it in a live test there?

tzachi-dar:
For the g4 only one device can connect at a time. So the first one that
connects will get the data, and others won't. If the device currently
connected will get out of range, other device might connect.

In our use case, xDrip is installed in the living room, while the boy might
go out. So, I had a program that was based on old version of xDrip that
connected to the BT and uploaded it to the mongo. But xDrip keeps going
forward, and I don't want to keep copying the new code to my program.

I'm also aware of other parent that has a device he built that works with
xDrip but not with my program, and now he can connect with any phone that
he wants.

As for the g5, I don't have experience with it, but as far as I know, only
one device can connect at any given time, so every time the g5 wakes up,
some phone will reach it and get the correct data.

We will try this for a few weeks in order to see how it is working.

tzachi-dar:
First experiments seems very promising. Two phones are 'fighting' over one BT device.
One of them seems to get connected, while the other one not, until the other connects. (I'm not sure what triggers the change).

tzachi-dar:
A small  update regarding this branch:
After talking with Adrian, it seems that for the g5 there will be (at least) two packets that are lost when switching from one reader to the other. So, g5 is not practical.
As for the G4 testing was good, and now I'm moving to testing it with xBridge that will close and open the BT service for each reading.
Will report how things work once I have more results.

tzachi-dar:
Looks like some of my data is not being uploaded correctly. (or actually missing).
Need to understand why (since SGV values are uploaded, this looks even stranger).

tzachi-dar:
Here is a short update regarding this issue:

So, I'm using an xBridge with a version that also retransmit packets on demand.
It seems that all packets are captured, but the BT stack on nexus 5 fails to connect many times. It seems that when there are a few packets to upload, there is a chance that they will not be stored in mongo.

I guess that this should not surprise us since, we do the transactions to the mongo with WriteConcern.UNACKNOWLEDGED which does not promise anything if I understand correctly. 
I have changed that to WriteConcern.ACKNOWLEDGED and all goes well, but the time for the operation to end is 4-10 seconds (used to be much shorter before). 

I would say that this did not work even before my changes (with WriteConcern.UNACKNOWLEDGED), but it does work for me (and probably for many other people). Maybe the problem only happens because I'm loading more than one packet at a time in this case.

I'll keep watching this and see what I'll get.

tolot27:
It looks like this is related to #1385 but supports the Dexcom datasource. Most of the code changes are related to data upload. Is this supersed by #1385?
If not, please rebase this PR to the current master.

tzachi-dar:
This is not related to #1385, but was merged on some other pr.

PR #238, opened 2017-12-13
Delay disabling Force Wear on low battery threshold by 5%

kskandispersonal:
Requested by @WanaGo and @old-square-eyes

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/CheckBridgeBattery.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/CheckBridgeBattery.java
@@ -2,7 +2,6 @@ package com.eveningoutpost.dexdrip.utils;
 
 import android.app.PendingIntent;
 import android.content.Intent;
-import android.util.Log;
 
 import com.eveningoutpost.dexdrip.Home;
 import com.eveningoutpost.dexdrip.Models.JoH;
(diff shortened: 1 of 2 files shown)


jamorham:
Can you give me a description of what this set of changes is designed to do? Thanks

kskandispersonal:
It postpones disabling force wear until a threshold of bridge_battery_alert_level - 5.  This delay should give the user a chance to recharge the watch battery before it gets switched to the phone.  It was requested on gitter.

app/src/main/java/com/eveningoutpost/dexdrip/utils/CheckBridgeBattery.java
```diff
@@ -63,6 +65,34 @@ public static boolean checkBridgeBattery() {
...
+            UserError.Log.e(TAG, "Got error parsing alert level");
+        }
+
+        final int this_level = Home.getPreferencesInt("bridge_battery", -1);
+        UserError.Log.d(TAG, "checkForceWearBridgeBattery threshold:" + threshold + " this_level:" + this_level);
+        if ((this_level > 0) && (threshold > 0)) {
+            if (this_level < threshold) {
+                lowbattery = true;
+                if (notification_showing) {
+                    cancelNotification(NOTIFICATION_ITEM);
```
jamorham: I don't understand why the notification is cancelled here?

kskandispersonal: I cancel the notification because the watch collector is being switched to the phone.  BUT maybe we should just let the user manually cancel it!  Sorry, I'll remove the cancellation.

PR #240, opened 2017-12-15
Tzachi oop phone scan 2

tzachi-dar:
Read first 43 blocks of sensor, and send it to oop algorithm if needed.

In this change, number of blocks being read was increased to 43 (actually in multiblock it will read even 45 blocks).
Since original buffer was already 360, there was no need to increase buffer size.
From my experiments on a dead sensor, data that is read is always the same. Still need to verify it on real sensors of course.

To make code simple, the same number of blocks is always read. If anyone sees a problem with it, please let me know.

All comments are welcomed.


jamorham:
This looks a lot cleaner and simpler to me. Reading through it it looks like it should work.

PR #243, opened 2017-12-17
blue reader intigration as class

MasterPlexus:
Changed the integration of blueReader to a Class approach:

- switch the blueReader processing from DexCollectionService to model blueReader-class.
- try to put bridge-battery in better place (out of statusfragment)
- try to establish a better check that blueReader is used from activeBluetoothDevice.name property
- handling if blueReader is in a ugly state

Hope this way fits more the requirements/coding of the project. If there are still stuff wich needs to be handled in a other way no problem to give me further advice ;)


jamorham:
Ok this looks good, I am reviewing it now

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -1317,6 +1322,14 @@ public void run() {
             }
         }
 
+        // blueReader
+        ActiveBluetoothDevice activeBluetoothDevice = ActiveBluetoothDevice.first();
```
jamorham: Can you use the static method `isblueReader()` here to avoid duplication?

jamorham:
Ok so this looks good. I would like some changes for the persistence storage. Basically it goes like this, historically xDrip has used the system preferences data to store some state information. This isn't good as it clutters preferences and can cause state problems when preferences are loaded/saved.

For all new persistent state stuff like this, instead use `PersistentStore` this works identically to what you are already using. The only difference is that it doesn't work with `int` but instead uses `long` for integers.

So where you have `  Home.setPreferencesInt("blueReader_Full_Battery", transmitterData.sensor_battery_level);` you would change this to: `PersistentStore.setLong("blueReader_Full_Battery", transmitterData.sensor_battery_level);` there are associated `get` methods to retrieve the data. `String` is also supported.

The only place you *don't* do this is for  `bridge_battery` because this is a legacy variable which we haven't moved to `PersistentStore` yet.

The only other thing is I need you to confirm that you have tested this and that it is working correctly for you.

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
MasterPlexus: Correct, seen that also too late after pull, will be changed in next commit.

app/src/main/java/com/eveningoutpost/dexdrip/Models/blueReader.java
```diff
@@ -0,0 +1,136 @@
...
+import com.eveningoutpost.dexdrip.Models.UserError.Log;
+import com.eveningoutpost.dexdrip.xdrip;
+
+/**
+ * Created by MasterPlexus on 11.12.2017.
+ */
+
+public class blueReader {
+    private static final String TAG = "blueReader";
+    private static SharedPreferences prefs;
```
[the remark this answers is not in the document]
MasterPlexus: Right, will be changed.

MasterPlexus:
OK, changed also the PersistentStore now,as I now also understand the difference between bridge_battery in that case (that was one point which puzzling me ;) )

Anyhow, the code works, right yesterday, but now I facing some mysterious time holes in the receiving of the glucose values. I've given the apk also to a other user which checks if this is really within my code, or because I set a new sensor Yesterday. 
Additionally I have made forme some Log.w's to check what really happend and will have a look this night. Come back tomorrow with last changes.

MasterPlexus:
I'm put your remarks into the code, and add some small enhancement for the ugly state.

For me, and one other the change works properly now for over one day. 

If someone see lost readings with the blueReader, then may switch off the "close GATT on BLE disconnect", as I see it for my phone that it helps (just to comment, don't know if this is well known in general) .

From my point of view ready to implement.

blocklist:
For me, the code (a966c261ace5c80f5dc08497ece2360d99722685) works in my xdrip Version.

jamorham:
See null pointer fix:

https://github.com/NightscoutFoundation/xDrip/commit/9b58353cf3e323af374d9f946db1b3cccccd768d

PR #249, opened 2017-12-29
Tzachi librealarm oop support

tzachi-dar:
Allow xDrip and libre alarm to work with OOP algorithm.

(a corresponding PR will be sent to librealarm in a day or two.)

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
@@ -10,10 +10,13 @@ import android.os.PowerManager;
 import android.preference.PreferenceManager;
 
 import com.eveningoutpost.dexdrip.Models.UserError.Log;
+import com.eveningoutpost.dexdrip.ImportedLibraries.usbserial.util.HexDump;
 import com.eveningoutpost.dexdrip.Models.BgReading;
 import com.eveningoutpost.dexdrip.Models.Forecast;
 import com.eveningoutpost.dexdrip.Models.GlucoseData;
 import com.eveningoutpost.dexdrip.Models.JoH;
+import com.eveningoutpost.dexdrip.Models.LibreBlock;
+import com.eveningoutpost.dexdrip.Models.LibreOOPAlgorithm;
 import com.eveningoutpost.dexdrip.Models.ReadingData;
 import com.eveningoutpost.dexdrip.UtilityModels.Constants;
 import com.eveningoutpost.dexdrip.UtilityModels.Intents;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
```diff
@@ -168,6 +171,15 @@ public void run() {
     }
 
     public static void processReadingDataTransferObject(ReadingData.TransferObject object) {
+    	Log.i(TAG, "Data that was recieved from librealarm is " + HexDump.dumpHexString(object.data.raw_data));
```
jamorham: What will happen with data from received from versions of LibreAlarm which do not support this field? Will this produce a crash or will the field not be initialized? If it is not initialized then it shouldn't be saved by `LibreBlock` (or otherwise processed) More checking should be done to ignore missing data.

app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreOOPAlgorithm.java
```diff
@@ -14,13 +14,20 @@
 	private static final String TAG = "LibreOOPAlgorithm";
 	
     static public void SendData(byte[] fullData) {
+    	if(fullData == null) {
+    		Log.e(TAG, "SendData called with null data");
+    		return;
+    	}
+    	
+    	if(fullData.length < 0x158) {
```
jamorham: 0x158 would be more readable I think if it were expressed in decimal, eg 344

jamorham:
Are you going to produce patches against my version of LibreAlarm or another?

What are the size limits for broadcast intents on various versions of Android? Will this additional data have the possibility to exceed the maximum size allowed? This needs to be tested I think. I can't offhand remember what the limits are.

tzachi-dar:
Already sent a PR against your version: https://github.com/jamorham/LibreAlarm/pull/9

tzachi-dar:
The exact size of broadcasts on android is not clear, but the numbers that are mentioned are between 90k and 1MB, so my 344 bytes are not a dangour.

https://www.neotechsoftware.com/blog/android-intent-size-limit
https://stackoverflow.com/questions/12496700/maximum-length-of-intent-putextra-method-force-close

app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
tzachi-dar: I was running with the older version, and the value that was passed for raw_data was NULL.

So, the program does not crash, and the only thing that the user will see in the log is:
LibreBlock - cannot save block with null data.
LibreOOPAlgorithm senddata called with null data.

I'll now add some code that will print in this case:
Please update librealarm to use OOP algorithm.

tzachi-dar:
Pushed the changes.

Please let me know if there is anything else that I should do/change.

app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
```diff
@@ -168,6 +171,20 @@ public void run() {
...
+    	// Save raw block record (we start from block 0)
+        LibreBlock.createAndSave("LibreAlarm", object.data.raw_data, 0);
+
+        if(Home.getPreferencesBooleanDefaultFalse("external_blukon_algorithm")) {
```
jamorham: Since you created this PR there has been a refactoring in xDrip master (moving the static preference handling to its own class) - this needs a change here:

Change `Home.getPreferencesBooleanDefaultFalse("external_blukon_algorithm")` to `Pref.getBooleanDefaultFalse("external_blukon_algorithm")`

jamorham:
Once the refactoring issue with the preferences is updated then I reckon this should be ready to merge.

tzachi-dar:
I have rebased the entire branch against current master and did the needed fix.

I have tested new code (against latest librealarm) and opened a new PR.

PR #258, opened 2018-01-05
Tzachi librealarm oop suport2

tzachi-dar:
This is the branch "Tzachi librealarm oop suport" checri-picked against head and tested again against latest version of librealarm.


tzachi-dar:
Thanks for the review.

Issue #244, opened 2017-12-17
SW3 Watchface Edit


Elvin6669:
Is it possible to edit/personalise the Sony SW3 watchface? 
Would need adding the date and maybe weather updates.

Any directions where to find the code?

jamorham:
Look in the `wear` folder of the main source tree. I would suggest any changes are either options or create an additional watchface from the base so any code changes are more easily accepted to the project as they wont change any existing behavior.

Elvin6669:
I looked throught the code and seems that the main watchface layout is here xDrip/wear/src/main/res/layout/rect_activity_home.xml
Which would be the Id of the date which could be sourced? 
I am thinking to have the date in the left of the time in the format:
MON
18/12
And maybe as a double-tap button to have it as
MON
12/18
And how about to have a small weather icon with temperature underneeth(simmilar to delta layout) on the right of the time.
For that needs to be created a new id with an outside weather-source based on location? Plus icon drawings?

Elvin6669:
I was trying to change this in the layout with Android Studio but the watch module was crashing when loading the home screen.
I think is because the colour layout code needs to be changed too? From wear\src\main\java\com\eveningoutpost\dexdrip\Home.java ?

This is now a bit too complicated for me.. is it possible  to help out with this feature?
For me it is a bit weird not to have these important features on the watch face of a smartwatch.
Or do you have a different way to add the date and weather forecast info onto the watch face?

jamorham:
Day / date etc is fairly standard but I am much less sure about weather. Based on your detailed diagram I am not sure why it would need custom colors changing. @kskandispersonal may have more ideas about this as she is more familiar with the wear watchface than me.

kskandispersonal:
Yes, this request should be doable, and it is a good idea, too:)  We can add these as preferences to the BaseWatchface.  As @jamorham mentioned, Day / date is standard.  For weather, http://openweathermap.org/appid provides an open weather API.  Have a look at https://code.tutsplus.com/tutorials/create-a-weather-app-on-android--cms-21587 for an example of how it is used.

I'm not sure why the watch module keeps crashing for you.  I don't think color layout should cause it to crash.

Elvin6669:
That sounds good, if is added to preferences it is even better, seems that every watchface has space for it on each side of the displayed time(on the round watchface on each side of the bg display) 
I see that even icons for weather are already available in the tutorial link!
[...]

PR #264, opened 2018-01-08
Add Date to Home and BIGChart watchface

kskandispersonal:
Enable this feature via the XDrip Prefs Show Date.  Format defaults to EEE mm/dd for current locale.


kskandispersonal:
This feature is in response to Issue #244 SW3 Watchface Edit.  This PR does not include the weather request in #244.

jamorham:
How does this change the look? Could you post a screenshot?

I don't really think weather is needed.

kskandispersonal:
screen

Yes, I agree, esp as Wear 1.5 does not support HttpURLConnection.  Best to just use the phone for weather in that case since Wear 1.5 watches will not be able to retrieve weather updates standalone anyway.  For Wear 2.0, a complication might be the better way to go using data from an existing weather app, assuming the weather app updates standalone on the Wear 2.0 device.

PR #271, opened 2018-01-14
Detect locale change to update date format

kskandispersonal:


jamorham:
Does the same also need to be done in `BaseWatchFace` ?

kskandispersonal:
I'm not sure what your question is.  The same is done in both BIGchart and BaseWatchFace.  I'm not sure why AU locale is not being supported.  Maybe as you suggest on gitter, we should just use final Locale locale = Locale.getDefault();

kskandispersonal:
Sorry, I forgot to commit the other changes for this PR (BaseWatchFace,java and *.xml which gives a bit of space between date and time).

PR #273, opened 2018-01-17
Override watch with phone locale

kskandispersonal:
Override watch with phone locale if wear XDrip overrideLocale pref is enabled.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/wearintegration/WatchUpdaterService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/wearintegration/WatchUpdaterService.java
@@ -2,8 +2,10 @@ package com.eveningoutpost.dexdrip.wearintegration;
 
 import android.app.ActivityManager;
 import android.app.PendingIntent;
+import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
+import android.content.IntentFilter;
 import android.content.SharedPreferences;
 import android.os.AsyncTask;
 import android.os.Bundle;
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/wearintegration/WatchUpdaterService.java
```diff
@@ -630,7 +634,23 @@ public void onCreate() {
         }
         setSettings();
         listenForChangeInSettings();
+        IntentFilter filter = new IntentFilter(Intent.ACTION_LOCALE_CHANGED);
+        registerReceiver(localeChangeReceiver, filter);
```
jamorham: Okay so I think this `registerReceiver` is going to cause problems.  There is no associated unregister and it is going to outlive the service which creates it. In my experience these get stacked up which will result in weird multiple sending and memory leaks even if you try to be careful and unregister in onDestroy etc.

The user isn't likely to change locale that often so I would say just send the current phone local when we send settings on the watch. That way if they happen to change locale it will get propagated eventually or they can do something like toggle xdrip wear off/on and cause the locale to get resent like that.

Let me know what you think.

kskandispersonal: I totally agree with you re: just send the locale with each sendPrefSettings() to watch, which this PR already does, so I will just remove the intent code.   sendPrefSettings() always gets sent on watch startup and I believe on reconnection as well.  I suppose we could implement a "Resend Prefs to Watch", similar to the request, "Resend glucose to watch" on the phone.

PR #276, opened 2018-01-22
Speak arrow trend and include Bridge_battery name in NightScoutUploader

MasterPlexus:
correct the speak of trend and the repeating of the spoken words, Including the translation of the Trend (initially in en and de).
Tested,and works for me.

Include also the right name of the Bridge into the NightScoutUploader for the Bridge_battery
Tested and works forme several days.


app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -941,7 +942,7 @@ private void postDeviceStatus(NightscoutService nightscoutService, String apiSec
                     break;
                 case "Bridge":
                     battery_level = Pref.getInt("bridge_battery", -1);
-                    battery_name = DexCollectionType.getDexCollectionType().name();
+                    battery_name = DexCollectionService.getBestLimitterHardwareName();
```
jamorham: This method currently only works for LimiTTer

app/src/main/java/com/eveningoutpost/dexdrip/utils/BgToSpeech.java
```diff
@@ -97,6 +96,7 @@ private static String calculateText(double value, String delta_name) {
 
         final boolean doMgdl = (Pref.getString("units", "mgdl").equals("mgdl"));
         final boolean bg_to_speech_repeat_twice = (Pref.getBooleanDefaultFalse("bg_to_speech_repeat_twice"));
+        final boolean bg_to_speech_trend = (Pref.getBooleanDefaultFalse("bg_to_speech_trend"));
```
jamorham: Isn't this conditional already handled in `realSpeakNow()` ?

jamorham:
This is a helpful PR but the parts of it which I would like to integrate are the translated strings and use of those when speaking the trend.

app/src/main/java/com/eveningoutpost/dexdrip/utils/BgToSpeech.java
MasterPlexus: Yes it is, but I see in my tests that on any reason, if I deselect the option of speaking trend, that a null value will be put into the string, also if the test condition did not allow that. So I made that as an double safety reason.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
MasterPlexus: The name of the function is misspelling. Getbestlimmitterhardwarename will return the real device name (like also bluereader, limmitter, bluecon, sweetreader)

MasterPlexus:
@jamorham , I've actualized the pull, and included also the translation of the Speak menu. 
Also I moved the two lines to the place wehre I think it is the right one.

Please ave a look in it. If it is to wired due to the updates, I could close it and generate a new one.

MasterPlexus:
will Close this pull request, and made a new one, with translations of trend and settings for Speech, due to better checkings.

PR #281, opened 2018-01-25
Fix layout of Missed Readings/Alert screens

jwoglom:
Slightly modifies the layouts of the Missed Readings and Edit/Create Alert settings pages to be more consistent. 

Missed Readings layout screenshot: 
image

Edit Alert layout screenshot: 
image

### DIFF
--- a/app/src/main/res/layout/activity_missed_readings.xml
+++ b/app/src/main/res/layout/activity_missed_readings.xml
@@ -29,31 +29,49 @@
                             android:layout_height="wrap_content"
                             android:text="Enable Missed Reading Alert"
                             android:textSize="15sp"
+                            android:padding="5dp"
                             android:layout_gravity="left"
-                            android:gravity="right" />
+                            android:layout_marginLeft="10dp"
+                            android:gravity="left" />
                         
                         <TextView
                             android:id="@+id/missed_reading_text_alert_time"
                             android:layout_width="wrap_content"
                             android:layout_height="wrap_content"
                             android:gravity="left"
-                            android:text="Alert if no data recieved in x minutes:"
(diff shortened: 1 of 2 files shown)


jamorham:
Hi there, thanks for submitting this PR.

Hybrid PRs such as this present a problem for review, it addresses two very different issues which should be split in to two PRs. I know what its like when editing code and you fix different things as you go but it makes the review process much harder and if one of the sets of changes is rejected there is no easy way for me to accept the other parts of the PR.

I would like to examine what you've done with the layout for the alarms because I'm sure it can be improved and I expect you have done so. But as far as changing the strings key goes this isn't something I really want to do because it has no effect on the end user and introduces incompatibility and merge conflicts with any translators who are half way through a translation.

Can you back out those changes and push an updated version so that I can review the changes to the screens as a single item.

Thanks!!

jwoglom:
No problem -- removed the translation commit from the PR.

app/src/main/res/layout/activity_missed_readings.xml
```diff
@@ -27,33 +27,51 @@
...
+                            android:layout_marginLeft="10dp"
+                            android:gravity="left" />
                         
                         <TextView
                             android:id="@+id/missed_reading_text_alert_time"
                             android:layout_width="wrap_content"
                             android:layout_height="wrap_content"
                             android:gravity="left"
-                            android:text="Alert if no data receieved in x minutes:"
+                            android:text="Alert if no data recieved in:"
```
jamorham: It looks like you've changed one spelling mistake for another in `received`

jamorham:
Overall this looks like an improvement. I will be happy to approve and merge but can you take care of that spelling mistake I noted above. Thanks

PR #283, opened 2018-01-28
Add GVI, PGS metrics to History and Stats per Nightscout glucosedistr…

kskandispersonal:
…ibution.js

https://github.com/NightscoutFoundation/xDrip/blob/kskandis-update1/app/src/main/java/com/eveningoutpost/dexdrip/stats/StatsResult.java ln 152 and FirstPageFragment.java152 differ from https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/report_plugins/glucosedistribution.js at ln 390.

GVI and PGS equations are discussed here:
https://www.healthline.com/diabetesmine/a-new-view-of-glycemic-variability-how-long-is-your-line

### DIFF
--- a/Documentation/WatchGuide.md
+++ b/Documentation/WatchGuide.md
@@ -28,7 +28,7 @@ xDrip+ Wear Setup and Troubleshooting Guide
     - [Debugging Android Wear](#debugging-android-wear)
 
 # Enable xDrip+ Android Wear Integration
-xDrip+ supports wear integration through its wear app.  The xDrip+ wear app is installed with the standard xDrip+ apk. On **Android Wear 1.x** watches, the wear component gets synced automatically to the watch. On **Android Wear 2.0** watches you need to install (activate) the wear component separately from the Play-Store on the _watch_ (under the section "Apps on your phone") after installing xDrip+ on the phone. The latest release supports a standalone mode which allows the wear app to communicate directly with the Bluetooth Collector, whether it is a Dexcom G5, a Dexcom G4 + xBridge, or a Libre LimiTTer.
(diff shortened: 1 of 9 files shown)


kskandispersonal:
Here is the original nih dexcom article describing these metrics:
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5467104/

jamorham:
What is the state of this? Has it been tested, ready for merge?

kskandispersonal:
Yes, it is ready for merge.

I tested the GVI and PGS algorithm and compared them to my Nightscout values but I do get different results due to a slightly different algorithm.  NS uses the full BG dataset but this version uses only those above 38 (the DBSearchUtils.CUTOFF ).  Also, this version uses a different equation from NS for the Ideal Line Length.

I tested the fix for the step counter sync (I.e., SYNC_ALL_DATA) but because my phone stopped working as a collector, I could not fully test with wear integration disabled.  I wanted to test that scenario to ensure steps were synced back to the phone when using the AAPS watchface while using the phone as collector.  It does sync as expected with Force Wear disabled though on Missed Readings, which is a similar scenario.

kskandispersonal:
Resolved a few addl issues re: disabled wear integration.  Updated watch doc w/ new watchface preferences.

kskandispersonal:
I may need to back the wear integration changes in this PR. Please hold on their merge 'til I test further.

doc77:
I had an endo appointment this afternoon and the doc was mentioning GVI as the most reliable standard of glycemic control. He had been to a seminar recently which he said explained it fully, and it makes a great deal of sense. 
Thank you @kskandipersonal for including this most helpful metric.

jamorham:
I look forward to integrating this whenever it is tested and ready

kskandispersonal:
This is ready for merge.  I've been using this version for the past week and it has been working as expected.  Step counts do get synced to xDrip+ phone when using the AAPS watchface w/ wear integration enabled, and Enable Wear/Force Wear disabled.

PR #285, opened 2018-01-30
Tzachi oop backfill and allow calibrations

tzachi-dar:
This PR allows the oop algorithm to back fill data for the last 8 hours.
It also allows to use the calibration mechanism on OOP results.

It still needs more testing, especially for the integration with the calibration mechanism.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
@@ -42,8 +42,8 @@ public class LibreAlarmReceiver extends BroadcastReceiver {
 
     private static final String TAG = "jamorham librereceiver";
     private static final boolean debug = false;
-    private static final boolean d = false;
-    private static final boolean use_raw = true;
+    private static final boolean d = true;
+    private static final boolean use_raw_ = true;
     private static final double segmentation_timeslice = Constants.MINUTE_IN_MS * 4.5;
     private static SharedPreferences prefs;
     private static long oldest = -1;
@@ -185,6 +185,10 @@ public class LibreAlarmReceiver extends BroadcastReceiver {
         	LibreOOPAlgorithm.SendData(object.data.raw_data);
         	return;
         }
+        CalculateFromDataTransferObject(object, use_raw_);
+    }
+        
+    public static void CalculateFromDataTransferObject(ReadingData.TransferObject object, boolean use_raw) {
     	
         // insert any recent data we can
         final List<GlucoseData> mTrend = object.data.trend;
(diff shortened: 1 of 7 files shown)


tzachi-dar:
I was working on adding a dialog that will allow one to see the last 15 minutes of data from the libre sensor. (In a place that this data was collected, for example, direct scanning, or libre alarm, or bluekon (with the oop algorithm).

The most compelling place to add this data was to use a click on the button. However, this place is already taken by the "add note snackbar".
Since a snakebar has only one action, I added both actions to one dialog, that ended looking like the following picture.

My questions are this:
1) Does this look good enough?
2) Should I put the dialog in another place? maybe in the bg data table. (although there is already a click there).
All input is welcomed.
screenshot_20180205-013453

jamorham:
The last 15 minutes numeric data looks like engineering mode debugging data. I can't think that normal users will be interested in this data and instead would just view the standard graph? Unless I am misunderstanding what it is for.

How about add a 3 dot menu on the right of the Bg Readings table activity and display the information there if it has been gathered? That would seem the proper place.

tzachi-dar:
Well, the user is many times interested to know if on the last 15 minutes bg is going up or down.
We have a problem with displaying this data in xDrip for the following reasons:
1) xDrip is showing only one dot per 5 minutes. More then that let's assume that one has scanned on 10:00, and then on 10:08. We will only show one point on 10:00 and the other on 10:08, so if BG was falling from 9:30 to 10:00, and he ate something and now from 10:05 to 10:08 bg is already stabilizing xDrip will not show it (and I guess that abbott reader will not as well - abbott reader is drawing without having the 15 minutes data at all).

2) Text in (1) above assumed a clean sensor reading. But if the sensor is noisy, there is no way for us to show this to the user (due to the accumulation of readings to 15 minutes, they will always look smooth).

All that said, I agree that the normal user will have a hard time understanding this.

And I'm 100% fine with adding this information to the 3 dot menu, if there is such information from the last 30 minutes.

jamorham:
Ok so lets keep the PR as it is at the moment which as far as I can tell contains no UI changes.

I would like to merge it but I don't completely follow what is going on in the receiver.

Can you confirm that this should have no impact on existing LibreAlarm users who are not using any of these new calibration features? I would like to merge it but I have to be careful not to break existing use cases.

As there are two separate issues here I will write about the UI stuff in the next comment.

jamorham:
For the UI element, if the idea is to assist users with more real-time data than xdrip supports. What I would suggest doing doing this graphically with an additional icon/text or mini curve displayed or some other representation of this data. A list of numbers is not very user friendly I don't think. Definitely not injected in to another unrelated UI element.

For implementation I would say you could store this possibly in an in-memory structure (maybe you already do this) and then it is possible to have extra optional UI components on the main home screen of xDrip.

I already do this with the Time Buttons which you can see by long pressing the xDrip title bar icon and selecting the option. I can improve this and make it more modular so that you could for example provide a fragment in a separate class to aid maintainability and separate the responsibilities. 

Then if the preference option is enabled we can give the user the option to switch on whatever extra mini UI element would best represent this data. That is the best way I can think about showing this extra information without altering the core data structures and related code ref the 5 minute reading interval.

Jon

tzachi-dar:
OK, we can go with this pr as is, but I would still want to do two minor changes before submitting it, so please wait with the submission. Since most of my checks were without calibration, I will still leave this option as grayed out, for now.

I have double verified that things should work for previous users as they are right now.

As for the UI, I agree, I have already created a dialog that displays last minute data. Still working on small changes and hope to submit this pr soon.

tzachi-dar:
closing this as https://github.com/NightscoutFoundation/xDrip/pull/298 is more recent.

PR #286, opened 2018-01-30
Update strings.xml

bordeb:
Just a small typo "unaknowledged" changed to "unacknowledged". Noticed this just while making a translation


bordeb:
You are very welcome. The pleasure is all mine.

PR #288, opened 2018-01-31
Bordeb transl hr Translation to Croatian (HR - Hrvatski) v 1.0

bordeb:
added strings-hr.xml file and made changes to app/src/main/res/values/locales.xml to add Croatian (Hrvatski) language


bordeb:
Glad to hear! 
I personally know (over phone) one native speaker who has openaps so I am pretty sure that he also uses xdrip, ns..., and I could maybe ask him to join the thread and provide a review if you think it is a good idea?

jamorham:
Yes that would be much appreciated. The changes will also be visible in today's nightly release.

bordeb:
Hi @jamorham , as discussed I have contacted @isuran who immediately accepted the proposal, so I am sure he will join in asap. I look forward to translation review and improvements!

bordeb:
I am so glad that you noticed this. While translating I was really torn between those two options while trying to translate the word "calibration". I finally opted for the "baždarenje" rather than "kalibracija" thinking that it is more literary language and that medical staff (and consequently maybe patients, parents etc) would use that more often than. Please suggest what to do. 
1.) Change "baždarenje" to "kalibracija"
2.) Add "kalibracija", so it might finally be: "baždarenje - kalibracija"
3.) Leave it as is for now - maybe future comments of different users will help...
4.) Speaking about thoroughly, what do you think between "kalibracija" vs. "kalibriranje", as performing calibration activity in the present time.

PR #298, opened 2018-02-10
Tzachi oop backfill2

tzachi-dar:
This is https://github.com/NightscoutFoundation/xDrip/pull/285 just against latest head.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -1057,11 +1057,19 @@ public class BgReading extends Model implements ShareUploadableBg {
 
             bgr.timestamp = timestamp;
             bgr.calculated_value = value;
+            
 
             // rough code for testing!
             bgr.filtered_calculated_value = value;
-            bgr.raw_data = value*1000;
-            bgr.filtered_data = value*1000;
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -1057,11 +1057,19 @@ public static void bgReadingInsertFromInt(int value, long timestamp, boolean do_
 
             bgr.timestamp = timestamp;
             bgr.calculated_value = value;
+            
 
             // rough code for testing!
             bgr.filtered_calculated_value = value;
-            bgr.raw_data = value*1000;
```
jamorham: Wont this break all existing calibrations from LibreAlarm etc?

tzachi-dar: bgReadingInsertFromInt is only called from the function CalculateFromDataTransferObject and only if use_raw==false.

For existing calibrations from LibreAlarm there is no way (and there wasn't before my change) to have use_raw==false.
So, the function CalculateFromDataTransferObject () was never called before my change, so I thought that changing it is save.

jamorham: ok great

PR #299, opened 2018-02-10
Tzachi trend 2

tzachi-dar:
This pr adds a graph with latest bg data for the latest 15 minutes.

screenshot_20180211-014207 1

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreBlock.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreBlock.java
@@ -7,6 +7,7 @@ import java.text.DecimalFormat;
 import com.activeandroid.annotation.Column;
 import com.activeandroid.annotation.Table;
 import com.activeandroid.query.Select;
+import com.eveningoutpost.dexdrip.Models.UserError.Log;
 import com.google.gson.annotations.Expose;
 
 /**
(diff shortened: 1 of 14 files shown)


tzachi-dar:
After playing with this PR for a few days, seems that there is one thing that still needs improvement:

Consider a scan that was done on time X. And then another scan that was done on time x+3 minutes.
The scan on x+3 minutes has a libreblock data, but no bg was created for it (xDrip is looking for 4 minutes difference).

In order to draw the correct trend, I need to read the BG data that was created on time x+3.

I can store it in memory, but I think that storing it on the LibreBlock is better.

Should I extend the libre block for this or do you think that another solution will be better?

app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreBlock.java
```diff
@@ -48,15 +51,15 @@
     public byte[] blockbytes;
 
     // if you are indexing by block then just * 8 to get byte start
-    public static LibreBlock createAndSave(String reference, byte[] blocks, int byte_start) {
```
jamorham: Really this signature method should remain and call the updated one with the current timestamp to preserve the original behavior and avoid having to refactor the callers.

jamorham:
As discussed add another field to `LibreBlock` just adjust the schema and it should handle it automatically. 

Also note that the PR fails integration checks, I think you forgot to commit the layout file.

app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreBlock.java
tzachi-dar: The problem with the current signature, is that we have a libreblock, and a bg-record that should have the same timestamp.
If each function will call JoH.tsl() by itself, it is not guaranteed that they will have the same value.
This is why I prefer to take the extra effort and change the signature of the function. (timestamp is used as the unique key between the two tables).

tzachi-dar:
Fixed the two issues above. Continuing to test it.

jamorham:
This one is superseded by: https://github.com/NightscoutFoundation/xDrip/pull/317 ?

tzachi-dar:
yes.

PR #310, opened 2018-02-18
Insert the last value as not quick.

tzachi-dar:

This fixes a bug that was introduced in my last commit.


jamorham:
Should this be added to: 

https://github.com/NightscoutFoundation/xDrip/pull/299  ?

tzachi-dar:
It is also on 299, but 299 is a nice to have feature.
This is a bug fix, so I wanted to have it fixed sooner.

PR #314, opened 2018-02-22
Update gradle files to remove deprecated freatures

ahaarrestad:
Updated gradle files to remove deprecated calls:

* compile -> implementation
* androidTestCompile -> androidTestImplementation
* testCompile -> testImplementation
* provided -> compileOnly
* '<<' -> doLast
* removed unneeded ';'


ahaarrestad:
@jamorham I added a small update for the gradle file. No functional changes, only replaced deprecated items.

To remove the last "compile" warning we need to upgrade firebase.

jamorham:
Ok thanks. Would this still work with Android Studio 2.x ? I assume that gradle is already updated so that is the issue here.

ahaarrestad:
@jamorham It is gradle that is complaining.

PR #326, opened 2018-03-02
Tzachi libre trend4

tzachi-dar:
Latest code of the libre trend, against current master.


jamorham:
Reading through this it looks like it should work. I see that this is a new version of the PR. Can you confirm that you've tested this locally?

Thanks. Jon

tzachi-dar:
This was tested locally.

PR #333, opened 2018-03-05
Gvi filter

AdrianLxM:
1) Use the 1st pass filter that fills the gaps like in the Nightscout report
2) Use the 2nd pass filter that eliminates single outliers with deltas > 8 (jump up AND down the next)
3) Use filtered data as micro-noise can artificially lengthen the trace

@kskandis (or @kskandispersonal ?) could you please take a look at this? Thanks!


kskandispersonal:
Looks good.  Thanks for incorporating these changes.

PR #344, opened 2018-03-10
Tzachi tomato 1

tzachi-dar:
This mainly addes checksum for tomato.

It also brings back the printing that have been removed in previous PR. I really hate to go them go away.

This also adds battery, and fw/hw versions to the status page.

I'll test it in the night making sure that bringing back printing (and not caching device name) does not create damage.

Once I have more time for testing, I'll move the checksum function for all libre sources.


tzachi-dar:
Worked well, through the night, can be merged.

jamorham:
I have merged this now, but just a question regarding the previous PR, why was the hexdump removed? Was it a fear about performance?

tzachi-dar:
In previous pr, the tomato device was sending data much faster, so I have removed all printing that were done per packet.
Now the device is sending data much slowlier, so we can handle the printing without an issue.

PR #355, opened 2018-03-17
Update for Tomato and Bluetooth handling

jamorham:
This is some experimental code that I've been running the last 24 hours. The idea is to improve the capture rate with Tomato.

Unfortunately studio for some reason reorganized the code and I didn't realize at the time so this change set looks massive when its not that big, however some of the bluetooth handling changes are pretty significant.

Key changes are summarized below:

* a new `BridgeResponse` data class to handle replies from delegated processors like `Tomato` this allows for error passing etc.
* state handling with `volatile` for extra thread safety
* failed checksum causes a retry request but not more than 2 per minute.
* reinitialize when reconnecting

Within `DexCollectionService` the way disconnections and connections are handled has changed. More of the connection management is handled within the service itself and the backoff timer is reduced.

The idea behind these changes is to try to ensure that when disconnected it reconnects as effectively as possible but without overloading the Android Bluetooth stack.

These changes are the most problematic because they could improve or harm a great many of other unrelated things (non-Tomato), however if they improve Tomato performance then there is good chance they would improve everything.

This new mechanism is likely to use quite a bit more battery when the bluetooth device is out of range because it is much less prone to "giving up" on a missing device and will keep hunting it. If this new method works then I may look at putting in some more backoffs after extended outages because for example after 8 hours of missing device it should be backing off more.

Basically, please try this branch out and see what you think. If it works for Tomato then also we should try it for Blucon and LimiTTer and xBridge to see how they look with it too.

If you don't already have the lombok android studio plugin installed then you will likely need it to properly navigate this code.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
@@ -1,14 +1,15 @@
 package com.eveningoutpost.dexdrip.Models;
+
+import com.eveningoutpost.dexdrip.ImportedLibraries.usbserial.util.HexDump;
+import com.eveningoutpost.dexdrip.Models.UserError.Log;
 import com.eveningoutpost.dexdrip.NFCReaderX;
+import com.eveningoutpost.dexdrip.UtilityModels.BridgeResponse;
+import com.eveningoutpost.dexdrip.UtilityModels.PersistentStore;
+import com.eveningoutpost.dexdrip.UtilityModels.Pref;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Arrays;
-import com.eveningoutpost.dexdrip.Models.UserError.Log;
-import com.eveningoutpost.dexdrip.UtilityModels.PersistentStore;
-import com.eveningoutpost.dexdrip.UtilityModels.Pref;
-import com.eveningoutpost.dexdrip.UtilityModels.StatusItem;
-import com.eveningoutpost.dexdrip.ImportedLibraries.usbserial.util.HexDump;
 
 /**
  * Created by Tzachi Dar on 7.3.2018.
@@ -17,33 +18,34 @@ import com.eveningoutpost.dexdrip.ImportedLibraries.usbserial.util.HexDump;
 public class Tomato {
     private static final String TAG = "DexCollectionService";//?????"Tomato";
 
-    private static enum TOMATO_STATES {
+    private static final String CHECKSUM_FAILED = "checksum failed";
+
+    private enum TOMATO_STATES {
         REQUEST_DATA_SENT,
         RECIEVING_DATA
     }
-    final static int TOMATO_HEADER_LENGTH = 18;
+    private final static int TOMATO_HEADER_LENGTH = 18;
 
     
-    static TOMATO_STATES s_state;
+    static volatile TOMATO_STATES s_state;
     
-    static long s_lastReceiveTimestamp;
-    private static byte[] s_full_data = null;
-    private static int s_acumulatedSize = 0;
-    private static boolean s_recviedEnoughData;
-    
-    
-    
+    private static volatile long s_lastReceiveTimestamp;
+    private static volatile byte[] s_full_data = null;
+    private static volatile int s_acumulatedSize = 0;
+    private static volatile boolean s_recviedEnoughData;
+
+
     public static boolean isTomato() {
         final ActiveBluetoothDevice activeBluetoothDevice = ActiveBluetoothDevice.first();
-        if(activeBluetoothDevice == null || activeBluetoothDevice.name == null) {
+        if (activeBluetoothDevice == null || activeBluetoothDevice.name == null) {
             return false;
         }
-        
+
         return activeBluetoothDevice.name.contentEquals("miaomiao");
     }
 
-    public static ArrayList<ByteBuffer> decodeTomatoPacket(byte[] buffer, int len) {
-        ArrayList<ByteBuffer> retArray = new ArrayList<ByteBuffer>();
+    public static BridgeResponse decodeTomatoPacket(byte[] buffer, int len) {
+        final BridgeResponse reply = new BridgeResponse();
         // Check time, probably need to start on sending
         long now = JoH.tsl();
         if(now - s_lastReceiveTimestamp > 10*1000) {
@@ -56,7 +58,7 @@ public class Tomato {
         s_lastReceiveTimestamp = now;
         if (buffer == null) {
             Log.e(TAG, "null buffer passed to decodeTomatoPacket");
-            return retArray;
+            return reply;
         } 
         if (s_state == TOMATO_STATES.REQUEST_DATA_SENT) {
             if(buffer.length == 1 && buffer[0] == 0x32) {
@@ -65,24 +67,25 @@ public class Tomato {
                 ByteBuffer allowNewSensor = ByteBuffer.allocate(2);
                 allowNewSensor.put(0, (byte) 0xD3);
                 allowNewSensor.put(1, (byte) 0x01);
-                retArray.add(allowNewSensor);
+                reply.send.add(allowNewSensor);
                 
                 // For debug, make it send data every minute (did not work...)
                 ByteBuffer newFreqMessage = ByteBuffer.allocate(2);
                 newFreqMessage.put(0, (byte) 0xD1);
                 newFreqMessage.put(1, (byte) 0x01);
-                retArray.add(newFreqMessage);
+                reply.send.add(newFreqMessage);
                 
                 //command to start reading
                 ByteBuffer ackMessage = ByteBuffer.allocate(1);
                 ackMessage.put(0, (byte) 0xF0);
-                retArray.add(ackMessage);
-                return retArray;
+                reply.send.add(ackMessage);
+                return reply;
             }
             
             if(buffer.length == 1 && buffer[0] == 0x34) {
                 Log.e(TAG, "No sensor has been found");
-                return retArray;
+                reply.setError_message("No sensor found");
+              return reply;
             }
             
             // 18 is the expected header size
@@ -95,22 +98,38 @@ public class Tomato {
                 InitBuffer(expectedSize);
                 addData(buffer);
                 s_state = TOMATO_STATES.RECIEVING_DATA;
-                return retArray;
+                return reply;
                 
+            } else {
+                Log.d(TAG,"Unknown initial packet makeup received");
+                return reply;
             }
         }
         
         if (s_state == TOMATO_STATES.RECIEVING_DATA) {
             //Log.e(TAG, "received more data s_acumulatedSize = " + s_acumulatedSize + " current buffer size " + buffer.length);
-            addData(buffer);
-            
+            try {
+                addData(buffer);
+            } catch (RuntimeException e) {
+                // if the checksum failed lets ask for the data set again but not more than once per minute
+                if (e.getMessage().equals(CHECKSUM_FAILED)) {
+                   if (JoH.ratelimit("tomato-full-retry",60)
+                           || JoH.ratelimit("tomato-full-retry2",60)) {
+                       reply.getSend().clear();
+                       reply.getSend().addAll(Tomato.resetTomatoState());
+                       reply.setDelay(8000);
+                       reply.setError_message("Checksum failed - retrying");
+                       Log.d(TAG,"Asking for retry of data");
+                   }
+                } else throw e;
+            }
 
-            return retArray;
+            return reply;
         }
         
-        Log.wtf(TAG, "Very strange, In an unexpeted state " + s_state);
+        Log.wtf(TAG, "Very strange, In an unexpected state " + s_state);
         
-        return retArray;
+        return reply;
     }
     
     static void addData(byte[] buffer) {
@@ -139,15 +158,15 @@ public class Tomato {
         s_recviedEnoughData = true;
         
         boolean checksum_ok = JoH.LibreCrc(data);
-        Log.e(TAG, "We have all the data that we need checksum_ok = " + checksum_ok + HexDump.dumpHexString(data));
+        Log.e(TAG, "We have all the data that we need " + s_acumulatedSize + " checksum_ok = " + checksum_ok + HexDump.dumpHexString(data));
         
         if(!checksum_ok) {
-            return;
+            throw new RuntimeException(CHECKSUM_FAILED);
         }
         
         PersistentStore.setString("Tomatobattery", Integer.toString(s_full_data[13]));
-        PersistentStore.setString("TomatoHArdware",HexDump.toHexString(s_full_data,14,2));
-        PersistentStore.setString("TomatoFirmware",HexDump.toHexString(s_full_data,16,2));
+        PersistentStore.setString("TomatoHArdware",HexDump.toHexString(s_full_data,16,2));
+        PersistentStore.setString("TomatoFirmware",HexDump.toHexString(s_full_data,14,2));
 
         NFCReaderX.HandleGoodReading("tomato", data);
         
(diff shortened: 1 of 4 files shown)


tzachi-dar:
It seems that on the file DexCollextion.java many functions have been moved around, although not being changed.

For example:

 private static String getDefaultPin() {
private static boolean shouldServiceRun() 
public static boolean isCollecting() 
private static void status(String msg) 
private static void error(String msg) 
private static String getStateStr(int mConnectionState) 
private static Integer convertSrc(final String Src) 
private static int getSrcValue(char ch) 
// Status for Watchface
public static boolean isRunning() 
public static DataMap getWatchStatus() 
public static void setWatchStatus(DataMap dataMap) 
public static String getBestLimitterHardwareName() 
    // data for MegaStatus
    public static List<StatusItem> megaStatus() 

There are probably many other places in code that have moved.
Can you split this to two parts of the code:
1) big one, only moving and arranging, with no real change at all.
2) The real change in code.

Otherwise review will be very difficult, and it will remain very difficult for anyone looking at the code when looking for a change in this file.

Or is the real change limited to tomato code only?

I know that it is a big request, but I can not see how the file can be reviewed otherwise.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
```diff
@@ -17,33 +18,34 @@
...
     
-    static TOMATO_STATES s_state;
-    
-    static long s_lastReceiveTimestamp;
-    private static byte[] s_full_data = null;
-    private static int s_acumulatedSize = 0;
-    private static boolean s_recviedEnoughData;
-    
-    
+    static volatile TOMATO_STATES s_state;
```
tzachi-dar: When I wrote this file, I was under the assumption that the class will not be accessed by two threads simultaneously. 
If this is not correct, we need to add locks on the function of this class.
Can we trust the dexcollectionservice to sync this readings?

app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
```diff
@@ -139,15 +158,15 @@ static void AreWeDone() {
         s_recviedEnoughData = true;
         
         boolean checksum_ok = JoH.LibreCrc(data);
-        Log.e(TAG, "We have all the data that we need checksum_ok = " + checksum_ok + HexDump.dumpHexString(data));
+        Log.e(TAG, "We have all the data that we need " + s_acumulatedSize + " checksum_ok = " + checksum_ok + HexDump.dumpHexString(data));
         
         if(!checksum_ok) {
-            return;
+            throw new RuntimeException(CHECKSUM_FAILED);
```
tzachi-dar: IMHO, using exception makes the code harder to understand (especially when this are not real exception, but something that might happen and can create a retry).
I guess that in this case, usage of it is correct, but most xDrip code is not using exceptions but return values.
Up to you if to change it or not.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BridgeResponse.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.Data;
+
+/**
+ * Created by jamorham on 16/03/2018.
+ */
+
+@Data
+public class BridgeResponse {
+
+    public final LinkedList<ByteBuffer> send;
```
tzachi-dar: Please make the send private and manipulate it through setters/getters.
Otherwise it is hard to say when this class is changing.

jamorham: The reason it is done like this is so you can do a cleaner syntax of:

` reply.send.add(allowNewSensor);`

instead of having to do:

`reply.getSend().add(allowNewSensor)`

I think it is easier to read like that myself.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
jamorham: I used an exception here so that we can catch the result from `AreWeDone()` in the grandparent `decodeTomatoPacket()` rather than using return codes or something like that. I considered creating an explicit Exception for the purpose but thought it best not to over-complicate things at this stage where we're only dealing with a single error type.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
jamorham: I don't know whether there is a situation where we might receive inbound bluetooth messages on multiple threads simultaneously, but even if they are different threads not simultaneously then I think there are possibilities for a state variable not declared volatile to cache a wrong value. 

I don't think there is a performance issue with the processing of this variable so marking it volatile seemed the safest way to deal with this. Its especially hard to simulate because different android implementations can handle bluetooth in different ways and have their own implementations. For example Samsung is often more picky about threads.

jamorham:
Its really annoying that AS3 has decided to reorganize the code. It isn't an option I selected and I had no idea it had done it until it came to time to commit. I will try to highlight the changes to see if that can make things clear.

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -794,39 +476,589 @@ public void onCharacteristicChanged(BluetoothGatt gatt, BluetoothGattCharacteris
...
-                    static_last_hexdump = hexdump;
-                }
-                Log.i(TAG, "onCharacteristicChanged entered " + hexdump);
+
                 if (data != null && data.length > 0) {
                     setSerialDataToTransmitterRawData(data, data.length);
+
+                    final String hexdump = HexDump.dumpHexString(data);
+                    //if (!hexdump.contains("0x00000000 00      ")) {
+                    if (data.length > 1 || data[0] != 0x00) {
```
jamorham: Here optimized to not search the string but look at the bytes for the same effect

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -794,39 +476,589 @@ public void onCharacteristicChanged(BluetoothGatt gatt, BluetoothGattCharacteris
...
+                        static_last_hexdump = hexdump;
+                    }
+                    if (d) Log.i(TAG, "onCharacteristicChanged entered " + hexdump);
+
                 }
                 lastdata = data;
-                setFailoverTimer(); // restart the countdown
-                // intentionally left hanging wakelock for 5 seconds after we receive something
-                final PowerManager.WakeLock wakeLock2 = JoH.getWakeLock("DexCollectionLinger", 5000);
+
```
jamorham: Here the creation of the `FailoverTimer()` is set by `Inevitable.task()` so only on the last received bluetooth packet where there is then 1 second of following silence.

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -794,39 +476,589 @@ public void onCharacteristicChanged(BluetoothGatt gatt, BluetoothGattCharacteris
...
+        l.add(new StatusItem("Phone Service State", lastState + (forced_wear ? " (Watch Forced)" : "")));
+
+        if (lastError != null) {
+            l.add(new StatusItem("Last Error", lastError, StatusItem.Highlight.NOTICE, "long-press", () -> lastError = null));
+        }
+
+        l.add(new StatusItem("Bluetooth Device", JoH.ucFirst(getStateStr(mStaticState))));
+
+        if (mStaticState == STATE_CONNECTING) {
```
jamorham: Here we show details about how long the current attempt to connect has been going on for

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -794,39 +476,589 @@ public void onCharacteristicChanged(BluetoothGatt gatt, BluetoothGattCharacteris
...
+        }
+
+        // blueReader
+        if (blueReader.isblueReader()) {
+            l.add(new StatusItem("blueReader Battery", Pref.getInt("bridge_battery", 0) + "%"));
+            l.add(new StatusItem("blueReader rest days", PersistentStore.getString("bridge_battery_days")));
+            l.add(new StatusItem("blueReader Firmware", PersistentStore.getString("blueReaderFirmware")));
+        }
+
+        if (Tomato.isTomato()) {
```
jamorham: Capitalized the Tomato name

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -967,52 +1179,71 @@ public synchronized boolean connect(final String address) {
         }
         Log.i(TAG, "connect: Trying to create a new connection.");
         setRetryTimer();
-        mBluetoothGatt = device.connectGatt(getApplicationContext(), true, mGattCallback);
+        mBluetoothGatt = device.connectGatt(getApplicationContext(), false, mGattCallback);
```
jamorham: Here we manage the connection manually

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -967,52 +1179,71 @@ public synchronized boolean connect(final String address) {
...
         mConnectionState = STATE_CONNECTING;
+        last_connect_request = JoH.tsl();
         return true;
     }
 
-    public void close() {
+    public synchronized void close() {
```
jamorham: add synchronized to close method because it changes state

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -967,52 +1179,71 @@ public synchronized boolean connect(final String address) {
...
             if (reply != null) {
                 Log.d(TAG, "Sending reply message from blueReader decoder");
                 sendBtMessage(reply);
             }
-        } else  if (Tomato.isTomato()) {
-            ArrayList<ByteBuffer> buffers = Tomato.decodeTomatoPacket(buffer, len);
-            for (ByteBuffer byteBuffer : buffers ) {
-                Log.d(TAG, "Sending reply message from tomato decoder");
-                sendBtMessage(byteBuffer);
+        } else if (Tomato.isTomato()) {
```
jamorham: Here converted to use `BridgeResponse` instead of array list of bytebuffer and able to send reply with delay to avoid cross-talk when Tomato is sending us a stream of data.

Error messages passed to the status UI elements.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
```diff
@@ -139,15 +158,15 @@ static void AreWeDone() {
...
         }
         
         PersistentStore.setString("Tomatobattery", Integer.toString(s_full_data[13]));
-        PersistentStore.setString("TomatoHArdware",HexDump.toHexString(s_full_data,14,2));
-        PersistentStore.setString("TomatoFirmware",HexDump.toHexString(s_full_data,16,2));
+        PersistentStore.setString("TomatoHArdware",HexDump.toHexString(s_full_data,16,2));
```
jamorham: order of firmware/hardware swapped but this has already been discussed, but just drawing attention to the change.

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -794,39 +476,589 @@ public void onCharacteristicChanged(BluetoothGatt gatt, BluetoothGattCharacteris
...
+                        status("Connecting" + (Home.get_engineering_mode() ? ": " + deviceAddress : ""));
+                        connect(deviceAddress);
+                        mStaticState = mConnectionState;
+                        return;
+                    }
+                } catch (IllegalArgumentException e) {
+                    Log.e(TAG, "IllegalArgumentException: " + e);
+                }
+            }
+        } else if (mConnectionState == STATE_CONNECTING) {
```
jamorham: Here we manually manage the connecting state and shut it down if we think its been trying for too long instead of letting the framework manage this aspect.

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -794,39 +476,589 @@ public void onCharacteristicChanged(BluetoothGatt gatt, BluetoothGattCharacteris
...
+            mStaticState = mConnectionState;
+            if (JoH.msSince(last_connect_request) > Constants.SECOND_IN_MS * 30) {
+                Log.i(TAG, "Connecting for too long, shutting down");
+                retry_backoff = 0;
```
jamorham: Also the backoff is reset to 0, I may want to change this in the future for battery saving when the phone and bluetooth device are apart for long periods but that is sort of an edge case and the intention here is maximum reliability.

jamorham:
I've added single line comments to the changed logic sections, there are not that many. Hopefully that makes it clearer. I don't have a copy of the code before the AS rearrange. I will check settings to see if that can be disabled, it definitely never did that with AS2 but its not the first time I've seen it happen when only using reformat on a small section.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BridgeResponse.java
tzachi-dar: I think that the best way to do it is to add a function add on the BridgeResponse 
in that case, adding data to it will only be done inside it's code. Otherwise it is not possible to know  who is manipulating it's data.

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
tzachi-dar: As far as I can see retry_backoff is never used in code. (Or did I miss it?)

tzachi-dar:
So: beside for the changes that I wrote, I have no further comments.

This code changes a few things that are tomato related, and I think that they are fine.
It also changes some very small parts on the code that will affect all devices running bluetooth with android.
For example:  changing autoConnect on connectGatt().
I'm afraid that I don't understand android bluetooth to tell if this will not hurt any android phone in the world. (but for being on this forum for many years, I know android BT code is a piece of Sxxx )

As for tomato, since there are only very few people using it, I'm 100% fine with this change.
As for other phones / sensors, I believe that this kind of change might cause complaints from some people, and in that case it will be very hard to know for sure what has changed.

If the big changes here have been caused by AS2, I would have start from letting it do it's changes on an unmodified version of DexCollectionService, and commit that first. In that case (assuming it has some persistent logic) your changes will be very visible.

All that said, you have much more experience with Dealing with android BT, so I guess it is LGTM from my side.
I'll start running with it tonight, and hope for the best.

tzachi-dar:
One last comment:
The code on onCharacteristicChanged() used to take the wakelock for 5 seconds, after each packet.
It does not do it anymore, although I understand that there is assumption here that after a second the timer will fire and take the lock.
Just making sure this is by design.

tzachi-dar:
This has run well through the night and continuing to run well.
One thing that seems to have changed is that in the past every time a new connection was made, it used to start an immediate data transfer. This was very important if the device goes out of range.
Now it seems that I have to wait 5 minutes (this is from my limited experience with new code).

jamorham:
@tzachi-dar thanks for your testing and comments. It should try to get the readings right after a connection - eg where it has disconnected and come back in to range. That was one of the areas I particularly wanted to make sure worked so I'm surprised if you're reporting it not doing that. In the tests here it looks solid in that aspect.

Regarding your comment about the changes to `onCharacteristicChanged` the `Inevitable.task()` will hold the wake lock during operation which will be at least one second and before it releases its wakelock it will enact the `DexCollectionLinger` wakelock so the amount of wake-locking should be the same or with at least an extra second (or more if data keeps streaming in)

Regarding AS3 formatting I have found that by pressing CTRL + SHIFT + ALT + L it is possible to then uncheck the `Rearrange code` feature. As far as I know this is a new default which has been added in AS3. I only ever intended to reformat selected lines which I had modified and never the entire file because I know how difficult that makes following change differences!

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BridgeResponse.java
jamorham: Yes this obeys object orientated principles. noted.

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
jamorham: Its used by the retry timer

jamorham:
I will merge this for wider testing and request users to observe any impact. Of course those reports are not always scientific!

PR #358, opened 2018-03-19
Common test config

ahaarrestad:
@jamorham - Extracted Robolectric test config to ease config change between CI and AndroidStudio.


jamorham:
Ok this is good, what about the ones with no manifest like:

```
@RunWith(RobolectricTestRunner.class)
@Config(manifest = Config.NONE)
```

To avoid confusion should there be another class that can be extended like `RoboelectricTestNoConfig` so there is always a similar pattern to enact?

I also assume its not possible to detect the misconfigured android studio? It was really annoying that they introduced this problem with AS3, I also don't know if it will get "fixed" in the future...

ahaarrestad:
I will addnother file for NoConfig. 

Agree that it is a bit anoying, and upgrading Roboletric seems to work, but that opens another can of worms. Will try to solve that further down the road. This is more of a bandaid to ease config and avoid submitting the wrong files.

ahaarrestad:
@jamorham Updated setup. Will in a later pull-request try to once again upgrade Robolectric and uptade config to http://robolectric.org/getting-started/ and see if that can work.

jamorham:
Mostly I am only really concerned as to whether the CI tests pass as those are the ones used most frequently. Anyone who wants to modify or run the tests locally in AS can adjust the environment variable.  If it becomes a common problem then I can create some better documentation for configuring AS3 other than line you've already seen.

Are you saying that this path bug doesn't occur with later versions of RoboElectric? Because as far as I was aware it was android studio 3 which introduced this issue and it isn't caused by roboelectric?

ahaarrestad:
I have yet to make it work, but later versions of Robolectric seems to have a different setup and I expect it to work, but I do at the moment run into various problems when upgrading and have not managed to trace them all down yet.

jamorham:
Ok thanks for the No Config addition. 

Shall I merge this now then?

ahaarrestad:
@jamorham please do.

PR #368, opened 2018-03-26
Tests for Initial Calibration

ahaarrestad:
Added test for Initial calibration, verifying slope and intercept.


ahaarrestad:
@jamorham Added one more test

PR #372, opened 2018-03-27
Fixed various performance errors reporte by codacy

ahaarrestad:
* All refactoring completed by AndroidStudio
* Cleaned up imports in affected files
* Fixed some issues reported by AndroidStudio such as
  simplifications and lambda suggestions

@jamorham No functional changes


wear/src/main/java/com/eveningoutpost/dexdrip/Simulation.java
```diff
@@ -382,14 +375,14 @@ private void promptSpeechInput() {
     }
     private void handleWordPair() {
         boolean preserve = false;
-        if ((thisnumber == -1) || (thisword == "")) return;
+        if ((thisnumber == -1) || (Objects.equals(thisword, ""))) return;
```
jamorham: This doesn't look correct. Should it just be `thisword.equals("")` ? 

Here the compiler is already automatically fixing it to that to avoid the incorrect syntax of object instance comparison.

wear/src/main/java/com/eveningoutpost/dexdrip/Models/UserNotification.java
```diff
@@ -179,27 +179,36 @@ public static UserNotification create(String message, String type, long timestam
...
-        } else if (type == "bg_rise_alert") {
-            userNotification.bg_rise_alert = true;
-        } else if (type == "bg_fall_alert") {
-            userNotification.bg_fall_alert = true;
-        } else {
-            Log.d(TAG,"Saving workaround for: "+type+" "+message);
-            Pref.setString("UserNotification:timestamp:" + type, JoH.qs(timestamp));
-            Pref.setString("UserNotification:message:" + type, message);
-           return null;
+        switch (type) {
```
jamorham: Are you certain all of these have been transcribed correctly? Was this done with an automated tool?

wear/src/main/java/com/eveningoutpost/dexdrip/Simulation.java
ahaarrestad: Objects.equals(a,b) is one step up from a.equals(b) since it is null-safe. other possible variants are ["".equals(thisword)] or StringUtils.isEmpty(thisword)

wear/src/main/java/com/eveningoutpost/dexdrip/Models/UserNotification.java
ahaarrestad: This was done automatically by AndroidStudio.

wear/src/main/java/com/eveningoutpost/dexdrip/Simulation.java
dabear: See line 70: String thisword = ""; and is never assigned to null. 
So thisword will always be initialized and never null. This change is strictly unneccessary?

ahaarrestad: using "==" to compare string is not the correct way of comparing strings. I was guided to this line by codacy which suggested fixing it. If you really don't approve of it, I can revert the change, but given previous comments that the code has been build by a lot of different people over time, and the quality is variable, I find that fixing some of the issues standard tools like codacy suggests is a good thing.

wear/src/main/java/com/eveningoutpost/dexdrip/Models/UserNotification.java
ahaarrestad: @jamorham Added test for this class in pull request 376

wear/src/main/java/com/eveningoutpost/dexdrip/Simulation.java
ahaarrestad: @jamorham Rewrote to "".equals(thisword)

ahaarrestad: @jamorham, @dabear: regarding == vs .equals, there is a difference in java. Some of it is explained at stackoverflow: https://stackoverflow.com/questions/513832/how-do-i-compare-strings-in-java 
The most important difference is that == is comparing instances and .equals is comparing value. The compiler is not doing any optimization to the code, but the current way string is implemented is by using an internal cache in String to try to keep one instance of each string. The correct way to compare strings is to use either String.equels() or Objects.equals().

ahaarrestad:
If you want som fun / interresting reading about Strings in java, please check out https://www.javaspecialists.eu/archive/Issue014.html :-)

CC: @dabear @jamorham

ahaarrestad:
@jamorham Pulled from upstream to include test for UserNotification. The test has now run green both before and after this change.

PR #376, opened 2018-03-29
Added tests for UserNotification in both app and wear

ahaarrestad:
@jamorham Added tests for UserNotificatoin in both app wear. 

* No functional changes, only added tests
* Found a "feature" shown in line 360 of the test when trying to use this class to snooze non-legacy notifications.


ahaarrestad:
@jamorham This is a standalone pull-request with a test. No other dependencies. Should be OK to submit.

PR #378, opened 2018-04-02
Tests for BlueReader model

ahaarrestad:
* Added tests for BlueReader model class
* Fixed "assert" error in BgReading (app and wear)

CC: @MasterPlexus  @jamorham @SandraK82

### DIFF
--- a/app/src/test/java/com/eveningoutpost/dexdrip/Models/BlueReaderTest.java
+++ b/app/src/test/java/com/eveningoutpost/dexdrip/Models/BlueReaderTest.java
@@ -24,7 +24,7 @@ import static com.google.common.truth.Truth.assertWithMessage;
 
 /**
  * Tests for {@link blueReader}. Verifying commands and bg readings.
- *
+ * <p>
  * Got great input from @SandraK82 on how the BlueReader works.
  *
  * @author Asbjørn Aarrestad, asbjorn@aarrestad.com - 2018.04
@@ -387,7 +387,11 @@ public class BlueReaderTest extends RobolectricTestWithConfig {
         // Setup
         String message = "150000 4300";
         Sensor mockSensor = createMockSensor();
-        createCalibration(mockSensor);
+
+        addMockBgReading(150, 15, mockSensor);
+        addMockBgReading(150, 10, mockSensor);
+        addMockBgReading(150, 5, mockSensor);
+        Calibration.initialCalibration(150, 150, RuntimeEnvironment.application.getApplicationContext());
 
         // Act
         byte[] bytes = blueReader.decodeblueReaderPacket(message.getBytes(), message.length());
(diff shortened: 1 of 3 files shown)


app/src/test/java/com/eveningoutpost/dexdrip/Models/BlueReaderTest.java
```diff
@@ -0,0 +1,556 @@
...
+        byte[] reply = blueReader.decodeblueReaderPacket(buffer, -1);
+
+        // :: Verify
+        assertThat(reply).isNull();
+    }
+
+    @Test
+    public void decodeBlueReaderPacket_IDR_noBgReading() {
+        // :: Setup
+        String inputString = "IDR0blue131-a1";
```
MasterPlexus: not sure what should be tested here, but normaly the inputstring should be with a '|' (IDR0|blue131-a1).

app/src/test/java/com/eveningoutpost/dexdrip/Models/BlueReaderTest.java
```diff
@@ -0,0 +1,556 @@
...
+        // :: Verify
+        assertThat(reply).isNotNull();
+        assertThat(new String(reply)).isEqualTo("l");
+        assertThat(PersistentStore.getString("blueReaderFirmware")).isEqualTo(inputString);
+    }
+
+    @Test
+    public void decodeBlueReaderPacket_IDR_OldBgReading() {
+        // :: Setup
+        String inputString = "IDR0blue131-a1";
```
MasterPlexus: here the same: not sure what should be tested here, but normaly the inputstring should be with a '|' (IDR0|blue131-a1).

app/src/test/java/com/eveningoutpost/dexdrip/Models/BlueReaderTest.java
```diff
@@ -0,0 +1,556 @@
...
+        // :: Verify
+        assertThat(reply).isNotNull();
+        assertThat(new String(reply)).isEqualTo("h");
+        assertThat(getLogs()).contains("E/blueReader: Found blueReader in a ugly State (1/5), send hibernate to reset! If this does not help in the next 5 Minutes, then turn the bluereader manually off and on!");
+    }
+
+    /**
+     * Found a bug in the code which increase hibernate count twice pr. invocation instead of once.
+     * This test shows how it works. If This test fails because the bug has been fixed, rewrite the
+     * test.
```
MasterPlexus: Thanks for the Hint. I seen that also in my last codereview, and will take care of that also for the Test within my next pull request. Also in future the count will raise only up to 3 as this is enough for checking the ugly state.

ahaarrestad: Only wanted to write the test in this iteration, but it is easier to understand and remember when adding some documentation :-)

MasterPlexus:
Looks fine for me, except the Firmware-String as commented. But anyhow, as the String is not seperated at the moment the test will have no influance. 
As I have no glue on the Test-Stuff, thanks for taking care of them! As I see now how they are implemented, I will try to adapt them in further pull requests if I change something in the bluereader class.

jamorham:
I can't tell from the comments above whether there will be further changes made to this or whether it is ready for merge?

ahaarrestad:
I believe the code can be merged. The string in question is a mock reply and could be anything. Changing it will not change any functionality in the test. 

I believe this can be merged.

CC: @jamorham

MasterPlexus:
From my side the Tests look ok. The changes wich will be need will comes in future when I change the Class. So I think It is finde to implement.

PR #379, opened 2018-04-03
added Samsung Galaxy S9+

PoweRGbg:


jamorham:
Does this mean you have tested this successfully? Was this with OB1 collector or the original one?

PoweRGbg:
Yes - I'm using it for more than 12 days and capture rate is above 97%. It's with the OB1 collector

jamorham:
If it is with OB1 you should add OB1 to the mode column, see the other examples towards the end.

jamorham:
I assume this is no longer valid in reference to: https://github.com/NightscoutFoundation/xDrip/issues/663

PoweRGbg:
My S9+ is not updated to the latest version, so no connection issues

Navid200:
@PoweRGbg Hi,
My understanding is that all Android phones can be used to run xDrip unless their operating system is extremely old (before Android 4.4?).  That is what I tell everyone on the facebook group.  
There was a time we maintained these files of phones that were compatible with xDrip.  But, I don't believe we do that any longer.

Thanks

PoweRGbg:
This PR is 3 years old :) closing it

PR #387, opened 2018-04-10
New test config

ahaarrestad:
Updated test config for unit tests. Now running with same config in both Travis and AndroidStudio

@jamorham Content of commit is fixing of junit config and some minor cleanup of tests
@MasterPlexus  Your tests should work when this pull-request is merged.


jamorham:
There is no manifest any more? Does that cause any problems with active android?

ahaarrestad:
@jamorham ActiveAndroid is set up using the TestingApplication and annotations using Robolectric. If ActiveAndroid did not work, very few of the tests would be green. This simplifies test-writing even more, since you can use the same config both for android studio and travis.

jamorham:
Yes it looks good. The only reason for needing two different test setups is because of a bug in android studio 3.0 that doesn't set the path properly. It is fine in 2.x and might get fixed in the future but this looks neater.

Can you also explain the removal of the `throws Exception` I thought that was supposed to be part of the test specification?

ahaarrestad:
@jamorham throws Exception is only needed if you actually throw an exception, and if you do, you should always throw the exact exception you are throwing. 

When no checked exception is thrown, you do not need to specify throws. If it is needed, you will get a compile-error :-)

PR #390, opened 2018-04-11
Updated test-config

ahaarrestad:
* Updated javadoc in RobolectricTestWithConfig
* Copied app test-config to wear


ahaarrestad:
@jamorham updated test config as promised.

ahaarrestad:
@jamorham When you have time, this should be quite easy to merge. Test-config is updated as we discussed a couple of weeks ago.

jamorham:
This looks very familiar. Is this a mirror of the test config changes for the wear module?

ahaarrestad:
@jamorham yes it is

PR #395, opened 2018-04-14
Tzachi libre wifi

tzachi-dar:
This are the changes that are needed to allow xDrip to read libre data from wifi readers, either directly or using a mongodb.

Basic tests passes, but will need some more experience to make sure it is working well.


tzachi-dar:
This was replaced by: https://github.com/NightscoutFoundation/xDrip/pull/418

PR #397, opened 2018-04-17
Removed empty statements (;) from files

ahaarrestad:
All touched files are also just formatted and imports are organized

CC: @jamorham

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/ImportDatabaseActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/ImportDatabaseActivity.java
@@ -81,8 +81,8 @@ public class ImportDatabaseActivity extends ListActivityWithMenu {
     }
 
     private void showWarningAndInstructions() {
-        LayoutInflater inflater= LayoutInflater.from(this);
-        View view=inflater.inflate(R.layout.import_db_warning, null);
+        LayoutInflater inflater = LayoutInflater.from(this);
+        View view = inflater.inflate(R.layout.import_db_warning, null);
         AlertDialog.Builder alertDialog = new AlertDialog.Builder(this);
         alertDialog.setTitle("Import Instructions");
         alertDialog.setView(view);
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/ImportDatabaseActivity.java
```diff
@@ -358,8 +359,10 @@ public void run() {
             });
 
             String result = DatabaseUtil.loadSql(xdrip.getAppContext(), dbFile.getAbsolutePath());
-            if (delete_file != null) delete_file.delete();
-            statusDialog.dismiss();;
```
ahaarrestad: example of "empty statement" which has been removed.

jamorham:
I'm very grateful for the work you are doing regarding code quality and especially testing, but here is my criticism and explanation regarding this PR. 

Generally speaking in a PR we only want the code which is modified to have its formatting updated. This makes any PR review easier and lessens the impact of a commit on anyone else who may have a fork open with other changes. Changing the white space also opens things up to unnecessary merge conflicts which slows development and reduces motivation of external developers. 

There are other potential pitfalls, for example in line 88 of `AlertType` the comment is moved to the line below where it was placed. In this instance it sits between an annotation and variable but in other situations it could mean that a reader would interpret the comment regarding the line below rather than the statement to the left.  This isn't hugely serious, but it is a degradation of the code as a result of auto-formatting. As a reviewer I might notice something like this and request a change. All of that takes up development bandwidth and if I miss something like that then it has the potential to sow confusion at some point in the future.

It is quite common for people to use `reformat code` on the whole file and where this happens as part of the development process (for developer sanity) - it isn't always desirable (on non-modified code) but it happens.. and a PR is weighed up on the benefit it gives vs the cost of integrating it. 

Formatted code looks great but there isn't a particular reason to reformat files unless there are other code changes.

Stylistically, changing `if` statements to always use braces can make the code more robust against accidental logic changes and can improve readability, but it isn't always the case. Code can be more compact and thus the intent of functions clearer when simple `if` statements only take a single line, for example `if (param == null) return` etc. I don't enforce this restriction on new code submitted so I don't really think it is a change worth making unless the code is already being edited.

The title of this PR is for removing empty statements, basically duplicate `;` characters. As far as I know these will have no impact on the final code as they will not be compiled in to byte-code and have very little impact as far as code readability goes.  Changing these would be a very low priority and not really worth creating a PR for in my opinion.

Taking all of the factors described above together I can't really rate this PR positively and wouldn't plan to merge it. I don't feel the changes are particularly desirable and the formatting changes have the potential to introduce an unnecessary burden on anyone who may have open modifications to those files at the moment.

I've been very happy with what you've done on the testing side and almost no-one is working on that aspect of things so all the improvements you've made have been great. 

Code quality is important (and applied more strictly to new code submitted) but more important is the functionality and user experience which is where I try to focus efforts generally.

ahaarrestad:
I will remove til pr and keep changes smaller in the future.

PR #418, opened 2018-05-03
Tzachi libre wifi 2

tzachi-dar:
This replaces tzachi libre wifi 1 that worked very well with raspberry pi receivers.

Merging this with current code worked well, but there is one thing that somehow frightening me:
As part of the changes to to tomato.java I have removed one buffer that was sent in tomato.resetTomatoState().
Theoretically, this should not change anything, but as android BT code is so sensitive, I wonder if this does not start a new testing cycle for every phone.
(my fear is that since now there is a 150 ms sleep between each two buffers, a change in number of buffers changes timing ). So jamorham@ if you think that this should be reverted let me know.

In any case I'm starting to check the new code.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -2251,9 +2251,12 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
         // port this lot to DexCollectionType to avoid multiple lookups of the same preference
         boolean isDexbridgeWixel = CollectionServiceStarter.isDexBridgeOrWifiandDexBridge();
         boolean isWifiBluetoothWixel = CollectionServiceStarter.isWifiandBTWixel(getApplicationContext());
+        boolean isWifiandBTLibre = CollectionServiceStarter.isWifiandBTLibre(getApplicationContext());
+
         isBTShare = CollectionServiceStarter.isBTShare(getApplicationContext());
         isG5Share = CollectionServiceStarter.isBTG5(getApplicationContext());
         boolean isWifiWixel = CollectionServiceStarter.isWifiWixel(getApplicationContext());
+        boolean isWifiLibre = CollectionServiceStarter.isWifiLibre(getApplicationContext());
         alreadyDisplayedBgInfoCommon = false; // reset flag
         if (isBTShare) {
             updateCurrentBgInfoForBtShare(notificationText);
@@ -2261,10 +2264,10 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
         if (isG5Share) {
             updateCurrentBgInfoCommon(collector, notificationText);
         }
-        if (isBTWixel || isDexbridgeWixel || isWifiBluetoothWixel) {
+        if (isBTWixel || isDexbridgeWixel || isWifiBluetoothWixel || isWifiandBTLibre) {
             updateCurrentBgInfoForBtBasedWixel(collector, notificationText);
         }
-        if (isWifiWixel || isWifiBluetoothWixel || collector.equals(DexCollectionType.Mock)) {
+        if (isWifiWixel || isWifiBluetoothWixel || isWifiandBTLibre || isWifiLibre || collector.equals(DexCollectionType.Mock)) {
             updateCurrentBgInfoForWifiWixel(collector, notificationText);
         } else if (is_follower || collector.equals(DexCollectionType.NSEmulator) ||DexCollectionType.isLibreOOPAlgorithm(collector)) {
             displayCurrentInfo();
@@ -2480,6 +2483,9 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
         if(DexCollectionType.isLibreOOPAlgorithm(collector)) {
         	// Rest of this function deals with initial calibration. Since we currently don't have a way to calibrate,
         	// And even once we will have, there is probably no need to force a calibration at start of sensor use.
+            displayCurrentInfo();
+            // JamorHam, should I put here something like:
+            // ?? if (screen_forced_on)  dontKeepScreenOn();
         	return;
         }
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
@@ -48,7 +48,7 @@ public class Tomato {
         final BridgeResponse reply = new BridgeResponse();
         // Check time, probably need to start on sending
         long now = JoH.tsl();
-        if(now - s_lastReceiveTimestamp > 10*1000) {
+        if(now - s_lastReceiveTimestamp > 3*1000) {
             // We did not receive data in 10 seconds, moving to init state again
             Log.e(TAG, "Recieved a buffer after " + (now - s_lastReceiveTimestamp) / 1000 +  " seconds, starting again. "+
             "already acumulated " + s_acumulatedSize + " bytes.");
@@ -102,7 +102,7 @@ public class Tomato {
                 
             } else {
                 if (JoH.quietratelimit("unknown-initial-packet", 1)) {
-                    Log.d(TAG, "Unknown initial packet makeup received");
+                    Log.d(TAG,"Unknown initial packet makeup received" + HexDump.dumpHexString(buffer));
                 }
                 return reply;
             }
@@ -159,7 +159,8 @@ public class Tomato {
         byte[] data = Arrays.copyOfRange(s_full_data, TOMATO_HEADER_LENGTH, TOMATO_HEADER_LENGTH+344);
         s_recviedEnoughData = true;
         
-        boolean checksum_ok = NFCReaderX.HandleGoodReading("tomato", data);
+        long now = JoH.tsl();
+        boolean checksum_ok = NFCReaderX.HandleGoodReading("tomato", data, now);
         Log.e(TAG, "We have all the data that we need " + s_acumulatedSize + " checksum_ok = " + checksum_ok + HexDump.dumpHexString(data));
 
         if(!checksum_ok) {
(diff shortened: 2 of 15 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
```diff
@@ -48,7 +48,7 @@ public static BridgeResponse decodeTomatoPacket(byte[] buffer, int len) {
         final BridgeResponse reply = new BridgeResponse();
         // Check time, probably need to start on sending
         long now = JoH.tsl();
-        if(now - s_lastReceiveTimestamp > 10*1000) {
+        if(now - s_lastReceiveTimestamp > 3*1000) {
             // We did not receive data in 10 seconds, moving to init state again
```
jamorham: Are we sure this will never take more than 3 seconds even in really bad RF situations?

Also the comment below still references 10 seconds, should be updated to something more generic

app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
```diff
@@ -212,12 +213,6 @@ static void InitBuffer(int expectedSize) {
 
         s_state = TOMATO_STATES.REQUEST_DATA_SENT;
 
-        // For debug, make it send data every minute (ERROR - We fail to send this message... needs more work,
-        // not sure it works at all)
-        ByteBuffer newFreqMessage = ByteBuffer.allocate(2);
-        newFreqMessage.put(0, (byte) 0xD1);
-        newFreqMessage.put(1, (byte) 0x05);
-        ret.add(newFreqMessage);
```
jamorham: What is the effect of removing this? I thought this enabled the required scheduler. This seems unrelated to adding Wifi support?

app/src/main/java/com/eveningoutpost/dexdrip/Services/LibreWifiReader.java
```diff
@@ -0,0 +1,620 @@
...
+    static void statusLog(String key, String msg) {
+        statusLog(key, msg, 0); // default no time since
+    }
+
+    // timestamp or 0 = don't use or -1 = now
+    static void statusLog(String key, String msg, long time) {
+        hostStatus.put(key, msg);
+        hostStatusTime.put(key, (time != -1) ? time : JoH.tsl());
+    }
+}
```
jamorham: This looks like a significant duplication of code. I don't propose any changes at this point, but what are the constraints that prevent us for reusing the existing wifi-reader for this purpose?

app/src/main/java/com/eveningoutpost/dexdrip/utils/DexCollectionType.java
```diff
@@ -59,13 +61,13 @@
             mapToInternalName.put(dct.internalName, dct);
         }
 
-        Collections.addAll(usesBluetooth, BluetoothWixel, DexcomShare, DexbridgeWixel, LimiTTer, WifiBlueToothWixel, DexcomG5, WifiDexBridgeWixel);
-        Collections.addAll(usesBtWixel, BluetoothWixel, LimiTTer, WifiBlueToothWixel);
-        Collections.addAll(usesWifi, WifiBlueToothWixel,WifiWixel,WifiDexBridgeWixel, Mock);
+        Collections.addAll(usesBluetooth, BluetoothWixel, DexcomShare, DexbridgeWixel, LimiTTer, WifiBlueToothWixel, DexcomG5, WifiDexBridgeWixel, LimiTTerWifi);
+        Collections.addAll(usesBtWixel, BluetoothWixel, LimiTTer, WifiBlueToothWixel, LimiTTerWifi); // Name is misleading here, should probably be using dexcollectionservice
```
jamorham: The `usesBtWixel` is to describe the simple text based protocol used by things like LimiTTer and the classic xDrip device. It doesn't include things like xBridge or blucon, miaomiao etc

jamorham:
Ok, so really change to tomato behavior within xDrip shouldn't be in a PR for adding wifi libre support. That should be in a separate PR. But what is the reason for that change anyway?

Can you also provide a summary as to what this PR is intended to do. Assume I know nothing about the two new collection methods being introduced.

app/src/main/java/com/eveningoutpost/dexdrip/utils/DexCollectionType.java
jamorham: Also see below where we have a specific set for whether something uses dexcollection service

app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
tzachi-dar: Well, from the tomato standpoint, this code is not needed.
When it comes from the factory, it sends data every 5 minutes. Also when we start a new code we tell it to send every 5 minutes.

That said, since code now sends each buffer in a 150 ms period, this has influence on different androids stack implementations (actually if only one buffer is being sent, the wait will not happen), so I'll revert it and we can think if we are brave enough to change it.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
tzachi-dar: I have fixed the comment.
The main reason that we need this change is that now (actually for some time now) there is a retry code after crc errors. That retry takes time after 10 seconds, and that might cause an issue, so I have changed that to be 3 seconds.
The distance between the packets should be ~100ms. So no reason to see things taking more than 3 seconds.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -2480,6 +2483,9 @@ public void onClick(DialogInterface dialog, int which) {
         if(DexCollectionType.isLibreOOPAlgorithm(collector)) {
         	// Rest of this function deals with initial calibration. Since we currently don't have a way to calibrate,
         	// And even once we will have, there is probably no need to force a calibration at start of sensor use.
+            displayCurrentInfo();
+            // JamorHam, should I put here something like:
+            // ?? if (screen_forced_on)  dontKeepScreenOn();
```
tzachi-dar: Please see if I need to add something like:
if (screen_forced_on)  dontKeepScreenOn();

app/src/main/java/com/eveningoutpost/dexdrip/Services/LibreWifiReader.java
tzachi-dar: So, the main reason that we were not able to reuse the existing code are:
1) The need to move to a new protocol (see in the main answer why).
2) The existing code was returning a list of <TransmitterRawData>. Since libre has other fields (see LibreWifiData), one would have to have a third class that both will inherit from. this will make code more complex.
3) The way multiple points are handled: When reading from the g4 each packet matters. when reading from libre, we can skip older readings. That said, if for example last packet contains something like "no sensor" we need to move to the previous packet. So, libre code has to be different.

app/src/main/java/com/eveningoutpost/dexdrip/utils/DexCollectionType.java
tzachi-dar: Ok, so I know what got me confused:
The tomato, does not have a type of it's own, rather it was using the limiter name.

So, I guess that there are two solutions here:
1) Change the line to be:
   Collections.addAll(usesBtWixel, BluetoothWixel, LimiTTer, WifiBlueToothWixel, LimiTTerWifi); //  describe 
    the simple text based protocol used by things like LimiTTer and the classic xDrip device
I'll still add the LimiTTerWifi since this is still limiter.
2) Add a new HashSet<DexCollectionType> usesLimiterBT and add limite, limiterWifi there.

What do you think is the better solution?

tzachi-dar:
This are the changes that are needed to allow xDrip to read libre data from wifi readers, either directly or using a mongodb.
Like the g4 has an all house coverage mode, now the libre (tomato) has such a mode.
The receivers listen on port 50005 and wait for requests to send their data.
Optionally, they also write the data to a mongodb. This is needed to allow the system to work with a more complicated network environment.

A new version of the protocol is introduced. This version is needed in order to allow improved behavior which includes:
1) Allow to configure the mac of the remote tomato device from xDrip.
2) Allow sending all data starting at a specific time. (In the current protocol, the time is being translated to number of readings, but one does not know if this will bring entries that are too old).
3) Tell receivers not to connect to tomato until a given time. (why is this needed? well, consider receiver-a had a reading on 12:00 and then on 12:05. On 12:08 the child went to another room, and a new receiver has connected to libre, and it will now send the data on 12:08. xDrip will reject the data on 12:08 (as it is too close to 12:05) so next reading will only be on 12:13. Telling the remote device not to connect until 12:10 will stop this problem).
4) better protocol versions. (allow the device to say what protocols it supports).
5) Sending device type, and a debug message.

As for changes in tomato code:

tzachi-dar:
(meant to say: As for changes in tomato code: - see above.)

app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
```diff
@@ -211,12 +212,11 @@ static void InitBuffer(int expectedSize) {
...
-        // not sure it works at all)
-        ByteBuffer newFreqMessage = ByteBuffer.allocate(2);
-        newFreqMessage.put(0, (byte) 0xD1);
-        newFreqMessage.put(1, (byte) 0x05);
+        
+        // Make tomato send data every 5 minutes
+        ByteBuffer newFreqMessage = ByteBuffer.allocate(2);    
+        newFreqMessage.put(0, (byte) 0xD1);    
+        newFreqMessage.put(1, (byte) 0x05);    
         ret.add(newFreqMessage);
```
tzachi-dar: Well, from the tomato standpoint, this code is not needed.
When it comes from the factory, it sends data every 5 minutes. Also when we start a new code we tell it to send every 5 minutes.

That said, since code now sends each buffer in a 150 ms period, this has influence on different androids stack implementations (actually if only one buffer is being sent, the wait will not happen), so I'll revert it and we can think if we are brave enough to change it.

Issue #430, opened 2018-05-10
xDrip+ hard crash libre oop algo on new sensor


dabear:
Xdrip+ will crash consistently when using libre as collector and out of process algorithm on a brand new sensor, when only one historic bg is available. This hit me in real life after starting a new sensor may 6th.

Please note that libre will add more historic bgs every five minutes. So this crash only happens when sensor is brand new and only one historic bg is available.

Crashlog:
```

05-10 12:30:58.271 20049 20716 E LibreBlock: Updating bg for timestamp 1525948257328
05-10 12:30:58.288 20049 20716 E BgReading: Invalid data fed to InsertFromInt
05-10 12:30:58.309 20049 20716 E AndroidRuntime: FATAL EXCEPTION: Thread-53
05-10 12:30:58.309 20049 20716 E AndroidRuntime: Process: com.eveningoutpost.dexdrip, PID: 20049
05-10 12:30:58.309 20049 20716 E AndroidRuntime: org.apache.commons.math3.exception.NumberIsTooSmallException: number of points (2)
05-10 12:30:58.309 20049 20716 E AndroidRuntime: 	at org.apache.commons.math3.analysis.interpolation.SplineInterpolator.interpolate(SplineInterpolator.java:74)
05-10 12:30:58.309 20049 20716 E AndroidRuntime: 	at com.eveningoutpost.dexdrip.LibreAlarmReceiver.CalculateFromDataTransferObject(LibreAlarmReceiver.java:269)
05-10 12:30:58.309 20049 20716 E AndroidRuntime: 	at com.eveningoutpost.dexdrip.Models.LibreOOPAlgorithm.HandleData(LibreOOPAlgorithm.java:112)
05-10 12:30:58.309 20049 20716 E AndroidRuntime: 	at com.eveningoutpost.dexdrip.NSEmulatorReceiver$1.run(NSEmulatorReceiver.java:105)
05-10 12:30:58.403  9209  9209 E FJD.ExternalReceiver: Unable to bind to com.google.android.apps.tachyon.jobs.DuoJobService
```

Please see this change in libreoopalgorithm to contsistently reproduce this bug: https://github.com/dabear/LibreOOPAlgorithm/commit/29e12b8a812234e2c75d67ea199a3c6c830e6a73

dabear:
@tzachi-dar maybe you could take a look at this

tzachi-dar:
The issue is the fact that the extrapolation library wants to have at least 3 values each time.

[...]

PR #431, opened 2018-05-10
Tzachi fix interpolation

tzachi-dar:
This fixes the bug when there is only one historic data.


jamorham:
Hi, thanks for this. Did you test it using the test condition described in https://github.com/NightscoutFoundation/xDrip/issues/430 or is this just based on your understanding of the code and error condition?

tzachi-dar:
Run the error test and saw the crash.
Then fixed and saw that it does not happen.

jamorham:
:+1:

PR #433, opened 2018-05-11
Bind the CGM web service to all IPs

sulkaharo:
Removed the localhost IP from the web service creation, so OpenAPS etc can load data directly from the app as the API is then exposed to devices connected to the phone hotspot

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/webservices/XdripWebService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/webservices/XdripWebService.java
@@ -149,10 +149,10 @@ public class XdripWebService implements Runnable {
                 final SSLServerSocketFactory ssocketFactory = SSLServerSocketHelper.makeSSLSocketFactory(
                         new BufferedInputStream(xdrip.getAppContext().getResources().openRawResource(R.raw.localhost_cert)),
                         "password".toCharArray());
-                mServerSocket = ssocketFactory.createServerSocket(listenPort, 1, InetAddress.getByName("127.0.0.1"));
+                mServerSocket = ssocketFactory.createServerSocket(listenPort, 1);
             } else {
                 // Non-SSL type
-                mServerSocket = new ServerSocket(listenPort, 1, InetAddress.getByName("127.0.0.1"));
+                mServerSocket = new ServerSocket(listenPort, 1);
             }
             while (isRunning) {
                 final Socket socket = mServerSocket.accept();


jamorham:
Simply changing this to listen on all interfaces changes the behavior in a way that is only suitable for a certain use case. 

Instead there should be an option for this so it is as widely suitable as possible especially given the security implications.

I have already done some work on this, see: https://github.com/NightscoutFoundation/xDrip/commit/6fd68da3ea189747c47f94b8fd1180f2c5c754fd

app/src/main/java/com/eveningoutpost/dexdrip/webservices/XdripWebService.java
```diff
@@ -213,13 +218,25 @@ private void handle(Socket socket) throws IOException {
             // Read HTTP headers and parse out the route.
             reader = new BufferedReader(new InputStreamReader(socket.getInputStream()));
             String line;
+            boolean authFailed = false;
+
             while (!TextUtils.isEmpty(line = reader.readLine())) {
+                if (line.startsWith(("api-secret"))) {
+                    String secret = Pref.getStringDefaultBlank("xdrip_webservice_secret");
+                    if (!secret.equals("")) {
+                        secret = sha1Hash(secret);
```
jamorham: Use something like ` final String hashedSecret = Hashing.sha1().hashBytes(secret.getBytes(Charsets.UTF_8)).toString();` instead to avoid duplicating code?

app/src/main/java/com/eveningoutpost/dexdrip/webservices/XdripWebService.java
```diff
@@ -213,13 +218,25 @@ private void handle(Socket socket) throws IOException {
...
+                        String requestSecret[] = line.split(": ");
+                        if (requestSecret.length < 2) continue;
+                        if (!secret.equals(requestSecret[1].toUpperCase())) authFailed = true;
```
jamorham: use `equalsIgnoreCase` to avoid having to convert to upper case?

app/src/main/res/xml/pref_advanced_settings.xml
```diff
@@ -375,6 +375,12 @@
                 android:key="xdrip_webservice_open"
                 android:summary="Accept connections from any network instead of just internally within this device. There are security implications to enabling this!"
                 android:title="Open Web Service" />
+            <EditTextPreference
+                android:defaultValue="ABC123"
```
jamorham: should default to `""` to unless we make it only require the secret when the server is set to open but I'm thinking its better if it defaults to no secret and then has consistent behavior whether localhost or not and also can allow restriction of the localhost interface.

jamorham:
This is good, see my in-line comments for some small changes.

My plan would be to merge this and then update it to require authentication if the passphrase is set because opening this up also exposes the tasker interface which is pretty powerful.

Issue #432, opened 2018-05-10
Predictive High alerts


jwoglom:
I just noticed that Nightscout has this, and given that I imagine it would be conceptually very similar to how low alerts are already implemented I think this could be really useful.

jamorham:
I would be interested to hear from people who have used this feature with Nightscout with a view on how useful it is?

From my household I can imagine it would go off all the time and I'm not sure how useful it is when rises in blood sugar are related to the delay of insulin working vs carb absorption. But others may have different experience/treatment regime.

Webifi:
I would like this feature, though would need the ability to silence it after confirmation.   I often eat, but forget to bolus.  A predicted high would remind me to bolus for what I recently consumed.

phicks1:
I definitely need this feature. My old Medtronic pump & sensor had both predictive lows & highs & I found both useful

Navid200:
Is the existing Forecasted Low Alert an extrapolation, or a simulation?
Does it take into account the insulin and carb intake?

Can I just copy the existing forecasted low alert and modify the copy to become a forecasted high?

Navid200:
@AdrianLxM Would you please state your opinion here too for the record?
I am working on this right now.  But, if you think it would be a mistake to add the capability, I will stop.

I hope we can make a balanced decision.  
I understand that some people are concerned that this can mislead an inexperienced person to take insulin when they shouldn't.
On the other hand, people with good control claim that they need this because it will take the slope into account.

We also have alarms that go off if the slope is too high.  Someone can misunderstand those too and take insulin when they shouldn't.

I think the most important element in the control system is the brain of the diabetic person.  If we cannot depend on that, all bets are off.  But, that's only my opinion.

PR #441, opened 2018-05-16
High predictions: graph and notifications

jwoglom:
Uses the same algorithm as low predictions to enable high predictions, both with a red prediction line on the graph and notification options. The prediction algorithm can be sort of overly aggressive if the lookahead time is too high, so some fine-tuning can probably be done to make high predictions better -- maybe a higher highMark set by the user instead of the system-wide one so that this can better catch more gradual highs that wouldn't be picked up by "increasing fast" check?

screenshot_20180516-164808
screenshot_20180516-164858

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
@@ -138,7 +138,7 @@ public class BgGraphBuilder {
     private static final ReentrantLock readings_lock = new ReentrantLock();
 
     private final List<Treatments> treatments;
-    private final static boolean d = true; // debug flag, could be read from preferences
+    private final static boolean d = false; // debug flag, could be read from preferences
 
     public Context context;
     public SharedPreferences prefs;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
@@ -109,8 +109,11 @@ public class BgGraphBuilder {
     public final static double NOISE_HIGH = 200;
     public final static double NOISE_FORGIVE = 100;
     public static double low_occurs_at = -1;
+    public static double high_occurs_at = -1;
     public static double previous_low_occurs_at = -1;
+    public static double previous_high_occurs_at = -1;
     private static double low_occurs_at_processed_till_timestamp = -1;
+    private static double high_occurs_at_processed_till_timestamp = -1;
     private static long noise_processed_till_timestamp = -1;
     private final static String TAG = "jamorham graph";
     //private final static int pluginColor = Color.parseColor("#AA00FFFF"); // temporary
@@ -135,7 +138,7 @@ public class BgGraphBuilder {
     private static final ReentrantLock readings_lock = new ReentrantLock();
 
     private final List<Treatments> treatments;
-    private final static boolean d = false; // debug flag, could be read from preferences
+    private final static boolean d = true; // debug flag, could be read from preferences
 
     public Context context;
     public SharedPreferences prefs;
@@ -1041,6 +1044,7 @@ public class BgGraphBuilder {
             final boolean interpret_raw = prefs.getBoolean("interpret_raw", false);
             final boolean show_filtered = prefs.getBoolean("show_filtered_curve", false) && has_filtered;
             final boolean predict_lows = prefs.getBoolean("predict_lows", true);
+            final boolean predict_highs = prefs.getBoolean("predict_highs", false);
             final boolean show_plugin = prefs.getBoolean("plugin_plot_on_graph", false);
             final boolean glucose_from_plugin = prefs.getBoolean("display_glucose_from_plugin", false);
             final boolean illustrate_backfilled_data = prefs.getBoolean("illustrate_backfilled_data", false);
@@ -1264,6 +1268,8 @@ public class BgGraphBuilder {
                     Log.e(TAG, "Error creating back trend: " + e.toString());
                 }
 
+                boolean lowPredicted = false;
+
                 // low estimator
                 // work backwards to see whether we think a low is estimated
                 low_occurs_at = -1;
@@ -1271,7 +1277,8 @@ public class BgGraphBuilder {
                     if ((predict_lows) && (prediction_enabled) && (poly != null)) {
                         final double offset = ActivityRecognizedService.raise_limit_due_to_vehicle_mode() ? unitized(ActivityRecognizedService.getVehicle_mode_adjust_mgdl()) : 0;
                         final double plow_now = JoH.ts();
-                        double plow_timestamp = plow_now + (1000 * 60 * 99); // max look-ahead
+                        int low_lookahead_mins = 99;
+                        double plow_timestamp = plow_now + (1000 * 60 * low_lookahead_mins); // max look-ahead
                         double polyPredicty = poly.predict(plow_timestamp);
                         Log.d(TAG, "Low predictor at max lookahead is: " + JoH.qs(polyPredicty));
                         low_occurs_at_processed_till_timestamp = highest_bgreading_timestamp; // store that we have processed up to this timestamp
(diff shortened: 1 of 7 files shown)


nathangiovannini3:
As a new function it would be comfortable

Webifi:
A predicted high alert would help remind me to bolus for a meal/drink.   I'm often at an event and begin snaking on some pretty high glycemic index foods, engaged in conversation, completely forgetting my condition.  (A condition I've had for over 35 years now and, like the rest of us, pretty f'ing sick of it.)   It would be very nice to have a predicted high warning to remind me of my condition as early as possible, especially if I haven't recently logged a bolus that is anywhere close to dealing with the prediction. 

Bury the setting for it and toss up an "Are you sure, you know you'll risk taking too much insulin yada yada yada..." warning for those that choose to enable it if you must.   But for people like me, it'll help.

kolosmv:
Hello! Request from Russia: Please, please make Russian in your assemblies, this is very important and necessary!
And I ask you to make step-by-step instructions on your assembly, if possible - with skarts and detailed descriptions of each menu item settings. Thank you!

nicoleloop:
There are situations where I also would benefit from an alarm for a high blood glucose prediction. Perhaps make the option only available in engineering mode, or create the option for a very short/unobtrusive alarm that only goes off one time so as to save on alarm fatigue for those people who often have high blood sugars. I feel the benefit would be to give enough advanced warning so the the user could then take more insulin before the bg gets high enough that large amounts of insulin are required to bring it back down - which can then start an hours-long yo-yo effect. Making that the user will have the tendency to over correct four the high bg, resulting in low bg, which the user then has the tendency to over correct again... 😉
Just a thought, but unfortunately I don't have any clear idea of how to configure an alarm of that type without potentially driving the user crazy, lol.

digiexchris:
Its also extremely useful for people on extremely fast insulin such as fiasp or inhaled insulins. Often if im going high its because of a mistake and its quite easy to arrest the high before it gets there with fiasp once youre used to it and have your pump dialed in well. Right now it depends on me looking at my phone at the right time to pull it off, but a predictive alert would make it easier to do successfully.

definitivequestion:
My use case is a little different. I have not been diagnosed with diabetes, but I have significant problems with low blood sugars. However, I want to receive alerts predicting high blood sugar because I like to do my workouts when I am near a high so that I actually have the energy to be productive. It really should be under the normal alarm settings so that it can be controlled on a schedule with options for whether or not to make a sound.

Navid200:
Closes https://github.com/NightscoutFoundation/xDrip/issues/432

Navid200:
@jwoglom I have a question, which is going to show how little I know about this.  But, that's why I would like to ask.
It has nothing to do with just this PR.  It is also related to predicted low.

Does momentum only perform a numerical analysis solely based on existing data points?
Or, does it also take into account insulin on board and carbs taken?

tolot27:
@jwoglom Please can you fix the merge conflicts?

jwoglom:
@tolot27 I've fixed the obvious merge conflicts from the GitHub UI. Let me know if you notice any other issues -- this PR is pretty old, thankfully it looks like not much code in these files has changed.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -1410,14 +1414,17 @@ private synchronized void addBgReadingValues(final boolean simple) {
...
+
                 // low estimator
                 // work backwards to see whether we think a low is estimated
                 low_occurs_at = -1;
                 try {
                     if ((predict_lows) && (prediction_enabled) && (poly != null)) {
                         final double offset = ActivityRecognizedService.raise_limit_due_to_vehicle_mode() ? unitized(ActivityRecognizedService.getVehicle_mode_adjust_mgdl()) : 0;
                         final double plow_now = JoH.ts();
-                        double plow_timestamp = plow_now + (1000 * 60 * 99); // max look-ahead
+                        int low_lookahead_mins = 99;
```
tolot27, line 1426: It looks like `low_lookahead_mins` is just used once. Hence, this variable extraction is not necessary.

tolot27 (changes requested):
Just a minor change and my repeating and annoying request to apply code formatting at least for the code part you have contributed. :smirk:

Navid200:
There is an open PR for lowering the FUZZER:
https://github.com/NightscoutFoundation/xDrip/pull/1760

The main concern about it is the impact of changing the FUZZER value on the predicted low alert.
So, the FUZZER PR was modified to move it to engineering mode only.  The intent is to use it like that for testing, for probably a long time, to verify the impact for different devices.  The long-term hope is to make the predicted low alert independent of the value of the FUZZER.

Is this alert also dependent on the FUZZER value?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
jwoglom: I wanted to keep this value easily identifiable/potentially changeable if necessary so I moved it to a static constant for high and low instead

jwoglom:
@tolot27 tried to code format a bit, I think recommending an autoformatter in the developer docs and enforcing a linter on any changed code might be a nice to have for the future, so that these minor code formatting things can get fixed automatically

FYI I haven't tested a compiled apk with this code in it yet, but later this week I hope to have time to test it out on my secondary device

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -1410,41 +1416,70 @@ private synchronized void addBgReadingValues(final boolean simple) {
...
-                            Log.i(TAG, "LOW PREDICTED AT: " + JoH.dateTimeText((long) low_occurs_at));
-                            predictivehours = Math.max(predictivehours, (int) ((low_occurs_at - plow_now) / (60 * 60 * 1000)) + 1);
                         }
+                        Log.i(TAG, "LOW PREDICTED AT: " + JoH.dateTimeText((long) low_occurs_at));
+                        predictivehours = Math.max(predictivehours, (int) ((low_occurs_at - plow_now) / (60 * 60 * 1000)) + 1);
+                        lowPredicted = true;
                     }
+                }
 
-                } catch (NullPointerException e) {
```
tolot27, line 1446: Hmm, this catch was added some time ago. Did you verify that it is not necessary at all?

jwoglom: Ah, oops. Saw the commented out block inside this and instinctively removed it, forgetting the side effect of this ignoring the exception from being thrown. Re-added

tolot27:
> FYI I haven't tested a compiled apk with this code in it yet, but later this week I hope to have time to test it out on my secondary device

I'm currently testing this PR and will report back any anomalies.

jwoglom:
I've been testing this for about a day now, and I think the feature could use some refinement. xDrip seems to be warning me of highs on the home page in situations where I don't think it should be happening. The "High predicted in: .." text on the home screen doesn't seem to go away when the trend drops back down below a level which would trigger the high prediction.

Also, the lookback interval (how far back xDrip looks to try and find a linear trend) and the time before we alert (how long until the linear prediction will hit the high interval) should be configurable, I think, or at least set lower.

e.g., over the last 30 minutes my BG incrementally went up from 78 mg/dl to 127, and the high prediction alert kicked in with a notification for "High predicted in 19 mins". Doing some napkin math, I was at a trend of +1.63 mg/dl/min over the entire 30 minutes, with a maximum delta peak of +2.6 mg/dl/min between two readings (from 86/dl to 99). 


| Time      | BG |
|---------|----|
| 1:02pm | 78 |
| 1:07pm | 86 |
| 1:12pm | 99 |
| 1:17pm | 104 |
| 1:22pm | 109 |
| 1:27pm | 119 |
| 1:32pm | 127 |
| 1:37pm | 128 |
| 1:42pm | 130 |

At 1:41pm I received a "High predicted in 19 mins" notification. I'm not incredibly familiar with the math that xDrip uses for low predictions but whatever's happening here looks totally wrong to me. Something must be messed up with the math.

Navid200:
**Observation:**  There are concerns about the concept/idea of this alert and its existence as an alert.  The following is from a facebook group post:
*A low is an urgent situation.
A high might not even be needed to be treated if it is an after-meal spike e.g.
For a high it imho therefore makes more sense to have an alarm "high longer than x time" and not forecast it (which basically would be a "you have eaten" detector).*

I was in favor of adding a forecast high alert, since we had a forecast low, until I read that post.  It made sense.  

**Concern:**  If I understand the forecast low alert behavior correctly, it is only a numerical analysis, not a simulation.  It completely ignores insulin on board or carbs taken.   If predicted high is the same, it can be extremely dangerous predicting a high, which is never going to happen because of insulin that has already been taken but not taken into account.  Someone may assume the alert takes the insulin on board into account and go ahead and take more insulin.

**Question:**  Are treatments (insulin and carbs) taken into account for the high prediction?
If not, can that be added?
If not, can you explain how you don't have a concern about people who may be taking insulin because of this alert when they shouldn't?

tolot27:
> I've been testing this for about a day now, and I think the feature could use some refinement.

@jwoglom Are you planning to work on this PR again? What do you think about Navids comments?

tolot27:
> @tolot27 tried to code format a bit, I think recommending an autoformatter in the developer docs and enforcing a linter on any changed code might be a nice to have for the future, so that these minor code formatting things can get fixed automatically

@jwoglom Do you have any suggestion for a linter tool or can you provide a PR to address linting and developer docs?

jwoglom:
@tolot27 sorry for the delayed response. I haven't had time to re-set up my xDrip testing environment recently and given both the bugs in the current implementation and some concerns as to improper use, I might end up rewriting this from scratch whenever I have the time, and try to build this in a way that will be more useful for end-users.

jwoglom:
As for linting, perhaps we could set up a GitHub Action to run on pull request changes which suggests formatting fixes. I haven't tried it personally, but https://github.com/marketplace/actions/checkstyle-for-java might work for this

suside:
I would like to propose spotless as code linter, it have following advantages over actions/checkstyle-for-java:
1. can be run on dev machine with `./gradlew spotlessCheck`
2. can format code with ` ./gradlew spotlessApply`
3. can be extended to pretty much all file types
4. it can lint only differences between PR and master branch (perhaps actions/checkstyle-for-java has that too?)

It could be something [like this](https://github.com/suside/xDrip/commit/58d3cbddbda0082a295a7a1684e67044639b46e9)
Would you like me to create PR?

Navid200:
@suside Is your message/proposal related specifically to this pull request, High prediction alert?
If not, and if it is a general suggestion affecting all PRs, would you please delete it and instead open a thread in the discussions (link below)?
https://github.com/NightscoutFoundation/xDrip/discussions

tolot27:
@suside @jwoglom Thanks for your hints regarding code formatting. We should continue the discussion in #1522.

Navid200:
This is something that I can benefit from.
I don't take my entire dinner insulin in one shot.  I take half of it first.  And then, half an hour or 45 minutes after, I take the remaining half.
And, I have forgotten the second half.  This would act as a reminder for me.

But, my worry is how this is going to affect many others who are not understanding the details.
As an example, there are many who have no idea that the existing forecast low alert completely ignores IOB and COB.
Can you imagine the disaster if someone thinks that the forecast high alert takes into account the IOB?
That's my main reason for not having pursued this.

Can we do anything to address this concern?

digiexchris:
I suggest an option that is off by default under an advanced settings with
strong warnings?

Issue #415, opened 2018-05-02
Increase Size of Glucose Value in Notification 


triplem:
In the Dex App the Notification Drawer shows a quite large BG Value and the Arrow. This is rather nice because you can read it at night without glasses. 

Please increase the Size of the Glucose Value in the Notification.

Navid200:
@triplem Would you please close this issue if it has been addressed?
Otherwise, would you please update?
Thanks

PR #443, opened 2018-05-17
Large number icon in notification

jwoglom:
Option for showing the current bg reading and optionally the trend in the largeIcon section of the notification, in place of a condensed version of the expanding graph. Partially addresses #415, though a more complete solution to that would probably involve a custom notification layout.

The text size can likely be a bit larger in the with-arrow case, I imagine that the bitmap generation code can be done a bit more intelligently to ensure the text fills the space.

screenshot_20180517-003425
screenshot_20180517-003344

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -614,7 +614,16 @@ public class Notifications extends IntentService {
                     }
                 }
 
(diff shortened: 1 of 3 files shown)


jamorham:
Can you confirm that you've tested reverting the settings to the defaults and that the previous xDrip chart behavior is preserved then? Eg it still works the same as before with this new feature disabled.

jwoglom:
Yes, the notification still shows the mini-graph on the right when it's disabled. And this is an option independent from the small-icon setting (so you can have this enabled but still have the default xDrip icon as the small icon, as in the first screenshot, and vice versa)

PR #444, opened 2018-05-17
Ongoing notification channel

jwoglom:
A couple notes here:
* NotificationCompat doesn't allow a bitmap/Icon object to be set as a smallIcon; unlike Notification, it only supports a drawable reference. This means that, as far as I can tell, the number icon functionality requires the ongoing notification to be constructed using a Notification.Builder instead of a NotificationCompat.Builder. This adds a fair amount of cruft (given that the notification channel code used NotificationCompat everywhere) and is unfortunate.
* On a Nexus 5X, with the ongoing notification channel enabled, I am not getting any alerts (sounds/vibrate/notification popup/etc) whenever the notification changes. However, inconsistently this is causing the ambient display to wake up and show each new bg reading when the screen is off. I'm not sure how to get this to stop, but honestly I kind of like it. I'm not sure how non-nexus/pixel devices react to this though, and this option should be tested.


jwoglom:
The most important thing for me is that it actually pushes the xDrip notification to the top of the notifications list in my experience. Without an ongoing notification channel, it's often pushed below email and news notifications which is super irritating because it's then hard to get to when I scroll down to view my notifications, and it also won't appear (with the number-icon mode enabled) in the front few set of icons that I can see from the top when an application is open or on my phone's always-on lock screen. This works out of the box on oreo with an ongoing channel enabled, as xDrip defaults to the "High" notification importance mode. 

Of the three additional "notification importance" modes which you can customize per notification channel, there is also Urgent, which causes the ongoing notification to pop onto the screen from the notification bar (similar to the incoming call dialog https://i.stack.imgur.com/YVP5w.png); Medium, which pushes the xDrip notification below apps like Gmail in the ordering (and appears to be roughly equivalent to when notification channels aren't enabled for the ongoing notification); and Low, which pushes it to the very bottom and doesn't display the xDrip icon in the status bar at all. The one thing to be aware of is that by default, if you modify the importance of the channel it will add back in a default sound, which you can then remove by setting the sound option underneath importance to "None".

I only just resumed using the build for this PR on my phone (after updating my branch with the latest nightly changes) but already for the first time in months xDrip is actually occupying the first spot in my notification bar so I don't have to keep hunting for it constantly, which is really awesome.

jamorham:
I do have a physical Oreo device for testing now. So are you saying that Oreo ignores the existing: `setPriority(Notification.PRIORITY_HIGH)` which should already push the ongoing notification to the top?

jwoglom:
As far as I can tell, yes. I haven't tested how the notification priority is affected with notification channels disabled entirely on Oreo, but with it enabled for everything but the ongoing it definitely lowers the priority.

jamorham:
Previously I thought it only did this if you used notification channels. Using the old system bypassed the new degraded behavior. 

Android reference talks about `PRIORITY_HIGH` being replaced with `IMPORTANCE_HIGH` https://developer.android.com/reference/android/app/NotificationManager#IMPORTANCE_HIGH
Is that significant here?

I hesitate because notification channel introduction in xDrip caused nothing but problems and android's design of the feature seems completely flawed based on my limited understanding of it.

gitanO0:
I'll be happy to test via Android 9 when ready.

PR #445, opened 2018-05-17
Tzachi sensor sn

tzachi-dar:
This code calculates the libre serial number and also sends it to the OOP algorithm.

As far as I saw, the data for the blukon was never used. If this is the case, then I would like to use one storage place for them both.


tzachi-dar:
This is replaced by 
https://github.com/NightscoutFoundation/xDrip/pull/448

PR #448, opened 2018-05-20
Tzachi sensor sn 2

tzachi-dar:
This code calculates the libre serial number and also sends it to the OOP algorithm.

It replaces https://github.com/NightscoutFoundation/xDrip/pull/445.
It will store both tomato and blukon on the same location.

### DIFF
--- a/wear/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
+++ b/wear/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
@@ -52,7 +52,7 @@ public class Blukon {
         INITIAL
     }
 
-    private static final boolean testWithDeadSensor = true; // never in production
+    private static final boolean testWithDeadSensor = false; // never in production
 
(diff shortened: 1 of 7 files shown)


wear/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
```diff
@@ -52,7 +52,7 @@
         INITIAL
     }
 
-    private static final boolean testWithDeadSensor = true; // never in production
+    private static final boolean testWithDeadSensor = false; // never in production
```
dabear: Is this change intentional?

tzachi-dar: This is a bugfix, actually already fixed in https://github.com/NightscoutFoundation/xDrip/pull/447

dabear:
Looks good to me. Sensor serial number is correct (verified against Abbot's libre reader handheld)
image

tzachi-dar:
Ready for merge.

PR #465, opened 2018-06-01
blukon: raw command strings converted to readable names

keencave:
added additional comments how the communication works.

It should be easier now to understand and/or change the java code for blucon. The raw strings were replaced by static String variables. Comments added where to find which byte in which block of the FRAM.

The code compiles without warnings. I made a line by line check but are not able to real test it within the next days.

Issue #174, opened 2017-09-08
Replace string "Limitter Battery" for Blukon


gregorybel:
With Blukon when shoying Battery statue string is "Limitter Batterie", could you please maybe use the new variable "HARDWARE" and display either Blukon, BlueReader, etc....?

jamorham:
Should be resolved in the Sept 14th nightly.

gregorybel:
yes yorking, thanks Jon!

PR #467, opened 2018-06-02
Allow history retrieval from a Blukon device...

trallen:
… when connecting to the device after a period of more than 15 minutes has elapsed, without using the LibreOOPAlgorithm.

(This includes a fix for a possible bug in the LibreOOPAlgorithm code which occurs when the sensor age is requested before the history. The sensor age is then updated, and then the history is handed off to LibreAlarmReceiver, which again checks the sensor age. When finding that the sensor age is current, it discards the historical data.)


app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
```diff
@@ -489,15 +503,41 @@ private static void handlegetHistoricDataResponse(byte[] buffer) {
...
-            LibreBlock.createAndSave("blukon", now, m_full_data, 0);
-            
-            Intent intent = new Intent(Intents.XDRIP_PLUS_LIBRE_DATA);
-            Bundle bundle = new Bundle();
-            bundle.putByteArray(Intents.LIBRE_DATA_BUFFER, m_full_data);
-            bundle.putLong(Intents.LIBRE_DATA_TIMESTAMP, now);
-            intent.putExtras(bundle);
-            intent.addFlags(Intent.FLAG_INCLUDE_STOPPED_PACKAGES);
-            xdrip.getAppContext().sendBroadcast(intent);
+
```
tzachi-dar: Can you replace this code block with NFCReaderX.HandleGoodReading and verify that it works with the oopalgorithm or without it?

trallen: Thanks, that's much neater. Hadn't wanted to make changes to the LibreOOPAlgorithm code that were too significant, but it seems to work just fine. I'm just leaving it overnight to do some history testing for both versions; will submit the corrected version tomorrow morning.

gregorybel:
thanks adding this feature! there's another PR on this code so you may have some conflict issues, sorry for this. I think we should merge first PR from keenkave.

trallen:
I'll be happy to review once #465 is merged and push any necessary changes then. (A brief look at the PR suggests it should apply cleanly.)

keencave:
Nice work! This was outstanding for 1/2 a year but always on my todo list. Thanks!

keencave:
The code looks good for me so far. 
But in line #174 of blukon.java you check:

m_minutesDiffToLastReading > 15

Have you tested if this can cause glitches? There is a calculation in the NFC scan java module how to "connect" the trend buffer to the history buffer without glitches. I remember an offset of 3 as the 15 min history values are calculated as the avarage of n-2,n-1,n,n-1,n-2 of the fitting trend buffer sequence.
May it be better to test for >17 as this is the uper border for the trend buffer backfilling? Anyway, I think we should start with this and do an intertation fixing this later on. A lot of people which used the SW3 / Librealarm are waiting for this.

jamorham:
You guys know this code much better than me so I am relying on you to check and test it and let me know when it looks good to merge. Thanks.

trallen:
I had intended to revisit the value of 15 there; thanks for spotting. I kept it at 15 since I had questions about the reason for the 17 check (the comment talks about double draws, which wasn't clear to me). I wasn't sure if 15, 16, 17, or 20 would be better values. I think you're correct, though -- 17 seems like it would work best with the current backfill code. I'll update this commit.

trallen:
Incidentally, I've left this code disabled by default, since I was assuming it would not be widely used. If it is expected to be widely used, you may wish to change the relevant defaultValue to true in app/src/main/res/xml/pref_advanced_settings.xml. But that can be done later on, I think.

keencave:
for the first run we should set it to false. After a longer test we should make it treu as default.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
```diff
@@ -592,15 +603,12 @@ private static void handlegetHistoricDataResponse(byte[] buffer) {
...
+            final String tagId = PersistentStore.getString("LibreSN");
+            NFCReaderX.HandleGoodReading(tagId, m_full_data, now);
+
```
dabear: What happens if tagId is empty here? As can be seen from other parts of blukon.java, we don't always decode the serialnumber:
```
if (JoH.pratelimit(BLUKON_DECODE_SERIAL_TIMER, GET_DECODE_SERIAL_DELAY)) {
                decodeSerialNumber(buffer);
}
```

trallen: I've run the code a couple of times with an empty tagId, and it doesn't appear to cause any problems (I had initially created a fallback to the string "blukon", and was testing its removal). But perhaps someone with more experience of the LibreAlarm and NFCReaderX code can comment?

jamorham:
Planning to merge this on Friday. Please let me know if it needs any further work.

PR #476, opened 2018-06-12
Libre - Don't handle reading that are marked as not ready.

tzachi-dar:

Still being tested.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/LibreUtils.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/LibreUtils.java
@@ -2,9 +2,9 @@ package com.eveningoutpost.dexdrip.UtilityModels;
 
 import android.util.Log;
 
-public class LibreCrc {
+public class LibreUtils {
 
-    private static final String TAG = LibreCrc.class.getSimpleName();
+    private static final String TAG = LibreUtils.class.getSimpleName();
 
     private final static long[] crc16table = {
             0, 4489, 8978, 12955, 17956, 22445, 25910, 29887, 35912,
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/LibreUtils.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/LibreUtils.java
@@ -1,10 +1,14 @@
 package com.eveningoutpost.dexdrip.UtilityModels;
 
+import com.eveningoutpost.dexdrip.Home;
+
 import android.util.Log;
 
 public class LibreUtils {
 
     private static final String TAG = LibreUtils.class.getSimpleName();
+
+    private static final boolean testWithDeadSensor = false; // never in production
 
     private final static long[] crc16table = {
             0, 4489, 8978, 12955, 17956, 22445, 25910, 29887, 35912,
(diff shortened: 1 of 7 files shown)


libxmike:
@tzachi-dar is this hard coded function or function that is by default on and can be turned on? to have it as option is impotent for testing transmitters like limitter etc.

dabear:
Maybe this function should be moved Into the nfcreaderx class instead of living inside the blucon class

libxmike:
also this function is already implemented in keencave's code for Limitter/RFDuino and Simblee

gregorybel:
@tzachi-dar we all appreciate how much time you invest in Xdrip and how fast you are but code quality is going to be worser and worser. 
module nfcreader is imported by Blukon, now nfcreader imports blukon. And blukon knows about ooap and nfcreader too etc...

I think we slowly need a big refactoring where we have a common Libre class and several modules which are only Transport specific class (like NFC, Blukon, MM, etc...). only the libre class should know about ooap algo. etc...

keencave:
The code is OK and should work. It should be integrated soon.

I would suggest to make the check configurable like @libxmike suggested. It is a pain to test without the possibility to use a dead sensor. I mostly test two system together at the same time. One life sensor, one dead sensor and two phones. Maybe we can place this under less common settings, misc options?

keencave:
@libxmike I provided the isSensorReady() function for the blukon.java file ;-)
 It is the same like the one in the LBridge code but now coded in java.

jamorham:
@gregorybel we have a `LibreCrc` class, perhaps there could be a `LibreUtils` class for some of these generic Libre data processing functions that don't require any dependencies.

tzachi-dar:
Ok, So I have renamed librecrc to libreutils, and added the function IsSensorReady to there.

If this direction is ok, than we can add more functions there.

As for allow testing with dead sensor.
So, this PR does not change anything in this regard. If people do want to test with a dead sensor, I think that it is safer to add to the ui a button saying "allow testing with a dead sensor for 4 hours".  (I must say that I don't know how to do it for the wear, but I guess someone will tell me).

All comments are welcomed.

Not tested at all.

gregorybel:
this looks clean now! thanks. good Idea to move it here. 


A next step could be to move the code for decoding serial number, calculate sensor days, etc.... should be all the same code for all Libre plugins.

tzachi-dar:
I guess that next step should be having one file for wear and app, but that will need a new pr...

gregorybel:
is it possible? how?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/LibreUtils.java
```diff
@@ -1,10 +1,14 @@
...
+
 import android.util.Log;
 
-public class LibreCrc {
+public class LibreUtils {
+
+    private static final String TAG = LibreUtils.class.getSimpleName();
 
-    private static final String TAG = LibreCrc.class.getSimpleName();
+    private static final boolean testWithDeadSensor = false; // never in production
```
jamorham: Can we add `@Getter` annotation to `testWithDeadSensor` so that it can be referenced elsewhere? For example to bypass the clock advancement check and so that something can be displayed in the UI to warn that dead sensors are being allowed. Then I could make this a run time configuration option for engineering mode.

tzachi-dar:
This was replaced by https://github.com/NightscoutFoundation/xDrip/pull/497

PR #420, opened 2018-05-05
Adding support for Smartwatch Amazfit Pace/Stratos

Klaus3d3:
for the watch app see https://github.com/Klaus3d3/XdripWidgetForAmazfit-Klaus

PR #477, opened 2018-06-12
Adding support for Smartwatch Amazfit Pace/Stratos (cleaned)

jamorham:
This is a cleaned up version of: https://github.com/NightscoutFoundation/xDrip/pull/420

I have resolved the merge conflicts, restored to breakages to the gradle files and minimized the changeset.

Additionally I have removed the bundled jar library and instead added a reference to my fork of the library which simply changes the minimum and target sdk values to be compatible with those which xDrip uses. 

As far as I can tell at the moment the library does not utilize anything in the higher sdk levels. See https://github.com/jamorham/AmazfitCommunication/commit/0c1d24c64e4c28aaa903fb6aeff35e6e798e38b4 for the only change I have made to the library. 

I have done absolutely no testing of this but I believe it is otherwise identical to the submitted PR.

For the watch app see https://github.com/Klaus3d3/XdripWidgetForAmazfit-Klaus

Please review this and let me know if it looks correct!


jamorham:
@dabear @Klaus3d3 please review this version thanks

PR #491, opened 2018-06-26
G5 Native backfill analytics

jwoglom:
Adds an option in Nightscout Sync>Extra Options to append the source_info for each BgReading to the device string. This lets the nightscout DB know whether or not a specific reading was a backfill, or whether it was captured with Ob1g5 since that is currently merged in with regular G5 as a collection method.

Nightscout doesn't explicitly make this user-visible in their web UI but this does mean that the data about the collection method for each record is stored outside of xDrip's database.

future todo is that source_info doesn't appear to be used outside of the ob1 g5 collector

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
@@ -1010,7 +1010,7 @@ public class Ob1G5StateMachine {
         DexTimeKeeper.updateAge(getTransmitterID(), glucose.timestamp);
         if (glucose.usable() || (glucose.insufficient() && Pref.getBooleanDefaultFalse("ob1_g5_use_insufficiently_calibrated"))) {
             UserError.Log.d(TAG, "Got usable glucose data from G5!!");
-            final BgReading bgReading = BgReading.bgReadingInsertFromG5(glucose.glucose, JoH.tsl());
+            final BgReading bgReading = BgReading.bgReadingInsertFromG5(glucose.glucose, JoH.tsl(), null);
             if (bgReading != null) {
                 try {
                     bgReading.calculated_value_slope = glucose.getTrend() / Constants.MINUTE_IN_MS; // note this is different to the typical calculated slope, (normally delta)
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
```diff
@@ -1218,8 +1218,7 @@ private static void processBacksies(List<BackFillStream.Backsie> backsies) {
                 UserError.Log.wtf(TAG, "Backfill timestamp unrealistic: " + JoH.dateTimeText(time) + " (ignored)");
             } else {
                 if (BgReading.getForPreciseTimestamp(time, Constants.MINUTE_IN_MS * 4) == null) {
-                    final BgReading bgr = BgReading.bgReadingInsertFromG5(backsie.getGlucose(), time);
-                    if (bgr != null) bgr.appendSourceInfo("Backfill");
+                    final BgReading bgr = BgReading.bgReadingInsertFromG5(backsie.getGlucose(), time, "Backfill");
```
jamorham: Why change the method signature when can just use the `bgr.appendSourceInfo("Backfill");` ?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -710,12 +710,22 @@ private static synchronized void handleRestFailure(String msg) {
         Log.e(TAG, msg);
     }
 
+    private String getDeviceString(BgReading record) {
+        String withMethod = "xDrip-" + prefs.getString("dex_collection_method", "BluetoothWixel");
```
jamorham: Wondering about using `DexCollectionType.getBestCollectorHardwareName()` instead of the raw `dex_collection_method` preference item.

app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
jwoglom: From what I could tell, using appendSourceInfo after the insertfromg5 method didn't save the entry in the database, only modified the local instance of bgreading. So by doing this it runs appendSourceInfo before the .save(), and I don't have to explicitly call save after which seems like something internal to the method that external accessors shouldn't have to worry about.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
jwoglom: I think that'd make sense... However because  "G5 Native" gets added to the source info already when that's in use as it is now this would lead to a duplicated "xDrip-G5 Native G5 Native" from the current "xDrip-DexcomG5 Native G5"/"xDrip-Dexcom G5 Native G5::Backfill"

app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
jamorham: I think you're right about that and its effectively a bug as the appended source info *is* intended to be persistent. Other than doing multiple saves, your refactoring to put it in the constructor seems to be a good way of doing it given that `bgReadingInsertFromG5` is only called in two places.

PR #501, opened 2018-07-10
Show tomato battery on UI.

tzachi-dar:

Show tomato battery on home screen as we do for all other devices.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
@@ -174,6 +174,7 @@ public class Tomato {
         byte[] serialBuffer = Arrays.copyOfRange(s_full_data, 2, 13);
         Blukon.decodeSerialNumber(serialBuffer);
         PersistentStore.setString("Tomatobattery", Integer.toString(s_full_data[13]));
+        Pref.setInt("bridge_battery", s_full_data[13]);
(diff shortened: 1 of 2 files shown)


dabear:
This should probably be set to 0 when initialized. This is done by the blukon class.

tzachi-dar:
Already done here: https://github.com/NightscoutFoundation/xDrip/pull/501/files#diff-cfb70c5b76b8099b7989b90ba5c75106L214

tzachi-dar:
By the way I wonder, should we really set the battery to zero? the way tomato works, we will not show the battery when there is no sensor and this is when people want the battery the most.

blocklist:
i think a "-" or "n/a" is better then "0", otherwise, when the miaomiao is not directly on the sensor, he can lost a more battery then normaly.

dabear:
The bridge_battery pref should really be set to 0, or else you will get false information when changing between bridges. For example, you could risk getting stale battery info from the blukon when changing to the miaomiao.

tzachi-dar:
Changing between bridges is very uncommon scenario.
A much more interesting scenario is what happens when one is charging (or before charging). 
Since the tomato does not give us information in that case, we will get it wrong in any case, so I guess that we should not show what is going on.
Will leave it as is (battery zeroed).

tzachi-dar:
This is ready from my side.

dabear:
This seems ok to me, this will also enable upload and tracking of miaomiao battery data to nightscout; so this is a very welcome two-liner fix indeed. Should be merged for wider testing.

PR #502, opened 2018-07-10
Allow testing with a dead sensor on engineering mode.

tzachi-dar:

This ades the option to allow testing bad sensors from UI based on engineering mode.

A safer option for this is to remove the UI, and allow to set it by editing the xml files.
@keencave @jamorham does removing UI sound reasonable.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/LibreUtils.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/LibreUtils.java
@@ -3,13 +3,10 @@ package com.eveningoutpost.dexdrip.UtilityModels;
 import com.eveningoutpost.dexdrip.Home;
 
 import android.util.Log;
-import lombok.Getter;
 
 public class LibreUtils {
 
     private static final String TAG = LibreUtils.class.getSimpleName();
-
-    @Getter private static final boolean testWithDeadSensor = false; // never in production
 
     private final static long[] crc16table = {
             0, 4489, 8978, 12955, 17956, 22445, 25910, 29887, 35912,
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/LibreUtils.java
```diff
@@ -111,7 +108,13 @@ public static boolean isSensorReady(byte sensorStatusByte) {
     
         Log.i(TAG, "Sensor status is: " + sensorStatusString);
     
-        if (testWithDeadSensor) return true;
+        
+        
+        
+        if (Pref.getBooleanDefaultFalse("allow_testing_with_dead_sensor")) {
```
jamorham: I think this should be in a public method, eg `boolean getTestWithDeadSensor()` so that it matches the removed getter and can be accessed externally by other things which wish to check this flag, then replace this code block with a call to that method and then the resulting logic for logging and returning.

Then update the method so it definitely requires engineering mode as well with something like `&& Pref.getBooleanDefaultFalse("engineering_mode")`

Then of course mirror the same for wear. Hope that makes sense.

tzachi-dar:
@jamorham I believe that if this feature will not have any UI, it should be very safe for our users. What do you think?
Tzachi

jamorham:
@keencave @tzachi-dar I would expect to provide a voice command to enable/disable the feature and for it only to be functional in engineering mode.

Please see my code comments above about making the `getTestWithDeadSensor()` public method.

tzachi-dar:
Moved the feature to work by speech recognition and added a getter.
Feature also depends on engineering mode.
Still not tested with dead sensor.

keencave:
Perfect! Ince availabke in nightly i will test it. Thx in advance!

tzachi-dar:
Verified with a dead sensor.
Can be merged.

PR #526, opened 2018-07-25
Tzachi trend6

tzachi-dar:
Increase the libre trend to show data for the entire sensor life.

Still needs to verify it works well sensor is being replaced.


tzachi-dar:
Replaced by https://github.com/NightscoutFoundation/xDrip/pull/540

PR #540, opened 2018-08-08
Libre trend 7

tzachi-dar:
This code creates the libre trend for the entire sensor life.
It allows to show the libre trend graph for longer periods (see below for examples).
As a side result it also settes the sensor sn in all places except for libre alarm (where the data does not exist).
Tested for a few weeks with two sensors.


tzachi-dar:
Here is an example of the improved dialog.
image

gregorybel:
don't think it will break Blukon class.

keencave:
Looks good for me too. I dont await any negative effects.

jamorham:
Side note: Should `decodeSerialNumber` be moved to `LibreUtils` class?

gregorybel:
would make sense!

tzachi-dar:
decodeSerialNumber should move to libreutils. If it is ok, I'll do it in the next pr.

PR #559, opened 2018-08-23
add 'G6 Native' combined with 'G5 Native' for source info label

jwoglom:


jamorham:
This PR will break interoperability with AndroidAPS as it stands.

jwoglom:
Does AndroidAPS have the "G5 Native" label hardcoded? I did a quick search on https://github.com/MilosKozak/AndroidAPS and didn't see the string "g5 native" in the source.

AdrianLxM:
The affected line is here: https://github.com/MilosKozak/AndroidAPS/blob/dev/app/src/main/java/info/nightscout/androidaps/plugins/Source/SourceXdripPlugin.java#L73

As Milos wants to release a new RC soon It should not be much of a problem though if we merge this to dev: https://github.com/MilosKozak/AndroidAPS/pull/1361

jamorham:
xDrip needs to maintain compatibility with previous AAPS versions. The best way to deal with it at the moment I think is to transmute G6 to G5 in the broadcast.

jwoglom:
Maybe when G6 mode is enabled the sourceInfo should just be appended with "G6" and that way the "G5 Native" part can stay as-is.

jwoglom:
didn't add a comment when I pushed an updated commit earlier, but when using a G6 "G5 Native" and "G6 Native" are now both appended to source info and the hardware name identifier reports as "G5 Native (G6 Native)". Ideally the G5 Native could eventually be dropped, but for now this will at least make individual readings identifiable as either G5 or G6.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -1113,6 +1114,10 @@ public static synchronized BgReading bgReadingInsertFromG5(double calculated_val
             bgr.calculated_value = calculated_value;
             bgr.raw_data = SPECIAL_G5_PLACEHOLDER; // placeholder
             bgr.appendSourceInfo("G5 Native");
+            // send both G5 and G6 native for legacy compat
+            if (Ob1G5CollectionService.usingG6()) {
```
jamorham: I don't think it is the source info which is used by legacy AAPS to make decisions about whether to enable SMB, it is the source description. Its only at the point of local intent broadcast that this compatibility needs to be introduced afaik which is why I did it there.

Navid200:
@jwoglom Hi,
The last comment by jamorham is on September 18, 2018.
After that, I see nothing from you.
May I ask your thoughts on this pull request?

Thanks

jwoglom:
It appears that xDrip still stores the "G5 Native" for G6 BgReading's in this one specific location, and that "G5 Native" source info is still published to Nightscout. So the BgReading.java changes in this PR are still relevant.

jwoglom:
@jamorham Please look at the newly opened pull request at https://github.com/NightscoutFoundation/xDrip/pull/1714.

Closing this PR.

PR #560, opened 2018-08-23
Correct Setting of LibreSN:

tzachi-dar:
decodeSerialNumber is moved to LibreUtils.
It is used also for manual scanning.
We set LibreSN for tomato (a bug fix) and also for manual scanning.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
@@ -34,6 +34,7 @@ import com.eveningoutpost.dexdrip.utils.DexCollectionType;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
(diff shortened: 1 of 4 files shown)


tzachi-dar:
Still needs some testing.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -272,10 +273,12 @@ protected void onPostExecute(Tag tag) {
                 if (succeeded) {
                     final String tagId = bytesToHexString(tag.getId());
                     long now = JoH.tsl();
+                    String SensorSn = LibreUtils.decodeSerialNumberKey(tag.getId());
                     boolean checksum_ok = HandleGoodReading(tagId, data, now);
                     if(checksum_ok == false) {
                         Log.e(TAG, "Read data but checksum is wrong");
                     }
+                    PersistentStore.setString("LibreSN", SensorSn);
```
gregorybel: why not moving this call within the decode Serial method? like this all collectors can show serial number in system Status.

keencave:
Perfect!

dabear:
Could anyone still using the blukon test to see if we can create a function that covers all use-cases (rather than having both decodeSerialNumber and decodeSerialNumberKey) ?

tzachi-dar:
Did some more testing, worked correctly. Can be merged.

PR #568, opened 2018-08-30
Treatment entry improvements

bassettb:
On treatment entry keypad, only show the green tick submit button if we have a non-zero blood, insulin, or carbs.

Allow entering time without punctuation.  On the treatment confirmation screen, if entered time is missing the dot, add it. If only time is entered, show the cancel button.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -1533,10 +1533,6 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
 
             case "time":
                 Log.d(TAG, "processing time keyword");
-                // Add the dot to the time if it's missing.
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -1533,6 +1533,10 @@ private void handleWordPair() {
 
             case "time":
                 Log.d(TAG, "processing time keyword");
+                // Add the dot to the time if it's missing.
+                if (thisnumber >= 100)
```
jamorham: Can this be moved to `PhoneKeyPadInputActivity` so that we don't have to worry about changes in behavior with the voice input?

bassettb: I thought it would be nice to be able to optionally voice input the time as "twelve hundred time".  Let me know what you think.  I can move it if you want.

PR #584, opened 2018-09-12
Optionally store g5/g6 transmitter battery information in nightscout

jwoglom:
When enabled, this will add an object like the following in nightscout's devicestatus whenever the voltages or resistance state of the battery changes:
```
{
  "_id": "5b98a7d3d2ea1d06adda76db",
  "device": "G6 Transmitter",
  "uploader": {
    "days": 20,
    "status": "OK",
    "battery": "298/282/959",
    "voltagea": 298,
    "voltagea_status": "GOOD",
    "voltageb": 282,
    "voltageb_status": "WARNING",
    "resistance": 959,
    "resistance_status": "NORMAL",
    "temperature": 35
  },
  "created_at": "2018-09-12T05:44:51.553Z"
}
```

In the nightscout UI, the G5/G6's voltagea, voltageb, and resistance will appear when hovering over the battery icon:

image

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
@@ -9,6 +9,7 @@ import android.os.Build;
 import android.os.PowerManager;
 import android.preference.PreferenceManager;
 
+import com.eveningoutpost.dexdrip.G5Model.Ob1G5StateMachine;
 import com.eveningoutpost.dexdrip.Home;
 import com.eveningoutpost.dexdrip.MegaStatus;
 import com.eveningoutpost.dexdrip.Models.BgReading;
@@ -23,6 +24,7 @@ import com.eveningoutpost.dexdrip.Models.UserError;
 import com.eveningoutpost.dexdrip.Models.UserError.Log;
 import com.eveningoutpost.dexdrip.Services.DexCollectionService;
 import com.eveningoutpost.dexdrip.Services.ActivityRecognizedService;
+import com.eveningoutpost.dexdrip.Services.Ob1G5CollectionService;
 import com.eveningoutpost.dexdrip.utils.CipherUtils;
 import com.eveningoutpost.dexdrip.utils.DexCollectionType;
 import com.eveningoutpost.dexdrip.xdrip;
@@ -1158,6 +1160,7 @@ public class NightscoutUploader {
 
 
     private static final String LAST_NIGHTSCOUT_BATTERY_LEVEL = "last-nightscout-battery-level";
+    private static final String LAST_G5TX_BATTERY_LEVEL = "last-g5tx-battery-level";
 
     private void postDeviceStatus(NightscoutService nightscoutService, String apiSecret) throws Exception {
 
@@ -1170,6 +1173,12 @@ public class NightscoutUploader {
             batteries.add("Bridge");
         }
         if (DexCollectionType.hasWifi()) batteries.add("Parakeet");
+
+        boolean isOb1 = Pref.getBooleanDefaultFalse(Ob1G5CollectionService.OB1G5_PREFS);
+        boolean sendG5TxBattery = Pref.getBooleanDefaultFalse("send_g5tx_battery_to_nightscout");
+        if (DexCollectionType.isG5G6() && isOb1 && sendG5TxBattery) batteries.add("G5 Transmitter");
+
+        TransmitterBattery txBattery = null;
 
         for (String battery : batteries) {
 
@@ -1188,17 +1197,43 @@ public class NightscoutUploader {
                     battery_level = Pref.getInt("parakeet_battery", -1);
                     battery_name = "Parakeet";
                     break;
+                case "G5 Transmitter":
+                    txBattery = Ob1G5TransmitterBattery.getTransmitterBattery();
+                    battery_level = txBattery.voltageA();
+                    battery_name = Ob1G5StateMachine.usingG6() ? "G6 Transmitter" : "G5 Transmitter";
+                    break;
                 default:
                     battery_level = -1;
                     break;
             }
             final long last_battery_level = PersistentStore.getLong(LAST_NIGHTSCOUT_BATTERY_LEVEL);
+            final String last_g5tx_level = PersistentStore.getString(LAST_G5TX_BATTERY_LEVEL);
 
             final JSONArray array = new JSONArray();
             final JSONObject json = new JSONObject();
             final JSONObject uploader = new JSONObject();
 
-            if ((battery_level > 0) && (battery_level != last_battery_level || always_send_battery)) {
+            if (battery_name.equals("G5 Transmitter") || battery_name.equals("G6 Transmitter")) {
+                if (txBattery != null && !last_g5tx_level.equals(txBattery.battery())) {
+                    PersistentStore.setString(LAST_G5TX_BATTERY_LEVEL, txBattery.battery());
+                    uploader.put("days", txBattery.days());
+                    uploader.put("status", txBattery.status().name());
+                    uploader.put("battery", txBattery.battery());
+                    uploader.put("voltagea", txBattery.voltageA());
+                    uploader.put("voltagea_status", txBattery.voltageAStatus().name());
+                    uploader.put("voltageb", txBattery.voltageB());
+                    uploader.put("voltageb_status", txBattery.voltageBStatus().name());
+                    uploader.put("resistance", txBattery.resistance());
+                    uploader.put("resistance_status", txBattery.resistanceStatus().name());
+                    uploader.put("temperature", txBattery.temperature());
+
+                    json.put("device", battery_name);
+                    json.put("uploader", uploader);
+
+                    array.put(json);
+                    postDeviceStatusAPI(json, nightscoutService, apiSecret);
+                }
+            } else if ((battery_level > 0) && (battery_level != last_battery_level || always_send_battery)) {
                 PersistentStore.setLong(LAST_NIGHTSCOUT_BATTERY_LEVEL, battery_level);
                 // UserError.Log.d(TAG, "Uploading battery detail: " + battery_level);
                 // json.put("uploaderBattery", battery_level); // old style
@@ -1209,6 +1244,8 @@ public class NightscoutUploader {
 
                 array.put(json);
 
+                postDeviceStatusAPI(json, nightscoutService, apiSecret);
+
                 // example
                 //{
                 //    "device": "openaps://ediscout2.local",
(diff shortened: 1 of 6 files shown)


TecMunky:
I merged this with my personal branch on my local repo.
It was working at first, but now I don't get the transmitter battery info anymore.
Is there anything I should do to get it pushed to nightscout again?
Yes, I have the option selected to Upload Ob1G5/G6 transmitter battery.

thanks in advance

jwoglom:
@TecMunky the only situations where it shouldn't upload a new entry should be when the previously uploaded entry was the same (the previous battery stat is stored in a local xDrip cache) or if the requirements aren't met (ob1g5 + that option enabled). This pr also doesn't change whatever interval xDrip actually polls the battery status (which is something you can manually do in the system status view).

How are you identifying that the transmitter battery info isn't displaying? Are you simply looking at the Nightscout UI? I'm not sure under what circumstances it displays in the UI, but values should still be added to nightscout's mongo database (devicestatus).

TecMunky:
> @TecMunky the only situations where it shouldn't upload a new entry should be when the previously uploaded entry was the same (the previous battery stat is stored in a local xDrip cache) or if the requirements aren't met (ob1g5 + that option enabled). This pr also doesn't change whatever interval xDrip actually polls the battery status (which is something you can manually do in the system status view).
> 
> How are you identifying that the transmitter battery info isn't displaying? Are you simply looking at the Nightscout UI? I'm not sure under what circumstances it displays in the UI, but values should still be added to nightscout's mongo database (devicestatus).

I am hovering over the battery status icon in my cgm website.  When I first enabled this, the G5 battery level was at the bottom of the list.  I have 3 phones reporting battery level - the collector and 2 followers - so the list shows 3 phones.  I turned off reporting from the 2 followers a couple days ago, but they are apparently still reporting battery status.

Navid200:
May I ask what the status of this is?
It would be very useful.  
What needs to be done for it to move forward?

Thanks

Navid200:
Will this allow a graph of Voltage B versus time to be plotted?

Navid200:
@jwoglom Any updates on this?
Can I create a graph (Voltage B versus time) with this?
If yes, this is very useful and I would recommend moving this forward and merging it.

Would you please respond?

tolot27:
> Can I create a graph (Voltage B versus time) with this?
> If yes, this is very useful and I would recommend moving this forward and merging it.

Yes, this should be possible if you pull the data from Nightscout. I don't expect to see a graph in Nightscout. You may ask in the Nightscout Discord channel if this is possible.

Navid200:
Would it be frowned upon if I work on a pull request that I did not open myself?  
Is there an unspoken code that one should only work on pull requests that he/she has opened themselves?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -1158,6 +1160,7 @@ private void checkGzipSupport(Response r) {
 
 
     private static final String LAST_NIGHTSCOUT_BATTERY_LEVEL = "last-nightscout-battery-level";
+    private static final String LAST_G5TX_BATTERY_LEVEL = "last-g5tx-battery-level";
```
tolot27, line 1163: This constant indicates that it is only valid for G5. Is there a generalized constant defined in Nightscout?
Even if not, we should name use a generalized name like DEXNT for new technology or simply DEX because the battery information was not available with G4.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -1188,17 +1197,43 @@ private void postDeviceStatus(NightscoutService nightscoutService, String apiSec
                     battery_level = Pref.getInt("parakeet_battery", -1);
                     battery_name = "Parakeet";
                     break;
+                case "G5 Transmitter":
+                    txBattery = Ob1G5TransmitterBattery.getTransmitterBattery();
+                    battery_level = txBattery.voltageA();
+                    battery_name = Ob1G5StateMachine.usingG6() ? "G6 Transmitter" : "G5 Transmitter";
```
tolot27, line 1203: Please use constants/enums instead of strings.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -1188,17 +1197,43 @@ private void postDeviceStatus(NightscoutService nightscoutService, String apiSec
...
             }
             final long last_battery_level = PersistentStore.getLong(LAST_NIGHTSCOUT_BATTERY_LEVEL);
+            final String last_g5tx_level = PersistentStore.getString(LAST_G5TX_BATTERY_LEVEL);
 
             final JSONArray array = new JSONArray();
             final JSONObject json = new JSONObject();
             final JSONObject uploader = new JSONObject();
 
-            if ((battery_level > 0) && (battery_level != last_battery_level || always_send_battery)) {
+            if (battery_name.equals("G5 Transmitter") || battery_name.equals("G6 Transmitter")) {
```
tolot27, line 1216: Same here: Please use constants/enums instead of strings.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Ob1G5TransmitterBattery.java
```diff
@@ -0,0 +1,81 @@
...
+import com.eveningoutpost.dexdrip.G5Model.VersionRequestRxMessage;
+import com.eveningoutpost.dexdrip.Services.G5BaseService;
+
+/**
+ * Holds information about a G5/G6 transmitter's battery status,
+ * used when sent to Nightscout and obtained via Ob1G5StateMachine.
+ *
+ * @author James Woglom (j@wogloms.net)
+ */
+public class Ob1G5TransmitterBattery implements TransmitterBattery {
```
tolot27, line 15: Class name could be more generalized like `Ob1DexTransmitterBattery`.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -1171,6 +1174,12 @@ private void postDeviceStatus(NightscoutService nightscoutService, String apiSec
         }
         if (DexCollectionType.hasWifi()) batteries.add("Parakeet");
 
+        boolean isOb1 = Pref.getBooleanDefaultFalse(Ob1G5CollectionService.OB1G5_PREFS);
+        boolean sendG5TxBattery = Pref.getBooleanDefaultFalse("send_g5tx_battery_to_nightscout");
```
tolot27, line 1178: Same here: more generalized name.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Ob1G5TransmitterBattery.java
```diff
@@ -0,0 +1,81 @@
...
+   }
+
+   public int resistance() {
+       return batteryRx.resist;
+   }
+
+   public ResistanceStatus resistanceStatus() {
+       int r = batteryRx.resist;
+
+       if (r > 1400) {
```
tolot27, line 62: Are these resistance values identical for G6?

app/src/main/java/com/eveningoutpost/dexdrip/utils/DexCollectionType.java
```diff
@@ -127,6 +127,8 @@ public static boolean hasSensor() {
 
     public static boolean usesClassicTransmitterBattery() { return usesTransmitterBattery.contains(getDexCollectionType()); }
 
+    public static boolean isG5G6() { return getDexCollectionType() == DexcomG5 || getDexCollectionType() == DexcomG6; }
```
tolot27, line 130: could be more generalized like `hasBatteryInfo` or `providesBatteryInfo`.

tolot27 (changes requested):

tolot27:
> Would it be frowned upon if I work on a pull request that I did not open myself?
> Is there an unspoken code that one should only work on pull requests that he/she has opened themselves?

That's indeed possible. It is best to inform the PR creator beforehand. Small changes can be made directly at GitHub.

jwoglom:
Apologies for the delayed response, I will look into this again. I suspect I might need to make a new CR given how long it has been since I published the initial CR and the potential for things like merge conflicts, but if it is mergeable as-is I can respond to the directly requested changes.

Navid200:
@jwoglom   Great to see you back.

The reason I find this valuable is that it comes up a lot when someone asks, on facebook, how is the battery, when they post a screenshot of their system status page showing the current battery voltages.  
Considering the voltages of all these transmitters are measured by themselves, to begin with that is on shaky grounds.

You could see the same voltage from two different transmitters.   And, you shouldn't be surprised if one goes for 2 more months and the other dies tomorrow.
It's the trend that matters, not the absolute values.

If you look at the xDrip logs, you can only see Voltage B for only 1 or 2 days.  The older values are deleted.  

If Voltage B was stored on Nightscout, it would be great so that we could see the trend.

I have been recording Voltage B for my transmitters for my past two and the current one.  But, not every individual can do that.  And the person who is asking the question has definitely not recorded the voltage previously.

Then, there are people who claim that xDrip has killed their transmitter by draining the battery.   Again, having the history, in a case like that, very likely would show that their battery was already on the way out when they started using it.

The disadvantage I see with this is that it is only going to work if the user has Nightscout.
So, in my opinion, we should also make a slight change to xDrip and force it to maintain a local history of Voltage B for more than just 2 days.  For example, let's say we would store Voltage B in xDrip to maintain the values for the last 2 months.  There are two measurements a day.  So, that's a total of 120 measurements.  This would be very useful even if the user has never set up Nightscout.
I'm not sure what everyone thinks about that.

Anyway, if you like, I can do the work and you just review.  I will be happy to do it if you supervise if you think it would help.  Of course, you don't have to agree.   The thing is even if we do the work, it may still not be merged.  There is no guarantee.  But, I don't mind because I am learning.  So, I look at it as training.  So, even if it is not merged, I am still OK with it.
Any way this is done, I will be very happy to have it.

TecMunky:
It would seem to me that the Voltage A value is more important than the
Voltage B value.  Anyway, that is what I have tracked in the past when I
was looking at such things.

On the oldest G5 transmitters, the trend was generally, that when Voltage A
fell below 300, it would take another 4 to 5 weeks to drop to 260.  When
Voltage A drops to 260, an internal status is set indicating the battery is
"dead".  When Voltage A is above 260, the data is still reliable.  When it
falls below 260, the data is increasingly inaccurate and unreliable.  The
"dead" battery status persistent once set - I think the only way to reset
it is to reset the transmitter, which also resets the time interval since
start.

On newer G5 transmitters, it seems it takes more like 7 to 10 days to drop
from 300 to 260.  I have not tracked voltages on the G6 transmitters.

I never analysed the data from the Voltage B values.  I always considered
the Voltage A to be more important, as this is more indicative of the
actual battery level.  As I understand, Voltage A is "unloaded" voltage
(probably a direct measurement of the battery), and Voltage B is "loaded"
voltage (probably a measurement at some internal circuit test point).

This is my $0.02 for today

Navid200:
@TecMunky The proposal is to save the voltages on Nightscout.  It is not to choose between Voltage A versus Voltage B.  At least, I didn't get that impression from the pull request.  I thought both voltages will be saved.  
So, you can use Voltage A, and I can continue to tell everyone on facebook that Voltage B is more representative of how the battery is doing.

If you like to convince me wrong, or ask me why I say that, we can have a conversation about that.  But, I don't think this pull request is the best venue for that.

jwoglom:
@Navid200 @tolot27 Given that I initially made this PR almost three years ago, I decided to try and re-implement it while cleaning up some code in xDrip along the way. I created a new PR at https://github.com/NightscoutFoundation/xDrip/pull/1709, please take a look.

I will close this CR since I don't anticipate any of the code visible here being merged.

Issue #492, opened 2018-06-26
Wear complication show wrong BG plugin


gregorybel:
Dont know if it an old issue I never saw or a new one but using datriasec as plugin for libre, wear complication shows value from xdrip plugin and not from datriasec plugin, which somtimes are not the same values!

Can you fix it or give my a hint where issue could be?

nsghub:
And also #468

gregorybel:
yes indeed!

gregorybel:
fixed in https://github.com/NightscoutFoundation/xDrip/commit/1d7a28be51c56eddc1c9aa1effdf77a731185f72

Issue #468, opened 2018-06-05
Plugin glucose value not pushed to android wear (complication & watchface)


nsghub:
I'm using a calibration plugin and the plugin value is shown on the phone as expected.
On Android Wear 2.x the complication always shows the xdrip+ default value but not the plugin value. The dedicated xdrip watchface is able to show the plugin values, but it's not reliable. After a while it switches back to default (display always on) and it takes another while until it changes back to plugin.

gregorybel:
fixed in https://github.com/NightscoutFoundation/xDrip/commit/1d7a28be51c56eddc1c9aa1effdf77a731185f72

Navid200:
@nsghub Would you please close this issue if it has been addressed?
Thanks

Issue #391, opened 2018-04-11
Data from calibration plugin not uploaded to nightscout.


Klaus3d3:
Hi there,

i'm still working on the interface for my amazfit smartwatch and while this testings i found out, that i get the corrected curve and the notification from the selected calibration plugin, but in the uploaded nightscout data and in the Glucose Data table i find the values from xdrip, wich are different.
latest nightly

gregorybel:
fixed in https://github.com/NightscoutFoundation/xDrip/commit/1d7a28be51c56eddc1c9aa1effdf77a731185f72

tolot27:
@Klaus3d3 Please can you confirm that it is working now and close the issue?

PR #595, opened 2018-09-24
Gregorybel: fix displayed BG value into wear complication

gregorybel:
This is an old bug, somehow til now wear complication shows the original xdrip BG value even if a plugin is activated and users wants to use BG value from plugin. This PR fixes this.

Successfully tested on my watch. 

see issues:
https://github.com/NightscoutFoundation/xDrip/issues/492
https://github.com/NightscoutFoundation/xDrip/issues/391
https://github.com/NightscoutFoundation/xDrip/issues/468

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -253,16 +253,23 @@ public class BgReading extends Model implements ShareUploadableBg {
         String unit = prefs.getString("units", "mgdl");
         DecimalFormat df = new DecimalFormat("#");
         df.setMaximumFractionDigits(0);
+        double usedBgValue = 0;
 
-        if (calculated_value >= 400) {
+        if (dg_mgdl > 0) {
+            usedBgValue = dg_mgdl;//from plugin
+        } else {
+            usedBgValue = calculated_value;//xdrip original
+        }
+
+        if (usedBgValue >= 400) {
             return "HIGH";
-        } else if (calculated_value >= 40) {
+        } else if (usedBgValue >= 40) {
             if (unit.compareTo("mgdl") == 0) {
                 df.setMaximumFractionDigits(0);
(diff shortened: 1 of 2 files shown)


gregorybel:
Hi Jon

no problems if you think it is better.

But where do you catch if plugin is enabled or not?

I am in vacation, so not easy to read Code from smartphones...

If you want you can push changes or I will do it when I am back.

Grégory

gregorybel:
@jamorham your suggestion has been undertaken.

Anything else?

gregorybel:
@jamorham can we merge or any concerns? I have other improvements in the queue but would like first to get a few things merged otherwise PR will become bigger and bigger.

jamorham:
These changes are already merged in commit: https://github.com/NightscoutFoundation/xDrip/commit/1d7a28be51c56eddc1c9aa1effdf77a731185f72

gregorybel:
as already commited, close this PR

yosepito:
Hi guys: although this is closed, I would like to reopen it. The question is that I'm using Libre-2 with OOP and enabled Xdrip plugin to enable calibrations, which are by far more accurate than any other option. At least in my current sensor.
However, my problem is that data showed in BG table is the raw data of Libre-2. If I upload that to NS, it will be wrong. For instance, now I'm 116 mg/dl, but BG table says 75. You can see the problem when I have a true 80 and the alters it will trigger in NS.

Is there an option to override this? I activated "use_pluggable_alg_as_primary" as true in settings, but it does not seem to have an effect. Any suggestion will be welcome

gregorybel:
Hi

did you try settings/interapp/send display glucose?

Gregory

yosepito:
Thanks, @gregorybel 
I have it enabled (the box is ticked) and no luck.
However, I'm referring to uploading data to a NS server in heroku
see #1350 for details.

I think the problem will affect thousand of people and a solution could be using the Xdrip plugin as a complement for calibration, provided that the data table generated can be uploaded to a NS server in the cloud.

PR #608, opened 2018-10-02
Tzachi libre uploader2

tzachi-dar:
This code allows xdrip to also upload the libre data that it reads to a mongodb.

This data then allows another xdrip to show results. 
Can be used for all house coverage or two parents running xdrip with two phones transparently.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
@@ -150,7 +150,7 @@ public class LibreAlarmReceiver extends BroadcastReceiver {
                                 try {
                                     final ReadingData.TransferObject object =
                                             new Gson().fromJson(data, ReadingData.TransferObject.class);
-                                    processReadingDataTransferObject(object, JoH.tsl(), "LibreAlarm");
+                                    processReadingDataTransferObject(object, JoH.tsl(), "LibreAlarm", false);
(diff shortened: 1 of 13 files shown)


tzachi-dar:
This was checked a few more days and works well.

jamorham:
Hi, can you explain why the uploads are enabled for `Tomato` but not `Blucon` or `LibreAlarm` ? What is the reason for this? thanks

tzachi-dar:
For librealarm, the main reason is testing. it should probably work. I can test it, if you think that someone is still using it.
For blukon, there are two modes:
1) where raw data is used. In this case, we simply just don't have all the data, so the reading part will now know how to parse it.
2) When oop is used, we have all the data, but we need to test it. If you think it is important, we can ask someone to test it.

jamorham:
Okay. If you create another signature for ` HandleGoodReading` which handles the arguments as they were and then calls your updated method with the default as `false` then you can eliminate a lot of the changes to unrelated classes. The same is probably true for `LibreBlock.createAndSave`. 

What I would like to see is the minimum amount of changes to unrelated code paths. Overall I am not a fan of expanding the methods to pass parameters down multiple levels, but due to the organic way the code has grown and its structure, this may be unavoidable. But I think it can be neatened up by preserving the previous signatures with compatibility methods as I mentioned above.

tzachi-dar:
Created the two old functions.
I'm doing final testing for the new version.

tzachi-dar:
Verified with the new code. works well.

tzachi-dar:
Thanks.

PR #610, opened 2018-10-03
 Allow wear sync if timestamp difference is older than zero.

nielsvanvelzen:
Every time when entering data using my smartwatch (Ticwatch E) the sync would fail because the timestamp difference (`since` variable) was around `-900`.

This PR fixes the issue by allowing timestamps up to five seconds older than the current time.


jamorham:
Overall this looks good. As far as I can work out, future treatments shouldn't cause a problem here especially by only a few seconds.

nielsvanvelzen:
I did not know if it would cause issues so therefore I restricted it to only 5 seconds.

PR #622, opened 2018-10-13
Some more Hebrew translations.

tzachi-dar:
Also support the possibility to write something like:
"Before 3 minutes" instead of "3 minutes ago"

This is needed in Hebrew, and according to google translate in mermen it is also "vor 3 Minuten".


jamorham:
Ideally it could be done with `MessageFormat.format()` and the value string can have a choice selector for 1 or more minutes ago. I can take a look at that.

tzachi-dar:
I guess that they all should work.
Whatever you fill feets the code better.

jamorham:
What I meant was we could change the minutes ago string generation to using `MessageFormat.format()` for example:

```
final String fmt = "{0, choice,0#{0, number, integer} Minutes ago|1#{0, number, integer} Minute ago|1<{0, number, integer} Minutes ago}";
        
System.out.println(MessageFormat.format(fmt,0));
System.out.println(MessageFormat.format(fmt,1));
System.out.println(MessageFormat.format(fmt,2));
System.out.println(MessageFormat.format(fmt,3));
```

Would give the output:

```
0 Minutes ago
1 Minute ago
2 Minutes ago
3 Minutes ago
```

I started trying to do this with the hebrew in your strings file, but my editor couldn't cope and started reversing the string order so I'm probably not the best person to attempt that.

Doing it this way would give flexibility for all languages and avoid using concatenation. `MessageFormat` choices are designed for this type of multi-language string processing.

tzachi-dar:
Replaced by: https://github.com/NightscoutFoundation/xDrip/pull/679

PR #632, opened 2018-10-18
Change hardcoded strings in prefs to translateable

MasterPlexus:
Searched all hardcoded strings in Preferences,and made them translatable. after accepting we can translate them over crowdin.


MasterPlexus:
@jamorham, I also started to analyze other places where hard coded English texts are used, to switch them also to translatable versions, knowing that maybe some of them are only for interims-stuff. But I think as translation is now much more easy, we should not exclude them. It is easy after a place was recoded and a text will be deleted to delete then also the translations. If you agree, I'll made additional pull's on that.

jamorham:
Generally I think about putting strings in to resource files after they have been in operation for about 3 months.  These preference ones should be fine.

PR #644, opened 2018-10-27
Rewrite of log-capture, print and assert for tests

ahaarrestad:
Utilizing ShadowLog which is part of Robolectric


ahaarrestad:
@jamorham is this what you were looking for regarding logs in tests?

app/src/test/java/com/eveningoutpost/dexdrip/RobolectricTestWithConfigAndPowerMockLogRedirect.java
```diff
@@ -1,112 +0,0 @@
-package com.eveningoutpost.dexdrip;
```
jamorham: Can you undelete this file. It is a useful example I think of how to get powermock and roboelectric actually working together. We don't need to use it but I don't think it should be deleted.

ahaarrestad:
Sure, but why would you need powermock? You can use robolectric shadows...

Asbjørn

Den tir. 30. okt. 2018, 21:52 skrev JamOrHam <notifications@github.com>:

jamorham:
Can RoboElectric shadows mock any static method?

ahaarrestad:
I believe so: http://robolectric.org/extending/

ahaarrestad:
Will change the commit (restore the file) as soon as possible.

jamorham (approved):
Thanks!

PR #646, opened 2018-10-28
Master plexus translation work

MasterPlexus:
Started to made all texts in Reminders translateable.
Please check, as there is some influence also in JoH class for the scalartime-functions which is often used.
Also I changed the handling to get the right property from the Snoozepanle buttons, as when they are translated, a switch case on the button-text will no longer work. So I've put the data to use in the Tag-Property.

Tested on my phone with German strings, and worked as expected.

If there is something which you would handle in a other way, please advice, and I will change it for sure.


app/src/main/java/com/eveningoutpost/dexdrip/Reminders.java
```diff
@@ -388,25 +388,25 @@ public void onBindViewHolder(final ActivityWithRecycler.MyViewHolder base_holder
...
                         String natural_due = JoH.niceTimeScalarNatural(duein);
                         if (natural_due.matches("^[0-9]+.*")) {
-                            natural_due = "in" + " " + natural_due;
+                            natural_due = xdrip.getAppContext().getString(R.string.in) + " " + natural_due;
                         }
-                        firstpart = "Due " + natural_due;
+                        firstpart = xdrip.getAppContext().getString(R.string.due) +" " + natural_due;
                     } else {
-                        firstpart = "DUE " + JoH.hourMinuteString(reminder_item.next_due);
+                        firstpart = xdrip.getAppContext().getString(R.string.due) + " " + JoH.hourMinuteString(reminder_item.next_due);
```
jamorham: This is no longer capitalized?

jamorham:
I'm thinking to merge this in the nightly after next. I think it should be okay

MasterPlexus:
@jamorham, as you said it should be ok, when it will be merged? No hurry, just to have it in mind to check that no other commit influence it...

PR #649, opened 2018-10-29
sclae event log items according to density

AdrianLxM:
This scales the event log items according to the display density.
`density` has MDPI as reference value (it exists since API Level1) - and I assume this was designed with at least HDPI in mind, I adjusted the scale factor by ~1.5.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/EventLogActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/EventLogActivity.java
@@ -564,7 +564,7 @@ public class EventLogActivity extends BaseAppCompatActivity {
 
         // reformat text size for long messages
         public float textSize(String message) {
-            final float scale = 4f;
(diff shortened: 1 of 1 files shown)


jamorham:
Does this make text larger or smaller? Which device were you having issues with? (I'm sure it does make sense to scale it vs the density)

jamorham:
The density on my phone which is 720x1280 screen 4.x inch screen is `2.0` and so this PR would increase the font size. Using a multiplier of 2 would make it the same for me, but what will that do for your screen?

AdrianLxM:
I do have no idea why my previous reply to your first comment was deleted/not stored?
On the Atom a factor of 2 would make the long text very very small - but still readable.

Without the density adaption the text would not fit the screen at all - so definitely an improvement with factor two as well.

jamorham:
Shall I just make the scale change to 2.0 locally or do you want to provide another commit with that?

jamorham:
I have done this internally.

PR #655, opened 2018-10-29
gregorybel: Blukon improvements (2)

gregorybel:
- BT watchdog timer value configurable
- add Libre SN for Blucon
- code optimisation for Blukon Module
- stop resetting bridge_battery and nfc_sensor_age after each initialisation to simply keep current value

successfully tested!


wear/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -208,8 +208,8 @@ public synchronized static void updateDB() {
     }
 
     public double getDg_mgdl(){
-        if(dg_mgdl != 0) return dg_mgdl;
-        return calculated_value;
+        if(dg_mgdl != 0) return dg_mgdl;//from plugin
```
jamorham: These comments are not correct. Display glucose covers much more than just plugins, it handles noise compensation and more. I would suggest removing the comment at least from `dg_mgdl` as it is misleading. `calculated_value` is also not guaranteed to be the original xDrip algorithm either.

jamorham:
This looks fine other than the item I referenced in my inline comment.

wear/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
gregorybel: ok, removed!

gregorybel:
@jamorham can you noy hopefully merge it?

gregorybel:
cool! you are welcomed, could you please merge now?

PR #665, opened 2018-11-04
gregorybel: do not allow BT watchdog timer smaller than 6min

gregorybel:
If user wants to set a timer smaller than 6min then default value will be applied,
This to avoid to frequent watchdog calls, which will be useless as most CGM send new value every 5min.


gregorybel:
as we use <= 5, smallest value is 6? I think it is good.

gregorybel:
@jamorham  can you merged or any thing else?

jamorham:
Yes you are correct about that smallest value being 6. Thanks.

PR #667, opened 2018-11-04
gregorybel: blukon optimisations, improvements

gregorybel:
- optimize logic and do not request sensor age when not wanted
@trallen @tzachi-dar @keencave :review is welcomed as I dont remember who does that code

- force reading serial after restart and after sensor age backward
- refuse too early new reading at end of com process, to avoid confusing blukon
- force reading sensor age after getting historic
- more debug
- reset nfc_age_problem

Successfully tested and squashed!

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
@@ -104,6 +104,7 @@ public class Blukon {
     public static void initialize() {
             Log.i(TAG, "initialize Blukon!");
             JoH.clearRatelimit(BLUKON_GETSENSORAGE_TIMER);
+            JoH.clearRatelimit(BLUKON_DECODE_SERIAL_TIMER);// set to current time to force timer to be set back
             m_getNowGlucoseDataCommand = false;
             m_getNowGlucoseDataIndexCommand = false;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
```diff
@@ -426,29 +424,42 @@ public static void unBondIfBlukonAtInit() {
...
-                } else {
-                    Log.e(TAG, "Do not set 'nfc_sensor_age'");
-                }
-                currentCommand = GET_NOW_DATA_INDEX_COMMAND;
-                m_getNowGlucoseDataIndexCommand = true;//to avoid issue when gotNowDataIndex cmd could be same as getNowGlucoseData (case block=3)
-                Log.i(TAG, "getNowGlucoseDataIndexCommand");
+                Log.e(TAG, "Do not set 'nfc_sensor_age'");
             }
 
+            currentSensorAge = Pref.getInt("nfc_sensor_age", 0);
```
trallen: I'm puzzled by the duplicate setting of currentSensorAge and currentCommand? If required, can you document?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
```diff
@@ -405,15 +395,23 @@ public static void unBondIfBlukonAtInit() {
...
             }
 
-            if (JoH.pratelimit(BLUKON_GETSENSORAGE_TIMER, GET_SENSOR_AGE_DELAY)) {
-                currentCommand = GET_SENSOR_TIME_COMMAND;
-                Log.i(TAG, "getSensorAge");
+            /* LibreAlarmReceiver.CalculateFromDataTransferObject, called when processing historical data,
+             * expects the sensor age not to be updated yet, so only update the sensor age when not retrieving history.
+             */
+            if (Pref.getBooleanDefaultFalse("external_blukon_algorithm") || getHistoricReadings) {
+                // Send the command to getHistoricData (read all blcoks from 0 to 0x2b)
```
trallen: Since you're moving things around, recommend changing "blcoks" to "blocks".

gregorybel:
@trallen thanks for remarks, will change. 

Request for review was especially regarding code to retrieve history, I think new code will do same but it is easier.

I use other changes since several weeks successfully.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
gregorybel: good catch, thanks!

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
gregorybel: done

gregorybel:
> I would suggest you add some tests for this code, since it is used by quite a few people. We added some tests for the BlueReader code which is in the same area as this one. It will help you gain confidence in how the changes work.

For sure a good idea! I use this code daily since several days and it works ao confidence level is high.

But to learn somethingm can you point to me how your tests look like?

ahaarrestad:
You can take a look at for example a unit test here:
https://github.com/NightscoutFoundation/xDrip/blob/master/app/src/test/java/com/eveningoutpost/dexdrip/Models/BlueReaderTest.java

We are using robolectic to do shadow implementstions.

Just ask if you have questions.

Asbjørn

Den fre. 9. nov. 2018, 21:06 skrev Grégory Beloncle <
notifications@github.com>:

gregorybel:
@jamorham any else or can you merge this?

gregorybel:
@jamorham anything else or can you merge? it will help Blucon users.

PR #670, opened 2018-11-05
gregorybel: improvements in dexcollectionservice

gregorybel:
- add lot of debug in system status, should help debugging without logs
- add a fix which might help getting reconnected faster

Happy to get reviews!

Look this:

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
@@ -185,6 +185,7 @@ public class DexCollectionService extends Service implements BtCallBack {
     // Experimental support for rfduino from Tomasz Stachowicz
     private static volatile BluetoothGattCharacteristic mCharacteristicSend;
     private byte[] lastdata = null;
+    private static int mStatus = -1; // for display in system status
     public SharedPreferences.OnSharedPreferenceChangeListener prefListener = new SharedPreferences.OnSharedPreferenceChangeListener() {
         public void onSharedPreferenceChanged(SharedPreferences prefs, String key) {
             if (key.compareTo("run_service_in_foreground") == 0) {
@@ -233,6 +234,12 @@ public class DexCollectionService extends Service implements BtCallBack {
         } else {
             Log.d(TAG, "Services already discovered");
             checkImmediateSend();
+        }
+
+        if (mBluetoothGatt == null) {
+            //gregorybel: no needs to continue if Gatt is null!
+            UserError.Log.e(TAG, "gregorybel: force disconnect!");
+            handleDisconnectedStateChange();
         }
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -185,6 +185,8 @@ public void onReceive(Context context, Intent intent) {
     // Experimental support for rfduino from Tomasz Stachowicz
     private static volatile BluetoothGattCharacteristic mCharacteristicSend;
     private byte[] lastdata = null;
+    private static BluetoothManager myBluetoothManager = null;//for display in system status
```
ahaarrestad: For readability it would be ok to use comments either before line or after line. If you choose to use after line, a space between ";" and "//" helps. But this is only a suggestion.

gregorybel: done, thanks

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -185,6 +185,8 @@ public void onReceive(Context context, Intent intent) {
     // Experimental support for rfduino from Tomasz Stachowicz
     private static volatile BluetoothGattCharacteristic mCharacteristicSend;
     private byte[] lastdata = null;
+    private static BluetoothManager myBluetoothManager = null; // for display in system status
```
jamorham: I'm not really sure this should be being held in a static field - or indeed at all. I would imagine the static reference could create issues with garbage collection as its a system service. Can't the instance of this just be retrieved where you are displaying connected devices in engineering mode megaStatus() ?

gregorybel: But problem that I need to call getSystemService() which cant be called from within a static function.
Ýou have more knowledge than me so I simply removed it.

jamorham: You can just do ` xdrip.getAppContext().getSystemService()` from a static context.

gregorybel: oh thanks for hints, I changed code accordingly and squashed

gregorybel:
@jamorham can you merge this or any concerns? it will help all Libre users

gregorybel:
any reasons not to merge?

app/src/main/java/com/eveningoutpost/dexdrip/Services/DexCollectionService.java
```diff
@@ -234,6 +235,12 @@ private synchronized void handleConnectedStateChange() {
             Log.d(TAG, "Services already discovered");
             checkImmediateSend();
         }
+
+        if (mBluetoothGatt == null) {
```
jamorham: What is the logic behind this change? When do you expect this condition to be met?

gregorybel: I saw a lot time that with blucon really often I don"t get any new data because xdrip thinks blucon is connected but in fact mBluetoothGatt is null (that is the reason why I expended system status) so xdrip doesn't try to reconnect and blucon can't send new data. This is more a workaround, I did not find how this can happens but it definitly happens.

So I assume, either it never happens and code doesn't break anything or it happens and xdrip gets quicker reconnected.

What do you think?

PR #672, opened 2018-11-05
gregorybel: create howto for blukon and fix missed reading issues

gregorybel:
- I took info from my wiki, moved it here and update it to current status.
- create a howto fix missed reading issues for Libre


gregorybel:
no I think Trust Auto connect should be kept default ON.
For Blukon, I think it is better to ON and I was reffering to Blukon how-to, I sugget putting it to OFF.
I corrected this, case users dont read blukon-howto, thanks for hint!

gregorybel:
@jamorham when you have decided to merge, would you mind to squash the commits? It seems github is offering this option by merging. Would save time for me and get history clean.

PR #679, opened 2018-11-09
Some more hebrew translations.

tzachi-dar:


tzachi-dar:
This replaces the previous pr.
It adds some more translations and brings file against the head.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -3159,13 +3162,35 @@ private void displayCurrentInfoFromReading(BgReading lastBgReading, boolean pred
...
+         String original_text = notificationText.getText().toString();
+        Log.d(TAG, "original_text = " + HexDump.dumpHexString(original_text.getBytes()));
+        if(original_text.length() >=1 && original_text.charAt(0) == 0x0a) {
+            Log.d(TAG,"removing first and appending " + display_delta);
+            notificationText.setText(display_delta + "  " + original_text.substring(1));
+        } else {
+            notificationText.setText(display_delta + "  " + original_text);
+        }
+    }
+
```
jamorham: This glitch is a bit worrying. Do we have any more information about it?

app/src/main/res/values-iw/strings-iw.xml
```diff
@@ -161,9 +161,11 @@
...
     <string name="added">Added</string>
     <string name="checking_for_update">Checking for update..</string>
-    <string name="signal_missed">Signal Missed</string>
-    <string name="space_minutes_ago">" Minutes ago"</string>
-    <string name="space_minute_ago">" Minute ago"</string>
+    <string name="signal_missed">אין קליטה</string>
+    <string name="space_minutes_ago">" דקות"</string>
+    <string name="space_minute_ago">" דקה"</string>
+    <string name="before_minutes_ago">"לפני "</string>
+    <string name="minutes_ago">{0, choice,0# לפני {0, number, integer} דקות |1# לפני דקה|1&lt; לפני {0, number, integer} דקות }</string>
```
jamorham: What is the situation for the other supported locales? Are we going to see a regression to english there unless we have the choice message format translated for them too?

tzachi-dar: Checked this under german and it seems that we are failing down to english.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
tzachi-dar: Not too much information, this is my observation of what is happening there.
I can try and remove this from this PR and send it in another pr, that will not have hebrew specific code.
Please also note that this is happening for the big tablets, so in any case only very few devices will be affected by this.

app/src/main/res/values-iw/strings-iw.xml
jamorham: If we merge this PR as is then it will regress all the other locales to english for the strings which are replaced by this format string. 

To make it complete we would need to probably create a small script to recreate the format string for each other locale using the existing text we have for the plurals for those locales.

tzachi-dar:
So, I have fixed all the minutes ago text in all languages.
The script that has created the fix is also attached in the case we will need to use it again.

Here are the german screenshots with the new code:

screenshot_20181129-015225_xdrip
screenshot_20181129-015332_xdrip
screenshot_20181129-015413_xdrip

tzachi-dar:
Thanks for review and helpful comments.

app/src/main/res/values/strings.xml
```diff
@@ -1121,7 +1128,19 @@
...
+
+    <string name="all_alerts_currently_disabled">ALL ALERTS CURRENTLY DISABLED</string>
+    <string name="low_and_high_alerts_currently_disabled">LOW AND HIGH ALERTS CURRENTLY DISABLED</string>
+    <string name="low_alerts_currently_disabled">LOW ALERTS CURRENTLY DISABLED</string>
+    <string name="high_alerts_currently_disabled">HIGH ALERTS CURRENTLY DISABLED</string>
+    <string name="bg_reading_missed">BG Readings Missed</string>
+    <string name="not_connected">Not Connected</string>
+    <string name="no_data">No data</string>
+    <string name="no_bluetooth">No bluetooth</string>
+    <string name="connection_status">Connection Status:   </string>
```
MasterPlexus: @tzachi-dar could you explain why you add three spaces at the end of the string? Normally something like that should be handled in the code? I'm just wondering, as this now has to take care in every translation.

tzachi-dar: I did not change anything here, it existed before.
I only moved this to the resource file.
See https://github.com/NightscoutFoundation/xDrip/pull/679/files#diff-c8160ebde93fb72b4dec407c2c46b779L115

MasterPlexus: Ah, OK. Just wondering, not a big thing...

PR #698, opened 2018-11-28
Change Medtronic 600-series pump image.

pazaan:
Change 5/7-series image to 600-series image


pazaan:
@jamorham - Added the image as a PNG, since that seems to be the convention for images in this repo. However, due to the detail in the pump UI, the PNG is 21KB. A JPG with the same visual fidelity is only 7KB. Let me know if you'd prefer the JPG.

PR #700, opened 2018-12-02
Correct formatting in strings-de.xml

MasterPlexus:
there was some mistakes in the format on the new strings in strings-de.xml, wich are corrected with that pull request.


MasterPlexus:
@jamorham This is just a format-correction in strings-de.xml. Maybe the missformat is also the cause why crowdin did not take the German translation into account...

tolot27:
Crowdin generates the translated xml files and it works flawlessly, so far.

PR #708, opened 2018-12-12
Update Phones_For_G5.md

fclingen:
Added Motoroloa Moto Z2 to list


Navid200:
@fclingen Hi,
My understanding is that all Android phones can be used to run xDrip unless their operating system is extremely old (before Android 4.4?). That is what I tell everyone in the facebook group.
There was a time we maintained these lists of phones that were compatible with xDrip. But, I don't believe we do that any longer.

Thanks

tolot27:

PR #717, opened 2018-12-18
Tzachi libre average

tzachi-dar:
This PR is needed in order to have more stable results, and more immunity to libre sensor noise.

The main idea is to use an average of 5 points instead of one point.
This promises that noise will be smaller and programs like openaps will get more stable results.

Still needs more testing.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
@@ -63,10 +63,15 @@ public class LibreAlarmReceiver extends BroadcastReceiver {
         return (lib_raw_value * LIBRE_MULTIPLIER); // to match (raw/8.5)*1000
     }
 
-    private static void createBGfromGD(GlucoseData gd, boolean quick) {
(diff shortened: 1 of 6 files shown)


tzachi-dar:
Here is an example from two xdrips, running on the same data, one with the new algorithm and one without it:

Without it:



screenshot_20181218-230729_xdrip 1

tzachi-dar:
Here are the pictures after the change:
screenshot_20181218-230842_xdrip 1
screenshot_20181218-231027_xdrip 1

tzachi-dar:
And here is a graph of both showing the smaller noise:
image

bcllivewire:
Where did you down load XDrip from , I have tried looking on Nightscout site but cannot find , I long in but no results ,
Can somebody help with this 
Brian

tzachi-dar:
@bcllivewire Is your question related to this pr?
If not, the latest stable version of xdrip can be downloaded from: https://bit.do/xdrip-plus

bcllivewire:
Thanks
I am up and running now with Xdrip 
Brian

tzachi-dar:
I have rebased this over master in https://github.com/NightscoutFoundation/xDrip/pull/828

PR #726, opened 2018-12-26
Prepared the start sensor dialog to be be translated.

tzachi-dar:
Also added the script that helped in the mission.
The script also replaced texts in some other files.


jamorham:
I would suggest instead statically importing `xdrip.gs()` and then replace `getApplicationContext().getString()` with just `gs()` 

This reduces the visual complexity of the code and also works from a static context.

I'm also not sure why you haven't included the body of the permission advice dialogs in the extracted strings.

tzachi-dar:
Thanks for the review.

I'm not sure how I can import xdrip.gs(). As far as I know I can import xdrip and use xdrip.gs()
[on python, I can import xdrip.gs, on java I can write a static function on each class].
In any case, did what you asked and opened https://github.com/NightscoutFoundation/xDrip/pull/736 against latest master.

If there is still a need to only use gs(), please let me know how.

tzachi-dar:
This was merged on a different pr.

PR #736, opened 2019-01-08
Prepared the start sensor dialog to be be translated.

tzachi-dar:
Also added the script that helped in the mission.
The script also replaced texts in some other files.

This PR also reply to comments from other prs.


jamorham:
In the code you can just put the cursor on the `gs()` and press `ALT` + `ENTER` and select the static import for gs and then you will reduce it to simply `gs()` instead of `xdrip.gs()`

tzachi-dar:
Well, every day I learn a new trick...

Created https://github.com/NightscoutFoundation/xDrip/pull/739

tzachi-dar:
Are you still looking for anything from me regarding this PR?
I believe I have fixed everything we have talked about in https://github.com/NightscoutFoundation/xDrip/pull/739

PR #739, opened 2019-01-08
Tzachi start sensor translation3

tzachi-dar:
Revised version of https://github.com/NightscoutFoundation/xDrip/pull/736.


tzachi-dar:
Fixed a build break.

jamorham:
Thanks for your patience with this. I've been overseeing the Jan 10th roll out and had a change freeze.

PR #745, opened 2019-01-14
Resolve 742 steps from watch not updating on phone

kskandispersonal:
Also resolved [CrashlyticsCore](https://stackoverflow.com/questions/49519316/crashlyticscore-expected-method-missing-registeronmeasurementeventlistener) with proguard rule.

### DIFF
--- a/wear/proguard-rules.pro
+++ b/wear/proguard-rules.pro
@@ -107,3 +107,5 @@
    public static boolean isCollecting();
    public static ** nanoStatus();
 }
+-keep class com.google.android.gms.measurement.AppMeasurement { *; }
+-keep class com.google.android.gms.measurement.AppMeasurement$OnEventListener { *; }
--- a/wear/src/main/java/com/eveningoutpost/dexdrip/ListenerService.java
+++ b/wear/src/main/java/com/eveningoutpost/dexdrip/ListenerService.java
@@ -1109,7 +1109,7 @@ public class ListenerService extends WearableListenerService implements GoogleAp
                     }
                     messageIntent.putExtra("data", dataMap.toBundle());
                     LocalBroadcastManager.getInstance(this).sendBroadcast(messageIntent);
-                    if (!mPrefs.getBoolean("enable_wearG5", false)) {
+                    if (mPrefs.getBoolean("enable_wearG5", false)) {
                         Inevitable.task("sync-all-data", 2000, () -> ListenerService.SendData(context, ListenerService.SYNC_ALL_DATA, null));
                     }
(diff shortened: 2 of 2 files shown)


jamorham:
Can you explain a little about what this change does. Did I break code when converting to `Inevitable.task` ?

kskandispersonal:
Looks like this code was added with [...sync Wear step counter and log data when force_wear/enable_wear is disabled...](https://github.com/NightscoutFoundation/xDrip/commit/44476f0d3eef8a74e64b7f6d0c21dc9bec665806) by me, and you added the Inevitable.task (which has no problem).  Hwr, this conditional fails when Force is disabled now.  Has handling of `enable_wearG5` and `force_wearG5` preferences changed?  If I disable `force_wearG5`, preference `enable_wearG5` remains enabled so this conditional never succeeds.  Maybe this has been a long time bug which I just never tested?:(  I think this fix is not correct (after looking at the original code).  Instead, we should check both of these preferences.  I'll make the change, test, and resubmit.  Thanks for pointing this out!!

kskandispersonal:
@jamorham Here is a filtered Wear log with Force disabled.  After the watch sends `SYNC_ALL_DATA` to the phone, steps are updated on the phone's home screen.  I'm wondering now though if a user discontinues use of the watch, is the Phone pebble steps table ever updated on its own?  It seems like it should initialize itself to 0 at midnight if Health Data is enabled.  Otherwise, old data may continue to display on the phone home screen.  Is that correct?  I haven't investigated.
[WearStepsForceDisabled.log](https://github.com/NightscoutFoundation/xDrip/files/2766471/WearStepsForceDisabled.log)

kskandispersonal:
@jamorham This is ready for your review.  I've been using it and see no issues w/ G5 & SWS watch.  Thank you!

dv1:
I think this causes xDrip to crash at startup if Google Play is not installed, which can happen on specially set up looper phones (like ones with LineageOS on it but without Google Play). I'll verify once I have a phone that I can test that with.

Navid200:
> I think this causes xDrip to crash at startup if Google Play is not installed, which can happen on specially set up looper phones (like ones with LineageOS on it but without Google Play). I'll verify once I have a phone that I can test that with.

Discussions (https://github.com/NightscoutFoundation/xDrip/discussions) are better for these conversations.   

Are you sure you have disabled Sync?  https://navid200.github.io/xDrip/docs/GooglePlayServices_NotInstalled
Please kindly open a discussion (not an issue) and we can go into detail there.

PR #781, opened 2019-02-14
made mostly all prefs translateable

MasterPlexus:
Putting all hardcoded strings of prefs into strings.xml.
After acceptance by @jamorham  and merging, we could use Crowdin to made the proper translation steps.

Not sure, but I think also the wear strings.xml should be included in crowdin?

PR #783, opened 2019-02-16
Tzachi translatoins4

tzachi-dar:
This is a trivial pr that prepares more strings to be translated.

After this pr, the setup process should be translatable, and also start sensor, stop sensor and calibrate.

Many of the toasts have also been added (although, there are so many, that another pr is needed).


MasterPlexus:
Checked the change compared to https://github.com/NightscoutFoundation/xDrip/pull/781, and it is as I feared: the Strings.xml will be mixed hardly due to our work. So I would suggest to merge at first the pr  from tzachi, after that I will then correct my pr to be in line.

@jamorham : this PR you could check for merging.

tzachi-dar:
Was replaced by https://github.com/NightscoutFoundation/xDrip/pull/788

PR #784, opened 2019-02-16
Copy the he translations to iw.

tzachi-dar:

A trivial pr that only copies the he data to the iw data.
This is needed since on most android phones, hebrew language is used as iw.


jamorham:
Is there any easy way to resolve the differences between `he` and `iw` locale? Are they always the same?

tzachi-dar:
'he' and 'iw' should always be the same, so no real reason to solve the issue.
If we can find an automatic way to do that, it would be great.

PR #801, opened 2019-02-25
Fix translation typos.

tzachi-dar:


tzachi-dar:
Error seems like a race that is not related to this pr.

Issue #795, opened 2019-02-22
backfill data not uploaded to Dexcom Follow


doc77:
Would it be possible to have the Dexcom Follow upload have a look back feature. This would be valuable for when the watch is used as a collector and it only shows on the uploader phone when it is back in proximity of the watch.  The phone will backfill but Follow will not look back.

doc77:
@kskandispersonal  Thank you!

Navid200:
@doc77 I am sorry to bother you again.  I am going through the old issues.  Was this ever resolved?

tolot27:
@Navid200 The PR is still not merged. Will have a look soon.

Navid200:
@doc77 Please update the current status of xDrip with respect to this.

Navid200:
xDrip offers follow capability based on Nightscut as well as xDrip sync.
The ability to follow using Dex share also exists.

Is an official Dexcom follower capable to do what is being requested here?

My concern is that if we provide too much redundancy in xDrip, we will only be spreading ourselves too thin considering we don't have many active developers.

We can always reopen this if someone is interested to add it.

PR #802, opened 2019-02-25
resolve issue 795 upload backfill bgs to dexcom share

kskandispersonal:
This PR resolves issue #795 . In particular, performs `uploadToShare()` for ALL BGs in watch sync queue when watch is collector, backfill when phone is collector. This issue was noticed when using sugarmate.io as a Dexcom Follower. sugarmate.io failed to show BGs that were synced from the watch to phone on re-connect, as well as BGs that were backfilled by the phone.

LOG:
**02-25 10:43**:46.283 1880-7846/? D/ShareRest: About to call ShareRest!! **2019-02-25 10:43:43 BG: 110.0**

Move phone out-of-range of transmitter, then re-connect phone (in-range with transmitter) after 10 minutes:

**02-25 10:53**:47.005 1880-16271/? D/ShareRest: About to call ShareRest!! **2019-02-25 10:53:43 BG: 109.0**
**02-25 10:53**:50.575 1880-16468/? D/ShareRest: About to call ShareRest!! **2019-02-25 10:48:44 BG: 111.0**

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NewDataObserver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NewDataObserver.java
@@ -124,8 +124,8 @@ public class NewDataObserver {
     // share uploader
     private static void uploadToShare(BgReading bgReading, boolean is_follower) {
         if ((!is_follower) && (Pref.getBooleanDefaultFalse("share_upload"))) {
-            if (JoH.ratelimit("sending-to-share-upload", 10)) {
-                UserError.Log.d("ShareRest", "About to call ShareRest!!");
+            if (JoH.ratelimit("sending-to-share-upload", 10) || Home.get_enable_wear()) {
+                UserError.Log.d("ShareRest", "About to call ShareRest!! " + JoH.dateTimeText(bgReading.timestamp) + " BG: " + bgReading.calculated_value);
                 String receiverSn = Pref.getString("share_key", "SM00000000").toUpperCase();
                 BgUploader bgUploader = new BgUploader(xdrip.getAppContext());
                 bgUploader.upload(new ShareUploadPayload(receiverSn, bgReading));
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NewDataObserver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NewDataObserver.java
@@ -38,18 +38,20 @@ public class NewDataObserver {
     // TODO move appropriate functions in to their responsible classes
 
     // when we receive new glucose reading we want to propagate
-    public static void newBgReading(BgReading bgReading, boolean is_follower) {
+    public static void newBgReading(BgReading bgReading, boolean is_follower, boolean quick) {
 
-        sendToPebble();
-        sendToWear();
-        sendToAmazfit();
-        sendToLeFun();
-        Notifications.start();
-        uploadToShare(bgReading, is_follower);
-        textToSpeech(bgReading, null);
-        LibreBlock.UpdateBgVal(bgReading.timestamp, bgReading.calculated_value);
-        LockScreenWallPaper.setIfEnabled();
-        TidepoolEntry.newData();
+        if (!quick) {
+            sendToPebble();
+            sendToWear();
+            sendToAmazfit();
+            sendToLeFun();
+            Notifications.start();
+            textToSpeech(bgReading, null);
+            LibreBlock.UpdateBgVal(bgReading.timestamp, bgReading.calculated_value);
+            LockScreenWallPaper.setIfEnabled();
+            TidepoolEntry.newData();
+        }
+        uploadToShare(bgReading, is_follower, quick);
 
     }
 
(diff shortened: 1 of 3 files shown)


kskandispersonal:
@jamorham I'm not sure if this fix is related to a newly discovered issue w/ `TidepoolUploader` or not, but I'm noticing that Tidepool `UploadChunk()` is not uploading the full payload (of BGs) once the watch (collector) reconnects to the phone.  I believe I've noticed this issue even before this fix.  This issue results in potentially a large gap of missing BGs in Tidepool - see image below.

For example, disconnect watch from phone for 2+ hours.  On reconnect - see image below, xDrip+ log shows Tidepool uploaded at `14:13 Syncing data between: 11:33 - 13:58 `.  Tidepool shows a gap though for `12:53 - 14:03`. So it uploaded only about one hour of data but misses the last hour of data. All the synced (watch to phone) BGs show in xDrip+, NS, and sugarmate though.

Perhaps `UploadChunk` attempts to upload data prior to the watch-sync'd BGs have been created?  The `Event View` shows `STEPS` being processed AFTER the Tidepool `UploadChunk()` process, which perhaps means that BGs, too, may not have been completely processed?

If related to this PR fix, perhaps this fix significantly increases time to process BGs, causing this issue, and perhaps the "fix" needs to be optimized?  

Even so, I think `TidepoolUploader` should ensure all BGs are being processed even if there is a delay.  It seems like `LAST_UPLOAD_END_PREF` is set based on the `session.end` which is set by `UploadChunk.maxWindow()`:

`TidepoolUploader.doUploader()` callback:
`UploadChunk.setLastEnd(session.end);`

I'm wondering if instead, `session` should be based on the last BG timestamp processed in `UploadChunk.getNext(session)`?  Of course, `session` is used by all synced records, bloodtests, basals, treatments, too.  Alternatively, `UploadChunk.getBgReadings()` could track last BG timestamp processed separately in a new pref?

Any advice??  Thanks!

screenshot_2019-02-26-17-47-29

screenshot_2019-02-26-20-39-13

kskandispersonal:
@jamorham I've tested this with my phone for backfill data and it is working as expected.  All data gets uploaded to Dexcom Share and Tidepool.  Unfortunately, my watch suddenly stopped working today so I am charging up my backup watch.  However, I did test the watch as collector and it is working as expected for Dexcom Share.  I have not tested watch with Tidepool, but as this PR uses the same code for both phone and watch, I don't expect there to be an issue.

This PR just resolves the issue.  No attempt was made to optimize the uploads, as you have done in other sections of the code.  Maybe you have some suggestions for optimization??:-) . Thank you!!

jamorham:
Regarding the Tidepool changes, instead shouldn't you just set your `Data Age Mins` in the Tidepool settings to something suitable (like 12 hours) ?

I think the changes you propose will prevent the upload sequencer skipping blocks of empty records which is particularly relevant when Tidepool support is first enabled by the user. There may be other side-effects. Changing the logic is risky.

The current upload mechanism is designed to absolutely prevent duplicated/changed data. As far as I understand it, the reason the historical watch synced data is not available at Tidepool is because it was not present in the phone database when the database was synced with Tidepool. With the current model of how the uploader is supposed to work, this is the correct operation.

If you set the data age minutes then that gives the xDrip database a grace time to be correct before the upload is made. This allows for backfilling or collector re-connection within whatever time period you specify. 

For my use case with Tidepool I am not needing the data to be real-time as it is for more general historical review. I am actually also using the "only on wifi + charging" options to sync at night as well. If there are more use cases I haven't considered then please let me know.

kskandispersonal:
@jamorham I agree, changing the logic is risky.  Yes, increasing the `Data Age Mins` could potentially resolve the issue for both the Backfill and the Watch Data Sync missing data.  

Hwr, if the user isn't aware of this issue, they may have their Age Mins set to the default 15 minutes and could have data gaps in their Tidepool graph.  I think Dexcom Backfill only supports 3 hours, so at a minimum for Backfill support, the user should use 3+ hours in their Age Mins preference.  The watch could, of course, be standalone for days w/out syncing w/ the phone, so data could be missing even with a 3 hour setting.  

Is there another way to resolve potentially missing data in Tidepool Uploader that is less risky?  Is this Tidepool Uploader implementation different to how you implemented Nightscout uploader which seems to work perfectly for both watch sync'd data and for backfill data, and Send Nightscout historical back-fill data?

>I think the changes you propose will prevent the upload sequencer skipping blocks of empty records which is particularly relevant when Tidepool support is first enabled by the user. 

Perhaps we could save the timestamp of the last processed BG separately from the session-end preference?  Then on next Upload, check to see if there are any BGs between last BG timestamp and last session-end (ie., the new session-start).  If BGs exist, process those as well?  Would this interfere with the upload sequencer when Tidepool is first enabled?

FWIW, I have been using this fix for the past week and it is working well, with no data gaps in Dexcom Share nor Tidepool.

Navid200:
@kskandispersonal Hi,
Is there still an issue that this will address?
I am just asking because of the age of this pull request.
Would you please update?

Thanks

Navid

app/src/main/java/com/eveningoutpost/dexdrip/tidepool/UploadChunk.java
```diff
@@ -143,7 +148,13 @@ static long getOldestRecordTimeStamp() {
     }
 
     static List<ESensorGlucose> getBgReadings(final long start, final long end) {
-        return ESensorGlucose.fromBgReadings(BgReading.latestForGraphAsc(15000, start, end));
+        List<BgReading> bgReadingList = BgReading.latestForGraphAsc(15000, start, end);
+        if (bgReadingList.size() > 0) {
+            BgReading lastBgReading = bgReadingList.get(bgReadingList.size() - 1);
+            lastBg = lastBgReading.timestamp + 1;
```
tolot27, line 154: Why do you add +1 here? This can lead to duplicate entries, AFAIK?

Navid200:
I suggest we close this pull request.  It can always be reopened if necessary.

tolot27:
I've discovered the same for both Dexcom Share and Tidepool. The problem with Tidepool can be fixed by adjusting `Data Age Mins` in the Tidepool settings. The problem persists with Dexcom share even in combination with BlueJay as the collector. Backfills are sent do Nightscout. Saving the last upload timestamp for each sync service separately, could solve the problem - at least partially.

PR #803, opened 2019-02-25
Remove unused version of  UserError.cleanup()

tzachi-dar:

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/UserError.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/UserError.java
@@ -113,16 +113,6 @@ public class UserError extends Model {
        new Cleanup().execute(deletable());
     }
 
-    public static void cleanup(long timestamp) {
-        List<UserError> userErrors = new Select()
-                .from(UserError.class)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Models/UserError.java
```diff
@@ -113,16 +113,6 @@ public static void cleanup() {
        new Cleanup().execute(deletable());
     }
 
-    public static void cleanup(long timestamp) {
```
jakobsandberg: This function is still being consumed here: https://github.com/NightscoutFoundation/xDrip/blob/master/wear/src/main/java/com/eveningoutpost/dexdrip/ListenerService.java#L1152

If that logic is no longer relevant, can you remove it as part of this PR?

We also reference this function in a comment that we can remove in this PR: https://github.com/NightscoutFoundation/xDrip/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/wearintegration/WatchUpdaterService.java#L957

tzachi-dar:
Seems not relevant anymore.

PR #815, opened 2019-03-07
Added LG G5 and reordered phone list

philippebourget:
Added the LG G5 to the high capture list as I have been using xDrip on this phone and I get a well above 90% capture rate. Took the time to reorder the list by brand forUX ;)


Navid200:
@philippebourget Hi,
My understanding is that all Android phones can be used to run xDrip unless their operating system is extremely old. That is what I tell everyone in the facebook group.
There was a time we maintained these lists of phones that were compatible with xDrip. But, I don't believe we do that any longer.

Thanks

tolot27:

Issue #752, opened 2019-01-21
after starting a new libre sensor, it's old values are uploaded to nightscout as well


tzachi-dar:
I had a sensor running (very close to the 14'th day). 
Started a new one. After a few hours, connected the MM to it.
NS was loaded with duplicate entries. (see picture):
![image](https://user-images.githubusercontent.com/7942372/51491987-d79ba300-1db8-11e9-89ad-3c15448a1227.png)

The most problematic point is at time: 1547930076608

tzachi-dar:
Here is some more info about the issue:

Tried to reproduce the issue many times without success. 
After reading the code, it seems that this was probably a very rare race code that I was not able to find.

That said, it is very easy to reach an ugly state where for 5 readings we are having the same data (this will happen every time on xDrip but will not go to nightscout).
Here is an example:
![image](https://user-images.githubusercontent.com/7942372/53845261-170d0000-3fb1-11e9-917c-60da34bed1ea.png)

xDrip has a mechanism that should block this kind of behavior, but it does not work in the case of a new sensor. This is because in the function createBGfromGD() we call BgReading.getForPreciseTimestamp with locktosensor=true. 
This means that after creating a new sensor, the same point can be entered again and the mechanism of getForPreciseTimestamp does not solve the issue.

One question that I was asking myself, is why it does not happen for all points, 8 hours back.
The answer that I have found is that after starting a new sensor, there are still no calibrations, so first bg readings are thrown.
So, after there is the first reading, new bg points are created, and this points are not shown, since there is no calibration. This readings already belong to the new sensor.

After the first calibration, new points are rejected because of the points that were created before the calibration. Still, as one can see in the attached picture, last 5 points get recreated with the new calibration data.

I'll soon send a simple fix for this problem.

PR #819, opened 2019-03-08
Make sure that xdrip will not have two sets of points on the same time (libre)

tzachi-dar:
Fix the issue that is widely explaind in https://github.com/NightscoutFoundation/xDrip/issues/752

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
@@ -46,7 +46,7 @@ public class LibreAlarmReceiver extends BroadcastReceiver {
     private static final boolean debug = false;
     private static final boolean d = true;
     private static final boolean use_raw_ = true;
-    private static final double segmentation_timeslice = Constants.MINUTE_IN_MS * 4.5;
+    private static final long segmentation_timeslice = (long)(Constants.MINUTE_IN_MS * 4.5);
     private static SharedPreferences prefs;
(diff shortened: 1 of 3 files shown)


tzachi-dar:
Still being tested.

tzachi-dar:
Tests have passed, can be merged.

jamorham:
By all means make `getForPreciseTimestamp(long timestamp, double precision, boolean lock_to_sensor)` public but beware that `precision` is likely to be refactored to `long` type in the future. 

Don't remove the helper signature that has `long` as the precision type as this is what it is supposed to be long term.

tzachi-dar:
Changed precision to long right now.
Still needs final testing.

tzachi-dar:
`I wonder what the failure in tests means:
com.eveningoutpost.dexdrip.Models.SensorSanityTest > checkLibreSensorChangeTest FAILED
    com.google.common.truth.Truth$AssertionErrorWithCause at SensorSanityTest.java:90

We don't seem to have such a test, and SensorSanityTest.java only has 46 lines.

jamorham:
You can ignore that test failure it is due to some other changes and is due to differences between local and CI test parameters. I will try to resolve that today. 

So this PR now conflates two different change sets. The one you originally intended to resolve this specific issue and now the refactoring to long type for `getPreciseTimestamp()` which means it is spilling over in to things like the medtrum collector. 

This isn't what I wanted but as the changes are still fairly minimal and should be harmless I will accept it anyway. The smaller and more isolated a changeset is, the easier it is for me to accept for merging.

PR #883, opened 2019-05-19
Blukon: baulk on unexpected sensor changes

gregorybel:
Reviews are welcomed as I cant test it

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Blukon.java
@@ -10,6 +10,7 @@ import android.view.WindowManager;
 import android.widget.EditText;
 
 import com.eveningoutpost.dexdrip.Home;
+import com.eveningoutpost.dexdrip.Models.SensorSanity;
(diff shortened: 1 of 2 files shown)


dabear:
One other possibility is to use footer crc as unique identifier for the sensor? We know that body is static, but header contents (and header crcs) change infrequently.

jamorham:
@tzachi-dar is right. I think the `pratelimit` needs to be completely removed as we have to check serial number every time.

gregorybel:
actually yes good catch from @tzachi-dar but the timer is force resetted after time is going backward so after a sensor change, Serial number will be requested again. so latest 5min later.

We can for sure request each time, this is not battery optimised but at least we have more security.

jamorham:
@gregorybel I think you understand the code flow here better than me, if the timer is reset then we would expect one bad reading if there is then a sensor serial mismatch?

gregorybel:
I checked code again and in fact sensor time is requested every 3 hours, so it could be that we catch new sensor after 3 hours.
So, I simply removed serial decode timer, which anyway doesn't cost that much battery, as it is decoded by phone (I thought we need to request it from sensor over BT).

tzachi-dar:
LGTM.

PR #638, opened 2018-10-23
Don't show values in case calibration data is not reasonable.

tzachi-dar:
This is a proof of concept value which fixes #633. 

Very likely will not work, but to give a clue of what I think that we should do.

PR #887, opened 2019-05-22
Restrict maximum positive intercept value

jamorham:
I think @tzachi-dar has made a good point in https://github.com/NightscoutFoundation/xDrip/pull/638 - that we should restrict the maximum intercept value. 

This would be for unusual edge cases where it might be possible to produce calibrations which cannot measure below a certain value due to raw values and slopes always being positive in all the existing models.

This PR uses `CalibrationAbstract` to enforce this sanity check over all calibration plugins and directly in `Calibration.calculate_w_l_s()` along with the existing slope sanity checking. 

`Calibration.lastValid()` is also targeted to ensure that even if invalid records are synced from an external source they will not be used.

This should propagate the restriction to the graph, alarms, display glucose, inter-app communication and basically everywhere I can think of.

The restriction currently cannot be turned off.

I would consider pegging this to the `dead sensor testing` preference variable, but the main reason a switch isn't provided at the moment is that I think the code to do that needs to be really performance focused, provide a caching mechanism, work from static context and in a multi-threaded situation. 

I thought the code to do all that would add needless complexity at this stage and the other question is whether there is a use case that would actually need to turn it off?

Proof-reading, comments and testing very welcome! The most important test being whether this breaks anything.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -602,8 +602,13 @@ public class BgReading extends Model implements ShareUploadableBg {
 
                 if ((plugin != null) && ((pcalibration = plugin.getCalibrationData()) != null) && (Pref.getBoolean("use_pluggable_alg_as_primary", false))) {
                     Log.d(TAG, "USING CALIBRATION PLUGIN AS PRIMARY!!!");
-                    bgReading.calculated_value = (pcalibration.slope * bgReading.age_adjusted_raw_value) + pcalibration.intercept;
-                    bgReading.filtered_calculated_value = (pcalibration.slope * bgReading.ageAdjustedFiltered()) + calibration.intercept;
+                    if (plugin.isCalibrationSane(pcalibration)) {
+                        bgReading.calculated_value = (pcalibration.slope * bgReading.age_adjusted_raw_value) + pcalibration.intercept;
+                        bgReading.filtered_calculated_value = (pcalibration.slope * bgReading.ageAdjustedFiltered()) + calibration.intercept;
+                    } else {
+                        UserError.Log.wtf(TAG, "Calibration plugin failed intercept sanity check: " + pcalibration.toS());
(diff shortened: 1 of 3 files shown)


tzachi-dar:
Generally speaking, this would put us in a batter place.
I think that there are a few issues that I can think of.

1) Behavior is not consistent:
If one is using a calibration plugin, then no data will be created.
On the other hand, without a plugin, the last valid calibration will be used. (so data will be created).
I think that not showing any data is better.

2) xdrip has a lot of ways to create data. I'm not sure that all of them are covered.
Specifically, the following are not:
https://github.com/NightscoutFoundation/xDrip/blob/23ed81bdc94750198690bc943932842ae2cb54a3/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java#L322
https://github.com/NightscoutFoundation/xDrip/blob/23ed81bdc94750198690bc943932842ae2cb54a3/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java#L328
https://github.com/NightscoutFoundation/xDrip/blob/23ed81bdc94750198690bc943932842ae2cb54a3/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java#L334
https://github.com/NightscoutFoundation/xDrip/blob/23ed81bdc94750198690bc943932842ae2cb54a3/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java#L360


I think that they are ok, since data comes from share, but let's just verify it.

jamorham:
The plugin calibrations are an overlay over the existing standard xDrip calibrations so there is already some inconsistencies about how things are handled and particularly the calibration is calculated dynamically so it isn't stored in the database, it is an alternate representation of previous calibration points. Hopefully this inconsistency isn't too big a deal.

When I traced those other places where bgReading instances are created they looked to me to be when syncing records with an official receiver (G4/G4 share) which I think either should already have been validated by the external device. As far as I know they are also rarely used.

jamorham:
:+1:

tolot27:
@karelinoleg Do you still get the NPEs? If yes, did you raised an issue for that?

PR #913, opened 2019-06-16
Replace read-only ui-controls by non-interactive-look controls at InitialStatusPopup layout

karelinoleg:
The problem is that the controls assuming user interaction are used on this screen. There are 4 radio-buttons. The one is selected and users click on the other trying to make some action instead of waiting of initial data collection.
I think using non-interactive controls on the wait-screen would be more user-friendly and more clearly for users.
I've tried to improve initialStatusPopup layout without using radio-buttons.

59563875-1839d880-9048-11e9-947a-e44420a7fe261

### DIFF
--- a/app/src/main/res/drawable/ic_check.xml
+++ b/app/src/main/res/drawable/ic_check.xml
@@ -1,0 +1,5 @@
+<vector android:height="24dp" android:tint="#009900"
+    android:viewportHeight="24.0" android:viewportWidth="24.0"
+    android:width="24dp" xmlns:android="http://schemas.android.com/apk/res/android">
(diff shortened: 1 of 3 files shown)


karelinoleg:
> We can't use progress bar or any ui animation at all or it will break Bluetooth reception on certain phones, eg Samsung

Lets try to use non animated controls. Hourglass icon, for example

Actual(old) layout | New layout
------------------------- | ---------------
old | new

karelinoleg:
> There is also a helper method: app:showIfTrue which you can use to avoid the ternary with android view visible etc.
Ok. I have use it. Thanks

jamorham:
I have merged this, but can you confirm that you tested the latest version? Thanks

karelinoleg:
Yes. I've tested merged version just now on physical device.
screenshot0234

PR #937, opened 2019-06-29
Added tests for Nighscout Follow Entry - getTimestamp

ahaarrestad:
* Added tests for NightscoutFollow Entry - getTimestamp
* Minor updates in DateUtil to remove warnings in code

@jamorham hope you can find time to take a look and merge if ok.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/DateUtil.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/DateUtil.java
@@ -32,14 +32,14 @@ public class DateUtil {
      */
     private static Date fromISODateString(String isoDateString)
             throws Exception {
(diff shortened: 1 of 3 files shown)


jamorham:
What effect will adding: `Locale.getDefault()` do to the date formatting vs how it is now?

ahaarrestad:
Nothing. If you are not comfortable adding it, I can remove it.... but the formatting wants a Locale specified.

ahaarrestad:
@jamorham removed the default locale setting... back to the way it was.

jamorham:
So is specifying `Locale.getDefault()` the same as not specifying it? Eg that is what the existing signature actually does. Should we be specifying a Locale?

ahaarrestad:
Almost the sane. In effect the same. Leaving it like it was is less change which i know is preffered :-) work fine and build is green. I easily get carried away ckeaning up code :-)

jamorham (approved):
Looks good, thanks for keeping the change set minimal.

PR #938, opened 2019-06-29
Adding tests for Anticipate. Also suggesting javadoc for "next"

ahaarrestad:
@jamorham adding tests to get to know this part of the code better

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
@@ -80,8 +80,8 @@ import static com.eveningoutpost.dexdrip.UtilityModels.ColorCache.X;
 import static com.eveningoutpost.dexdrip.UtilityModels.ColorCache.getCol;
 
 public class BgGraphBuilder {
-    public static final int FUZZER = (1000 * 30 * 5); // 2.5 mins?
-    public final static long DEXCOM_PERIOD = 300000;
+    public static final int FUZZER = (1000 * 30 * 5); // 2.5 minutes
+    public final static long DEXCOM_PERIOD = 300_000; // 5 minutes
     public final static double NOISE_TRIGGER = 10;
     public final static double NOISE_TRIGGER_ULTRASENSITIVE = 1;
     public final static double NOISE_TOO_HIGH_FOR_PREDICT = 60;
@@ -105,7 +105,7 @@ public class BgGraphBuilder {
     public double start_time = end_time - ((60000 * 60 * 24)) / FUZZER;
 
 
-    private final static double timeshift = 500000;
+    private final static double timeshift = 500_000;
     private static final int NUM_VALUES = (60 / 5) * 24;
 
     // flag to indicate if readings data has been adjusted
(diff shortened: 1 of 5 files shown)


ahaarrestad:
Need to investigate a bit more. The code does not behave as I expected. More tests to arrive soon

ahaarrestad:
@jamorham Updated Anticipate as discussed. 

Added quite a bit of documentation and loads of tests. 

I have tested the new version and it reduces the lag to 2 minutes instead of about 8 minutes for Nightscout follow when Nightscout is in Dexcom Follow bridge mode.

Xdrip will end up asking every 20 seconds for new data until it arrives, when last is between 1 and 2 periods old. This will also apply to Dexcom Follow setup since they share the same Anticipate class.

jamorham:
Can you describe the changes to behavior with this new vs the original code? Thanks

ahaarrestad:
@jamorham 

The only change in behaviour is that if last reading is between 1 and 2 periods old, xdrip queries again after the grace peroid. For Nightscout that is 2x grace (20 seconds) since it adds another grace in the NighscoutFollowService.

All other parts of the Anticipate file is the same. Only line 26 to 29 is changed. The rest is documentation / comments and tests.

ahaarrestad:
The test line 90 to 172 shows the 3 different scenarios in Anticipate

ahaarrestad:
Nigtscout is updated every 2.5 minutes in bridge mode.

jamorham:
With the new code, in a situation where for example the receiver is missing every other capture and uploading backfilled data every 10 minutes: how many more times will this end up making a network call compared to the old code?

From my first reading of this it looks like you're just making it poll every x seconds, eg 20 when readings are missed. This is not what the anticipate class is supposed to do. The idea is that it is anticipating when readings will appear and making a minimalist amount of network calls given that the exact time can vary a little with network propagation delays etc. We're trying to give the user the lowest latency on the data we can, but avoiding constantly polling for data to conserve battery life and mobile data costs.

From what you have said it seems that this model doesn't work correctly with Nightscout in bridge mode because (I am guessing) that we receive data from Nightscout with a timestamp of X but that data was not actually available until, for example, X+1 minute and so `Anticipate` selects the wrong time for the polling window.

If that is the case then is there any data available in the Nightscout records that we can use to choose a better time window for polling? Does it record a timestamp of when the data was loaded in to the database as well as the timestamp for the glucose record itself?

Beyond that, should Nightscout be using something like Anticipate to reduce latency for data in bridge mode itself?

ahaarrestad:
I have searched the data from nightscout but did not manage to find any data in nightscout regarding when the data was stored. The entry only contains data on when the data was read. 

The only thing I found was the interval of polling with is 2.5 minutes by default (hardcoded, not a variable). From this, it is possible to change the code to query after 2.5 minutes if just missed, but I don't like to base my code on external variables, thus a more frequent poll seems to be a better way to do it.

In my tests now I have seen that it queries every 20 seconds for about 2 - 2.5 minutes before receiving a new reading. This results, for now, in 5-6 extra readings if data is late, but the downside is that this will happen for every reading.

Another idea I've had to improve on this is to store when xdrip sees the last reading, but this is also a bit dodgy since any skew will give you a constant lag.


Suggestions to improve:
* Increase to 30 seconds between polls, or even 60 seconds?
* Investigate more on a way to detect when previous reading was received by xdrip and use that to anticipate when the next reading will be available.

In addition, I am going to look into nightscout to see if we can improve the polling setup.

ahaarrestad:
PR now only contains updated javadocs and testing for Anticipate.

ahaarrestad:
@jamorham  I have now removed the changed functionality. This pull request only contains javadocs and tests.

jamorham (approved):
Looks good. I  assume you found another way to improve latency in bridge mode at the nightscout end.

PR #941, opened 2019-07-02
Cleanup of code in StatusItem

ahaarrestad:
* Inlined colorHint in enum
* Removed unused code


ahaarrestad:
@jamorham Some minor code cleanup. Found this code when writing the MegaStatus for Nightscout Follow

jamorham:
Thanks. I don't remember the history exactly but do you think `colorHint` is excessively complex because it can't ever actually receive a value not present in the enum?

ahaarrestad:
I believe it was written by someone who didn't know that you can add properties to enums in java. I have seen that before. All calls are to the same methods now as they were.

jamorham (approved):
Looks good

PR #942, opened 2019-07-02
Added MegaStatus to Nightscout Follow

ahaarrestad:
* Added some statistics data to NightscoutFollowService
* Added MegaStatus to NightscoutFollow containing some info about
  when data was received, delay of sent vs received, previous and next
  poll and some config info.

@jamorham First verson of the MegaStatus page for Nightscout Follow


jamorham:
This looks good but what is the rationale for moving `treatmentDownloadEnabled()` in to the service class? Although the borders may be a bit blurred, I tried abstract the preference handling details in to a different class.

ahaarrestad:
The two classes are a bit too interconnected, but I did not want to start on that now.... If you preffer it the other way around, I can move it back where it was. The Nightscout follow class is cross-dependant on the service as it is now.... the service knows about follow and the service knows about follow.... Ideally they should either be completely separated or merged.

want me to move the treatmentDownloadEnabled back? it is now used in both classes.

ahaarrestad:
@jamorham moved the "treatmentDownloadEnabled" back to NightscoutFollow

ahaarrestad:
@jamorham care to retrigger the build for me?

ahaarrestad:
Added treatments to the status page as well.

jamorham (approved):
Looks good, I haven't tested it but assume you have. Thanks.

PR #944, opened 2019-07-07
Nightscout Follow MegaStatus for Treatments fix

ahaarrestad:
Nightscout follow mega status checked for all treatments, but should
only check for none-xdrip treatments. Added a new query method
to keep the query quick, and also added tests to verify new code works
as expected.


ahaarrestad:
@jamorham minor fix for the NightscoutFollow megastatus when there are local treatments, not only those downloaded from Nightscout

ahaarrestad:
@jamorham any chance you will get time to merge this one?

jamorham (approved):

PR #945, opened 2019-07-07
Fix snoozing 'until you re-enable'.

jotomo:
Integer overflow meant 'for 10 years', were actually '3 weeks'.


ahaarrestad:
Any chance of adding a test for this?

jamorham:
This constant is also used elsewhere. Should we actually change the definition of the constant to be a long rather than casting it in this one place?

jotomo:
Good point. Fixed.

jamorham:
I don't trust force pushes so I wouldn't merge from this branch, but I have dealt with this issue in this commit: https://github.com/NightscoutFoundation/xDrip/commit/94989cfddf1ae02ed43ec375ec7adae18b977e46

Thanks for spotting it.

PR #946, opened 2019-07-07
Preemptive restart improvements

jotomo:
When a G6 is preemptively restarted and time travels to 1h50m, the native algo starts correcting for insertion trauma, which usually leads readings jumping up 20-30mg, without warning the user that a calibration might be necessary.

Over the last few weeks I've experimented with resetting the time from 9 days back to 3 days (which works if attempted after 9 days but not before). This lead to no jumps happening on restart. After 3 days the algo shouldn't be compensating for insertion trauma anymore, so the data isn't interpreted differently on the new 3 days age compared to the actual 9 days age.

Together with @th122 we've tested about 7 restarts with the provided change, without seeing any jumps. At one time for three readings after a restart sensor errors occurred (after which xDrip recovered). To avoid this, restarts are deferred up to 12h if the current delta is 5 or higher. After 12h the restart is forced regardless of delta.

Additionally - which was my initial approach - a notification is shown, informing the user of the restart and asking him to check the upcoming readings. For documentation purposes a note is also created for each restart. This is up for discussion, see below

- [ ] Reduce time to just 2 days (or even 1)? I started with 3 as a defensive first value
- [ ] Code currently runs for both G5 and G6. I've only tested with G6. Code needs either to be updated to apply to G6 only or G5 needs to be tested. For G5 restarts would occur every 3 days.
- [ ] Remove notification or make it optional. It might not be needed anymore.
- [ ] Keep automatic creation of note?
- [x] Undo unnecessary imports changes


app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
```diff
@@ -627,10 +620,21 @@ public static boolean doGetData(Ob1G5CollectionService parent, RxBleConnection c
                                 DexSessionKeeper.clearStart();
                             }
                             if (Pref.getBooleanDefaultFalse("ob1_g5_preemptive_restart")) {
-                                if (txtime.getSessionDuration() > Constants.DAY_IN_MS * (usingG6() ? 9 : 6)
+                                int restartDaysThreshold = usingG6() ? 9 : 6;
```
ahaarrestad: Is it possible to add tests for this change?

ahaarrestad:
Like the short commit, but is it possible to add tests for the changed behaviour?

jamorham:
I haven't looked at the code yet, but can you tell me what happens if there is a calibration recorded on the transmitter during the time-travel period? This is where I would expect a problem to occur and why I haven't previously attempted to implement this approach.

Can you also detail which firmware versions this has been tested against. Thanks.

jotomo:
@jamorham Transmitter is Rev002 wit firmware version 1.6.5.25.
I have one calibration after 6 days, which the jump back from 9 days to 3 days crossed which doesn't appear to have had any negative impact (and I'll watch what happens if this calibrations (which is in the future at the moment) crosses back into _now_ again). 
Do you mean such a calibration, or one that was send to the transmitter within the first 3 days, prior to any time travel?
What kind of problem would you expect? The sensor reporting an error, or calibration being off?

jotomo:
@ahaarrestad 
> is it possible to add tests for the changed behaviour?

Practically no. Given there are no existing tests for the Ob1 state machine, nor infrastructure as far as I've seen (accessing shared prefs, static calls, PowerMock?/Robolectric) adding a test would mean setting those things up and either mocking a lot and producing a brittle test with questionable value, or refactoring the state machine to be more testable. The latter would mean more changes without backing tests. So it's out of scope for this change and should be done separately as a pure refactoring. That of course requires having or acquiring non-trival knowledge of the codebase and lot of time to do the refactoring and then also testing the result manually (AFAIK the whole G5/G6 access is reverse-engineered, so there aren't exactly specs available to base tests on).

jamorham:
@jotomo what I would expect to see is that the restart can't be at a place in time earlier than where there is a calibration.

th122:
I happen to have the same transmitter version as @jotomo. The procedure went very smooth each time it was initiated. I definitely would advocate keeping the notification to the user - I'd even go as far as giving the user the choice of having a preferred timespan for these things, so he can schedule to have the restart happen during a time of day he is comfortable with (for me, the safest time would seem in the morning between waking up and breakfast - stable glucose readings for a while).

th122:
Is the format and reference of the transmitter-internal imestamp used for calibrations known? Does it count from 'start of transmitter' or 'start of sensor'? The latter might Indeed lead to interesting effects - until we're very certain  what's happening in these cases, I'd rather restrict the  preemptive restart to uncalibrated sensors.

jamorham:
The current system for the timing of restarts works that the restart would occur around the same time the sensor was started. So if that was 10am then the restart would be at 10am. I did this to avoid situations where the restart might occur during the middle of the night in a more unexpected way for the user.  I'm not sure what the best way to handle this vs the situation with restarting only during flat trend  that is included in this PR.

Also the `FirmwareCapability` class can be used to list known good firmware versions that extended time travel works with and check that when choosing the time travel window.

jotomo:
Before deciding on how automatic (without user intervention) a restart can be, let's see if a restart to 3 days work if there are calibrations.
I just ran a test: with the session 4.8 days old (overall age 11.4d, so restarted once already), 1h after a calibration I manually (voice command) restarted the sensor back to 3 days. This caused no missed readings and no jumps:

Screenshot_20190709-112109_xDrip+

After triggering restart, before command is send to transmitter:
Screenshot_20190709-112046_xDrip+

Restarting ...
Screenshot_20190709-112005_xDrip+ (1)

Restarted:
Screenshot_20190709-113057_xDrip+

(So unlike I previously stated, restarting before the 9th day also works. Must've messed something up before).

th122:
@jotomo so now we would wait out 1.8 days, to see if the calibration is stored relative to sensor age, and applied at 4.8 days?

jotomo:
No jumps or other issues crossing 4.8d (time of last calibration), so time travelling to a time before a calibration, as well as crossing the calibration timestamp again in realtime appear to do no harm to the transmitter algo (at least n=1).

Screenshot_20190711-121909_xDrip+

app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
```diff
@@ -627,10 +614,21 @@ public static boolean doGetData(Ob1G5CollectionService parent, RxBleConnection c
...
+                                if (txtime.getSessionDuration() > Constants.DAY_IN_MS * restartDaysThreshold
                                         && txtime.getSessionDuration() < Constants.MONTH_IN_MS) {
-                                    UserError.Log.uel(TAG, "Requesting preemptive session restart");
-                                    restartSensorWithTimeTravel();
+                                    BestGlucose.DisplayGlucose displayGlucose = BestGlucose.getDisplayGlucose();
+                                    // Defer restart up to 12h if current delta is high, which can lead to sensor
+                                    // errors if session is restarted during times of high fluctuation
+                                    if (displayGlucose != null && Math.abs(displayGlucose.delta_mgdl) <= 4
```
jamorham: This potential to shift the restart time by 12 hours is something I am nervous about as people have been told the restart will be at the same time as the original start.  I would like to have an additional preference option in the G5/G6 debugging options to control this and just for now to default this to `false` but with a plan to make it default to `true` in the near future. That way nightly users who update will still get the old behavior even once we change the default but new users will get the new behavior.

I also want to do the same for the longer time-travel as we know the 2 hours works but this is potentially a bit more risky. So we would want a new option for that as well to say that the preemptive restarts can adjust the time to minimize reading jumps.

jotomo:
Hm, the _allow extended time-travel (with any firmware)_  option sucks, it's entirely unclear to the user what options will actually be applied ..

jamorham:
Hi, thanks for making all these changes. I didn't realize this had already been done and I've introduced a conflict by adding a helper method to `FirmwareCapability` to do the same as yours does. I'm quite happy to use yours instead except that I also added another firmware to the rev2 list. See https://github.com/NightscoutFoundation/xDrip/commit/60785250188d935daecaff28ccb22a91343db5b8 if you can mirror this ((`2.18.2.88` addition) in your branch then I can just accept your version as I think its identical otherwise in terms of function.

The option you've added about applying extended time travel for any firmware is bound by engineering mode isn't it? Users enabling that should know what they're doing. We could add a line to mega-status to indicate based on the options whether extended time travel will actually be in effect? Would that help?

I think this PR is important and I would like to get it integrated soon. Thank you for all your work on it so far.

jotomo:
@jamorham Conflicts are resolved, I toke over your version which aligned better (naming) with existing code.

> The option you've added about applying extended time travel for any firmware is bound by engineering mode isn't it? Users enabling that should know what they're doing. We could add a line to mega-status to indicate based on the options whether extended time travel will actually be in effect? Would that help?

Correct, engineering mode only. My concern was about the regular user which enables the extended time-travel but has no idea whether his firmware is whitelisted. I've added the status to the mega status as you suggested.

About deferring a restart: this seemed to be a nice idea initially, then again I only observed it once that high fluctuations caused 3 missed readings after which the sensor resumed normally. Adding this option might add a can of problems: should there be an alarm if a jump is detected and occurs at night? Should the restart be limited to daytime or a user-set time as @th122 suggested? Should it be silent in general unless there's a jump? .. so having restarts at a fixed time - where the user can handle sensors errors if needed - might be a simpler and safer solution after all.
Also, would it be practical to remove the option again if it doesn't work out? Users like new features, but can be rather disgruntled when features are removed (even for valid safety concerns). I've labelled it experimental at least.

> I think this PR is important and I would like to get it integrated soon. Thank you for all your work on it so far.

Definately needs a proper review from someone more awake  than me before merging

jamorham:
It could be tricky to detect "jumps" as they could be quite subjective. More useful I think would be for things like AAPS to detect the restart note and add skepticism to their algorithm when there is a restart in the timeline without a subsequent blood test/calibration. I'm not really sure what the best thing to do would be but I think even just having the note (and notification) will be big improvements in transparency for what is occurring.

PR #947, opened 2019-07-07
Add poll delay for Nightscout Follow

### LINKED ISSUE #929: nightscout follower delayed by 5 to 10 min.
using xdrip to follow my kids bg. prefer xdrip as follower app because of the prediction display. but the bg data gets delayed by 5 to 10 min. have to restart collector everytime. and gettin error like 'nightscout follower.  woke up early'. is there a way to get xdrip to ping nightscout site si it gets latest bg value always?

ahaarrestad:
Enabling a poll delay for Nightscout follow to reduce the time from
a new value is made available on Nightscout until xdrip polls for new
data. Depending on the setup of nightscout this might be anything
between 0 and 150 seconds. Minimum value is set to 0s. Max is 300s.

This pull request fixes #929 by allowing custom poll delay.

@jamorham tested in AVD and works fine. There are sanity checks for input values.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollow.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollow.java
@@ -132,6 +132,26 @@ public class NightscoutFollow {
         return Pref.getBooleanDefaultFalse("nsfollow_download_treatments");
     }
 
+    static long pollDelay() {
+        try {
+            long delay = Long.parseLong(Pref.getString("nsfollow_delay", "0"));
+            if (delay < 0) {
+                msg("Only positive values for delay allowed. Illegal value [" + delay + "]");
+                return 0;
+            }
+            if (delay > DEXCOM_PERIOD / 1000) {
+                msg("Delay is longer than poll period. Try a value less than [" + (DEXCOM_PERIOD / 1000) + "] seconds");
+            }
+
+            msg("");
+            return delay * 1000;
+
+        } catch (NumberFormatException nfe) {
+            msg("Unable to parse [" + Pref.getStringDefaultBlank("nsfollow_delay") + "] as number");
+            return 0;
+        }
+    }
+
     // TODO make reusable
     public static Retrofit getRetrofitInstance() throws IllegalArgumentException {
         if (retrofit == null) {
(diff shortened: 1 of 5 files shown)


ahaarrestad:
Don't merge yet. Found a bug when changing collector. The new property is not removed / only shown for nsfollow.

ahaarrestad:
Found bug and corrected it. Poll delay now only shows for Nightscout Follow

jamorham:
Can you explain in more detail what this is for?

ahaarrestad:
I can add more explanation in code / docs too.

The goal is to let the user try to tune xdrip into fetching data after the last data has arrived in nightscout. For dexcom bridge the default poll interval on nightscout is 2.5 minutes, but not linked directly to when the last data was received. This interval can be changed in nightscout, but still it is very hard to automatically find the right interval for each persons individual setup. The simplest way is to let the user choose a poll-delay, which makes use of less cpu and bandwidth (ask once), and at the same time receive the update as soon as possible after the data was made available in nightscout.

ahaarrestad:
As the issue too states: "Experienced delay is about 5 minutes". This might be OK for some and others will want to reduce the time though the bridge and poll.

ahaarrestad:
@jamorham I will try to get a fix to Nightscout, but until then it is your call if you want this added to xdrip or if we should abandon the pr :-)

FredMK:
This PR would be a big help for Nightscout users, please accept it. The fixed 10 seconds poll delay is not enough.

seletok:
Hello,
I have Latest BG in XDrip+ (Follower from Nightscout) up to 9 mins. Is it possible to make it shorter? Max 5 minute, that is the BG measure period?

Thank you.
Regards,

ahaarrestad:
How much delay is there in nightscout for bg data? You can reduce that time and xdrip will pick it up. Check the poll interval for dexcom follow in nightscout docs and add it to your config for nightscout.

ahaarrestad:
Config variables can be found here https://github.com/nightscout/cgm-remote-monitor#bridge-share2nightscout-bridge

seletok:
I have Medtronic 640g with CNL 2,4. CNL 2.4 connected to a Xperia x Compact to upload data to Nightscout. In Nightscout data refresh in a few second.
From Nightscout download to Xdrid+ installed on a S10.

ahaarrestad:
Contact me on gitter and I can try to have a look at it.

seletok:
Sorry... I am a newbie... How can I do this?

seletok:
I write to you. Do you see?

> 
> 
> Contact me on gitter and I can try to have a look at it.

I write to you. Do you see?

PR #949, opened 2019-07-07
Code cleanup of class "Home"

ahaarrestad:
*No functional changes!*

* Created lambdas of a lot of calls with auto change in Android Studio)
* Removed a lot of un-needed Double.toString(...) with auto change in Android Studio
* Removed a lot of un-needed Integer.toString(...) with auto change in Android Studio
* Removed a lot of double new-lines which just made the file even longer
* Removed some un-needed generics in class creation.

Result: File has less boilerplate code and should be easier to read.


ahaarrestad:
@jamorham trying once more to do a cleanup commit. This time only 1 file. OK or close?

ahaarrestad:
@jamorham is this cleanup ok?

PR #965, opened 2019-07-27
Preemptive restarts fixups

jotomo:
Various fixes to preemptive restarts based on user feedback.
See the commit message of each commit for the reason the change was made.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
@@ -678,9 +678,10 @@ public class Ob1G5StateMachine {
      */
     private static boolean deferPreemptiveRestart(long sessionDuration, int restartDaysThreshold) {
         BestGlucose.DisplayGlucose displayGlucose = BestGlucose.getDisplayGlucose();
+        boolean highDelta = displayGlucose != null && Math.abs(displayGlucose.delta_mgdl) > 4;
+        boolean maxDelayExceeded = sessionDuration > DAY_IN_MS * (restartDaysThreshold + 0.5);
         return Pref.getBooleanDefaultFalse("ob1_g5_defer_preemptive_restart_if_needed")
-                && (displayGlucose != null && Math.abs(displayGlucose.delta_mgdl) <= 4
-                || sessionDuration > DAY_IN_MS * (restartDaysThreshold + 0.5));
+                && highDelta && !maxDelayExceeded;
     }
 
     private static void glucoseRxCommon(final BaseGlucoseRxMessage glucose, final Ob1G5CollectionService parent, final RxBleConnection connection) {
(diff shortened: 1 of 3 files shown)


jamorham:
Hi, thanks for this, can you describe the rationale for this PR?

jotomo:
I thought the commits would do that :-)
* The notification was wrong in talking about 3 days if regular restart to 2h is used.
* The note in the graph was also wrong when talking about _transmitter session_, rather than _sensor session_.
* Restart time to 3 days is increased by 2 hours to keep restarts at the same time of day and not move 2h every time (same as with the regular restarts)

jotomo:
@jamorham On the topic of fix-ups; is _BestGlucose_ a reliable way to get a delta here? https://github.com/NightscoutFoundation/xDrip/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java#L680
Is it calculated at that point in time? It seems a restart was triggered despite a high delta.

Sidenode: the travis build for this PR failed. We had the same thing over at AAPS. Travis switched the default Ubuntu VM from trusty to xenial if no distribution is requested. We explicitly request trusty for the moment like so: https://github.com/MilosKozak/AndroidAPS/blob/dev/.travis.yml#L3
Travis post on the switch: https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

jotomo:
Added a pref to make alert on restart configurable (user feedback).

jamorham:
Thanks for the heads up on travis

BestGlucose should be a good way to get the delta especially if it is for the delta "now"

The rationale for the PR is so I can understand the purpose and intended function. A single character can dramatically alter the code flow and so to review I have to understand the purpose to be able to determine if the code does what is intended.

So the restart time previously would have got two hours earlier each restart?

jotomo:
> The rationale for the PR is so I can understand the purpose and intended function. A single character can dramatically alter the code flow and so to review I have to understand the purpose to be able to determine if the code does what is intended.

Sorry, I meant to refer to the commit _messages_, not the code of the commits. You're absolutely right in asking for a rationale. I should have put a proper note in the PR to begin with. I'll update that now.

> So the restart time previously would have got two hours earlier each restart?

Yes, the restart is triggered if age is > 9 days, so it triggers after 9.1 days, which roughly works out to 9 days and 2 hours. With normal restart travelling to 2h, the jump is -9.0 days. Before this PR, extended time-travel jumps from 9.1 to 3.0, hence -6.1 days, With this PR, the jump is to 3.1 days, so -6.0 days.

jotomo:
> BestGlucose should be a good way to get the delta especially if it is for the delta "now"

Thanks, the issue was indeed not with BestGlucose but the condition, which i fixed now. I tested the preference for disabling the alarm and the updated condition in prod last night, both worked as expected. That might just complete fixing up the things I've screwed up in the original PR.

jamorham:
Is this ready for merge? I haven't studied your logic changes for the deferring. Are you confident they perform as desired now?

jotomo:
> Is this ready for merge? 

Yes, with the last commit added.

> I haven't studied your logic changes for the deferring. 
> Are you confident they perform as desired now?

No. 

However, I've removed the option. It was meant for an event I (or anyone else) haven't observed since, so I'm removing the complexity of unpredictably shifting the restart hour around. Should an error occur due to a restart it will be at an predictable hour where the user can attend to it.

PR #971, opened 2019-08-05
Libre2 weighted average working

keencave:
Libre2 weighted average calculation now working.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
@@ -69,7 +69,7 @@ public class LibreReceiver extends BroadcastReceiver {
                                 Libre2RawValue currentRawValue = processIntent(intent);
                                 if (currentRawValue == null) return;
                                 Log.v(TAG,"got bg reading: from sensor:"+currentRawValue.serial+" rawValue:"+currentRawValue.glucose+" at:"+currentRawValue.timestamp);
-                                if(!BgReading.last_within_minutes(5)) {
+                                if(!BgReading.last_within_millis(270000)) {
                                     List<Libre2RawValue> smoothingValues = Libre2RawValue.last20Minutes();
                                     smoothingValues.add(currentRawValue);
                                     processValues(currentRawValue, smoothingValues, context);
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
@@ -70,7 +70,7 @@ public class LibreReceiver extends BroadcastReceiver {
                                 if (currentRawValue == null) return;
                                 Log.v(TAG,"got bg reading: from sensor:"+currentRawValue.serial+" rawValue:"+currentRawValue.glucose+" at:"+currentRawValue.timestamp);
                                 // period of 4.5 minutes to collect 5 readings
-                                if(!BgReading.last_within_millis(4.5 * 60 * 1000 )) {
+                                if(!BgReading.last_within_millis(45 * 6 * 1000 )) {
                                     List<Libre2RawValue> smoothingValues = Libre2RawValue.last20Minutes();
                                     smoothingValues.add(currentRawValue);
                                     processValues(currentRawValue, smoothingValues, context);
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
@@ -69,7 +69,8 @@ public class LibreReceiver extends BroadcastReceiver {
                                 Libre2RawValue currentRawValue = processIntent(intent);
                                 if (currentRawValue == null) return;
                                 Log.v(TAG,"got bg reading: from sensor:"+currentRawValue.serial+" rawValue:"+currentRawValue.glucose+" at:"+currentRawValue.timestamp);
-                                if(!BgReading.last_within_millis(270000)) {
+                                // period of 4.5 minutes to collect 5 readings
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
```diff
@@ -69,7 +69,7 @@ public void run() {
                                 Libre2RawValue currentRawValue = processIntent(intent);
                                 if (currentRawValue == null) return;
                                 Log.v(TAG,"got bg reading: from sensor:"+currentRawValue.serial+" rawValue:"+currentRawValue.glucose+" at:"+currentRawValue.timestamp);
-                                if(!BgReading.last_within_minutes(5)) {
+                                if(!BgReading.last_within_millis(270000)) {
```
AdrianLxM: I usually write this as `4.5 * 60 * 1000` or `4.5 * 60000` to make it easier to read.

keencave: Good idea, will change it.

keencave: Done.

jamorham: `4.5 * MINUTE_IN_MS` also makes it easier.

keencave: The travis check went wrong after changing the ms calculation. Any idea?

keencave: last_within_millis() only accepts LONG. Calculations with a fraction results in an conversion error.

keencave: The current travis check for 45 * 6 * 1000 still runs for 20 min++ now.

app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
```diff
@@ -69,7 +69,8 @@ public void run() {
                                 Libre2RawValue currentRawValue = processIntent(intent);
                                 if (currentRawValue == null) return;
                                 Log.v(TAG,"got bg reading: from sensor:"+currentRawValue.serial+" rawValue:"+currentRawValue.glucose+" at:"+currentRawValue.timestamp);
-                                if(!BgReading.last_within_minutes(5)) {
+                                // period of 4.5 minutes to collect 5 readings
```
MasterPlexus: Not sure, but if you decrease the time frame of collecting the bg-Data to less then 5 minutes bevoe smoothing and storing, why you expect that you will get safely 5 readings? As far as I have seen the intends are every minute (between 59 up to 62 seconds), so in my mind, if you decrease the time frame to lower then 5, you will maximum work only with the last 4 readings in addition to the 20 minutes of smoothing?

app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
keencave: and failed again. I have absolute no idea what went wrong here.

app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
keencave: It was proofed via extensive testing. With the 5 minutes frame the valuecount per interval was changing between 4 and 6. Going to 4.5 I got mostly 5 readings.

app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
jamorham: Unfortunately just ignore travis errors at the moment, as far as I can work out part of fabric's cloud is down causing build failures within the wear module. Its unrelated to this PR.

keencave: OK. Anything I can do? Is the PR usable in that form?

jamorham: If you are convinced that you have tested these changes and it is working then I can approve it.

keencave: The changes are live tested and well working. Please approve.

app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
MasterPlexus: Ok, so then it should be tested in the wild ( I also did not think that it will make a big difference to have 4 or 5 readings ;) ) we will see.

AdrianLxM: Having a new value every 5 minutes instead of jumping between 5 and 6 minutes intervals makes a difference in looping. So making sure it is recorded exactly every 5 minutes by setting the time to not 5 minutes but slightly less has a real advantage here!

AdrianLxM: @MasterPlexus: if you wait more than 4, you will get the 5th reading in. If you wait 5 minutes, you may still reject the 5th reading and not create a GlucoseValue.

MasterPlexus: OK, so the comment is also slight misunderstoodable, as it suggests to get 5 readings, but means to be save to have a calculated value every 5 minutes... So sorry for the notice, was just wondering, but I'm clearly fine with it.

jamorham:
Should I be merging this now?

There is no problem for me whether the code is doing `270000` millis or 5 minutes or 4.5 minutes with a constant cast to long or whatever. It is just a question of readability how it is presented but the functionality comes first so don't compromise functionality over a small cosmetic point.

PR #974, opened 2019-08-07
add flexible Libre2 smooth filter range

MasterPlexus:
This PR will add the possibility to change the time frame of smoothing the Libre2 data in the settings.
- standard Timeframe is set to 20 as initial was chosen
- using it for half a day and works for me
@AdrianLxM and @keencave , can you have a closer look?


AdrianLxM:
Could you please add this to the Broadcast sent to AndroidAPS with the broadcast? Then we can accept it but exclude it from looping if 15 minutes or less are used e.g.

AdrianLxM:
Alternatively we could set the minimum to 15 or better 20 minutes and only allow higher values if people experience higher noise.

MasterPlexus:
Ok, I'll analyse the point to implement it to the broadcast. To limit it to 15 will be too hard in my eyes, because this should also give the possibility to see the effectiveness depending from the timeframe.

AdrianLxM:
If 15 minutes are to hard (I fully agree, as with the weoghting this means the last 5 minute value would count for 55% so noise and the over-interpretation of the libre algo would be too strong),  we should definitely set a hard limit that people can't go below 20 minutes.

MasterPlexus:
After some discussions, I will close the pr, as we have to put more intelligence in this function to ensure some things...

PR #975, opened 2019-08-09
Libre2 add raw graph

MasterPlexus:
This Pull-Request adds the raw-data from Libre2 into the graph.
- including advanced Pref to switch on/off (standard off) 
- Pref only visible if Libre2 is selected as collector
- pref-texts are not jet included in Strings.xml, as it is a new setting. Will made later a additional PR for active translation, because I also want to add other settings here later with other separated PR's

What is it good for: makes visible how noisy the data are, and how effective the smoothing algorithm works. Also it will be visible if there are lost intends if dots are not shown in the graph.

It will looks like this:
Screenshot_20190809-084510__01

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
@@ -208,6 +208,9 @@ public class BgGraphBuilder {
             loaded_start=start;
             loaded_end=end;
             bgReadings = BgReading.latestForGraph(numValues, start, end);
+            if (DexCollectionType.getDexCollectionType() == DexCollectionType.LibreReceiver) {
+                Libre2RawValues = Libre2RawValue.latestForGraph(numValues * 5, start, end);
+            }
             plugin_adjusted = false;
         } finally {
             readings_lock.unlock();
@@ -1304,7 +1307,6 @@ public class BgGraphBuilder {
             }
 
             if (DexCollectionType.getDexCollectionType() == DexCollectionType.LibreReceiver && prefs.getBoolean("Libre2_showRawGraph",false)) {
-                Libre2RawValues = Libre2RawValue.latestForGraph(numValues * 5, start, end);
                 for (final Libre2RawValue bgLibre : Libre2RawValues) {
                     if (bgLibre.glucose > 0) {
                         rawInterpretedValues.add(new PointValue((float) (bgLibre.timestamp / FUZZER), (float) unitized(bgLibre.glucose)));
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
@@ -137,7 +137,7 @@ public class BgGraphBuilder {
     private final int loaded_numValues;
     private final long loaded_start, loaded_end;
     private final List<BgReading> bgReadings;
-    private List<Libre2RawValue> Libre2RawValues;
+    private final List<Libre2RawValue> Libre2RawValues = new ArrayList<>();
     private final List<Calibration> calibrations;
     private final List<BloodTest> bloodtests;
     private final List<PointValue> inRangeValues = new ArrayList<>();
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
@@ -137,7 +137,7 @@ public class BgGraphBuilder {
     private final int loaded_numValues;
     private final long loaded_start, loaded_end;
     private final List<BgReading> bgReadings;
-    private final List<Libre2RawValue> Libre2RawValues = new ArrayList<>();
+    private final List<Libre2RawValue> Libre2RawValues;
     private final List<Calibration> calibrations;
     private final List<BloodTest> bloodtests;
     private final List<PointValue> inRangeValues = new ArrayList<>();
@@ -208,9 +208,8 @@ public class BgGraphBuilder {
             loaded_start=start;
             loaded_end=end;
             bgReadings = BgReading.latestForGraph(numValues, start, end);
-            if (DexCollectionType.getDexCollectionType() == DexCollectionType.LibreReceiver) {
                 Libre2RawValues = Libre2RawValue.latestForGraph(numValues * 5, start, end);
-            }
+
             plugin_adjusted = false;
         } finally {
             readings_lock.unlock();
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
@@ -137,7 +137,7 @@ public class BgGraphBuilder {
     private final int loaded_numValues;
     private final long loaded_start, loaded_end;
     private final List<BgReading> bgReadings;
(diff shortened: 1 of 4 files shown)


MasterPlexus:
Identified that the count of numbers was too less for the raw values, as they are set every minute, and was not shown over the hole diagram. Increased the amount with the last commit.

keencave:
Looks good for me! How long did you test it? Do you plan to add the a seperate Libre Trend Graph like @tzachi-dar did for the MM?

MasterPlexus:
Not at the moment, but I can have a look to it... So or so this should be then a separate or. At the moment I work at an additive mega status page for libre2-app collector...

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -1300,6 +1303,14 @@ private synchronized void addBgReadingValues(final boolean simple) {
                 }
 
             }
+
+            if (prefs.getBoolean("Libre2_showRawGraph",false)) {
```
AdrianLxM: It would be an option to additionally check if Libre2 is the currently selected source.
I  have no strong opinion here.

app/src/main/res/xml/pref_advanced_settings.xml
```diff
@@ -902,6 +902,17 @@
...
                 android:title="@string/title_Batterylog" />
         </PreferenceScreen>
+
+        <PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android"
+            android:key="xdrip_libre2_advanced_settings"
+            android:title="Advanced settings for Libre 2">
+            <SwitchPreference
+                android:defaultValue="false"
+                android:key="Libre2_showRawGraph"
+                android:summary="switch on to show the raw values on the Graph"
```
AdrianLxM: Those strings should be extracted to strings.xml and referenced to facilitate localisation.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
MasterPlexus: I agree, I leave it, as I think that this setting is only possible if a libre is used. The menu will be switch off if the collector is a other one. But if someone change later the colector then the setting is wrongly on (nevermind that then no data will be avalible ;) ) anyhou I will implement it

app/src/main/res/xml/pref_advanced_settings.xml
MasterPlexus: Will be done later, as Jamorham told me in the past to put those strings to translation after the new functionality will be ok and also used (like 4 weeks after implementation), I have them on my list to switch them later...

jamorham:
My main concerns about this would be the performance impact on the graph drawing.

MasterPlexus:
Up to now I have it active permanently, and see no impact on slow graph building. I have active what is possible in the graph, and all is fast. But surely maybe if someone could have a look on a older phone could test the speed...

keencave:
As @tzachi-dar implemented exactly the same for MiaoMiao without any problems for me I have no real performance concern on that. If the arise problems the user can switch it off again.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
MasterPlexus: Solved out with commit https://github.com/NightscoutFoundation/xDrip/pull/975/commits/79e519ba03cd80c81b151c3adbddca0b68569cf6

MasterPlexus:
Added the recommended proof that Libre2 is set as collector. The implementation to made strings in settings translatable will be done later (there will be additional PR wich I have in mind wich will increase the menu).
@jamorham , @AdrianLxM everything in addition? from my side it is ready to merge.

keencave:
I would recommend to integrate this PR soon without the translatable strings. It turns out that more people seem to have problems to detect fast rising/falling BG readings on time. It would help if one can check the raw readings in parallel.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -206,6 +208,7 @@ public BgGraphBuilder(Context context, long start, long end, int numValues, bool
             loaded_start=start;
             loaded_end=end;
             bgReadings = BgReading.latestForGraph(numValues, start, end);
+            Libre2RawValues = Libre2RawValue.latestForGraph(numValues * 5, start, end);
```
jamorham: Does this make a database query even when not using Libre?

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -1452,6 +1453,10 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
 
             }
 
+            if (DexCollectionType.getDexCollectionType() != DexCollectionType.LibreReceiver) {
```
jamorham: I would advise wrapping this in a try/catch block in case `libre2settings` is null or other problem.

jamorham:
Overall I would plan to merge this but I think it introduces some overhead even when Libre is not enabled (see my inline comments) and you have to be very careful manipulating preferences as changes to the XML can then result in crashes at times in the future. See my recommendation to wrap that section with a try/catch block to ensure any problem doesn't bring down the whole of the settings section.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
MasterPlexus: You are right. I'll correct that this evening.

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
MasterPlexus: Will have a look on that this evening, and try as you recommend.

MasterPlexus:
I'm under testing that the last changes will work. I'll come back after all tests have been done.

jamorham:
Please let me know when testing is complete so I can review again. Thanks

MasterPlexus:
I've some problems with my build environment, and I'm on the way to reestablish it. I'll keep you updated.

MasterPlexus:
@jamorham , Now I'm back in a status, that I could work and Test. Up to now the requested Changes are implemented, so you can review. The build is up and running well on my phone up to now.

k2s:
@MasterPlexus I am now testing this feature. Please look at #1037. Is there anything I should focus on ? thx for you work

MasterPlexus:
> @MasterPlexus I am now testing this feature. Please look at #1037. Is there anything I should focus on ? thx for you work

@k2s, I'm not sure what exactly you requested. I don't see a direct relevance to your Feature-Request https://github.com/NightscoutFoundation/xDrip/issues/1037 (even that it is interesting, but not easy to handle I fear)

k2s:
@MasterPlexus wanted to let you know that I installed your feature on multiple devices and that I am testing it. 

Referenced #1037 because it could be interesting also for your testing on multiple devices. It is not difficult to handle, I think. You modify and run the gist script I reference there and setup own JSON gist file pointing to location of your own APK. Then you have own distribution channel with features mixed in which are not in master.

**Question:** should I see the raw points also on follower if not enabled on master ?

MasterPlexus:
> **Question:** should I see the raw points also on follower if not enabled on master ?

@k2s: not sure on that. I did not add anything, that the data of the Libre2RawValue2-Table will be distributed, so the question is, if the Follower process automatically include all available data (which is not the case I fear).

PR #982, opened 2019-08-17
Added Moto G4 to high-capture-rate list 

kwyd:


Navid200:
@kwyd This file is now for archive purposes only. We are not updating it any longer.
Thanks

tolot27:

Issue #842, opened 2019-04-04
Tidepool integration - upload errors


pazaan:
The Tidepool integration can fail with the error "400 Bad Request", leading to "stuck" uploads.

This occurs when there are invalid values in xDrip that the Tidepool servers will not accept, resulting in the 400 response (such as negative CGM values).

We need to add a data validation step before uploading to Tidepool

pazaan:
@jamorham - you can assign this one to me :)

imakeinterfaces:
@pazaan The way I experienced was that the first phone i enabled tidepool on had no problems whatsoever. However that phone died and i had to switch to an interim phone which experiences this problem. I also bought a third phone to replace the interim one and now that one experiences the error as well.

The problem both of them exhibit is an error saying 400, `Found Existing Remote Dataset` There's a big thread on this here https://forum.fudiabetes.org/t/xdrip-integration-with-tidepool/5653/5 I suspect this is an issue with multiple devices.

tzachi-dar:
I have seen a case where upload fails, probably because there is a negative value on calculated_value.
Data is: 

"51926"	"51926"	"0.0"	"0.012"	"0.0"	"0.0"	"-26.6857484754007"	"0.0"	"78"	"0"	"81b473fe-70c8-48ad-869c-46b1eb93280d"		"0.0"	"0.0"	"-26.6857484754007"	"0.012"	"1"	"0"	"0"	"0.0"	"0.0"	"0.012"	"0.0"	"0.0"	"10"	"626ce774-4bd1-4488-b73e-83cd61835654"		"1256437385.0"	"1560865244888"	"2489687d-cb1c-43b3-8bec-963662f4e0de"

See the -26 there. I'll create a fix and see if it will help.
Fix has to be in ESensorGlucose

tzachi-dar:
@pazaan I'm looking for the checks that tidepool is doing to verify the entries.
For example, i have noticed that some BG values are too small/big.
Bolus values also have some limitations. 

Can you tell me what this limitations are, so I can add them to xdrip?

pazaan:
Thanks @tzachi-dar. I obviously haven't got around to this yet (soooo many open source projects to work on!).
[...]

PR #983, opened 2019-08-17
Tzachi tidepoll2

tzachi-dar:
Fix issues that caused tidepool data not to be uploaded.

In the case of an object with bad data, tidepool drops the entire request, and this causes that data stops going up to tidepool.

First two  commits were based on issues found for a local user.

The last commit is based on the tidepool specs. See https://github.com/NightscoutFoundation/xDrip/issues/842#issuecomment-520195276

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/tidepool/EBloodGlucose.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/tidepool/EBloodGlucose.java
@@ -3,13 +3,16 @@ package com.eveningoutpost.dexdrip.tidepool;
 // jamorham
 
 import com.eveningoutpost.dexdrip.Models.BloodTest;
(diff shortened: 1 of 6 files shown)


tzachi-dar:
@pazaan can you take a look at it?

jamorham:
Thanks for this. In the wild was it a case of negative values causing a problem?

pazaan:
@jamorham: in the Tidepool server logs I checked, it was negative CGM values, but any of these conditions would return a 400 status.

tzachi-dar:
There were two issues for that specific user. One was bg of -26 (caused by libre with 0.012 reading).
The second is the user wanted to enter bg reading of 119, but for some reason entered this as a bolus of insulin with value=119 (which is also too big).

LongboatAline:
@tzachi-dar could we do some sanity checking on input data? For example ask back if a single dose of insulin exceeds previous tdd...

PR #1002, opened 2019-09-07
Reducing file sizes of PNG images by optimizing images with OptiPNG

westermarck:
Optimized images in the project with OptiPNG to reduce the size of the project. The app also benefits since the app size and it's memory footprint is somewhat reduced.


jamorham:
Can you add some more detail for this PR? Specifically the questions that come to mind are:

1) What is the total space saving from this PR?

2) Which devices/android versions has this been tested on? App + Wear? 

As far as I remember, there were issues with indexed color PNG images and transparency/alpha channel on (older) handsets which is why png true color mode may be being used.

tolot27 (changes requested):
@westermarck Please provide the parameters you have used with OptiPNG and explain why you have chosen OptiPNG rather than the other PNG optimizers.
I do not want to start a discussion about image optimization. I just want to know the reason for your decision.

It would be also great the get jamorhams question regarding to the overall size reduction answered.

Navid200:
@tolot27 
The individual who opened this pull request has not responded to the questions posted almost 2 years ago.  
Is there a reason for leaving this open?

tolot27:
There is no need to optimize images except you want to speed-up the build process - slightly. [Drawables overview](https://developer.android.com/guide/topics/graphics/drawables#drawables) mentioned that during the the build process all images below `res/drawable/` are optimized automatically.

PR #1003, opened 2019-09-07
Tzachi us sensors

tzachi-dar:
This adds support for us 14 days sensors (currently only when scanning).


jamorham:
Could you briefly describe how these changes work? What the impact is? Thanks.

tzachi-dar:
On the libre1 sensors, BG was based only on a scan of 344 bytes from the farm.

With libre-us 14 days, the results also depend on 2 more things:

1) is the sensor tag (what we calculate the sensor s/n from).
2) Another command that is being executed on the sensor.

So, xdrip now executes the extra command and the s/n data and sends them to the oop.

It will also store then in the libreblock table, and once they will be retrieved from wifi or BT devices, we can also use them there.

Please let me know if there is anything more to explain.

Issue #840, opened 2019-03-28
TicWatch pro collector with libre + MM - worked perfectly - after wear OS update March 19 not anymore


Funfly39:
As stated in headline - ticwatch pro was working perfectly as standalone collector with MM and libre for 6 month. Worked e.g. on stable Jan 2019 and all nightly xdrip+ updates till March.

After Wear OS update around 20th March everything has stopped and watch can't be used as standalone collector - even factory reset, reset of MM, phone app etc. doesn't help.

It seems that one reading could be achieved from watch when overall searching for new bluetooth devices in overall watch connectivity settings -> bluetooth -> available devices and pressing on mm bluetooth device found.

Hyuril:
Hello, I have also problems with the stand alone feature of the Ticwatch Pro. It happens during the nights so maybe it's some standby mode but I'm not sure at all and trying some options on phone and on watch...

dwpotter:
Funfly39 I experienced the same issue around the same time with my ticwatch Pro. It has been working flawlessly for 6 months.

My daughter has a Fossil Gen4 and it was also working flawlessly for 2-3 months. She received a replacement miaomiao the 3 rd week in January and it would not work as a collector. The phone always had to be nearby and force collector off.  I thought it was the miaomiao until it happens with my Ticwatch Pro I’m March.

I have switched both to Glimp and both are working fine. I would love to switch back to Xdrip

kubatom:
Same issue with my Fossil/Misfit watch, I think this is a problem with the upgraded Wear Os. Has anyone found another thread (I'm sure this must have been reported already) and can link it here or merge tickets?

Funfly39:
As a new observation - I am currently able to run Ticwatch Pro again as standalone collector on 28.3. nightly. It is not as perfectly running as before mid March update (stopped suddenly 2 times within a week and hard to get watch connected again as standalone collector). Also making this run again as standalone was a lot of trial and error - before it was smooth and straight foreward. I did a lot of installation, de-installation, different settings, cache cleaning etc. It could be that collector ramdomly stops and it's luck to get it running again with exactly the same setup. As stated before - till mid March it was running smoothly for 6 month without any problems.

kubatom:
Thanks for sharing @Funfly39 , has anyone tried the 1.04 release? 

I will try to give it a go later this week

busken:
[...]

PR #1023, opened 2019-09-21
Remove the scanning with MAC addresses filter since it does not work with TicWatch pro.

tzachi-dar:
See https://github.com/NightscoutFoundation/xDrip/issues/840

We have made many experiments to find what the issue is.
It seems that for this version of the watch, scanning with an IP filter does not work.
The code always returns with a scanerror of timeout.

When scanning without the filter, the scan does succeed.

Since we verify the ip addresses when the scan finishes, I have removed the filter all together.

Other options might be use a flag for setting this filter or try to detect this specific watch, but I think that the first method is not good enough since it will be a bad user experience and for the second option I'm not sure what is the list of watches that I should give.

### DIFF
--- a/wear/src/main/java/com/eveningoutpost/dexdrip/utils/bt/ScanMeister.java
+++ b/wear/src/main/java/com/eveningoutpost/dexdrip/utils/bt/ScanMeister.java
@@ -54,7 +54,6 @@ public class ScanMeister {
     protected volatile boolean stopOnFirstMatch = true;
     protected volatile String address;
     protected volatile List<String> name;
-    protected volatile ScanFilter customFilter;
     protected boolean wideSearch = false;
     private static String lastFailureReason = "";
 
@@ -83,11 +82,6 @@ public class ScanMeister {
             this.name = new ArrayList<>();
         }
         this.name.add(name);
-        return this;
-    }
-
-    public ScanMeister setFilter(final ScanFilter filter) {
-        this.customFilter = filter;
         return this;
     }
 
@@ -147,27 +141,11 @@ public class ScanMeister {
         stopScan("Scan start");
         UserError.Log.d(TAG, "startScan called: hunting: " + address + " " + name);
 
-        ScanFilter filter = this.customFilter;
-        if (filter == null) {
-            final ScanFilter.Builder builder = new ScanFilter.Builder();
-            if (address != null) {
-                try {
-                    builder.setDeviceAddress(address);
-                } catch (IllegalArgumentException e) {
-                    UserError.Log.wtf(TAG, "Invalid bluetooth address: " + address);
-                }
-            }
-            // TODO scanning by name doesn't build a filter
-            filter = builder.build();
-        } else {
-            UserError.Log.d(TAG,"Overriding with custom filter");
-        }
-
         scanSubscription = new Subscription(rxBleClient.scanBleDevices(
                 new ScanSettings.Builder()
                         .setCallbackType(ScanSettings.CALLBACK_TYPE_ALL_MATCHES)
                         .setScanMode(ScanSettings.SCAN_MODE_LOW_LATENCY)
-                        .build(), filter)
+                        .build())
                 .timeout(scanSeconds, TimeUnit.SECONDS) // is unreliable
                 .subscribeOn(Schedulers.io())
                 .subscribe(this::onScanResult, this::onScanFailure));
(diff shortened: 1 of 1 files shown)


parapenT1sta:
I can confirm that works with Ticwatch E (Android 8.0.0) and MiaoMiao 1.

jamorham:
I assume this PR should say MAC address rather than IP address.

Can you confirm which software version is running on the Ticwatch Pro that has been tested with this PR?

thanks.

tzachi-dar:
Obviously, this is a mac addresses and not an IP addresses.

parapenT1sta:
@jamorham I tested with a Ticwatch E with Android 8.0.0 which I was facing the same issues as Ticwatch Pro and with this fix everything went back to normal as before the Feb. 4th 2019 release.

jamorham:
The Ticwatch Pro can scan for mac addresses. 

I'm unable to reproduce the issue with TicWatch Pro Build number `PWDD.190617.007` Wear OS version `2.8` Google play services `19.2.75`  System version `H`, Patch level `June 2019` using the standard published xDrip+ Sept 21st Nightly.  (all latest versions afaik)

Settings `Use Scanning = enabled` & `Trust Auto-Connect = disabled`

It seems to work fine as is on TicWatch Pro.

```
09-23 20:49:24.597  5076  5076 D ScanMeister: stopScan called from: Scan stop
09-23 20:49:24.716  5076  5076 D ScanMeister: stopScan called from: Scan start
09-23 20:49:24.716  5076  5076 D ScanMeister: startScan called: hunting: EF:A7:89:12:D0:FF null
09-23 20:49:26.815  5076  5076 D ScanMeister: Found a device: EF:A7:89:12:D0:FF  rssi: -90  -> MATCH
09-23 20:49:26.816  5076  5076 D ScanMeister: stopScan called from: Got match
09-23 20:49:26.817  5076  5076 D ScanMeister: stopScan stopped scan
09-23 20:49:27.318  5076  5076 D ScanMeister: Processing callbacks for EF:A7:89:12:D0:FF SCAN_FOUND
09-23 20:49:27.318  5076  5076 D ScanMeister: Callback: DexCollectionService
09-23 20:49:27.815  5076  6594 D ScanMeister: stopScan called from: Scan stop
```

parapenT1sta:
@jamorham From the feedback that I got from members of diabet1cos.pt, Ticwatch and any WearOs Watch uing Android H version is working perfectly fine. The worst is any WearOS watch using android 8.0.0 version, like Ticwatch E ou S version. E2 or S2 have Android H and they are OK.

tzachi-dar:
Two small comments:
1) The issue was seen by a few other people and reported on https://github.com/NightscoutFoundation/xDrip/issues/840.

2) We are still testing this fix, trying to understand if it is better to remove the filter at all, or put a filter that is empty.

jamorham:
So does this issue only affect Ticwatch E ?

Removing the filter will break scanning on AOSP based Android 8.1
Using an empty filter will break on Qualcomm based Android 9+

On Android 8.0 I would expect that Scanning can simply be disabled via the preference option. Has that been tested? Collector may need to be restarted to be sure.

tzachi-dar:
I think that we are left with the option to set this by a UI preference.
Does anyone see something better?

parapenT1sta:
> I think that we are left with the option to set this by a UI preference.
> Does anyone see something better?

Be able to choose on the watch / phone "Alternative Pair mode", I think it's a good idea.

parapenT1sta:
@jamorham With a new version sent by @tzachi-dar the feedback so far: https://github.com/NightscoutFoundation/xDrip/issues/840#issuecomment-536193471

jamorham:
@parapenT1sta Could you get the model names of the affected models for E and S via `adb shell` use `getprop ro.product.model` - I am expecting they are just `Ticwatch E` and `Ticwatch S` but to be absolutely certain this should be verified. We can use this to enable conditional code to work around issues with those models.

jamorham:
See https://github.com/NightscoutFoundation/xDrip/commit/f2c33e3e550c0f3b38f11559392199e43940a657 for what I think is a much better way to implement this which wont break scanning on other devices as it is selective for these two Ticwatches.

I'm interested to know whether this works, it is included in the Sept 30th nightly which is available now.

parapenT1sta:
> See [f2c33e3](https://github.com/NightscoutFoundation/xDrip/commit/f2c33e3e550c0f3b38f11559392199e43940a657) for what I think is a much better way to implement this which wont break scanning on other devices as it is selective for these two Ticwatches.
> 
> I'm interested to know whether this works, it is included in the Sept 30th nightly which is available now.

I wrote on Gitter, that is working perfectly fine with Ticwatch E, I don't have a Ticwatch S to test. Other devices I reported had Android H and in fact I don't know if they were having issues with the newer versions of xDrip.

Do you want me to ask in my group who is using WearOS watchs and what Android version do they have? To add all with Android 8.0.0 or lower to the list?

tzachi-dar:
Issue was solved on another pr.

PR #1024, opened 2019-09-21
widget improvements

k2s:
* change: add 30% transparent black color to widget
* fix: use minutes_ago string to render localized information in widget


k2s:
goals of my work in this pull requests:

* it was not possible to translate correctly `x minutes ago` in widget
* default look of widget after installation was good only for home screen with black background 
* if background color for widget was set in application settings (still difficult to find) it was applied only to the graph not to the whole widget
  * this is confusing on screen when the widget size was set big by user
  * on some widget sizes the text information was flowing out of the background
* the smallest widget size was wrongly wrapping text info for `bg>=10`

before vs. with this pull request:

compare-widgets

jamorham (approved):
I will merge this for wider testing.

tolot27:
@k2s As you can seen in your 2nd screenshot, the graph is missing in the smallest widget (#1560).

tolot27:
@k2s Please can you explain why you introduced rendering the graph only if the widget has a minimum size of 337?

Issue #1037, opened 2019-09-29
feature: how to build xDrip APK from multiple pull requests and with custom updater


k2s:
**Motivation:** We are xDrip+Libre2 and 2 followers. I am capable to do some minor improvements to xDrip and work also on translation. I like work that is done in some pull requests which are not merged to master and for sure need some testing. I know other families that could also be interested in testing, but don't have the knowledge to build own xDrip version.

**Solution:**
- [x] add configurable updater to xDrip code base
- [ ] disable the current xDrip updater if custom updater is included
- [x] improve generated version number to work with multiple builds in the same day
- [x] create example updater configuration
- [x] create example build script
- [ ] update the build script to build release APK from command like (with `gradlew`)
- [ ] create script to deploy release APK and to update the updater configuration
- [ ] find out how to use Github hooks to build this release APK when any of the included pull requests receive new commit
- [ ] allow to customize where to send the crash reports
- [x] publish own APK with features I am interested in as proof of concept

k2s:
Current state of implementation:

* custom updater - #1038
* version number change - #1038
* update configuration for my APK - https://gist.github.com/k2s/aa7928197d8ea508135bc20bb9bdd664
* example build script - https://gist.github.com/k2s/2feac7fb5ca8319a8bd198a731015a81
* my personal APK build for testing of new features - http://bit.do/xdrip-k2s

Current show stoppers:

* build with latest Gradle doesn't work - #1012
* what is @jamorham opinion on this

k2s:
#1160 build APK signed with data in `signing.properties` by calling `./gradlew :xdripupdater:assembleRelease` so no manual Android Studio invokation is neede

tolot27:
I'm not really sure whether we should include this into the public version. IMHO, a good documentation of the basic command chain should be enough.

PR #1038, opened 2019-09-29
add: custom updater required for #1037

k2s:
see #1037 for more details

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -113,7 +113,6 @@ def generateVersionName = { ->
 }
 
 
-
 android {
     compileSdkVersion 26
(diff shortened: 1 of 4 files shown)


Navid200:
Why are there two failed tests?

tolot27:
> Why are there two failed tests?

Because there are merge conflicts which must be resolved. See below.

Navid200:
The following is a comment by a developer in the issue this PR is supposed to address:
I'm not really sure whether we should include this into the public version. IMHO, a good documentation of the basic command chain should be enough.

Is there a reason this PR and the corresponding issue should remain open?

PR #1043, opened 2019-10-03
Allow the tomato code to work with the new and old versions of mm2 fw.

tzachi-dar:
The new fw sends an extra 6 bytes with the last command data.
This patch allows to receive both version of the protocol.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
@@ -29,6 +29,7 @@ public class Tomato {
         RECIEVING_DATA
     }
     private final static int TOMATO_HEADER_LENGTH = 18;
+    private final static int TOMATO_PATCH_INFO = 6;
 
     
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Models/Tomato.java
```diff
@@ -170,6 +175,12 @@ static void AreWeDone() {
         final String SensorSn = LibreUtils.decodeSerialNumberKey(Arrays.copyOfRange(s_full_data, 5, 13));
         byte []patchUid = null;
         byte []patchInfo = null;
+        if(s_acumulatedSize >= extended_length) {
+            // patchUid = Arrays.copyOfRange(s_full_data, 5, 13);
+            // patchInfo = Arrays.copyOfRange(s_full_data, TOMATO_HEADER_LENGTH+ 344 + 1 , TOMATO_HEADER_LENGTH + 344 + 1+ TOMATO_PATCH_INFO);
```
jamorham: Are these supposed to be commented out? patchUid is always null?

PR #1052, opened 2019-10-07
Libre2 addStatus

MasterPlexus:
adding Status for Libre2:
- Used Sensor-Number
- last reading time
- last Calculation for Smoothing
- Last used Sensors with Times, and quality on minutes level


MasterPlexus:
Screenshot_20191007-104802

app/src/main/java/com/eveningoutpost/dexdrip/Models/Libre2Sensor.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.sql.SQLException;
+import java.util.List;
+
+@Table(name = "Libre2Sensors", id = BaseColumns._ID)
+public class Libre2Sensor extends PlusModel {
+    static final String TAG = "Libre2Sensor";
+
+    static final String[] schema = {
+            "DROP VIEW IF EXISTS Libre2Sensors;",
```
jamorham: Wont this result in the view being recreated each time the app is reloaded in to memory?

app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
```diff
@@ -146,11 +153,33 @@ private static void saveSensorStartTime(Bundle sensor, String serial) {
...
+
+        l.add(new StatusItem("Libre2 Sensor",Sensor.currentSensor().uuid + "\nStart: " +  DateFormat.format("dd.MM.yyyy kk:mm",Sensor.currentSensor().started_at) ));
+        String lastReading ="";
+        try {
+            lastReading = DateFormat.format("dd.MM.yyyy kk:mm:ss", last_reading).toString();
+            l.add(new StatusItem("Last Reading", lastReading));
+        } catch (Exception e) {
+            Log.e(TAG, "Error readlast: " + e);
+        }
+        l.add(new StatusItem("Last Calc.",libre_calc_doku));
```
jamorham: Last calc looks like so much technical information it should be hidden behind engineering mode. I'm not sure whether the Sensors section would want the same or not.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Libre2Sensor.java
MasterPlexus: Correct. this was an artefact from the startingphase where I designed the View in several ways. Will delete the drop this afternoon.

app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
MasterPlexus: The lastcalc could be a good idea to put it to engenieering mode. Will do that the evening.

For the Sensors I think this is a valuable information for everyone of the users, to see if the Sensors go often in a bad state, and also to have the Serial-Number in one Place. What do you think to add this as switchable in preferences to make it easy accessible for a normal user, but standard put to off?

k2s:
I think all the information is important for user.
What about moving some parts of this information as a page of Statistics activity ?

MasterPlexus:
> I think all the information is important for user.
> What about moving some parts of this information as a page of Statistics activity ?

I think there could be a Differentiation. The Calculation is not really something where a normal user is interested in. If he is more technical, the engeneering mode is sufficient, also not to confuse the standard user. The Sensor list is more interested, as this could be also interesting in terms of problems like bluetooth and so on. Maybe it would be helpfull to set up a more understandable approach, like "Data qualety last 20 minutes" or something like that? 
The Statistics are not the right place in my eyes, as there are really general Information on therapy needed data, which are not direct linked to the kind of collector. But maybe I'm wrong...

I could understand the remark of @jamorham , because we must also be carefully that the CPU-Time will not increase to much for xDrip, also if nobody needs really this information. So I think, to make it toggleable will be helpfull I think.

MasterPlexus:
@jamorham , I've delete the Drop, set the calculation information to engineering mode, and the Sensor information was toggleable in Prefs now.
I've tested it on my side, and it works.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Libre2Sensor.java
```diff
@@ -0,0 +1,55 @@
...
+    @Column(name = "ts_from", index = false)
+    public long ts_from;
+
+    @Column(name = "ts_to", index = false)
+    public long ts_to;
+
+    @Column(name = "readings", index = false)
+    public long readings;
+
+    public static String Libre2Sensors() {
```
jamorham: I am thinking this will create a database query every second when the megastatus display is in progress. How often does this data change? Would it be better to cache it in a static field in the class, for example:

```
private static volatile String cachedString = null;

if ((cachedString == null) || (JoH.ratelimit("libre2sensor-report",60)) {
...
cachedString = <result from calcs>;
}
return cachedString;
```

MasterPlexus: Will check that, and test it. I come back after testing.

MasterPlexus: @jamorham, I have set the query to 2 minutes as recommended. I tested it, and it works as expected.

PR #1075, opened 2019-11-01
add PachtUID and PatchInfo

bubbledevteam:
Hello，as you know Bubble can transmit the Libre sensor PachtUID and PatchInfo which is needed for support Libre2 and Libre US 14 days,so we update the intepretation about how to use Bubble with it.

The PachtUID and PatchInfo are available in Bubble with FW 1.33 and above,so it's better upate Bubble to the newest FW first


jamorham:
Thank you for this. Two questions:

1) Has this been tested and confirmed to work correctly?

2) What will happen if this new code is run when connected to a Bubble which has an older firmware? Will it work as before or will there be some kind of failure?

bubbledevteam:
> Thank you for this. Two questions:
> 
> 1. Has this been tested and confirmed to work correctly?
> 2. What will happen if this new code is run when connected to a Bubble which has an older firmware? Will it work as before or will there be some kind of failure?

Hello

We had tested it with both older and the newest FW with the code ,it works well;

As your concern ,with older FW which does't transmit the PatchInfo and PatchUID but just transmits 344 bytes from sensor FRAM, so it can be used to generate glucose reading as before.

here I can provide the newest FW for you: https://send.firefox.com/download/d3e89a5625920644/#PGrXE20zaejUoIiU9WJjIg

PR #1083, opened 2019-11-09
finished my review and i'm really satisfied with your rework - merge request

gruoner:
i finished my review of @jamorham's rework and applied just very small patches:
- added a default bolus profile to be defined in insulin-profiles.txt (currently its novorapid as the "old" curve is exactly the same as novorapid). This will result in a default insulininjection with the default-bolus and the injected dose even when multiple-insulin is disabled. So we will get nightscout / tidepool or any reporter working with the new injection-schema (multiple injections in the same treatment) without requirement for a differentiation on their side - insulinJSON is always filled correctly
- bugfixing InsulinInjection: there seems a getUnits function missing
- resized dial-no-text: in german the text is too long - so reisized to 18pt

from my point of view we can merge to master branch

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/InsulinInjection.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/InsulinInjection.java
@@ -1,0 +1,41 @@
+package com.eveningoutpost.dexdrip.Models;
+
+import com.eveningoutpost.dexdrip.insulin.Insulin;
+import com.eveningoutpost.dexdrip.insulin.InsulinManager;
+import com.google.gson.annotations.Expose;
+
+import lombok.Getter;
+
+public class InsulinInjection {
+    private Insulin profile;
+
+    @Expose
+    @Getter
+    private double units;
+
+    @Expose
+    @Getter
+    private String insulin;
+
+    public InsulinInjection(final Insulin p, final double u) {
+        profile = p;
+        units = u;
+        insulin = p.getName();
+    }
+
+
+    public Insulin getProfile() {
+        // populate on demand
+        if (profile == null) {
+            profile = InsulinManager.getProfile(insulin);
+        }
+        return profile;
+    }
+
+    // This is just a rough way to decide if it is a basal insulin without user needing to set it
+    // question as to whether this should be here or call to encapsulated method in Insulin
+    public boolean isBasal() {
+        return getProfile().getMaxEffect() > 1000;
+    }
+
+}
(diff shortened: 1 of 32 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Models/InsulinInjection.java
```diff
@@ -38,4 +38,5 @@ public boolean isBasal() {
         return getProfile().getMaxEffect() > 1000;
     }
 
+    public double getUnits() { return units; }
```
jamorham: This is not needed as there is a `@Getter` lombok notation above

k2s:
**BUG:** 
* I click adding treament
* I add *8U Lantus* (long lasting insuling) and *20g sacharids* 
* I submit the dialog
* I see overview where booth lantus and sacharids are listed
* I confirm

It adds only the long lasting insulin icon, but not the sacharids. It works fine if I use Actrapid as insulin.

app/src/main/java/com/eveningoutpost/dexdrip/Models/InsulinInjection.java
gruoner: its interesting - Android Studio's syntax inspection reported an error
Error
but when i compile the code i didn't get any....

So just decline the add or leave it as it is to make android studio happy

gruoner:
> **BUG:**
> 
>     * I click adding treament
> 
>     * I add _8U Lantus_ (long lasting insuling) and _20g sacharids_
> 
>     * I submit the dialog
> 
>     * I see overview where booth lantus and sacharids are listed
> 
>     * I confirm
> 
> 
> It adds only the long lasting insulin icon, but not the sacharids. It works fine if I use Actrapid as insulin.

Can confirm but it's just a problem of visualization. In threatments both lantus and sacharids are saved and used for prediction.
Maybe @jamorham can fix the issue. I couldn't find the piece of code defining the icon and the label.....

app/src/main/res/layout/keypad_activity_phone.xml
```diff
@@ -17,13 +17,13 @@
...
         android:layout_marginLeft="5dp"
-        android:layout_marginRight="5dp"
         android:layout_marginTop="10dp"
+        android:layout_marginRight="5dp"
         android:background="@drawable/tick_thing"
         android:fontFamily="sans-serif-condensed"
         android:gravity="center"
         android:textColor="@android:color/white"
-        android:textSize="25sp"
+        android:textSize="18sp"
```
k2s: could you pls. sync with master so this conflict will be resolved ?

gruoner: merged with @jamorham's commit c7b74a5

app/src/main/java/com/eveningoutpost/dexdrip/Models/InsulinInjection.java
jamorham: Just enable the lombok plugin in android studio

gruoner:
done - transferred to 2nd xdrip train (gruoner/Feature-MultipleInsulin - https://github.com/gruoner/xDrip/tree/Feature-MultipleInsulin)

Issue #475, opened 2018-06-12
Rewrite History - Delete and reupload to Nightscout


Tornado-Tim:
The Re-write history option is awesome, it would be cool if this could extend to the Nightscout Uploader.

So upon calibration, xdrip will delete all the re-calibrated records in the Nightscout Database and reupload the new re-calibrated values (so the graph will be smooth upon calibration).

That way it would prevent the Nightscout graphs from jumping when a CGM Calibration as been applied and it gives better data back to closed loopers for Autotune.

Navid200:
@sulkaharo Is it technically possible to delete something from Nightscout?

sulkaharo:
Good question! No simple answer here:

1) A Nightscout instance can have any number of data sources uploading data into the database and apps loading the data, so any data sync implementations need to consider the semantics of how the other applications see the data changes and in case of data deletion, any app deleting the data would have to be careful to not delete data uploaded by other apps. Note there are even users who are uploading CGM data from multiple sources.

2) Nightscout REST API uses upserts by default (if an object being inserted matches an existing record, it's updated instead), so if xDrip wants to update a CGM value that's already been sent to Nightscout, it should be enough to just re-insert the CGM values with the same values as before, with the exception of the SGV value updated.

3) The V1 API xDrip uses currently doesn't clearly define the semantics of how updating / deleting data should be handled by apps, where for example data deletion will cause the deleted data to just disappear from the database and it's up to clients to notice the data is gone and in case of having local cached copy of the data, deleting the local data has to be handled by the app. We're solving this in the new V3 REST API, which keeps record of data that's been updated or deleted and clients can get well defined updates on data deletion / updates, thus making it easier for the clients to keep local data in sync. If xDrip is extended to data updates / deletion, it'd be great if the team would look at porting the app over to V3 in the process and at the same time work with the AAPS team to address the data duplication issues we're seeing with both AAPS and xDrip uploading the same data (but xDrip rewriting the records in the process), causing issues.

tolot27:
@sulkaharo Thanks for pointing me to the v3 API. That could also solve some of the remaining tasks in #1403.

Navid200:
@Tornado-Tim Rewrite history is a graphics setting.  None of the graphics settings affect Nightscout.  Neither should this one.
Please let me know if you disagree.
There is a previous open issue requesting similar behavior as you are asking.  https://github.com/NightscoutFoundation/xDrip/issues/366

[...]

PR #1090, opened 2019-11-13
On Calibration, Sync Updated Readings

boc-the-git:


app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
```diff
@@ -875,7 +877,7 @@ private double calculateWeight() {
     }
 
 
-    public static void adjustRecentBgReadings(int adjustCount) {
+    public static void adjustRecentBgReadings(int adjustCount, SharedPreferences prefs) {
```
jamorham: I haven't really examined any other part of this PR, but don't adjust the method signature to pass in an instance of `SharedPreference`s, just use the `Pref` class static methods instead.

boc-the-git: Solid tip Jam, thanks. I've made that change.

Feel free to hold off fully reviewing this, I had left it as a draft PR so I could provide some additional info - I'm guessing the creation of PR notified you, where I assumed it would be on submission!

boc-the-git:
This is my first PR for xDrip, so I welcome all feedback :)

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
```diff
@@ -906,9 +912,15 @@ public static void adjustRecentBgReadings(int adjustCount) {
                             bgReading.filtered_calculated_value = new_calculated_value;
                         }
                         bgReading.calculated_value = new_calculated_value;
+                        
+                        Log.v(TAG, "Previous value: " + oldYValue + ", New value: " + bgReading.calculated_value);
```
jakobsandberg: It looks like this log was used for development. Can we remove it before we merge? Logging in each step of the iteration will pollute our logs.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
```diff
@@ -896,6 +898,10 @@ public static void adjustRecentBgReadings(int adjustCount) {
             try {
                 final Calibration latestCalibration = Calibration.lastValid();
                 int i = 0;
+                boolean uploadModified = Pref.getBoolean("upload_modified_bgreadings", false);
```
jakobsandberg: We currently use `adjusted` in our naming throughout this file. I'm concerned that introducing a new word `modified` may cause confusion to other maintainers.

What do you think about using `adjusted` instead of `modified` for these new additions? That way we can align with existing naming conventions

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
```diff
@@ -906,9 +912,15 @@ public static void adjustRecentBgReadings(int adjustCount) {
...
                         }
                         bgReading.calculated_value = new_calculated_value;
+                        
+                        Log.v(TAG, "Previous value: " + oldYValue + ", New value: " + bgReading.calculated_value);
 
                         bgReading.save();
                         BgReading.pushBgReadingSyncToWatch(bgReading, false);
+                        if (uploadModified) {
+                            UploaderQueue.newEntry("update", bgReading);
+                            SyncService.startSyncService(3000); // sync in 3 seconds
```
jakobsandberg: Do we need to sync in every step of the iteration? What do you think about doing a single sync after we are done iterating over `bgReadings`

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
boc-the-git: I'm on-board with removing it if that's the consensus of the maintainers.
Personally I like it, because to actually get these logs you have to specifically add calibration logging, so 99% of people would never see it.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
boc-the-git: Happy with alignment, will make changes :)

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
boc-the-git: Yea, there's no fantastic reason for why I did this, other than it mimics the watch syncing behaviour. I'll revisit it and almost certainly make the change.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
boc-the-git: Changes made, still need to test - will be building shortly and will start running immediately.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
boc-the-git: Changes made, still need to test - will be building shortly and will start running immediately.

jamorham:
What happens at Nightscout when multiple readings are uploaded with the same timestamp. I thought it produced duplicated data?

jk1993:
> What happens at Nightscout when multiple readings are uploaded with the same timestamp. I thought it produced duplicated data?

I was worried about this as well which is why I'd not used the backfill data functionality before because I thought it would mess up my NS data but I've just tested and it overwrites it which is brilliant. Having this automated would be a great feature

boc-the-git:
> What happens at Nightscout when multiple readings are uploaded with the same timestamp. I thought it produced duplicated data?

Sorry for the delay. As @jk1993 mentioned, there doesn't appear to be any issues, it updates the reads rather than inserting a new one. Attached are screenshots of some of the readings being updated as per xdrip and seeing an ordered search of my NS entries in MongoDB (note the reads in xdrip are reverse order, first in the log is the newest reading).
image
image

I've not seen any display issues within Nightscout either.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
tolot27: I agree with @boc-the-git but suggest to print two decimal places at max. Please try to fit into one line.

tolot27:
@boc-the-git Does his PR addresses an existing issue?

Navid200:
Can we close this PR please?  
It can always be reopened if the developer answers the question posed in November of 2020 (7 months ago).

boc-the-git:
To answer the question (sorry, missed it); no this does not address an issue to my knowledge. Purely something I added to allow Nightscout to have a more accurate history.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
tolot27: Please use `Log.v(TAG, String.format("Previous value: %.2f, New value: %.2f", oldYValue, bgReading.calculated_value));`

tolot27 (changes requested):
Please implement the last suggestion to get this PR ready for approval.

boc-the-git:
> Please implement the last suggestion to get this PR ready for approval.

Change made, thanks.

Note I've not extensively tested this change to logging myself, as I'm not calibrating at present (just moved onto G6 Native so letting it do what it wants)!

tolot27:
> To answer the question (sorry, missed it); no this does not address an issue to my knowledge. Purely something I added to allow Nightscout to have a more accurate history.

It looks like it addresses #475. Please can you confirm this, @Tornado-Tim?

tolot27 (changes requested):
This PR is a great contribution and allows tracing changed values.

As commented in https://github.com/NightscoutFoundation/xDrip/issues/475#issuecomment-850774431 please implement [API v3 UPDATE](https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/api3/doc/tutorial.md#update) if it is supported by the Nightscout instance (can be checked by [querying the version](https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/api3/doc/tutorial.md#version)).

@sulkaharo API v3 UPDATE does not support an eventType or note for svg entries. Does it matter?

tolot27 (approved):

tolot27:
> As commented in [#475 (comment)](https://github.com/NightscoutFoundation/xDrip/issues/475#issuecomment-850774431) please implement [API v3 UPDATE](https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/api3/doc/tutorial.md#update) if it is supported by the Nightscout instance (can be checked by [querying the version](https://github.com/nightscout/cgm-remote-monitor/blob/master/lib/api3/doc/tutorial.md#version)).

It is not necessary to implement it in this PR. I've created a separate issue for it: #1765

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
```diff
@@ -917,13 +924,22 @@ public static void adjustRecentBgReadings(int adjustCount) {
...
+                        if (uploadAdjusted) {
+                            UploaderQueue.newEntry("update", bgReading);
+                            syncAdjustedReadings = true;
+                        }
                         i += 1;
                     } else {
                         Log.d(TAG, "History Rewrite: Ignoring BgReading without calibration from: " + JoH.dateTimeText(bgReading.timestamp));
                     }
                 }
+                if (syncAdjustedReadings) {
```
tolot27, line 940: use `if (uploadAdjusted && i > 0) {`

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
```diff
@@ -906,6 +908,11 @@ public static void adjustRecentBgReadings(int adjustCount) {
             try {
                 final Calibration latestCalibration = Calibration.lastValid();
                 int i = 0;
+                boolean syncAdjustedReadings = false;
+                boolean uploadAdjusted = Pref.getBoolean("upload_modified_bgreadings", false);
+                if (uploadAdjusted) {
+                    Log.v(TAG, "Adjusted readings will be uploaded to NS");
```
tolot27, line 914: I think the modified BG values will not just be uploaded to NS. They will be synced with any active sync service.

tolot27: More precisely, BG values will be synced to the following services, if active: https://github.com/NightscoutFoundation/xDrip/blob/38fee03be56da6ce9cbe90e08dbe8e4b651c6edc/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/UploaderQueue.java#L87-L91

tolot27 (changes requested):
After more discussions in the dev group, a few lines have to be adjusted.

Navid200:
When you perform a calibration, it is because the CGM is wrong.  That's the reason you calibrate.
You may have taken food or insulin due to the incorrect values before the calibration.
If you change the past readings, your behavior before the calibration may become questionable when you look back at the database.
It may look like you took insulin when your readings were not high enough to justify it, or it may look like you took carbs when your readings were not low enough.

What is the point of this setting?
What is the point of rewriting history?

tolot27:
@Navid200 This is the wrong place to discuss the "rewrite history" feature of xDrip. Please discuss this at Gitter or elsewhere and discuss here only PR related topics.

Navid200:
This is not about history.  
This is about this PR.  I am asking why we need it.

jk1993:
People often use nightscout as their primary data source for their CGM reading data (I do). It has far more in depth reporting tools than xdrip. If I want to look at historical readings I look at nightscout as it's in the cloud and I can access it from any device or app via integration. When I do a calibration on xdrip, it rewrites the history on xdrip automatically to correct readings that were inaccurate. Since these have already been uploaded to nightscout at this point the data in xdrip is correct but it is incorrect in nightscout. You can manually re-upload the data to xdrip using the "Backfill data" function but this is a pain to do on every calibration. This PR simply gives us the option to make this process automatic, so our nightscout data is kept in sync with the xdrip data.

Navid200:
Is there going to be a way to find out in the future that the readings were modified?
I mean let's say you have rewrite history enabled and all your calibrations also affect a few readings before each cal.
And those modified readings are also uploaded to Nightscout.
Then, let's say you change your mind about using rewrite history and disable it.  From now on, none of the readings prior to calibrations are being modified.

Later, you look back at Nightscout readings.  Will you have any way of figuring out which readings are real readings from the CGM and which readings are readings that have been manipulated by xDrip because of rewrite history?

If one of your readings, on the record in Nightscout, is high but your loop did not inject to correct it, will you be able to tell if this is because you calibrated after and the readings were raised because of the rewrite history so the loop at the time did not fail, or did the loop fail to inject when it really should have?

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
```diff
@@ -906,6 +908,11 @@ public static void adjustRecentBgReadings(int adjustCount) {
             try {
                 final Calibration latestCalibration = Calibration.lastValid();
                 int i = 0;
+                boolean syncAdjustedReadings = false;
```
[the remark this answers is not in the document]
boc-the-git: They are doing different things. Happy to name variables something else if any suggestion..

uploadAdjusted = do your preferences indicate we should upload adjusted readings?
syncAdjustedReadings = are there any adjusted readings to upload? Only start the sync service if so.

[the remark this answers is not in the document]
boc-the-git: Will resolve per later comment re the i variable.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
boc-the-git: Updated (commit soon).

PR #880, opened 2019-05-18
Improved the collapsed notifications layout for enlarge BG graph

karelinoleg:
[Show BG graph on MIUI LockScreen](https://github.com/NightscoutFoundation/xDrip/issues/879)

What do you think about this solution?
I change collapsed notification layout to enlarge image on it.

Actual(old) layout | New layout
------------------------- | ---------------
Screenshot from 2019-05-19 20-52-42 | Screenshot from 2019-05-19 20-51-20

Issue #879, opened 2019-05-18
Show BG graph on MIUI LockScreen


karelinoleg:
There is a problem on Xiaomi 6 with MIUI 10.3: it has not supported widgets on lockscreen. I have to unlock device every time I want to see BG graph. Using of xDrip widget on phone lockscreen would be a better solution for doing this. But it is not supported on MIUI.

Another solution might be to use "Number Wall on lockscreen".. but... terrible MIUI does not support it too. It work on another devices but on device with MIUI calling WallpaperManager#setBitmap(wallpaper, null, false, WallpaperManager.FLAG_LOCK); does not have any effect without any exceptions. (with WallpaperManager.FLAG_SYSTEM it works fine. But that's not what I need).

So now I'm thinking to make BG graph displayed in collapsed notification. (Because it is impossible to expand notifications on lock screen).

karelinoleg:
What do you think about this solution: [Pull-requets#880](https://github.com/NightscoutFoundation/xDrip/pull/880)? 
I have tried to change collapsed notification layout to enlarge image on it.

This is actual layout | New layout
------------------------- | ---------------
![Screenshot_1557956467](https://user-images.githubusercontent.com/8264069/57975912-ce27ed80-79dc-11e9-9413-7dc1ad862a9c.png) | ![Screenshot_1557956378](https://user-images.githubusercontent.com/8264069/57975914-d122de00-79dc-11e9-91cf-4ce666447885.png)

Navid200:
@karelinoleg What is the status of this?

karelinoleg:
It may be closed I think.
Because https://github.com/NightscoutFoundation/xDrip/pull/1101 was been merged. And that was an sufficient solution to my problem

PR #1101, opened 2019-11-22
 Improved the collapsed notifications layout for enlarge BG graph

karelinoleg:
Reopen the old pull-request https://github.com/NightscoutFoundation/xDrip/pull/880

[Show BG graph on MIUI LockScreen](https://github.com/NightscoutFoundation/xDrip/issues/879)

What do you think about this solution?
I change collapsed notification layout to enlarge image on it.

Actual(old) layout | New layout
------------------------- | ---------------
Screenshot from 2019-05-19 20-52-42 | Screenshot from 2019-05-19 20-51-20

Sorry. Looks like I missed the backwards compatibility issue.
DecoratedCustomViewStyle is not supported under API26.
I've fixed it. And renew review patch.
Now on androids under 8.0 will work the same as before

Screenshot from 2019-09-12 00-50-35
I have tested at Android 4.4 emulator. It works as before


markvader:
Thanks for this change. More value in the lockscreen/notification graph

Issue #1061, opened 2019-10-18
Android 10 BLE + G6 - Stuck in "scanning"


nshtg:
I have a reproduceable problem with XDrip being stuck in Scanning.

1) Reboot the device on Android 10
2) Brain state is stuck in scanning

Solution:
1) Forget device in G5/G6 Status screen
2) Next cycle it sucessfully reconnects and works from that point on... Until I reboot again/or the transmitter fails to connect once because of range etc.

Oneplus 7 Pro on the official Android 10 Version. Newest XDrip Nightly from October 11.

17gubbar:
Same problem

nshtg:
Does my fix ("Forget device" in System status screen of XDrip) work for you too? For me it's an 100% success rate so far!

17gubbar:
It works, but i found a workaround for the problem now read #1067 :)

17gubbar:
Also besure to turn of bt watchdog and untick the ob1 collector allowed to unbond.

ThoR294:
mine keeps asking for calibrations right after I hit forget device.  im on a pixel 4 xl

borijess:
I have the same issue

borijess:
> Also besure to turn of bt watchdog and untick the ob1 collector allowed to unbond.

Can you post a pic of your settings here?

![Screenshot_20191029-160131__01](https://user-images.githubusercontent.com/18737358/67809544-e2c9e080-fa66-11e9-88c7-1221636252df.jpg)
![Screenshot_20191029-160136](https://user-images.githubusercontent.com/18737358/67809546-e2c9e080-fa66-11e9-976d-7a3ac7de0d0c.jpg)

yumtime:
Same issue with pixel 3 android 10. Getting this error:

![Screenshot_20191101-151301](https://user-images.githubusercontent.com/37127272/68050002-abec0880-fcba-11e9-98ca-f6557a350b7b.png)

edisiENt:
Same issues on OnePlus 7 Pro Android 10.0.1 with G6.  Reverted phone back to Android 9.5 & saw good operation.  OnePlus refreshed my BT firmware, returned to 10.0.1 and BT problems returned.  Other BT devices (audio & car) work fine.  XDrip to G6 is the only BT problem I'm seeing.  G6 drops occur even while BT audio device is operating.

[...]

PR #1108, opened 2019-11-30
Fix Android 10 connection issues

nshtg:
See #1061.

I read some pages in the android documentation and tried to enable automatic reconnection for BLE devices.
With the changes from this PR my Android 10 phone sucessfully reconnected on its own after being out of range and losing connection.

I can't test for backward compatibility though.

@jamorham 
Why was auto-reconnect disabled in the first place?


tzachi-dar:
For some phones this works better and for others the opposite.
For other devices (for example MM) it is configured based on a preference settings.
I think you should put it based on getTrustAutoConnect()
or maybe we should add a UI for this in g5 debug settings.

jamorham:
I am 99.9% sure this wont work especially as it is placing auto-connect options in to services which are designed for polling reconnections.

BenMc19:
Any more luck @nshtg ?  I've been using the regular Dexcom receiver, and looking forward to getting back on the 'Drip!

PR #1114, opened 2019-12-03
Allow two versions of xDrip to work simultaneity on the same phone (using the OOP) .

tzachi-dar:
xDrip and OOP are using broadcast to send data to each other.
As a result, if there were two variants of xDrip installed on one phone, then data returned from the OOP was received by both of them.

With this change xDrip adds it's process id to the message sent, and verifies on the response that the same process id is used.

OOPs is passing this data for some time now.

A word about backward compatibility:
=======================================
Old versions of xDrip will not be affected by the change, as they don't check row_id when it returns.
New versions of xDrip with old OOP will not be affected since if row_id is not present in the message data is still accepted by xDrip.
So, only new versions of xDrip and new versions of the OOP will be able to use it.


jamorham:
I can see the sense in this, but wouldn't it be more standardized to transmit the application package id rather than the process id? 

Can't the process id change during the lifetime of the of the application, if it is say removed from memory or force killed? The application id is guaranteed unique and also allows targeted directing of android broadcasts which is often needed on Android 9+ ? 

Is OOP just returning the received identifier or does it actually want the process id?

thanks

tzachi-dar:
The oop will pass whatever it receives, but it expects an int.

During the life of an application the process id might change, but since I expect the OOP to respond in less then a second, I don't think it is a practical limitation.

What I can do is pass the hash code of the package id.
Whatever way you want it.

PR #1116, opened 2019-12-05
Ns follower patch second xdrip

gruoner:
We use two xdrip's (one as companion for G6 com and a tablet for reporting and notes etc.) and we are using NS to sync svg's, bg's and treatments between the xdrip's.
But xdrip doesn't sync treatments when they are coming from another xdrip.
So we defined a new pref to override this prohibition.


gruoner:
as i see now this PR also includes all the commits for the new insulin profile mode - i'm afraid we need to first merge #1083 and #1080 beforehand.....

gruoner:
done - transfered to 2nd xdrip train gruoner/Feature-2nd-xDrip-NSFollower (https://github.com/gruoner/xDrip/tree/Feature-2nd-xDrip-NSFollower)

Issue #865, opened 2019-04-26
Tidepool Upload Duplication


DunePlodder:
I usethe xDrip+ upload to Tidepool.
I now have a a t:slim x2 pump & upload from it.
The only problem is that data gets duplicated - carbs & boluses.
I want to upload CGM data & notes from xDrip, especially as the x2 is not yet integrated with the Dexcom G6.
Would it be possible to choose whether to upload carbs & boluses?

PaperT1D:
I have a similar problem (using Medtronic's 723), except only boluses seem to be duplicated. The only reason I manually upload  the pump to Tidepool at all is because xDrip doesn't upload basal data but, when I upload from the pump, my bolus amounts are duplicated and this makes my reported daily insulin doses way more than it is. I'd like to be able to select what information is uploaded to Tidepool as well if it's at all possible.

psonnera:
Same problem with manual uploading from the PDM (Omnipod). I guess we need a flag to choose treatments upload for those of us who use xDrip+ to log them. Either this or get it done directly from Nightscout the same way Dexcom is integrated with Dex Share.

Navid200:
@psonnera Is there anything we can change in xDrip to remedy this?
I'm not 100% understanding the problem.
If the cause of duplication is that two apps upload the same data, can we not disable uploading in one of them so that only one uploads?

psonnera:
It's been fixed. This is a duplicate. You can now select to upload only BG if you upload pump treatments manually.

Navid200:
@DunePlodder Considering what @psonnera has stated, is your issue resolved now?
If yes, would you please close it, or let me know and I will.

Otherwise, would you please explain why what has been explained does not work for you?

Navid200:
We can reopen if/when more data is provided.

PR #1132, opened 2019-12-15
Tidepool Upload Duplication #865

psonnera:
https://github.com/NightscoutFoundation/xDrip/issues/865
Quick fix. Please review and let me know if there is an impact on the Tidepool system I haven't thought of.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/tidepool/EBolus.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/tidepool/EBolus.java
@@ -27,7 +27,11 @@ public class EBolus extends BaseElement {
     }
 
     private static boolean IsBolusValid(Treatments treatment) {
-        if(treatment.insulin >= 0 && treatment.insulin <= 100) {
+        if(Pref.getBooleanDefaultFalse("tidepool_no_treatments")) {
+            UserError.Log.i(TAG, "Ignoring treatment " + treatment.toS());
+            return false;
+        }
+        else if(treatment.insulin >= 0 && treatment.insulin <= 100) {
             return true;
         }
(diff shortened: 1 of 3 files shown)


jamorham:
Can you explain why this is targeting low level methods relating to validity of boluses etc instead of wrapping the preference option around lines like `records.addAll(getTreatments(start, end));` in `UploadChunk` ? 

This also adds a dependency on `Pref` to the lower level classes which causes the unit tests to fail.

psonnera:
@jamorham you're right. And it's not even working since I forgot it's also pushed by EWizard. This is probably because I had no clue where to start in the code. Using your question I will try to find something more consistent. Sorry. Learning. And need that feature. Please discard.

jamorham:
Would you like me to try to implement something now that I understand the issue a little better?

psonnera:
Yes, if you can spare a moment that would be fantastic. And I'd return supporting users, where I'm more useful and efficient. Thanks.

jamorham:
This is included in the Feb 7th Nightly, please test and confirm. thanks

psonnera:
@jamorham  Installed now. Will test for a few days. Again, thanks so much.

psonnera:
image
I can't add treatments, they don't update in xDrip and don't upload to NS

psonnera:
@jamorham working great. No treatments in Tidepool but CGM and BG are uploaded fine. Nightscout gets all data as expected. Thanks again.

tolot27:
@psonnera, @jamorham It looks like this PR can be closed, right?

psonnera:
@tolot27 it's working perfectly. Agree. Thanks!

tolot27:
Merged with 94bee6e .

PR #1136, opened 2019-12-17
Sensor age alert

danielhers:
Add a configuration option to alert the user when the sensor age reaches a pre-defined value (in days).
Serves as a reminder to plan the sensor replacement, or place a sensor in advance (e.g. Libre).
The alternative would be to set a reminder elsewhere (using another app), but since xDrip must be informed when a sensor is started anyway, this is a convenient and easy option.


k2s:
tried to incorporate this pull request into my dev build, but found issues:

1. you should not commit changes to `gradle-wrapper.properties`
2. strings referenced in `pref_notifications.xml` are not commited so it is not possible to build this pull request

danielhers:
Thanks, @k2s! I hope I'll have time soon to fix the issues.

danielhers:
@k2s pushed an update. Hope it's good now. Happy new year ;)

k2s:
@danielhers thank you. build into my [custom build](https://gist.github.com/k2s/aa7928197d8ea508135bc20bb9bdd664) and testing it.

Happy new year.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
```diff
@@ -401,6 +402,21 @@ private boolean notificationSetter(Context context) {
...
+                OtherAlert(context, "sensor_age_alerts", "Sensor age: " + JoH.niceTimeScalar(sensorAge),
+                        sensorAgeAlertNotificationId, NotificationChannels.SENSOR_AGE_CHANNEL, true,
+                        Integer.MAX_VALUE / 100000);
+                prefs.edit().putBoolean("sensor_age_alerts_notified", true).apply();
+            }
+        } else { // Sensor is still young and fresh
+            notificationDismiss(sensorAgeAlertNotificationId);
+            prefs.edit().putBoolean("sensor_age_alerts_notified", false).apply();
+        }
+
```
jamorham: For state storage use `PersistentStore` instead of `prefs`
Use `Constants.DAY_IN_MS` instead of manually calculating the value for code clarity
There is a msSince() helper method so you don't need to manually calculate date subtractions etc
As you are storing the notification state, don't dismiss the notification unless the state is true to reduce overhead

From a user interface design point of view, I think that really we should be setting a preference of "notify hours before sensor expiry" and then automatically determining when this is. The problem is of course that that information isn't exposed in a unified way at the moment so it wouldn't be implemented easily. This approach can be used for now.

k2s: I had problem last time that `prefs.getLong("sensor_age_alerts_days", 13)` has crashed xdrip. When debugging the value was stored as string in prefs. Was not investigating further for the reason. @jamorham should it be safe to use getLong() ?

k2s: I had to solve the crashing so I created this push request https://github.com/danielhers/xDrip/pull/1

danielhers:
Thanks @k2s ! Sorry I hadn't had time to fix it myself.

wear/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
```diff
@@ -103,6 +103,7 @@
     public static final int lowPredictAlertNotificationId = 013;
     public static final int parakeetMissingId = 014;
     public static final int persistentHighAlertNotificationId = 015;
+    public static final int sensorAgeAlertNotificationId = 017; // ob1SessionRestartNotificationId would be 016
```
tolot27, line 106: Please remove the comment.

tolot27 (changes requested):
@danielharrelson Please can you incorporate jonorhams suggestions?

There is a similar feature request for Dexcom sensors: #1467.

danielhers:
> @danielharrelson Please can you incorporate jonorhams suggestions?
> 
> There is a similar feature request for Dexcom sensors: #1467.

Did you mean to tag me? In any case, I'm afraid I don't have time for it in the near future...

Navid200:
We now have this feature.  User can choose to be notified when sensor is about to expire.

Navid200:
Thanks for your contribution.
xDrip now has the option to show when sensor expires on the main screen as well as to alert at certain time periods remaining to expiry:
https://navid200.github.io/xDrip/docs/Sensor-Expiry.html 

Thanks

PR #1140, opened 2019-12-19
fix #1139: libre status crash when sensor stopped

### LINKED ISSUE #1139: libre status crash when sensor stopped
* use Libre
* stop sensor
* open Status page
* the application will crash

k2s:
Fixes #1139

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
@@ -167,11 +167,17 @@ public class LibreReceiver extends BroadcastReceiver {
         return Math.round(sum / weightSum);
     }
 
-
     public static List<StatusItem> megaStatus() {
         final List<StatusItem> l = new ArrayList<>();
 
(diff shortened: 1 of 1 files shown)


master3395:
Not sure if this works, as i haven't tested this change.
But i found it strange, that when the libre stopped, the xdrip+ app crashed, not sure if this is relevant here. (when clicking status)

k2s:
@jamorham This patch is needed otherwise xDrip is crashing.

I am managing 2 real life devices. One with nightly where it crashes and my own xDrip mix containing this patch and that is not crashing.

jamorham:
@k2s thanks for finding this issue. I think a better fix to avoid confusion between xDrip sensor sessions and an actual physical sensor and to display the remainder of mega status when a session has just been stopped, the fix would be better like this to simply exclude the line when the sensor session within xDrip is in stopped state:

```
        final Sensor sensor = Sensor.currentSensor();
        if (sensor != null) {
            l.add(new StatusItem("Libre2 Sensor", sensor.uuid + "\nStart: " + DateFormat.format("dd.MM.yyyy kk:mm", sensor.started_at)));
        }

jamorham:
The way of doing it I mentioned above is included in the Feb 7th Nightly.

tolot27:
Fixed with commit c0b1e91 and merged into nightly [2020.02.07](/NightscoutFoundation/xDrip/releases/tag/2020.02.07).

PR #1144, opened 2019-12-21
Allow to load us sensors data.

tzachi-dar:
This is a fix for the case that xdrip is used to capture data of us sensors and upload it to a mongodb.
It is needed on phones that do not have support for the OOP.
There have been two things that prevented this from working:
1) For us 14 days sensors we don't know how to calculate the checksum (It will later be calculated by the oop).
2) Since no BG is created locally for this libre-data the uploader did not start.


jamorham:
Does this change only impact mongo direct uploading?

tzachi-dar:
Yes, only mongo direct uploading.

PR #1150, opened 2020-01-02
Insulin trapezoidal calculation is fixed for long-acting insulin

stephencmorton:
Long acting insulins with a >instantaneous time at the peak value, had an incorrect calculation, which is now fixed.
A calculation with a reversed sign is corrected. This was currently a no-op as it was corrected either immediately or later in the code, but without fixing, that would not be guaranteed to be the case forever.

@gruoner

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/insulin/LinearTrapezoidInsulin.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/insulin/LinearTrapezoidInsulin.java
@@ -37,18 +37,18 @@ public class LinearTrapezoidInsulin extends Insulin {
         else if ((t1 <= time) && (time < t2))
             return 1.0 + 0.5 * max * t1 - max * time;
         else if ((t2 <= time) && (time < t3))
-            return 0.5 * (time - t3) * (time - t3) * max / (t3 - t2);
+            return 0.5 * (t3 - time) * (t3 - time) * max / (t3 - t2);
         else return 0;
     }
 
     public double calculateActivity(double t) {
         double time = t - onset;
         if ((0 <= time) && (time < t1))
-            return concentration * time * max / t2;
+            return concentration * time * max / t1;
         else if ((t1 <= time) && (time < t2))
             return concentration * max;
         else if ((t2 <= time) && (time < t3))
-            return concentration * (time - t3) * max / (t3 - t2);
+            return concentration * (t3 - time) * max / (t3 - t2);
         else return 0;
     }
 


app/src/main/java/com/eveningoutpost/dexdrip/insulin/LinearTrapezoidInsulin.java
```diff
@@ -37,18 +37,18 @@ public double calculateIOB(double t) {
         else if ((t1 <= time) && (time < t2))
             return 1.0 + 0.5 * max * t1 - max * time;
         else if ((t2 <= time) && (time < t3))
-            return 0.5 * (time - t3) * (time - t3) * max / (t3 - t2);
+            return 0.5 * (t3 - time) * (t3 - time) * max / (t3 - t2);
```
gruoner: as is is taken to square this change seems useless

app/src/main/java/com/eveningoutpost/dexdrip/insulin/LinearTrapezoidInsulin.java
```diff
@@ -37,18 +37,18 @@ public double calculateIOB(double t) {
...
         else return 0;
     }
 
     public double calculateActivity(double t) {
         double time = t - onset;
         if ((0 <= time) && (time < t1))
-            return concentration * time * max / t2;
+            return concentration * time * max / t1;
```
gruoner: thats already fixed in Features/MultipleInsulin (but not in this branch). Thanks for finding and fixing

app/src/main/java/com/eveningoutpost/dexdrip/insulin/LinearTrapezoidInsulin.java
```diff
@@ -37,18 +37,18 @@ public double calculateIOB(double t) {
...
         else if ((t1 <= time) && (time < t2))
             return concentration * max;
         else if ((t2 <= time) && (time < t3))
-            return concentration * (time - t3) * max / (t3 - t2);
+            return concentration * (t3 - time) * max / (t3 - t2);
```
gruoner: as we take activity in abs() this change seems useless

gruoner:
the t2/t1-fix MUST be applied as it fixes a real typo. The rest seems useless as it doesn't has an effect......

app/src/main/java/com/eveningoutpost/dexdrip/insulin/LinearTrapezoidInsulin.java
stephencmorton: Two points
1. Why is that `abs()` there anyway?
2. What this function returns in this case is **wrong**. The fact that later in the code, a wrong value is fixed does not make it any less wrong. This is medical software that people trust their health to, we should have a much much higher standard than "it's ok, the mistakes cancel each other out".

app/src/main/java/com/eveningoutpost/dexdrip/insulin/LinearTrapezoidInsulin.java
stephencmorton: Agreed, except that you never seen in a textbook that the area of a triangle can be calculated as `( -b) * ( -h) / 2`. And you would think it very strange if you ever saw somebody wrote it that way, even if it works. This is medical software and every "i" should be dotted and every "t" should be crossed. (Not sure if that expression translates well, but I just mean we should be extremely picky about every single little thing.)

jamorham:
I would like to see some varied unit tests for these calculations included with the PR so that this and any future refactoring can be fully tested and verified that this class is functioning correctly.

jamorham:
Does this include any changes regarding the recent discussion of concentration values?

gruoner:
> Does this include any changes regarding the recent discussion of concentration values?

no it can't because discussion #1080 ended up with awareness that there is no real bug to be fixed. BUT we still have an open issue "how to manage different insulin products with the same profile but different concentrations". I will rise an issue on that, but i'm not really sure how we can solve it without stressing "normal users" (even not developers & co) out.....

gruoner:
rised #1192

stephencmorton:
No, this PR fixes a different functional bug.

gruoner:
can be closed as those changes are made in #1194

peeyes-dev:
On a related issue, I identified an issue with inconsistent application of the boundary condition involving "onset" that skews the IOB and activity calculations for large "onset" values, if particular, when the ratio of onset/life value is high [Fiasp is okay, Humalog/Toujeo are NOT]. I have attached an excel spreadsheet with plots (in the first post) as well as the modified code (in the second post). Please review and pull it in as needed.
#1200

Thanks.

jamorham:
Is this PR still needed or is it superseded?

stephencmorton:
Superceded. Let me withdraw it.

PR #1151, opened 2020-01-02
English strings.xml improvements.

stephencmorton:
These are things that confused me when first using xDrip. I admit I'm new to xDrip so if these changes are not quite right, please let me know and I'll revise them.

**English strings.xml improvements.**
- "barcode" is changed to "QR code".
- Clarified meaning of some graph colour settings.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -39,8 +39,8 @@
     <string name="pref_header_cloud_storage">Cloud Storage</string>
     <string name="auto_configure_title">Auto configure</string>
     <string name="prefs_auto_config_summary">Auto configure using a QR code.</string>
-    <string name="scan_share2_barcode">Scan Share QR code</string>
-    <string name="pref_share2_scan_barcode_summary">Or scan the QR code on the share receiver</string>
+    <string name="scan_share2_barcode">Scan Share Receiver barcode</string>
+    <string name="pref_share2_scan_barcode_summary">Or scan the barcode on the Dexcom G4 Share Receiver</string>
 
     <!-- MongoDB Settings -->
     <string name="pref_title_mongodb">MongoDB</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -40,7 +40,7 @@
     <string name="auto_configure_title">Auto configure</string>
     <string name="prefs_auto_config_summary">Auto configure using a QR code.</string>
     <string name="scan_share2_barcode">Scan Share Receiver barcode</string>
-    <string name="pref_share2_scan_barcode_summary">Or scan the barcode on the Dexcom G4 Share Receiver</string>
+    <string name="pref_share2_scan_barcode_summary">Or scan the barcode on the Share Receiver</string>
 
     <!-- MongoDB Settings -->
     <string name="pref_title_mongodb">MongoDB</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -38,9 +38,9 @@
     <string name="title_activity_calibration_override">Calibration Override</string>
     <string name="pref_header_cloud_storage">Cloud Storage</string>
     <string name="auto_configure_title">Auto configure</string>
-    <string name="prefs_auto_config_summary">Auto configure using a barcode.</string>
-    <string name="scan_share2_barcode">Scan Share Barcode</string>
-    <string name="pref_share2_scan_barcode_summary">Or scan the barcode on the share receiver</string>
+    <string name="prefs_auto_config_summary">Auto configure using a QR code.</string>
+    <string name="scan_share2_barcode">Scan Share QR code</string>
+    <string name="pref_share2_scan_barcode_summary">Or scan the QR code on the share receiver</string>
 
     <!-- MongoDB Settings -->
     <string name="pref_title_mongodb">MongoDB</string>
@@ -490,10 +490,10 @@
     <string name="bad_glucose_values">Bad Glucose Values</string>
     <string name="revert_to_default">Revert to Default</string>
     <string name="filtered_values">Filtered Values</string>
-    <string name="treatment_color">Treatment Color</string>
-    <string name="treatment_color_dark">Treatment Color Dark</string>
-    <string name="predictive_color">Predictive Color</string>
-    <string name="predictive_color_dark">Predictive Color Dark</string>
+    <string name="treatment_color">Insulin On Board Color</string>
+    <string name="treatment_color_dark">Insulin Activity Color</string>
+    <string name="predictive_color">BG Prediction Color</string>
+    <string name="predictive_color_dark">COB Prediction Color</string>
     <string name="average_and_target_lines">Average and Target Lines</string>
     <string name="eight_hour_average_line">8-Hour Average Line</string>
     <string name="twenty_four_hour_average_line">24-Hour Average Line</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -492,7 +492,7 @@
     <string name="filtered_values">Filtered Values</string>
     <string name="treatment_color">Insulin On Board Color</string>
     <string name="treatment_color_dark">Insulin Activity Color</string>
-    <string name="predictive_color">BG Prediction Color</string>
+    <string name="predictive_color">Glucose Prediction Color</string>
     <string name="predictive_color_dark">COB Prediction Color</string>
     <string name="average_and_target_lines">Average and Target Lines</string>
     <string name="eight_hour_average_line">8-Hour Average Line</string>


tolot27:
@jamorham Please decide you if these suggestions are in line with your intention.

Navid200:
Where are the the options, in xDrip, related to lines 42 and 43?
I can find everything else.

jamorham:
Thanks for the contribution, but these strings are intended to be generic so I wouldn't propose to merge this.

app/src/main/res/values/strings.xml
```diff
@@ -483,10 +483,10 @@
     <string name="low_glucose_values">Low Glucose Values</string>
     <string name="revert_to_default">Revert to Default</string>
     <string name="filtered_values">Filtered Values</string>
-    <string name="treatment_color">Treatment Color</string>
-    <string name="treatment_color_dark">Treatment Color Dark</string>
-    <string name="predictive_color">Predictive Color</string>
-    <string name="predictive_color_dark">Predictive Color Dark</string>
+    <string name="treatment_color">Insulin On Board Color</string>
+    <string name="treatment_color_dark">Insulin Activity Color</string>
+    <string name="predictive_color">BG Prediction Color</string>
```
tolot27, line 488: BG -> Glucose

tolot27:
> Thanks for the contribution, but these strings are intended to be generic so I wouldn't propose to merge this.

@jamorham Which strings are generic? The color related strings does not look like to be generic. See following graph shows my interpretation of the color values which corresponds to stephencmorton string changes.

sample graph

Regarding to the barcode strings: I've provided a more detail string the Dexcom G4 Receiver Share barcode, which is indeed a barcode and not a string. But "Show Settings QR code" is indeed a QR code and not a barcode. Hence "Auto configure" scans a QR code as well.

Your are right that a QR code is somehow a barcode, namely a matrix or two-dimensional barcode. But it is clearly distinguishable from a barcode. Hence, the proposed string change is more precise and improves the understanding.

Navid200:
@jamorham Users ask what treatment color or treatment color dark is.
This PR will update the legend so the users will have the answer in xDrip.
If some day we decide to change the color to be used for something else, we can update the legend accordingly.

If you don't like to merge this, would you please explain how best we can address the question that comes up by the users?

app/src/main/res/values/strings.xml
tolot27: Fixed it with 6ff7792

tolot27:
The graph preview is based on the actual data which sometimes do not contain any customizable color. In the future we should provide a mock graph containing all kind of customizable data.

tolot27:
Here is a side-by-side comparison with the new QR code icon:
Screenshot_1628020238Screenshot_1628017710

tolot27:
@jamorham Please can you explain your concerns or merge this PR?

app/src/main/res/values/strings.xml
```diff
@@ -38,9 +38,9 @@
     <string name="title_activity_calibration_override">Calibration Override</string>
     <string name="pref_header_cloud_storage">Cloud Storage</string>
     <string name="auto_configure_title">Auto configure</string>
-    <string name="prefs_auto_config_summary">Auto configure using a barcode.</string>
-    <string name="scan_share2_barcode">Scan Share Barcode</string>
-    <string name="pref_share2_scan_barcode_summary">Or scan the barcode on the share receiver</string>
+    <string name="prefs_auto_config_summary">Auto configure using a QR code.</string>
```
jamorham, line 41: I don't think we are limited to QR codes with the auto-configuration feature so I wouldn't think this would be good to change.

app/src/main/res/values/strings.xml
```diff
@@ -38,9 +38,9 @@
...
+    <string name="scan_share2_barcode">Scan Share Receiver barcode</string>
+    <string name="pref_share2_scan_barcode_summary">Or scan the barcode on the Dexcom G4 Share Receiver</string>
```
jamorham, line 43: I'm not sure that any new users will be using G4 Share (as far as I know it is no longer available) and I would prefer to avoid adding any manufacturer trademarks to the wording wherever possible so I don't see any need to change text for this feature.

app/src/main/res/xml/pref_data_sync.xml
```diff
@@ -7,7 +7,7 @@
         android:key="dataSync"
         android:title="@string/data_sync">
         <PreferenceScreen
-            android:icon="@drawable/ic_scan"
+            android:icon="@drawable/ic_qrscan"
```
jamorham, line 10: The barcode scan icon is present in screenshot documentation and as mentioned in another comment I don't think the auto-configuration feature is limited to QR codes. From an aesthetic point of view the red coloring is the only highlight color on the settings top level which matches the xDrip drop icon.

app/src/main/res/values/strings.xml
```diff
@@ -483,10 +483,10 @@
...
     <string name="revert_to_default">Revert to Default</string>
     <string name="filtered_values">Filtered Values</string>
-    <string name="treatment_color">Treatment Color</string>
-    <string name="treatment_color_dark">Treatment Color Dark</string>
-    <string name="predictive_color">Predictive Color</string>
-    <string name="predictive_color_dark">Predictive Color Dark</string>
+    <string name="treatment_color">Insulin On Board Color</string>
+    <string name="treatment_color_dark">Insulin Activity Color</string>
+    <string name="predictive_color">Glucose Prediction Color</string>
+    <string name="predictive_color_dark">COB Prediction Color</string>
```
jamorham, line 489: It is true to say that these descriptions for the colors could help users in their understanding and that they are currently only used for these features afaik. If I need to add new graph element colors I can extend the preferences available as needed. Users have already accepted the disclaimer and warning when using the app for the first time so they are already aware of the limitations of the app and any predictive simulation features. I would accept these color naming string changes.

tolot27: Great, thanks!

app/src/main/res/values/strings.xml
tolot27: @jamorham Hmm, I currently don't now where the barcodes should came from. As far as I know, xDrip only provides a QR code if "Show Settings QR codes" is selected. Please can you clarify, when and how a barcode is created/displayed?

@Navid200 Do you know of any real barcode in xDrip?

app/src/main/res/xml/pref_data_sync.xml
tolot27: > From an aesthetic point of view the red coloring is the only highlight color on the settings top level which matches the xDrip drop icon.

Shall I change the color of the edge bars from green to red? Removing it all together makes the icon indistinguishable to the "Show Settings QR codes" icon.

app/src/main/res/values/strings.xml
Navid200: I don't.  But, there is a lot I don't know.

It may be best to break this up into two different PRs.
I already have this (see image) in the [documentation](https://navid200.github.io/xDrip/docs/Legend.html).

Screenshot 2022-02-23 161230

If we/I open a PR with only that, it may pass.  Then, the conversation about the bar code can continue and may resolve also.

stephencmorton:
I made this PR *two years* ago. I thought it would be a nice little un-controversial way to get into this SW. I was just trying to help out, to clarify some obviously incorrect strings. I have 30 years working in embedded SW and I thought I might get into this project.

Without in any way doubting the fact that for life-and-death software the utmost care must be taken, and I also admit that I don't understand some of the legal hurdles where perhaps you do not want to name certain things what they absolutely are in actual fact used for, I've got to say that watching the baloney that's gone on with this makes me know that I will never bother contributing to this project again.

I'm out.

(This is a great tool, and I applaud everybody who works on it. I don't want to leave insulting any of that, kudos to all of you. I just have no desire to work in this environment.)

app/src/main/res/values/strings.xml
Lupum1001: Splitting off the QR change seems like overkill. I looked into the scanner the other day as part of a separate discussion:
https://github.com/NightscoutFoundation/xDrip/discussions/1991#discussioncomment-2237437
The **TL;DR** is that there are two separate scanners in the UI, and the proposed alterations affect the one that _was_ intended for QR codes.

I support the QR-related changes in this PR, especially since they don't replace the labels for BOTH scanners.
At this point in time, "QR scanner" is probably the more relevant/user-friendly label, even if it _can_ secretly do legacy barcodes under the hood.

app/src/main/res/xml/pref_data_sync.xml
Lupum1001: @tolot27 What if you were to change it to red, _and_ add a horizontal line across the middle, similar to this:
image

That would address both of @jamorham's concerns, since the horizontal line implies barcode support. I also found these icons as well, which we could potentially use as a reference for a new icon:
image
image

tolot27: I've moved the QR icon to #2022.

tolot27: @Lupum1001 Thanks for your suggestions. Let us continue this discussion at #2022.

app/src/main/res/values/strings.xml
jamorham, line 41: As previously mentioned I think that `barcode` is a better word because a QR code is a type of `barcode` and so it better represents the feature which can support QR or other barcode types. Changing it to QR code makes it more restrictive regarding what features can be included via the auto-configure scanning and I believe that the word barcode is widely understood to mean all types of barcodes, 1D, 2D etc.

app/src/main/res/values/strings.xml
```diff
@@ -490,10 +490,10 @@
...
```
jamorham, line 496: These probably could be changed if that would help users. I think if I'm expanding what appears on the graph in future then I would most likely add additional colors for that purpose. @Navid200 what do you think of these wording changes?

Navid200: I find the word "Color" included at the end redundant as this is a legend and it is only a color legend and nothing else.  
So, in [documentation](https://navid200.github.io/xDrip/docs/Legend.html), I have the exact same words minus color at the end.
But, below the color legend, on the same page, I have added an image that let's the user also see the patterns for these 4 items.  If you scroll down, you will see the graph.

If you look at the insulin activity curve, you can see that it is not easy to see.  Only clarifying what the color is may not be enough to help  a user what the curve is.  Looking at the curve, it's not easy to associate it with the square of color in the legend.  That's why I thought it's best to show a typical image.

So, I guess that's a long way of saying I don't love those wording changes.

stephencmorton: Yes, but "color" was in the strings before. It was in the strings before it is in the strings after. So you don't object to the *change* you'd like to make _additional_ changes. You could choose to hold up this PR in order to get everything perfect, I see the benefit in that, although it seems to me that the discussions would likely go on for ever if that was the aim; or you could see if this change at least made things better and didn't make anything worse and realize that you have no objections and will likely create a further PR yourself in the future.

(Ok, this time I'm actually going to unsubscribe from this PR. It just makes me too frustrated.)

Navid200: You keep saying that and again come back.  Is someone twisting your arm?
You may be surprised to know that you are not the only one who has precious time.  Except, some people work hard without whining too much.  
We have people here like Tzachi who have worked hard consistently and have had many many PRs rejected, not just one.  And he has never whined about it.

If you don't like how we work here, get out and don't let the door hit you on the way out.

Navid200: I'm sorry about that.  I could have been less harsh there.

It's a lot of work to have to read all proposals and approve or reject.  You still have to read it to reject it.  But, the problem is that if you have to worry about what is going to be said to you if you reject and for that reason have to worry about rejecting a proposal, that would be a very unhealthy work environment.

I want the reviewers to be able to state their opinion with no fear of how their opinion is going to be criticized.
Just because someone doesn't say anything doesn't mean they don't see your complaints.
You are welcome to make proposals.  But, if you are going to be unhappy if your proposal is rejected, please don't make any proposals.

I agree 100% that sometimes it is better to take one step forward rather than taking a leap forward.  
I was going to say that instead of COB, I have Carbs ON Board.  And by removing the word color from the end, it still fits and is not a very long phrase.

jamorham: @Navid200 a reminder that incivility is not appropriate here.

In summary:

This PR started as changes suggested by @stephencmorton as to their personal opinion as to how to improve some parts of the user interface.  It wasn't to solve any particular raised issue that I know of.

I rejected the graphical changes for aesthetic reasons. @tolot27 redid the QR code graphic with red highlight which I accepted.

I've stated my reasons for not wishing to change the generic term barcode because I feel it constrains what that menu can be used for and so limits future options. I believe the term barcode to be widely understood but if not we can continue to discuss that.

The only remaining parts to this PR is the wording for the predictive simulation colors which were originally left intentionally vague and with a view to future expansion. That time has probably passed and so it is reasonable to consider improving that wording to be better understood by users and I'm simply asking for other's opinion on this to come to a decision. I asked @Navid200 about this as they are likely far closer to understanding how users react to these settings with their experience providing support to users.

This is quite an involved process to change a few words, but I'm simply responding to repeated requests to examine this PR and user interfaces are important to get right. Normally a suggestion to make a change would arise from a specific need. I might be blind to the need to change this text as I of course already know what it means which is why I would ask others for their opinion. I don't think it is a particularly high priority but I'm willing to consider it if others do think its worthwhile.

Navid200: I would use these words:

Insulin on Board
Insulin Activity
Glucose Prediction
Carbs on Board

In addition, many don't know what "Flair Colors" mean.
So, I think it's a good idea to edit those too.  But, that could be something else:
https://github.com/NightscoutFoundation/xDrip/pull/2031

app/src/main/res/values/strings.xml
Lupum1001: @jamorham I noticed that you merged https://github.com/NightscoutFoundation/xDrip/pull/2022 which changes the icon to a QR code. Given that, I'm ok with keeping the _text_ as "barcode".
At the very least, the juxtaposition clarifies to the users that WE consider the terms interchangeable. 👍

Explicitly calling it a 'barcode' scanner (with the old icon) was making people like myself assume that it explicitly _didn't_ work with QR codes. This was especially confusing because it's only used for QR codes in the current state of XDrip.

app/src/main/res/values/strings.xml
Lupum1001: > This is quite an involved process to change a few words, but I'm simply responding to repeated requests to examine this PR and user interfaces are important to get right. Normally a suggestion to make a change would arise from a specific need. I might be blind to the need to change this text as I of course already know what it means which is why I would ask others for their opinion. I don't think it is a particularly high priority but I'm willing to consider it if others do think its worthwhile.

@jamorham a completely understandable reaction and I doubt the OP would have bothered if he'd known it would ultimately explode into such a big discussion. I don't believe it's his (or anyone's) fault that it ended up consuming so much time though.

Are there changes to the _process_ that can be implemented to reduce the overhead to the repo managers? As a new contributor, I've been instructed to keep my PRs as limited in scope/incremental as possible, but that does increase the per-change overhead somewhat. New contributors are also inherently going to start with more "trivial" changes, due to lack of familiarity with the code base.
When it comes to things like translation for example, how do we decide what's "too trivial" to change?

app/src/main/res/values/strings.xml
```diff
@@ -38,9 +38,9 @@
     <string name="title_activity_calibration_override">Calibration Override</string>
     <string name="pref_header_cloud_storage">Cloud Storage</string>
     <string name="auto_configure_title">Auto configure</string>
-    <string name="prefs_auto_config_summary">Auto configure using a barcode.</string>
-    <string name="scan_share2_barcode">Scan Share Barcode</string>
-    <string name="pref_share2_scan_barcode_summary">Or scan the barcode on the share receiver</string>
+    <string name="prefs_auto_config_summary">Auto configure using a QR code.</string>
+    <string name="scan_share2_barcode">Scan Share Receiver barcode</string>
```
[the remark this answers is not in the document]
tolot27: In that case, we should remove this feature completely from xDrip. This will increase maintainability and also avoids work for translators of new languages like Estonian.

tolot27 (commented):
Please can we proceed with either this PR or #2031 and finally decide which wording is more appropriate?

Issue #1012, opened 2019-09-14
Project doesn't build


jsPl:
Can someone please update the project setup so that it builds?
I tried that myself for the past 2 days with Andorid Studio 3.5, and incrementally fixing errors that occured and causing other errors, updating gradle dependencies - to no avail....
Would really like test some of my ideas with this wonderful project.

MasterPlexus:
At the moment, please stay at Android studio 3.4, there are several projects which have problems with the newest version. With 3.4 everything works smooth.

jsPl:
Ok, tried with 3.4.
Cloned the project, first build failure is that the dependencies in wear:
```
    compileOnly 'com.google.android.wearable:wearable:2.0.1'
    implementation 'com.google.android.support:wearable:2.0.1'
```
no longer exist in any declared repository. So I changed it to 2.1.0.
Then there's another build error in wear manifest:

```
Execution failed for task ':wear:processDebugManifest'.
> Manifest merger failed : Attribute meta-data#android.support.VERSION@value value=(26.0.2) from [com.android.support:support-v4:26.0.2] AndroidManifest.xml:28:13-35 is also present at [com.android.support:support-core-utils:26.1.0] AndroidManifest.xml:28:13-35 value=(26.1.0).
Suggestion: add 'tools:replace="android:value"' to <meta-data> element at AndroidManifest.xml:26:9-28:38 to override.
```
This suggestion doesn't really help... Is something wrong with the manifest file?

jamorham:
It isn't a problem with manifest file. It is regressions introduced by Android Studio and poor stewardship of the maven repositories by google.

jamorham:
There are a couple of ways to work around this. I will see if I can integrate one of them.

RudieKat:
Managed to get things working (Note, below info is related to OS X so locations will vary)

Uninstall previous versions and remove all related stuff in 
~/Library/Application Support/AndroidStudio3.*
~/Library/Preferences/AndroidStudio3.*
~/Library/Caches/AndroidStudio3.*
~/Library/Logs/AndroidStudio3.*

also rm -rf ~/.android for good measure.
3.4.2 didn't work, 3.4.1 did so there is that too.

wkochFPV:
These are my first steps in android development... I managed to build the app using the current android studio version, based on tipps found on various  pages on the internet. This is what I have done:

1. I did not update  Gradle Plugin

[...]

Issue #878, opened 2019-05-18
Is there documentation on how to import and build xDrip


bert1973:
I pulled the project from Github, opened it with the latest version from IntelliJ community (IntelliJ IDEA 2018.2.8).
After build, I chose: Build Artifacts, but this keeps resulting in:
Error:java: invalid target release: x (where x is a version, e.g. 11)
Questions about import:
Is there any documentation?
Should I use the standard Gradle wrapper or custom?

tzachi-dar:
I think that most people use android studio to build xdrip.
You can also write gradlew assembledebug from command line to build it.

bert1973:
I decided to go with Android Studio, but even this one fails:
error: cannot find symbol class Models
error: package UtilityModels does not exist
error: package Models.ProcessInitialDataQuality does not exist
error: package UtilityModels does not exist
etc.

Do I have to do clone more then master from https://github.com/NightscoutFoundation/xDrip.git?

k2s:
the problem you have is related to comment `// use specific version of data binding until google resolves the bugs with mixed case package names` in `app/build.gradle`

when you open xDrip (after clean clone) in Android Studio it will ask you `Plugin Update Recommended: Android Gradle Plugin is ready to update.`. Ignore that otherwise you will get into troubles.

@jamorham do you know what is the problem ? by quick try there is no reason to wait until `google resolves the bugs with mixed case package names`. Is it needed to rename something manually in xDrip code to get it fixed and to upgrade to latest Gradle plugin ?

k2s:
related ticket #1012

tolot27:
As mentioned already, use Android Studio. I'm on 4.1.1 and it works pretty well.

PR #1160, opened 2020-01-07
Upgrade gradle

k2s:
working modifications to compile with new gradle and with upgraded libraries
* Gradle wrapper `5.1.1`
* `compileSdkVersion 29`
* `targetSdkVersion 23`
* `wearable:2.5.0`

~~also possible to sign debug/release APKs with personal `signing.properties` file~~

also less problems with memory problems (`javaMaxHeapSize "3g"`)


k2s:
* fixes #878
* fixes #1012 
* fixes task in #1037 
* makes possible to build APK with `./gradlew :app:assembleRelease`, so no manual Android Studio invokation is neede

jeromecc:
Hello!
I am using Android Studio 3.5.3
You branch throws an error:
```
Could not get unknown property 'release' for SigningConfig container of type org.gradle.api.internal.FactoryNamedDomainObjectContainer.
```
It refers to line 147 in xDrip/wear/build.gradle:
            signingConfig signingConfigs.release
when i comment it out, the project builds.
I couldn't build upstream, so I guess your PR should be considered, maybe after fixing the signingConfig stuff.
What is the puprose of signingconfig? (i am relatively new to android dev)
Thanks.
(in reference to https://github.com/NightscoutFoundation/xDrip/issues/1012)

k2s:
It was my mistake to add the signing.properties support into this branch. I removed it so this pull requests focuses on:

* upgrading of the Gradle version
* working with latest Android Studio
* and working command line `./gradlew :app:assembleRelease`

The signing changes were moved to my other pull request #1038 aka feature #1037.

I' am sorry for the inconvenience!

jamorham:
`./gradlew assembleRelease` already works, we use this in the continuous integration scripts?

wear/build.gradle
```diff
@@ -157,26 +157,27 @@ allprojects {
...
-    implementation 'com.google.android.support:wearable:2.0.1'
-    implementation 'com.android.support:appcompat-v7:23.1.1'
+    compileOnly 'com.google.android.wearable:wearable:2.5.0'
+    implementation 'com.google.android.support:wearable:2.5.0'
+    implementation 'com.android.support:appcompat-v7:26.1.0'
     // implementation 'com.google.android.support:wearable:1.4.0'
     //implementation 'com.google.android.gms:play-services-wearable:9.4.0'
-    implementation 'com.google.android.gms:play-services-wearable:10.2.1'
+    // FIX upgrage to play-services-wearable:17.0.0 requires upgrade to AndroidX libraries, https://developers.google.com/android/guides/releases#june_17_2019
+    implementation 'com.google.android.gms:play-services-wearable:16.0.1'
```
jamorham: Different versions of play services work in very different ways and may not be backwards compatible. How much testing has been done with the bi-directional syncing with Android wear and this version?

Also I'm pretty sure this will break rooted SWR50s but these may be far less common these days.

jamorham: Does this make the `play-services-wearable` library a different version between app and wear?

app/build.gradle
```diff
@@ -283,11 +289,11 @@ dependencies {
     implementation(name: 'hellocharts-library-release', ext: 'aar')
     implementation(name: 'search-preference', ext: 'aar')
     implementation 'org.apache.commons:commons-text:1.3'
-    implementation 'com.google.dagger:dagger:2.11'
+    implementation 'com.google.dagger:dagger:2.25.4'
```
jamorham: Are all these library version changes required to support gradle 5 ?

app/build.gradle
```diff
@@ -112,17 +112,22 @@ def generateVersionName = { ->
...
-    compileSdkVersion 26
-    buildToolsVersion '27.0.3'
+    compileSdkVersion 29
+
+    dexOptions {
+        javaMaxHeapSize "3g"
+    }
 
+    configurations.all {
+        resolutionStrategy.force 'com.google.code.findbugs:jsr305:1.3.9'
```
jamorham: Can you explain what this does? thanks

app/build.gradle
k2s: yes, it will not compile with 2.11

app/build.gradle
k2s: https://stackoverflow.com/a/37357786/645030

k2s:
> `./gradlew assembleRelease` already works, we use this in the continuous integration scripts?

it doesn't work with latest AndroidStudio which asks to upgraded gradle

it is related to https://github.com/NightscoutFoundation/xDrip/issues/1012#issuecomment-531602989 

this pull requests makes it possible to use latest Android Studio ... it makes base for next improvements to be compatible also with next versions with Gradle, where `android.databinding.enableV2=false` is not supported anymore and will need fixing of naming convention in some xDrip classes

wear/build.gradle
k2s: I am testing this with Sony Smartwatch 3 which was not syncing data to watchface for me and with this pull request it does. I am not able to test it as collector.

My version for testing build with this pull request is here http://bit.do/xdrip-k2s (https://gist.github.com/k2s/aa7928197d8ea508135bc20bb9bdd664) if somebody has the possibility to test it as collector.

You are right I didn't upgraded the play-services-wearable in app, because there are more play-services libraries at version 10.2.1. Maybe they should be and also firebase-messaging be upgraded too ?

It compiles also with the older version of play-services-wearable in wear. I have to test if it will not break syncing with my watch, becasue that was the reason why I upgraded it.

jamorham:
I would like to get this merged soon but am still doing some testing with it. Upgrading so many libraries makes me nervous as there are often regressions in newer versions.

wear/build.gradle
th122: > Also I'm pretty sure this will break rooted SWR50s but these may be far less common these days.

I'm using one of those quite happily, readily recommended it to others in the past, and am supporting at least one other happy user in my vicinity with that setup. So I'd be sad to see support for the SW50 dropped.

parapenT1sta:
> I would like to get this merged soon but am still doing some testing with it. Upgrading so many libraries makes me nervous as there are often regressions in newer versions.

If you want to share an APK with all updates I can test it.

sulkaharo:
FWIW I've now tried to build xDrip using fresh installs of Android Studio 3.4 and 3.5 and getting the build errors on both. Wondering if someone here would know how to get this working / if you've tried to get builds working on fresh AS installs recently? (Motivation is to get a branch working to test a change.)

tolot27:
I'm on the most recent version of Android Studio and it works flawlessly. It is important not to upgrade gradle.

sulkaharo:
Is gradle installed globally? I didn't upgrade it when prompted, but I guess it might be updated through something else, so wonder how to downgrade if that's the case

sulkaharo:
Ok found out how to clear gradle; if missing Android studio 3.5 installs version 5.4.1. Trying to downgrade that to 5.1.1 with a local install just causes AS to refuse to run with error "ERROR: Minimum supported Gradle version is 5.4.1. Current version is 5.1.1.". So I'm slightly confused on how you're able to run 5.1.1?

PR #1172, opened 2020-01-14
update bubble hardware

bubbledevteam:
update bubble hardware


bubbledevteam:
update bubble hardware

jamorham:
Can you provide a brief summary of what this actually does? thanks

gui-dos:
The last two bytes of the packet 0x80 contain the correct hardware version, currently 0x0200.

The Bubble advertises it also in the Bluetooth manufacturer data: you can already see that the current version is “2.0” and not  the “1.0” displayed currently by xDrip.

bubbledevteam:
> Can you provide a brief summary of what this actually does? thanks

As @gui-dos described, this request just fixes the hardware version,I hard code as 1.0 before.

For now DiaBox and other Apps implemented this by the last two bytes of the packet 0x80, so it's a little bit confusing for the users, we just need to fix it.

PR #1191, opened 2020-01-21
Update WatchGuide.md

TrevorScheitrum:
There should be a list of watches that work, so that people can plan what watch to buy without resorting to asking around for info from others.  Probably should also put a link to this documentation from the main README so people see it.


ANDREAPIPPI:
Hi. 
I absolutely agree on this request, I'm trying to chose among a zillion of existing watches, but I'm lost among the many ones discussed here, so.... welcome to the list of working ones!
Pippo

PR #1194, opened 2020-01-21
Hotfix4activityprediction - needs to be applied urgently

gruoner:
fixed 2 bugs and a poor line of code:
- removed concentration from activity calculation as discussed in #1080
- fixed a typo in activity calculation leading to a wrong prediction

### DIFF
--- a/Documentation/WatchGuide.md
+++ b/Documentation/WatchGuide.md
@@ -1,6 +1,7 @@
 xDrip+ Wear Setup and Troubleshooting Guide
 ======================================
 **Table of Contents**
+- [Choosing a Watch](#choosing-a-watch)
 - [Enable xDrip+ Android Wear Integration](#enable-xdrip-android-wear-integration)
     - [Patching Watch Firmware](#patching-watch-firmware)
     - [Collector](#collector)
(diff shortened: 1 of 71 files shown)


jamorham:
@stephencmorton can you provide a review of these modified calculations. thank you

stephencmorton:
The changes themselves look good and are necessary. (I think they're exactly what I put in my pull request #1150 )

One bigger-picture issue and a couple of bigger-picture questions for the original commit remain in my mind.
 - This patch does not fix the additional concentration issue identified last week.
- I also have some questions about an `abs()` call elsewhere in the code that seems unnecessary but is not harmful.
 - I also do not know why the original pre-PR1080 code did not directly use the IOB calculated but derived it elsewhere. The new code now uses the IOB calculated directly. I don't know if there is a bug, I just don't understand that part of the code yet.
 - For the original #1080 PR, I only looked at the insulin calculations. I know nothing about Android programming so I can't comment on that. There have been enough other issues discovered with this feature, that perhaps all of 1080 should be gone over with a fine-toothed comb.

jamorham:
Okay good. What I am looking for is incremental improvements without negative impacts. So if this improves the existing situation then I would plan to integrate it. 

Regarding the concentration parameter, that can be improved later when there is certainty about the implementation. As far as I know it is unnecessary and a risk if there is anything other than U100 defined. 

The multiple insulins feature is still on the nightly channel and is not enabled by default so there is still time for refinements and fixes.

Thanks to both of you for your attention to this feature.

PR #1204, opened 2020-01-26
Fixed problem with nightscout follow - Entry.date

ahaarrestad:
xDrip had Entry.date set as long, but nightscout is allowing
double values to be added. For example loop is uploading bg entries
with dates as doubles with decimals. This makes it impossible to use
the nightscout follow feature when uploading bg through loop.

This commit changes the datatype to double and rounds it back to long
when asking Entry for the timestamp. This way no changes are made
to the internal workings of xdrip, but nightscout follow started working
again.

The change is tested with nightscout sites which are both in dexcom
follow mode and loop upload mode.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/messages/Entry.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/messages/Entry.java
@@ -14,7 +14,7 @@ public class Entry extends BaseMessage {
     @Expose
     public String _id;
     @Expose
-    public long date;
+    public double date;
     @Expose
     public String dateString;
     @Expose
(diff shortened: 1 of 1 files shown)


ahaarrestad:
@jamorham Please consider merging this pull request. 

Error message without this fix is: 
NS entries download Failed:
com.google.gson.JsonSyntaxException:
java.lang.NumberFormatException: Expected a long but was "157982098720.356 at line 1 column 144 path$[0].date

Tested with live data to verify it works with new data and is backwards compatible, although I have only tested it with latest master of nightscout.

dabear:
This will also allow people that use the tomato app(libre) for uploading to nightscout have xdrip+ 
 ns folllowers

PR #1206, opened 2020-01-26
Minor code cleanup nightscout follow

ahaarrestad:
* Added _ separator in numbers to ease reading
* Removed unused imports

@jamorham No functional changes in this commit, only minor tweaks for better reading


jamorham:
I think whatever happened with your merge from master would break the recent changes to `AndroidManifest.xml` - Can you resolve?

PR #1207, opened 2020-01-27
Enable hex color editing for all color settings

cominixo01:
This allows the user to input manually a specific color for better theming convenience and precision, instead of having to select something approximate from the color wheel.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/ui/dialog/ColorPreferenceDialog.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/ui/dialog/ColorPreferenceDialog.java
@@ -29,7 +29,6 @@ public class ColorPreferenceDialog {
         }
         picker.setColor(color);
         picker.showAlpha(true);
-        picker.showHex(false);
         builder.setTitle(title)
                 .setView(picker)
                 .setPositiveButton(R.string.ok, new DialogInterface.OnClickListener() {
--- a/app/src/main/res/xml/xdrip_plus_prefs.xml
+++ b/app/src/main/res/xml/xdrip_plus_prefs.xml
@@ -36,14 +36,14 @@
                         android:title="@string/high_glucose_values"
                         app:colorpicker_noneSelectedSummaryText="@string/default_color_selected"
                         app:colorpicker_selectNoneButtonText="@string/revert_to_default"
-                        app:colorpicker_showHex="false" />
+                        app:colorpicker_showHex="true" />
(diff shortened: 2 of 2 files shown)


jamorham:
The hex value is disabled to simplify the user interface. Do you think it is possible to enable hex editing only when engineering mode is enabled?

cominixo01:
Yeah I considered doing that, I'll do it when I'm home.

cominixo01:
I'm not really sure how possible this is, the showHex variable is final and I'm not really sure what can be done without making different preferences with hex enabled and hide/show them, that would be really unoptimal.

app/src/main/java/com/eveningoutpost/dexdrip/ui/dialog/ColorPreferenceDialog.java
```diff
@@ -29,7 +29,6 @@ public static void pick(final Activity activity, final String pref, final String
         }
         picker.setColor(color);
         picker.showAlpha(true);
-        picker.showHex(false);
```
tolot27, line 32: use `picker.showHex(Home.get_engineering_mode());` here to enable it only in engineering mode.

app/src/main/res/xml/xdrip_plus_prefs.xml
```diff
@@ -36,30 +36,30 @@
                         android:title="@string/high_glucose_values"
                         app:colorpicker_noneSelectedSummaryText="@string/default_color_selected"
                         app:colorpicker_selectNoneButtonText="@string/revert_to_default"
-                        app:colorpicker_showHex="false" />
+                        app:colorpicker_showHex="true" />
```
tolot27, line 39: The default value is `true`. Hence, this attribute can be removed completely, also in the following preferences.

tolot27 (changes requested):
The review contains hints to implement jamorhams request.

Navid200:
It sounds like it is not easy to enable this only for engineering mode.
So, I suggest closing the PR.  If some day we can figure out a way to enable this only in engineering mode easily, we can reopen it.

tolot27:
> It sounds like it is not easy to enable this only for engineering mode.
> So, I suggest closing the PR. If some day we can figure out a way to enable this only in engineering mode easily, we can reopen it.

It is quite easy to implement this as I commented in https://github.com/NightscoutFoundation/xDrip/pull/1207#discussion_r523771775. @cominixo just has to implement it.

Navid200:
How long are we going to wait for him/her to implement it?
The last time he/she commented was 1.5 years ago. 
Or, is this going to be left open indefinitely?

PR #1245, opened 2020-02-18
extended range for Libre2 intercept -40 ... 20

keencave:
User claims that the intercept limitation from -20 to 20 is too small. Having tested with a few sensors it seems to be safe to extend the upper limit from 20 to 40 to have the ablility to use L2 sensors with a bigger difference between blood and readings. As far as we know the L2 sensors will not vary over time. The difference will be constant over the whole 14 days.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/Calibration.java
@@ -143,7 +143,7 @@ class Li2AppParameters extends SlopeParameters {
 
     @Override
     public double restrictIntercept(double intercept) {
-        return Math.min(Math.max(intercept, -20), 40);
+        return Math.min(Math.max(intercept, -40), 20);
     }
 }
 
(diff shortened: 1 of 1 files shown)


tzachi-dar:
I think that this change is very dangerous.
People will have a bg of 30, while xDrip will continue to show 80.
They will think that all is ok, while it is not.
If they are connected to some kind of a loop system it is even more dangerous. That system will not know what is going on.

keencave:
sorry, inverse logic. Will create a new PR.

keencave:
User claims that the intercept limitation from -20 to 20 is too small. This will be extended to -40 ... 20
Range exstimation:
Raw reading of 50 will be corrected with -40 to 10 -> save, the user will be warned early
Raw reading of lowest 39 will be corrected with +20 to 59 -> unchanged limit like before
Remark: As far as we know the L2 sensors will not vary over time. The difference will be constant over the whole 24 days.

jamorham:
@tzachi-dar what is your view on the updated PR?

tzachi-dar:
After the change the PR is 100% fine.
I'll send another pr to make sure that once BG is 59 we will show low, to make sure that people are aware that we can not measure anymore, but this is a different problem that also exists today.

Issue #1304, opened 2020-04-14
please include HTTP "Date:" header in xDrip Web Service responses


ecc1:
1. It's required by the HTTP spec.
2. It's a useful way for a rig to get the phone's time.

Please consider supporting the HTTP HEAD method too, which would be useful when just the time is desired, and the actual BG entries aren't needed.

Navid200:
Has this been addressed?  
Can we close this issue?

ecc1:
Yes, it works fine now, at least in the old 2020.11.27 version I'm running.

PR #1305, opened 2020-04-14
Add HTTP Date header and support HEAD method in xDrip Web Service

ecc1:
Here is a patch that addresses https://github.com/NightscoutFoundation/xDrip/issues/1304.  Lightly tested using "curl -i" and "curl -i --head".


ecc1:
Yes, I tested it using the "HTTP Client" app on the phone to localhost, and also using the open mode from my Linux desktop using "curl".

One thing that might be improved is my choice of DateFormatter for the Date header -- I just found a seemingly standard approach by googling, but I suspect there are other places in the code that already do date/time formatting that could be reused and/or conformed to this one so we don't use multiple libraries.

jamorham:
The use of the DateTimeFormatter is incompatible with Android versions prior to 8 resulting in app crashes on those platforms.

ecc1:
Sorry, that's my Android naivete showing. Is there a recommended approach that's backwards compatible?

Issue #1302, opened 2020-04-14
xDrip+ refreshrate


markxoe:
Hello,

i am using xDrip with Nightscout, but my Sensor sends BG-data every 5 minutes, but xDrip refreshes every 10 minutes...
Is there any way to change this?
I looked into the code, but didn't find any way to change this...

Thanks in advance

markxoe:
After some reverse engineering and testing, [`final long grace = Constants.SECOND_IN_MS * 10;`](https://github.com/NightscoutFoundation/xDrip/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollowService.java#L131) is (in my case) too short, because my uploader is sending the data after about 20-30 seconds after that...
A fix would be to let this value be changed by the User in the UI...
Could some Dev integrate this please?

Thanks

akadismas:
I also found that time delay on xDrip when refreshing BG-data may be sensible to difference between system time on Nightscout and on the phone with xDrip+

markxoe:
> 
> 
> I also found that time delay on xDrip when refreshing BG-data may be sensible to difference between system time on Nightscout and on the phone with xDrip+

I fixed it myself ( adjustable ): https://github.com/NightscoutFoundation/xDrip/pull/1307
But i have to wait, until someone looks at it

akadismas:
> I fixed it myself ( adjustable ): #1307
> But i have to wait, until someone looks at it

Suppose, it should be useful improvement, because eliminating system time gap won't solve the problem with uploader delay.

markxoe:
> 
> 
> > I fixed it myself ( adjustable ): #1307
> > But i have to wait, until someone looks at it
> 
> Suppose, it should be useful improvement, because eliminating system time gap won't solve the problem with uploader delay.

In my case it helps tho....
My Uploader uploads the data to nightscout, but xdrip+ syncs before that...

markxoe:
Soo this won't be fixed?
I really don't like an app that doesn't refreshes before the value is uploaded

tolot27:
Looks like a duplicate of #929.

Navid200:
Closing because it is very similar to #929.  Please add any comments you may have to that issue.
Thanks

Navid200:
[...]

PR #1307, opened 2020-04-15
Adjustable Grace (Interval added to refresh Interval)

markxoe:
Hello,
i had an [issue](https://github.com/NightscoutFoundation/xDrip/issues/1302) which i solved with this, you can now change the [grace](https://github.com/NightscoutFoundation/xDrip/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollowService.java#L131) (which is an addition to the refresh interval)
The value can be set via preferences in the app, and can be seen in the System Status
My issue was, that xdrip polled the data from Nightscout, before my Uploader could upload it, and this "update" can delay the poll, so that the uploader uploads the data and xdrip can poll it after that

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollowService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollowService.java
@@ -128,7 +128,7 @@ public class NightscoutFollowService extends ForegroundService {
         final BgReading lastBg = BgReading.lastNoSenssor();
         final long last = lastBg != null ? lastBg.timestamp : 0;
 
-        final long grace = Constants.SECOND_IN_MS * 10;
+        final long grace = Constants.SECOND_IN_MS * getGrace();
         final long next = Anticipate.next(JoH.tsl(), last, SAMPLE_PERIOD, grace) + grace;
         wakeup_time = next;
         UserError.Log.d(TAG, "Anticipate next: " + JoH.dateTimeText(next) + "  last: " + JoH.dateTimeText(last));
@@ -140,7 +140,9 @@ public class NightscoutFollowService extends ForegroundService {
     private static boolean shouldServiceRun() {
         return DexCollectionType.getDexCollectionType() == NSFollow;
     }
-
+    private static int getGrace(){
+        return Integer.valueOf(Pref.getString("nsfollow_grace", "10"));
+    }
     /**
      * MegaStatus for Nightscout Follower
      */
@@ -202,6 +204,7 @@ public class NightscoutFollowService extends ForegroundService {
         statuses.add(new StatusItem());
         statuses.add(new StatusItem("Last poll", lastPollText + (lastPoll > 0 ? " ago" : "")));
         statuses.add(new StatusItem("Next poll in", JoH.niceTimeScalar(wakeup_time - JoH.tsl())));
+        statuses.add(new StatusItem("Interval addition",getGrace()+" Seconds"));
         if (lastBg != null) {
             statuses.add(new StatusItem("Last BG time", JoH.dateTimeText(lastBg.timestamp)));
         }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
@@ -1028,6 +1028,7 @@ public class Preferences extends BasePreferenceActivity implements SearchPrefere
 
             final Preference nsFollowDownload = findPreference("nsfollow_download_treatments");
             final Preference nsFollowUrl = findPreference("nsfollow_url");
+            final Preference nsFollowGrace = findPreference("nsfollow_grace");
             try {
                 nsFollowUrl.setOnPreferenceChangeListener((preference, newValue) -> {
                     NightscoutFollow.resetInstance();
@@ -1388,6 +1389,7 @@ public class Preferences extends BasePreferenceActivity implements SearchPrefere
                 try {
                     collectionCategory.removePreference(nsFollowUrl);
                     collectionCategory.removePreference(nsFollowDownload);
+                    collectionCategory.removePreference(nsFollowGrace);
                 } catch (Exception e) {
                     //
                 }
@@ -2014,6 +2016,7 @@ public class Preferences extends BasePreferenceActivity implements SearchPrefere
                     if (collectionType == DexCollectionType.NSFollow) {
                         collectionCategory.addPreference(nsFollowUrl);
                         collectionCategory.addPreference(nsFollowDownload);
+                        collectionCategory.addPreference(nsFollowGrace);
                     }
 
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1254,6 +1254,7 @@
     <string name="title_nsfollow_url">Nightscout Follow URL</string>
     <string name="summary_nsfollow_download_treatments">Also download treatments from Nightscout as follower</string>
     <string name="title_nsfollow_download_treatments">Download Treatments</string>
+    <string name="title_nsfollow_grace">Interval addition</string>
     <string name="title_ob1_options">OB1 G5/G6 Collector Settings</string>
     <string name="summary_use_ob1_g5_collector_service">Complete re-write, should work on Android 4.4-9, supports native mode and more</string>
     <string name="title_use_ob1_g5_collector_service">Use the OB1 Collector</string>
--- a/app/src/main/res/xml/pref_data_source.xml
+++ b/app/src/main/res/xml/pref_data_source.xml
@@ -115,6 +115,12 @@
             android:summary="@string/summary_nsfollow_download_treatments"
             android:title="@string/title_nsfollow_download_treatments"
             />
+        <EditTextPreference
+            android:defaultValue="10"
+            android:inputType="number"
+            android:key="nsfollow_grace"
+            android:summary="Delay after 5 minute Interval in seconds"
+            android:title="@string/title_nsfollow_grace" />
 
         <EditTextPreference
             android:defaultValue=""


mutable-dan:
where you able to determine why it was missing BG readings?  it did not do that in the beginning. i have noticed that the xdrip UI gets very slow when you keep BG readings for more than 90 days.  Just speculating if that slow down might be contributing to the missed readings.

markxoe:
No, the Nightscout Uploader is uploading the Data at the same time as xdrip is checking for new Data. So It makes sense to check a bit later, which is possible with that

mutable-dan:
i wrote a dexcom client, i have it check every 5m and a few seconds and when there is a missed reading, i try x times every 10s and then fallback so i don't 'annoy' dexcom

mutable-dan:
@jamorham 
this is actually a very useful fix.  using xdrip as a follower, it consistently missed data every other 5 minuets because of the grace.  for example, i need to do a 'restart collector' almost anytime i need to see a number (or wait until it reaches 10 min).

Please consider merging this pul request.
thanks

tolot27:
After my observations w/o this patch I suspect a bug in `Anticipate.java` because every time a bg value is received, the next poll time is set to "last + SAMPLE_PERIOD" regardless of grace.
Please try setting verbose log for NightscoutFollowService (`NightscoutFollow:v` in extra logging) to verify this or monitor the Nightscout Follow status.

mutable-dan:
hi
I set the tag and will upload soon

thank you

-------- Original Message --------

mutable-dan:
I uploaded a log file from xdrip and ref'd this pull.  odd thing is that after i added the additional logging, it was working pretty good.  took a while until i saw the issue.

thank you

mutable-dan:
@tolot27 @markxoe 
Hi,
I upload the logs last week with a ref to the pull using the upload facility.  Did you happen to see them?

Interesting that the phone with verbose logging now rarely misses data and the phone that does not have verbose logging, misses data often.

tolot27:
@mutable-dan Unfortunately, I don't have access to the uploaded logs. Please can you post an extract here?

mutable-dan:
> @mutable-dan Unfortunately, I don't have access to the uploaded logs. Please can you post an extract here?
hi @tolot27 
I don't see how to get the logs off of phone.  only option i see is for emailing from app. can't find xdrip logs in filesytem

please advise.

tolot27:
> I don't see how to get the logs off of phone.  only option i see is for emailing from app. can't find xdrip logs in filesytem
> please advise.

Attaching a screenshot is the simplest way but not prefered by me because I cannot copy text like exceptions from it. Another way is to export the database, copy it to a computer, unzip it and using i. e. [DB Browser for SQLite](https://sqlitebrowser.org) to extract the newest entries from table UserErrors.

mutable-dan:
```
33903 | Woke up Early.. | 2 | NightscoutFollow | 1605216358887
-- | -- | -- | -- | --
33904 | Aggressively restarting collector service due to lack of reception: backoff: 180 | 2 | MissedReadingService | 1605219706722
33905 | Woke up Early.. | 2 | NightscoutFollow | 1605219783217
33906 | Woke up Early.. | 2 | NightscoutFollow | 1605220258680
33907 | Woke up Early.. | 2 | NightscoutFollow | 1605221160453
33908 | Woke up Early.. | 2 | NightscoutFollow | 1605222960924
33909 | Woke up Early.. | 2 | NightscoutFollow | 1605223259984
33910 | Aggressively restarting collector service due to lack of reception: backoff: 240 | 2 | MissedReadingService | 1605224314517
33911 | Woke up Early.. | 2 | NightscoutFollow | 1605224406292
33912 | Aggressively restarting collector service due to lack of reception: backoff: 300 | 2 | MissedReadingService | 1605224675531
33913 | Woke up Early.. | 2 | NightscoutFollow | 1605224737254
33914 | Woke up Early.. | 2 | NightscoutFollow | 1605224748800
33915 | BG Readings Missed  (@18:49) | 6 | Other Alert | 1605224977983
33916 | Aggressively restarting collector service due to lack of reception: backoff: 120 | 2 | MissedReadingService | 1605225025767
33917 | Woke up Early.. | 2 | NightscoutFollow | 1605225028649
33918 | BG HIGH ALERT: 141  (@19:35) | 6 | Alerting | 1605227742833
33919 | bg rising fast  (@19:35) | 6 | Other Alert | 1605227744168
33920 | setting next alert to 1 minute from now (no problem right now, but needs a fix someplace else) | 1 | AlertPlayer | 1605228944517
33921 | Woke up Early.. | 2 | NightscoutFollow | 1605229259973
33922 | BG HIGH ALERT: 133  (@20:36) | 6 | Alerting | 1605231362636


```

@tolot27  pls let me know if this provides enough info

mutable-dan:
```

34144 | Woke up Early.. | 2 | NightscoutFollow | 1605321662771
-- | -- | -- | -- | --
34145 | Woke up Early.. | 2 | NightscoutFollow | 1605321662818
34146 | Aggressively restarting collector service due to lack of reception: backoff: 180 | 2 | MissedReadingService | 1605321957591
34147 | Woke up Early.. | 2 | NightscoutFollow | 1605321962976
34148 | Aggressively restarting collector service due to lack of reception: backoff: 240 | 2 | MissedReadingService | 1605322325914
34149 | Woke up Early.. | 2 | NightscoutFollow | 1605322328078
34150 | BG Readings Missed  (@21:55) | 6 | Other Alert | 1605322508201
34151 | Woke up Early.. | 2 | NightscoutFollow | 1605322563894
34152 | Aggressively restarting collector service due to lack of reception: backoff: 120 | 2 | MissedReadingService | 1605322689471
34153 | Woke up Early.. | 2 | NightscoutFollow | 1605322739187
34154 | Woke up Early.. | 2 | NightscoutFollow | 1605322861461
34155 | Aggressively restarting collector service due to lack of reception: backoff: 180 | 2 | MissedReadingService | 1605323050501
34156 | Woke up Early.. | 2 | NightscoutFollow | 1605323139644
34157 | Woke up Early.. | 2 | NightscoutFollow | 1605323153434
34158 | Aggressively restarting collector service due to lack of reception: backoff: 240 | 2 | MissedReadingService | 1605323405902
34159 | Woke up Early.. | 2 | NightscoutFollow | 1605323408077
34160 | Aggressively restarting collector service due to lack of reception: backoff: 300 | 2 | MissedReadingService | 1605323709612
34161 | BG Readings Missed  (@22:15) | 6 | Other Alert | 1605323710245
34162 | Woke up Early.. | 2 | NightscoutFollow | 1605323711843
34163 | Aggressively restarting collector service due to lack of reception: backoff: 120 | 2 | MissedReadingService | 1605324133141
34164 | Woke up Early.. | 2 | NightscoutFollow | 1605324135305
34165 | Aggressively restarting collector service due to lack of reception: backoff: 180 | 2 | MissedReadingService | 1605324502324
34166 | Woke up Early.. | 2 | NightscoutFollow | 1605324547720
34167 | Woke up Early.. | 2 | NightscoutFollow | 1605324642525
34168 | Woke up Early.. | 2 | NightscoutFollow | 1605324662082
34169 | Aggressively restarting collector service due to lack of reception: backoff: 240 | 2 | MissedReadingService | 1605324858754
34170 | Woke up Early.. | 2 | NightscoutFollow | 1605324860917
34171 | BG Readings Missed  (@22:35) | 6 | Other Alert | 1605324911945
34172 | Aggressively restarting collector service due to lack of reception: backoff: 120 | 2 | MissedReadingService | 1605325227756
34173 | Woke up Early.. | 2 | NightscoutFollow | 1605325243286


```

mutable-dan:
it looks like the latest release 2021-03-08 fixed this issue

markxoe:
In my case it didn't fix it 😕

mutable-dan:
@markxoe 
i noticed that phones with faster processors seem to do a little better.
also, how many months of data do you save?  it does a lot of db queries and they are not too snappy.  wonder if that effects this?

markxoe:
It'd be really unfortunate if you need an Intel i9 in your phone to use xDrip without having an 5 minutes Delay...
The Queries to the Nightscout API for 1000 entries need in my case about 83ms. I can't imagine that this is the problem

If I watch the system status in xDrip closely I can see that it tries do download data at the time the uploader should have uploaded the data, then about 30 seconds later the uploader queries the pump and uploads the data to nightscout. But xDrip already checked nightscout before the upload happened and then xDrip waits for another 5 minutes because the download was unsuccessful, only to run into the same problem and only downloading the previous entry.

psit:
@mutable-dan this is definitively not a performance issue. I have the same problem following Nightscout with 600SeriesUploader (xDrip requesting data too early as described above).
An adjustable grace would be perfect for my needs aswell.

Navid200:
I have seen cases where the problem was resolved as soon as both the master and follower were set (Android Settings -> System -> Date & Time) to sync time with the network.

mutable-dan:
@Navid200 android defaults to network time.  and this happens when the source is dexcom

@tolot27 any thoughts?

Navid200:
What is the maximum possible delay from the moment the master has a new reading until the corresponding data point is uploaded to Nightscout?
Is it more than 1 second?  Is it more than 5 seconds?
It cannot be a minute, can it?

If the maximum delay is say 1 second, why not just change the constant 10ms to a 2s?
No user is going to notice a difference of 10ms with 2s.  And, there will be no need for creating a new setting.

markxoe:
The regular delay of data being parsed from my 670G pump until uploaded into the database of Nightscout is in my case about 45 seconds. The default Grace of 10 Seconds (not 10ms) won't do its job with that. And if I measured long enough, I will surely measure a full minute if the Pump is busy.

Also the timestamp of the datapoint is (in my case) set to the original value, so that if it gets uploaded 1m later, the timestamp of the value will be 1m before that. Idk if that's a good or bad thing though, or if it is just a setting for the 600SeriesUploader...

I've thought about this problem a few times since creating this PR, and I think, that an algorithm that looks at past delays and adjusts the grace accordingly is one of my favorite Solutions.

Navid200:
What is the pump uploading to the Nightscout that the follower needs to see with minimal delay?
I thought this was about seeing blood glucose.
I mean if someone's child is going to have a low blood glucose, I understand the parent would want to know right away.
But, what is the pump uploading?

Why does it have so much delay?
Can the delay be reduced?

psit:
Its not that easy, as far as I understand:
The sensor reads SG values every 5 minutes and pushes to/is pulled by the pump. That takes time.
The uploader pulls data from the pump with some grace delay (more time). This pulling happens over some proprietary 2.4 GHz radio protocol (takes some time). The uploader not only fetches SG values, but also "all" other data from the pump (status, battery, treatments, bolus, basal, ...). It takes time to process the data. Then it pushes the data with the original timestamp (from the sensor) to Nightscout. 

You see, there are many hops, all wireless, and each device adds some grace delay. IIRC you can set the delay in the uploader, but lowering that could result in missed readings because the pump might not be ready for sending, resulting in even more delay when retrying and/or missing readings at all.

Navid200:
That wasn't the question.  
What is this system that only has a pump as a collector, and not a CGM?
Can you provide a link to it?
I am aware of pumps that can collect, like a t:slim.  But, it collects from a Dexcom transmitter.  So, the user can have xDrip also collect from the same transmitter.

I just want to understand what the problem is before we can talk about what the solution may need to be.

psit:
I'm wonderung why this is a problem to add this single configuration parameter for NS follower data source. NS follower has 2 (i.e., two) settings. Why not add a 3rd? This would solve this problem once for all. No need for an algorithm that brings complexity and takes CPU. Keep it simple.
It feels like there are 200 parameters in xDrip, 95 % of which are not needed by NS follower (at least I don't need them for following). And you argue about one more? I don't get the point. Sorry.
Instead I always have to switch to NS webinterface when my T1D is getting low to see the live data. This is annoying and IMHO not necessary.

psit:
Sorry, the pump is medtronic 640G in my case.
Guardian 3 sensor --> 640G pump --> contour next link USB dongle --> mobile phone with uploader app --> NS --> xDrp NS follower

Navid200:
@psit This request has been sitting here for more than a year.
I didn't post in it to have an argument.  Definitely not with you who are not the person who opened it.
My intent is to understand what the need is.  As one of the volunteers, one of my tasks is to contribute to the review process, which is why I posted here.
You don't have to contribute to the review process.

But, please don't make life difficult for the volunteers.  

In general, if there is no not a valid reason to add something, it will not be added.  The only way to convince the developers that something is needed is through a healthy conversation and understanding rather than antagonizing them.

psit:
Sorry, I'm a developer myself. I didn't want to be rude. I just have had the discussion about this feature some time ago and I it looks like I mixed up the threads... sorry for that.
To get nack to work: For more information about the possible setting I described above, see the uploader repo at https://github.com/pazaan/600SeriesAndroidUploader
I hope that helps in understanding.

Navid200:
Thanks for that link.

What is this "upload to xDrip"?
https://github.com/pazaan/600SeriesAndroidUploader/commit/06439c9d138a9638b2329470ccc84fd5405e1e2c

Does that mean the app can directly communicate with xDrip?
Why is the Nightscout delay a concern then?

psit:
This only works on the same phone and only for SG readings, not for treatments.

Navid200:
So, do I understand correctly that if someone's child is using this system, they can set up xDrip on the master phone to receive the blood glucose readings directly and use a follower scheme on the parent's phone to follow with no 5-minute delay?
It's only the treatments that benefit from this PR?
But, why does it matter if there is an extra 5-minute delay in receiving the treatments on the follower as long as the blood glucose alerts are on time?

psit:
So you recommend using xDrip (master) to share SG values to another xDrip (follower) and the uploader to share the rest?
Can xDrip be set up to follow both NS and another xDrip? That seems btw. rather cumbersome to me.

Navid200:
I am wondering what the delay will be if the glucose readings are directly transferred to the master.  If they are transferred fast enough, the master can still upload the glucose readings to Nightscout.  
The follower will get both glucose readings and treatments from Nightscout.

Assuming the glucose readings reach Nightscout early enough, the follower will receive glucose readings with no delay.  But, it will still receive the treatment data with a 5-minute delay, through the same Nightsccout download.  So, some pipelining will be going on.  But, as far as setting the follower up, it will be just like before.

markxoe:
Just to clarify some points:
In the Medtronic 600 Series Pump with the Guardian sensor world you cannot send the data directly to the uploader as you won't have the benefits of the 670G auto mode.
The data has to go through the pump to the uploader and then to Nightscout.

Secondly in my case there's no difference between the treatments and glucose readings, as it gets uploaded at the same time and xdrip gets this data at the same time.

The problem is that xdrip tries to download data and fails because the data will be there 20 seconds after that. Xdrip then waits another 5 minutes and downloads the old data and fails to load the new data as it needs 20 more seconds to download from the pump. And so on until infinity or until there is a better solution to this problem

Navid200:
@markxoe 
What is the point of the thumbs down emoji?

tolot27:
> I'm wondering why this is a problem to add this single configuration parameter for NS follower data source. NS follower has 2 (i.e., two) settings. Why not add a 3rd?

Because there is an existing algorithm which should address this issue but it looks like it is buggy or incomplete, as I commented in https://github.com/NightscoutFoundation/xDrip/pull/1307#issuecomment-696408857.

Hence, it is advisable to fix this anticipation instead of adding a new setting.

markxoe:
I was not able to see this behavior in my case.
After observing the official release of the app and [playing around](https://gist.github.com/markxoe/f066c62ab0187ff47a7a1a5798d3f42b) with [`Anticipate.java`](https://github.com/NightscoutFoundation/xDrip/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/utils/Anticipate.java) I can't confim your `last + SAMPLE_PERIOD` theory.

Also I don't really know why [here](https://github.com/NightscoutFoundation/xDrip/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollowService.java#L132) is a `+ grace` though.

In the meantime I believe that the new setting can perhaps be sold as a further "advanced setting" if we do not come up with a better solution ...

mutable-dan:
perhaps a compromise to fixing the algo would be to add an optional refresh button or pull down on the screen to refresh

psit:
> perhaps a compromise to fixing the algo would be to add an optional refresh button or pull down on the screen to refresh

No, it's not.
To get the most current data manually, you can already "restart collector" in the status window.
But we need the most recent data on time especially for hypo or other alerts, when we are not constantly looking at the app.

olorinmaia:
Where can I adjust Grace in xDrip+ settings? I'm using Dexshare follow as data source on my xDrip+ and every 5 minutes or so it doesnt update alltho the new BS-data is available in Dexcom Follow app. Having to manually hit "restart collector" in system status on my phone is a bit irritating.

mutable-dan:
> > perhaps a compromise to fixing the algo would be to add an optional refresh button or pull down on the screen to refresh
> 
> No, it's not. To get the most current data manually, you can already "restart collector" in the status window. But we need the most recent data on time especially for hypo or other alerts, when we are not constantly looking at the app.

I know a pull down is not ideal. but perhaps until this is resolved.  it's so much quicker (esp wen you have to do it many times a day) to do a pull down refresh than:
1. menu
2. system status
3. restart collector
4. back arrow

mutable-dan:
Hi @tolot27,

do you have time to review this change or suggest another solution?

thanks

olorinmaia:
Would it be possible to add a hard coded delay of 5-10 seconds in addition to this improvement until a setting can be added to adjust grace per user?

Jackenmen:
I think it would be great if you could also set up a retry amount (would be a simple multiplier for the `grace` operand in the `if (modulus < -grace) {` line of the `Anticipate.next()` method) as this way one could minimize the delay by lowering the grace period and increasing the retry amount (personally I'm using 20 seconds grace with 3 retries in my *hardcoded* patch). It would also help if the situation with `grace` was sensibly resolved as right now `grace` gets added to the return value of `Anticipate.next()` in `NightscoutFollower` which only makes sense for some of the situations where it does that (cases where it's going to do the "long wait"/"wait after receiving BG successfully") while it doesn't for others (cases where it's going to do the "short wait"/"wait before retrying after failure"). See below for my interpretation of the issue with that.

Alternatively (I think this may be more painful to implement?), it could make sense to add a parameter for a bigger *first* offset (say, the rather consistent 40 or 45 seconds) while either keeping the grace at 10 seconds or maybe still allowing to configure it to something else.

---

> Also I don't really know why [here](https://github.com/NightscoutFoundation/xDrip/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollowService.java?rgh-link-date=2021-11-17T16%3A54%3A02Z#L132) is a `+ grace` though.

It's probably done because `grace` is subtracted from the returned wake-up time in the `Anticipate.next()` method and this makes it "zero out" (except `grace` isn't subtracted from the wakeup time in the `else` branch there so I don't think it works entirely as intended). I'm guessing that it doesn't make as much sense for other followers (probably the ones implemented earlier?) which is why the subtraction is part of the `Anticipate` class rather than individual followers which would seem like a better idea but I could be very wrong here.

Navid200:
How can I recreate this problem?

When I set up a follower of Nightscout, the follower receives readings once every 5 minutes.  The delay is something like a few seconds to sometimes one minute.
I never get a missed reading resulting in receiving one reading every 10 minutes.

I hope we can understand a problem before we fix it.

Jackenmen:
@Navid200 The problem is that xDrip tries to fetch a new reading from Nightscout five minutes after the last BG reading time (the time at which the reading was actually taken by the sensor) rather than say, last time the reading was added to Nightscout. 

To reproduce this, you'll need to consistently upload BG data to Nightscout one minute (40 seconds should also be enough since that's the kind of delay we experience when using 600 Series Uploader) after the time of the reading of the BG. For example, you would upload an entry with `dateString` set to 2023-05-01 16:50:00 but actually upload it through the API at 2023-05-01 16:51:00. I'm explaining this in terms of API calls since I'm not sure what uploader you are using and whether you have a way to delay it enough with your uploader to reproduce this like that. If you have one of Medtronic 600 pumps then you most likely can as the whole communication process through pump's proprietary protocol does take a while.

I think this should be enough to understand and reproduce the problem but let me know if you have any more questions.

Navid200:
I use xDrip to upload to Nightscout.
Thanks for the explanation.  So, this only is for individuals whose uploader has a problem.
Should this not be addressed on the uploader side?

Jackenmen:
> Should this not be addressed on the uploader side?

I don't think this can reasonably be addressed on the uploader's side, the delay related to the connection with the pump just adds up really fast. This protocol is meant for single synchronization with a PC maybe once every 2 weeks, not doing the sync every 5 minutes so it isn't exactly optimized for fast transmission (I'm not sure if you've ever had a Medtronic pump with CareLink but if you did then think about how long the CareLink sync took). It has to do the whole authorization handshake every single time, for example.

On top of that, Android forces some amount of additional delay for having to request USB permission every time and then there's the actual amount of time taken by the synchronization - SGV sync itself takes maybe 2-3 seconds (which isn't a lot but is still not insignificant) but we also want to sync other data from pump history (not every 5 minutes perhaps but few times an hour would be good) such as the data about taken bolus, temporary basal, pump alerts or sometimes also the less often changed data such as pump settings and all of that takes time as well. It doesn't take as long as CareLink sync as there is less data to sync but it's still taking a bit.

There's also a need for radio clash avoidance which means that the uploader actually has to wait 10 seconds *after* the value is expected to be received on a pump (the time could perhaps be minimized somewhat but it couldn't be eliminated) as otherwise the data either won't already be on a pump or worse, the pump won't be able to actually receive the data from the sensor due to the uploader radio clashing with the sensor transmitter which in some situations (without proper retry delay) could even lead to prolonged lost signal (I haven't played with the default delay values in the uploader and so I didn't have these issues but I'm trusting that it isn't a baseless statement).

Navid200:
I believe I have been able to recreate this problem: https://github.com/NightscoutFoundation/xDrip/discussions/3164
I intend to open a PR to address it.

Navid200:
The latest Nightly release (https://github.com/NightscoutFoundation/xDrip/releases/tag/2023.11.21) has a new setting when Nightscout follower is selected as the data source that lets you delay when the follower reads from Nightscout.
Please use it to see if it addresses this.

psit:
Thanks Navid200, this new settings now works for me. New entries form nightscout appear now within 40 seconds delay on my following xdrip. 
I dont fully understand why you offer only fixed delay values in the settings and not a free text field, but it's ok it is now and much better than before.
Thank you very much for this feature implementation!

Navid200:
@psit The only reason I did that was that I thought it would be easier to explain in the guide, and it would be less likely to cause a controversy.
You will not believe the stories you read on facebook as advice on how xDrip should be used.
Yesterday, someone claimed that the latest releases of xDrip only work with Android 14 and for Android 11, one should only use an old release from July of 2023.

Anyway, I am happy it is working for you, and we can always improve xDrip.  You can consider my change as a step forward.  There is no reason we cannot take more steps if/when needed.

Navid200:
Thanks for your submission.
This issue has now been addressed.  Would you please close this PR?

Thanks

PR #1312, opened 2020-04-20
Date header, take 2: Use SimpleDateFormat for better backwards compatibility

ecc1:
This version uses SimpleDateFormatter so should work on earlier Android versions.


tolot27:
Fixed with 4711331.

PR #1325, opened 2020-04-30
Tzachi libre all house coverage.

tzachi-dar:
This pr does the following:
1) Allow more than one phone to be used as a collector in a whole house network for libre (miaomiao/blucon/bubble). This is based on (almost) zero setup on xdrip normal channels.

2) Allow xDrip followers to see libre trend and bridge battery level.

3) Stop the sensor in a case a new sensor is detected also for bubble/direct scan/and libre wifi.

To add another receiver one has to do the following:
On the new phone:
1) Use source wizard to connect to libre. (until the new sensor has been scanned).
2) Use qr code to get security settings from current master.
3) Set the phone to master.
4) One should not start a new sensor on the new phone.

There is one improvement that I'm thinking of doing:
If the user will start a new sensor, and the user has also a follower phone on the network, the follower phone will receive BG messages from both phones.
To prevent this I'm thinking of adding a new button that will say something like: This phone is a libre receiver only. If this is set, the phone will only send messages of libreblock, and snooze alerts but not other messages (such as calibrations and BG).
Does this improvement looks good? Any suggestions for a better method?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -331,4 +331,12 @@ dependencies {
     annotationProcessor "org.projectlombok:lombok:1.18.10"
 }
 
+tasks.withType(Test) { 
+  testLogging {
+    exceptionFormat "full"
+    events "started", "skipped", "passed", "failed"
+    showStandardStreams true
+  }
+}
+
 apply plugin: 'com.google.gms.google-services'
--- a/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
@@ -795,7 +795,7 @@ public class GcmActivity extends FauxActivity {
         if (!Home.get_master()) {
             return;
         }
-        if (!JoH.ratelimit("libre-allhouse", 5)) {
+        if (!JoH.pratelimit("libre-allhouse", 5)) {
             // Do not create storm of packets.
             Log.e(TAG, "Rate limited start libre-allhouse");
             return;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
@@ -688,6 +688,13 @@ public class GcmActivity extends FauxActivity {
             GcmActivity.sendMessage(myIdentity(), "cal2", json);
         }
     }
+    public static void pushLibreBlock(String libreBlock) {
+        Log.i(TAG, "libreBlock called: " + libreBlock);
+        if (!Home.get_master()) {
+            return;
+        }
+        GcmActivity.sendMessage(myIdentity(), "libreBlock", libreBlock);
+    }
 
     public static void clearLastCalibration(String uuid) {
         sendMessage(myIdentity(), "clc", uuid);
--- a/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
@@ -279,6 +279,10 @@ public class GcmActivity extends FauxActivity {
             UserError.Log.wtf(TAG, "Cannot sync null bgreading - should never occur");
             return;
         }
+        if(Home.get_is_libre_whole_house_collector()) {
+            Log.w(TAG, "Not sending syncBGReading since this is libre allhouse collector.");
+            return;
+        }
         Log.d(TAG, "syncBGReading called");
         if (JoH.ratelimit("gcm-bgs-batch", 15)) {
             GcmActivity.sendMessage("bgs", bgReading.toJSON(true));
@@ -292,6 +296,10 @@ public class GcmActivity extends FauxActivity {
     // called only from interactive or evaluated new data
     public synchronized static void syncBloodTests() {
         Log.d(TAG, "syncBloodTests called");
+        if(Home.get_is_libre_whole_house_collector()) {
+            Log.w(TAG, "Not sending syncBloodTests since this is libre allhouse collector.");
+            return;
+        }
         if (Home.get_master_or_follower()) {
             if (JoH.ratelimit("gcm-btmm-send", 4)) {
                 final byte[] this_btmm = BloodTest.toMultiMessage(BloodTest.last(12));
@@ -308,6 +316,10 @@ public class GcmActivity extends FauxActivity {
     private synchronized static void processBgsBatch(boolean send_now) {
         final byte[] value = PersistentStore.getBytes("gcm-bgs-batch-queue");
         Log.d(TAG, "Processing BgsBatch: length: " + value.length + " now:" + send_now);
+        if(Home.get_is_libre_whole_house_collector()) {
+            Log.w(TAG, "Not sending processBgsBatch since this is libre allhouse collector.");
+            return;
+        }
         if ((send_now) || (value.length > (RELIABLE_MAX_BINARY_PAYLOAD - 100))) {
             if (value.length > 0) {
                 PersistentStore.setString("gcm-bgs-batch-queue", "");
@@ -332,6 +344,10 @@ public class GcmActivity extends FauxActivity {
         Log.i(TAG, "syncSensor called");
         if (sensor == null) {
             Log.e(TAG, "syncSensor sensor is null");
+            return;
+        }
+        if(Home.get_is_libre_whole_house_collector()) {
+            Log.w(TAG, "Not sending syncSensor since this is libre allhouse collector.");
             return;
         }
         if ((!forceSend) && !JoH.pratelimit("GcmSensorCalibrationsUpdate", 300)) {
@@ -386,18 +402,30 @@ public class GcmActivity extends FauxActivity {
     }
 
     static void sendLocation(final String location) {
+        if(Home.get_is_libre_whole_house_collector()) {
+            Log.w(TAG, "Not sending sendLocation since this is libre allhouse collector.");
+            return;
+        }
         if (JoH.pratelimit("gcm-plu", 180)) {
             GcmActivity.sendMessage("plu", location);
         }
     }
 
     public static void sendSensorBattery(final int battery) {
+        if(Home.get_is_libre_whole_house_collector()) {
+            Log.w(TAG, "Not sending sendSensorBattery since this is libre allhouse collector.");
+            return;
+        }
         if (JoH.pratelimit("gcm-sbu", 3600)) {
             GcmActivity.sendMessage("sbu", Integer.toString(battery));
         }
     }
 
     public static void sendBridgeBattery(final int battery) {
+        if(Home.get_is_libre_whole_house_collector()) {
+            Log.w(TAG, "Not sending sendBridgeBattery since this is libre allhouse collector.");
+            return;
+        }
         if (battery != last_bridge_battery) {
             if (JoH.pratelimit("gcm-bbu", 1800)) {
                 GcmActivity.sendMessage("bbu", Integer.toString(battery));
@@ -407,6 +435,10 @@ public class GcmActivity extends FauxActivity {
     }
 
     public static void sendParakeetBattery(final int battery) {
+        if(Home.get_is_libre_whole_house_collector()) {
+            Log.w(TAG, "Not sending sendParakeetBattery since this is libre allhouse collector.");
+            return;
+        }
         if (battery != last_parakeet_battery) {
             if (JoH.pratelimit("gcm-pbu", 1800)) {
                 GcmActivity.sendMessage("pbu", Integer.toString(battery));
@@ -416,6 +448,11 @@ public class GcmActivity extends FauxActivity {
     }
 
     public static void sendNotification(String title, String message) {
+        if(Home.get_is_libre_whole_house_collector()) {
+            Log.w(TAG, "Not sending sendNotification since this is libre allhouse collector.");
+            return;
+        }
+
         if (JoH.pratelimit("gcm-not", 30)) {
             GcmActivity.sendMessage("not", title.replaceAll("\\^", "") + "^" + message.replaceAll("\\^", ""));
         }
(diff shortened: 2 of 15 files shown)


jamorham:
In principle I think this could be good but the implementation needs to be carefully handled. 

I haven't fully examined the PR but the most obvious concern is that it isn't controlled by a preference switch. So for example in a non-whole-house situation with master/follower you are still relaying the LibreBlock data which will create an overhead. In order to avoid changing the existing default behavior I think this needs to be controlled by a preference switch that defaults to off.

app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
```diff
@@ -688,6 +688,13 @@ static void pushCalibration2(double bgValue, String uuid, long offset) {
             GcmActivity.sendMessage(myIdentity(), "cal2", json);
         }
     }
+    public static void pushLibreBlock(String libreBlock) {
```
jamorham, line 691: This should also have a `pratelimit` call to ensure that network traffic cannot increase much more than would be expected. So the rate limit should prevent the same or duplicate data being sent multiple times. I assume that this push should not occur more than once every approx 5 minutes?

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -3039,7 +3039,7 @@ public boolean onCreateOptionsMenu(Menu menu) {
         menu.findItem(R.id.showreminders).setVisible(Pref.getBoolean("plus_show_reminders", true) && !is_newbie);
 
         LibreBlock libreBlock = null;
-        if (DexCollectionType.hasLibre()) {
+        if (DexCollectionType.hasLibre() || get_follower()) {
```
jamorham, line 3042: I think this additional use of get_follower needs to also be conditional on a preference switch to avoid this being activated for all followers. eg non-libre etc

app/src/main/java/com/eveningoutpost/dexdrip/Services/LibreWifiReader.java
```diff
@@ -57,6 +58,9 @@
     private static final Gson gson = JoH.defaultGsonInstance();
 
     private final static long DEXCOM_PERIOD = 300000;
+    
+    // An object that is used to sync the readData
+    private static final Object readDataLock = new Object();
```
jamorham, line 63: Can you just do: `synchronized (LibreWifiReader.class)` to avoid creating an additional object?

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
tzachi-dar: Is it OK if I'll have this checked once per hour (with a rate limiter) or do you still prefer that I'll add a switch?

tzachi-dar:
What about the extra preference to control the whole house coverage?
Should I add a preference that will only allow sending libreblocks?

app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
tzachi-dar: Yes, moved this to a more visible place.

app/src/main/java/com/eveningoutpost/dexdrip/Services/LibreWifiReader.java
tzachi-dar: Done.

tzachi-dar:
I have added my changes from the code review:

1) Send the sensor age to the follower.
2) Use sync on classes and not on separate objects.
3) Move the rate limiter on new libreblock objects to GcmActivity.java
4) Only notify for bad sensor if the sensor has not advanced, and there are no new readings.
5) Cache the results of the function hasLibreblock().
6) Fix a bug in the function checkLibreSensorChange.
7) Add a preference key to control if this host will be a whole house collector.

Still doing final tests, but so far things look well.

app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
```diff
@@ -279,6 +279,10 @@ public synchronized static void syncBGReading(BgReading bgReading) {
             UserError.Log.wtf(TAG, "Cannot sync null bgreading - should never occur");
             return;
         }
+        if(Home.get_is_libre_whole_house_collector()) {
```
jamorham, line 282: Why is this conditional here? What is it for?

app/src/main/java/com/eveningoutpost/dexdrip/Models/SensorSanity.java
```diff
@@ -101,30 +102,109 @@ public static boolean checkLibreSensorChangeIfEnabled(final String sn) {
...
+                // This is the first time we see this sensor, update our uuid.
+                Log.i(TAG, "This is the first time we see this sensor uuid = " +  this_sensor.uuid);
+                PersistentStore.setString(PREF_LIBRE_SENSOR_UUID, this_sensor.uuid);
+                return false;
+            }
+        }
+    }
+
+    // This is a test for the CheckLibreSensorChange function.
+    public static void testCheckLibreSensorChange() {
```
jamorham, line 159: Should this be in its own unit test class?

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -218,10 +219,32 @@ public static boolean HandleGoodReading(String tagId, byte[] data1, final long C
...
+        }
+        // Create the object to send
+        libreBlock = LibreBlock.create(tagId, CaptureDateTime, data1, 0, patchUid, patchInfo);
+        if(libreBlock == null) {
+            Log.e(TAG, "Error could not create libreBlock for libre-allhouse");
+            return;
+        }
+        final String json = libreBlock.toExtendedJson();
+        
+        GcmActivity.pushLibreBlock(json);
```
jamorham, line 240: This should be wrapped either within the function or the caller with both a preference switch that it only becomes active when libre all house is in use and a rate limit. It looks like you have neither from reading the PR?

app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
```diff
@@ -688,8 +790,25 @@ static void pushCalibration2(double bgValue, String uuid, long offset) {
             GcmActivity.sendMessage(myIdentity(), "cal2", json);
         }
     }
+    public static void pushLibreBlock(String libreBlock) {
+        Log.i(TAG, "libreBlock called: " + libreBlock);
+        if (!Home.get_master()) {
+            return;
+        }
+        if (!JoH.ratelimit("libre-allhouse", 5)) {
```
jamorham, line 798: Use `pratelimit` for critical rate limiting as `ratelimit` isn't safe on all platforms.

app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
```diff
@@ -552,7 +557,8 @@ public void run() {
                             UserError.Log.wtf(TAG, "Exception processing rsom timestamp");
                         }
                     }
-
+                } else if (action.equals("libreBlock")) {
+                    HandleLibreBlock(payload);
```
jamorham, line 561: Doesn't this call `NFCReaderX.HandleGoodReading` which then in turn calls `GcmActivity.pushLibreBlock` which causes the received data to be duplicated? Its very important that inbound message doesn't generate the same outbound message.

tzachi-dar: Messages should be unique by their timestamp.
In the function HandleLibreBlock we do the following:
1) We check if we have already seen this message (based on it's timestamp). If we did, we ignore it.
2) We save it in our DB.

3) On the function SendLibrereading() we check for a message with this timestamp again. if it is stored in our DB, we will not send it.
So, a message received from the network should not be sent again.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
tzachi-dar: Libreblocks have the data to needed to show libretrend, see sensor age, and device battery.
This information should pass to normal followers as well. So I did not limit this on a preference swithc.

In the function pushLibreBlock() we have the rate limiter.

app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
tzachi-dar: In order for an xDrip to use BT to get data it needs to have hardware_data_type == librebt
In order for him to send data it needs to be marked as master.

The question here is what will happen if the user will start a new sensor (and add calibrations).
If he will then the network will have two masters, each sending BG data (as well as other messages).

In order to only have one master on the network, I have limited this master to send messages of LibreBlock, alarm snoozes and 'desert mode' messages. This is why we need the get_is_libre_whole_house_collector condition.

An alternative to this is to write in documentation that the users will not start a new sensor, but then users will find it hard to tell if this xDrip is realy communicating with the miaomiao.

app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
tzachi-dar: Done.

app/src/main/java/com/eveningoutpost/dexdrip/Models/SensorSanity.java
tzachi-dar: Yes it should.
I was not aware of this unit test, but moved it now.

tzachi-dar:
I have addressed all comments.
Please take a look again.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Bubble.java
```diff
@@ -82,6 +82,11 @@ public static BridgeResponse decodeBubblePacket(byte[] buffer, int len) {
             patchUid = Arrays.copyOfRange(buffer, 2, 10);
             String SensorSn = LibreUtils.decodeSerialNumberKey(patchUid);
             PersistentStore.setString("LibreSN", SensorSn);
+            
+            if (SensorSanity.checkLibreSensorChangeIfEnabled(SensorSn)) {
+                Log.e(TAG, "Problem with Libre Serial Number - not processing");
+            }
```
jamorham, line 88: Should this do more than just log the problem? Was more intended?

app/build.gradle
```diff
@@ -331,4 +331,12 @@ dependencies {
     annotationProcessor "org.projectlombok:lombok:1.18.10"
 }
 
+tasks.withType(Test) { 
+  testLogging {
+    exceptionFormat "full"
+    events "started", "skipped", "passed", "failed"
+    showStandardStreams true
+  }
+}
```
jamorham, line 340: Can you explain why this change is needed?

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -218,10 +219,32 @@ public static boolean HandleGoodReading(String tagId, byte[] data1, final long C
         return HandleGoodReading(tagId, data1, CaptureDateTime, false, null, null);
     }
 
+
+    public static void SendLibrereading(final String tagId, byte[] data1, final long CaptureDateTime, byte []patchUid,  byte []patchInfo){
+        if(!Home.get_master()) {
+            return;
+        }
```
jamorham, line 226: Can we add a check for the librewholehouse preference switch here and return if it isn't set?

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -282,6 +309,10 @@ protected void onPostExecute(Tag tag) {
                 if (succeeded) {
                     long now = JoH.tsl();
                     String SensorSn = LibreUtils.decodeSerialNumberKey(tag.getId());
+                    
+                    if (SensorSanity.checkLibreSensorChangeIfEnabled(SensorSn)) {
+                        Log.e(TAG, "Problem with Libre Serial Number - not processing");
```
jamorham, line 314: This logs that it isn't processing but does processing continue? should it?

app/build.gradle
tzachi-dar: Without this change, when one is running the unit tests, he gets a summery at the end of what has ahppened.
With this change, one can also see the test results as they are running.

It will now look like (also with green color for passing tests.):

com.eveningoutpost.dexdrip.watch.thinjam.utils.BitmapToolsTest > unpackMonoBytesToRGB565Test STARTED

com.eveningoutpost.dexdrip.watch.thinjam.utils.BitmapToolsTest > unpackMonoBytesToRGB565Test PASSED

com.eveningoutpost.dexdrip.watch.thinjam.utils.BitmapToolsTest > packRGB565bytesToMonoTest STARTED

com.eveningoutpost.dexdrip.watch.thinjam.utils.BitmapToolsTest > packRGB565bytesToMonoTest PASSED

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
tzachi-dar: This information is needed for all followers (not related to all house coverage).
It allows them to have per minute information (libre trend), sensor life and battery life.
Even Followers who will not use all house need this information.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
tzachi-dar: Sensor was stopped, so even if processing continues,  BG will not be created.
I can add a return in all this casses.

app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
```diff
@@ -688,8 +790,25 @@ static void pushCalibration2(double bgValue, String uuid, long offset) {
             GcmActivity.sendMessage(myIdentity(), "cal2", json);
         }
     }
+    public static void pushLibreBlock(String libreBlock) {
+        Log.i(TAG, "libreBlock called: " + libreBlock);
+        if (!Home.get_master()) {
+            return;
+        }
+        if (!JoH.pratelimit("libre-allhouse", 5)) {
```
jamorham, line 798: What is the expected frequency of these packets do you think?

tzachi-dar: Well, if the user is continuously connected to the sensor, then probably every 5 minutes.
I guess that if he is near the device and then goes away, and then reconnects it reasonably should be every minute.
The goal was to block storms, so I figured out that 5 seconds should be enough.

tzachi-dar:
This was replaced by:
https://github.com/NightscoutFoundation/xDrip/pull/1385

PR #1345, opened 2020-05-19
Added Levemir to insulin types.

vguttmann:
For now I added Levemir. The numbers in there are based on the following websites:
https://www.ema.europa.eu/en/documents/product-information/levemir-epar-product-information_en.pdf (page 10)

www.diabetes-news.de/uebersicht-insulin-praeparate (It's in german, soory for those of you who have to use google translate)

https://www.diabetesde.org/ueber_diabetes/therapie_bei_diabetes/alles_ueber_insulin/verschiedene_insuline (Also in German, sorry for that).

Don't be scared that I may have mixed up some numbers there: I am a native german speaker, and the EMA document I've read in german. In german it's available at https://www.ema.europa.eu/en/documents/product-information/levemir-epar-product-information_de.pdf

### DIFF
--- a/app/src/main/res/raw/insulin_profiles.txt
+++ b/app/src/main/res/raw/insulin_profiles.txt
@@ -40,6 +40,25 @@
       }
     },
     {
+          "name": "Levemir",
+          "displayName": "Levemir (long acting)",
+          "concentration": "U100",
+          "PPN":
+          [
+            "LevemirU100"
+          ],
+          "Curve":
+          {
+            "type": "linear trapezoid",
+            "data":
+            {
+              "onset": "60",
+              "peak": "480",
+              "duration": "1200"
+            }
+          }
+        },
+    {
       "name": "Humalog",
       "displayName": "Humalog (rapid acting)",
       "concentration": "U100",


mx11de:
Please add also Tresiba 100U and 200U
https://www.ema.europa.eu/en/medicines/human/EPAR/tresiba

vguttmann:
I really would like to, but according to NovoNordisk, Tresiba has a different acting profile than the other insulins, meaning, that it would require more modification of the app than I am currently able to do. Also I will post a link to Google Drive with the document sent by NovoNordisk where the acting times and curve for NovoRapid are listed.

tolot27:
@vguttmann Thx for your contribution. PR #1388 also adds Levemir. Does it fit your needs or are there an advantages of your implementation?

vguttmann:
Well, if Levemir is added, then that's all I wanted. I don't think that my (admittedly pretty basic implementation) offers any benefits, as I only changed the insulin type file to add Levemir. 

By the way, what is the official way of communicating for the development? I know there is a subreddit, and a facebook group, but what is the preferred way you guys talk about the project? Is there a mailing list, or something else I overlooked?

tolot27:
> By the way, what is the official way of communicating for the development? I know there is a subreddit, and a facebook group, but what is the preferred way you guys talk about the project? Is there a mailing list, or something else I overlooked?

Good question. Currently, we only have GitHub and Gitter (see f2e98daa). We started to restructure things because development and support fusioned, somehow.

vguttmann:
I can tell you why they fused: Most users are at least somewhat tech-savvy, since it still isn't a one-click solution, thanks to uncooperative companies [insert long rant here]. And this means that they are often also GitHub users, and one thing lead to another. By the way, I'll check Gitter out.

Thanks for the info!

PR #1357, opened 2020-06-03
fix bug 

bubbledevteam:
fix bug buffer length


bubbledevteam:
> I assume the buffer length can never be less than 4

Yep

Bubble FW 1.x transmit 22 pockets with 20 bytes each pocket(4 bytes header and 16 bytes Libre sensor block contents),and the pocket 22th with 8 bytes sensor contents and append 8 bytes zero to fill it up.

In order to make Bubble transmit data more efficiently, Bubble FW 2.0 changed the data contents in the pocket 22th with only 8 bytes sensor contents without zero append.

That's why we fix it here.

psonnera:
@bubbledevteam  You forgot wear (I think) https://github.com/NightscoutFoundation/xDrip/blob/master/wear/src/main/java/com/eveningoutpost/dexdrip/Models/Bubble.java

PR #1385, opened 2020-07-16
Tzachi libre multi receivers 3

tzachi-dar:
This replaces https://github.com/NightscoutFoundation/xDrip/pull/1325

1) It is resynchronized against latest master.
2) All changes to gcm activity have been removed.
3) Sensor will automatically be stopped when running in all house mode.

Still under testing.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
@@ -26,6 +26,7 @@ import com.eveningoutpost.dexdrip.Models.JoH;
 import com.eveningoutpost.dexdrip.Models.LibreBlock;
 import com.eveningoutpost.dexdrip.Models.RollCall;
 import com.eveningoutpost.dexdrip.Models.Sensor;
+import com.eveningoutpost.dexdrip.Models.SensorSanity;
 import com.eveningoutpost.dexdrip.Models.TransmitterData;
 import com.eveningoutpost.dexdrip.Models.Treatments;
 import com.eveningoutpost.dexdrip.Models.UserError;
@@ -584,8 +585,13 @@ public class GcmListenerSvc extends JamListenerSvc {
         Pref.setInt("bridge_battery", lb.bridge_battery);
         PersistentStore.setString("Tomatobattery", Integer.toString(lb.Tomatobattery));
         PersistentStore.setString("Bubblebattery", Integer.toString(lb.Bubblebattery));
+        PersistentStore.setString("LibreSN", lb.reference);
         
         if(Home.get_master()) {
+            if (SensorSanity.checkLibreSensorChangeIfEnabled(lb.reference)) {
+                Log.e(TAG, "Problem with Libre Serial Number - not processing");
+            }
+            
             NFCReaderX.HandleGoodReading(lb.reference, lb.blockbytes, lb.timestamp, false, lb.patchUid,  lb.patchInfo);
         }
     }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
@@ -1,5 +1,7 @@
 package com.eveningoutpost.dexdrip;
 
+import android.R.integer;
+
 /**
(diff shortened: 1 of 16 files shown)


tzachi-dar:
I have tested it on 3 phones and it works well.
Can be merged.

app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
```diff
@@ -565,6 +571,27 @@ public void run() {
...
+            // We already seen this one.
+            return;
+        }
+        LibreBlock.Save(lb);
+        
+        PersistentStore.setString("LibreSN", lb.reference);
+        
+        if(Home.get_master()) {
+            if (SensorSanity.checkLibreSensorChangeIfEnabled(lb.reference)) {
+                Log.e(TAG, "Problem with Libre Serial Number - not processing");
```
jamorham, line 589: Is this missing a return as it reports that it isn't processing?

app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
```diff
@@ -210,9 +210,15 @@ public static void CalculateFromDataTransferObject(ReadingData.TransferObject ob
...
                 Log.d(TAG, "Sensor age advanced to: " + thisSensorAge);
             } else if (thisSensorAge == sensorAge) {
-                Log.wtf(TAG, "Sensor age has not advanced: " + sensorAge);
-                JoH.static_toast_long(gs(R.string.sensor_clock_has_not_advanced));
-                Pref.setBoolean("nfc_age_problem", true);
+                // This is only a problem if we don't have a recent reading. It could happen that we have a recent
+                // reading, and then BT disconnects and connects after a few seconds, and we have a new reading (where 
+                // sensor did not advance). This does not mean that a sensor is not advancing. Only if this is happening
+                // for a few minutes, this is a problem.
+                if(BgReading.getTimeSinceLastReading() > 11 * 60 * 1000) {
```
jamorham, line 217: This looks like it should work, by the way you can use the `msSince()` helper and also `Constants.MINUTE_IN_MS` to improve readability.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Bubble.java
```diff
@@ -82,6 +82,11 @@ public static BridgeResponse decodeBubblePacket(byte[] buffer, int len) {
             patchUid = Arrays.copyOfRange(buffer, 2, 10);
             String SensorSn = LibreUtils.decodeSerialNumberKey(patchUid);
             PersistentStore.setString("LibreSN", SensorSn);
+            
+            if (SensorSanity.checkLibreSensorChangeIfEnabled(SensorSn)) {
+                Log.e(TAG, "Problem with Libre Serial Number - not processing");
+            }
```
jamorham, line 88: This logs that it isn't processing but still continues. Should the message be changed?

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -218,10 +219,32 @@ public static boolean HandleGoodReading(String tagId, byte[] data1, final long C
...
+        GcmActivity.pushLibreBlock(json);
+    
+    }
+    
     // returns true if checksum passed.
     public static boolean HandleGoodReading(final String tagId, byte[] data1, final long CaptureDateTime, final boolean allowUpload, byte []patchUid,  byte []patchInfo ) {
-
-
+        Log.e(TAG, "HandleGoodReading called");
+        SendLibrereading(tagId, data1, CaptureDateTime, patchUid, patchInfo);
```
jamorham, line 247: I am concerned that this change not being restricted to whole house coverage is an unrelated change to functionality which has a much wider scope and could increase the data load for followers somewhat but I understand why its desired so I think we will see how it goes.

I'm also concerned that this is the only code path where an incoming message can generate an outgoing message of the same time (as mentioned with the previous PR) because that has the potential for a loop. I can see that it is restricted with a database check but when phones get out of storage they can fail to write database records. I think for a loop to occur that would have to happen to multiple devices in the follower group which is less likely. 

It could be mitigated with an in memory check when sending the message that the message isn't the same one that was sent last time.

Issue #985, opened 2019-08-20
Libre 2 NFC scan 


adenis76:
Hi,

Currently xdrip can't scan libre 2 with Nfc and also not retrieve history from patched librelink app. Please add Nfc scan for libre 2 and possibility to retrieve missed reading.

Thank you.

MasterPlexus:
@adenis76, unfortunately there is at the moment no way to add this functionality to xdrip. In Libre 2 everything is encrypted, and the keys are not known. If nobody can get the keys, a direct scan will not work.

nsghub:
There is a new Blucon Nightrider for the Libre 2. So somebody already found the keys. Maybe someone could talk to the guys from Ambrosia.

chaosbiber:
Ambrosia wasn't the first. There was a group that found a first solution around April(?) this year and offered Ambrosia, MiaoMiao and alike the libraries if they make a donation to the Nightscout foundation. Don't know if any company uses that lib or if they did something on their own. But the problem is: those readings aren't as accurate as the data that comes from the patched libre app. They'll differ quite a bit and are much more noisy. So the question is: is it worth the effort of implementing that function in xDrip until the protocols of libre2 are fully reverse engineered?

adenis76:
My real "problem" here is gap on xdrip and nightscout graph when phone lose connection with sensor. With blucon and libre 1 sensor xdrip retrieve missed libre history. Can we add possibility to xdrip to retrieve missed readings from patched librelink app?

mysticprog:
So, Libre 2 over Bluecon is not supported ? Is MiaoMiao 2 supported i xdrip ?
Thanks.

tzachi-dar:
We are working on this.

mysticprog:
Ok. Thanks

dideldum:
The patched LibreLink sends also an Intent "com.librelink.app.ThirdPartyIntegration.SENSOR_SCAN" with all the scanned values (historic and realtime). I tried it in a small test app and all the scanned values seem to be there.
It's just xDrip+, which doesn't take this intent and fills the gap with the scanned values.
Couldn't xDrip+ implement the reception of this patched LibreLink intent?

mysticprog:
> The patched LibreLink sends also an Intent "com.librelink.app.ThirdPartyIntegration.SENSOR_SCAN" with all the scanned values (historic and realtime). I tried it in a small test app and all the scanned values seem to be there.
[...]

PR #1401, opened 2020-08-11
Libre2 improvements

smuething:
# Intro

This MR improves the current support for Libre2 data based on the patched Libre app.
It contains the following changes:

- Adds Kotlin support to the project (required for the math utilities in the MR)
- Adds an alternative smoothing algorithm based on a Savitzky-Golay filter, optionally mixed with a weighted average
- Adds a noise estimator for Libre2 data which is based on the estimation idea of the Dexcom code, but takes advantage
  of the higher-frequency raw data of the Libre2 sensor.

# Filtering

The Savitzky-Golay filter is based on the Wikipedia description at https://en.wikipedia.org/wiki/Savitzky%E2%80%93Golay_filter. The filtering coefficients are recalculated every time we generate a smoothed value, because the standard Savitzky-Golay coefficients assume a regular spacing of the data points, which we cannot guarantee as points will often be missing. The filter works by fitting the data in a window to a given polynomial. Standard Savitzky-Golay uses a centered window, i.e. it produces a value for the middle of the window. We are more interested in a value to the right of the window (as that is closer to the current time). I have done a bit of data science on my own raw Libre2 values as stored in the app, and for me the optimum parameters are a window length of 25 minutes, a cubic polynomial and a delay of 2 minutes. If someone is interested in testing the parameters on their own data, I can share the Jupyter notebook that directly works on the exported xDrip db.
My experiments have also shown that it might be better to calculate a linear combination of the Savitzky-Golay filter and a weighted average like the one that is currently in use, as that provides the best combination of quick response and smooth curves, at least for my data. As a result, the code as posted mixes 2/3 of Savitzky-Golay with 1/3 of a weighted average with a horizon of 15 minutes.
The Savitzky-Golay algorithm can be enabled in the preferences under advanced Libre2 settings and is disabled by default. All of its parameters can also be adjusted, but are gated by the engineering mode.

# Noise Estimation

People are interested in using Libre2 data with SMB in AndroidAPS (and I might be too, once I get my pumped approved by the insurance and have collected enough experience), but doing so requires noise filtering of the data sent to AAPS. This MR also contains a noise estimator built on the same principles as the one for Dexcom, but it makes use of the raw data in addition to the filtered data. Looking at my own Libre2 data again, the noise estimator tends to react quicker than the Dexcom approach of just operating on the filtered data, but it is of course different. Again, I can provide a Jupyter notebook so that people can verify the algorithm on their own data.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
@@ -19,10 +19,12 @@ import com.eveningoutpost.dexdrip.UtilityModels.Pref;
 import com.eveningoutpost.dexdrip.UtilityModels.StatusItem;
 import com.eveningoutpost.dexdrip.utils.DexCollectionType;
 import com.eveningoutpost.dexdrip.utils.math.AdaptiveSavitzkyGolay;
+import com.eveningoutpost.dexdrip.utils.math.PolynomialFitErrorEstimator;
 
 import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 import java.util.concurrent.TimeUnit;
 
 import static com.eveningoutpost.dexdrip.Home.get_engineering_mode;
@@ -34,13 +36,26 @@ import static com.eveningoutpost.dexdrip.Models.Libre2Sensor.Libre2Sensors;
 
 public class LibreReceiver extends BroadcastReceiver {
 
-    private static final String TAG = "xdrip libre_receiver";
+    private static final String TAG = "xdrip_libre_receiver";
     private static final boolean debug = false;
     private static final boolean d = false;
     private static SharedPreferences prefs;
     private static final Object lock = new Object();
     private static String libre_calc_doku="wait for next reading...";
     private static long last_reading=0;
+
+    // default parameters for adaptive Savitzky-Golay smoother
+    // These defaults will normally be overridden by the values from shared preferences
+    private final static int ASG_HORIZON = 25;
+    private final static int ASG_LAG = 2;
+    private final static int ASG_POLYNOMIAL_ORDER = 3;
+    private final static double ASG_WEIGHTED_AVERAGE_FRACTION = 0.333;
+    private final static int ASG_WEIGHTED_AVERAGE_HORIZON = 15;
+
+    // default parameters for noise estimator
+    // These defaults will normally be overridden by the values from shared preferences
+    private final static int NOISE_HORIZON = 19;
+    private final static int NOISE_POLYNOMIAL_ORDER = 2;
 
     @Override
     public void onReceive(final Context context, final Intent intent) {
@@ -135,21 +150,19 @@ public class LibreReceiver extends BroadcastReceiver {
 
         }
 
-        if (false) {
-            List<Libre2RawValue> smoothingValues = Libre2RawValue.last20Minutes();
 
-            double value = calculateWeightedAverage(smoothingValues, currentValue.timestamp);
+        if (Pref.getBooleanDefaultFalse("Libre2_useSavitzkyGolay")) {
 
-            BgReading.bgReadingInsertLibre2(value, currentValue.timestamp, currentValue.glucose);
+            int horizon = Pref.getStringToInt("Libre2_sgHorizon",ASG_HORIZON);
+            int lag = Pref.getStringToInt("Libre2_sgLag",ASG_LAG);
+            int polynomialOrder = Pref.getStringToInt("Libre2_sgPolynomialOrder",ASG_POLYNOMIAL_ORDER);
+            double curvaturePenalty = Pref.getStringToDouble("Libre2_sgDiffJumpPenalty",ASG_CURVATURE_PENALTY);
+            double weightedAverageFraction = Pref.getStringToDouble("Libre2_sgWeightedAverageFraction",ASG_WEIGHTED_AVERAGE_FRACTION);
+            int weightedAverageHorizon = Pref.getStringToInt("Libre2_sgWeightedAverageHorizon",ASG_WEIGHTED_AVERAGE_HORIZON);
 
-        } else {
-
-            int horizon = 45;
-            int lag = 3;
-            int polynomialOrder = 3;
             List<Libre2RawValue> smoothingValues = Libre2RawValue.lastMinutes(horizon);
 
-            AdaptiveSavitzkyGolay asg = new AdaptiveSavitzkyGolay(3,3);
+            AdaptiveSavitzkyGolay asg = new AdaptiveSavitzkyGolay(lag,polynomialOrder, weightedAverageFraction,weightedAverageHorizon);
             for (Libre2RawValue rawValue : smoothingValues) {
                 if (!rawValue.serial.equals(Sensor.currentSensor().uuid)) {
                     Log.v(TAG,"Skipping raw measurement from old sensor at t=" + rawValue.timestamp);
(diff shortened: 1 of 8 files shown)


k2s:
hello, thank you for your work. Could you send me Jupyters to test what I could expect if I compile this patch to my build ? Please leave message here or use email 5qe8beuei@relay.firefox.com.

smuething:
Here is the Jupyter notebook that lets you test both features and also tweak the parameters. The default parameters match what the defaults in Xdrip+.


[libre-improvements.ipynb.zip](https://github.com/NightscoutFoundation/xDrip/files/5073629/libre-improvements.ipynb.zip)

k2s:
Wonderful usage for Jupyter notebook. 

I found out that I need to use export from xDrip that is on mobile with patched libre (my sons). On follower (my phone) the table `Libre2RawValue2` is empty. Will this improvement work for followers too ?

k2s:
small comments to Jupyter notebook:
* `start_ts = datetime.datetime.fromisoformat(start_date).timestamp()` should be changed to `start_ts = datetime.datetime.fromisoformat(start_date).timestamp() * 1000`, otherwise it processes always all data
*  `2020-07-20` occurrences could be replaced by `start_date`

k2s:
added blood test values:

```python
# sql part
tests = pd.read_sql(f'select timestamp, mgdl from BloodTest where timestamp >= {start_ts} order by timestamp',conn,parse_dates={'timestamp':'ms'},index_col='timestamp')

# in plot_data
shorttests = tests.loc[start:end]
fig.add_trace(go.Scatter(x=shorttests.index,y=shorttests.mgdl,name="tests",marker=dict(color="crimson", size=12),mode="markers"))
```

smuething:
> Wonderful usage for Jupyter notebook.
> 
> I found out that I need to use export from xDrip that is on mobile with patched libre (my sons). On follower (my phone) the table `Libre2RawValue2` is empty. Will this improvement work for followers too ?

I'm not really all that well-versed in the source code, but I assume that Xdrip sends the smoothed value to the follower phone, so your phone would see the same values as what is shown own your son's. And I have no idea about the noise value - it gets sent to Nightscout, but I don't know about a follower phone.

smuething:
> added blood test values:

I didn't include those, mostly because filtering and noise estimation happens before calibration (which for LIbre2, applies a fixed offset of -40 to +20 I think, the slope is fixed at 1).

Artemdrip:
Will your code solve the problem of Xdrp + leaving a small gap
(due to the lag by getting raw values) on the graph every 3-5 hours?

That bothers me a lot.

Screenshot_2020-08-15-17-45-27-606_com eveningoutpost dexdrip
Verlauf nach Steiner Toast
Screenshot_2020-08-15-17-56-35-811_com eveningoutpost dexdrip

smuething:
Interesting, that might happen for me as well, but I never really noticed a single missing value. The curve would look less smooth, e.g. around 14:00 and 17:00 in the top picture. In that case, the filtered values would react more quickly to the steep gradient in the raw values.
If your are looping, the noise estimation would also block some values from being sent to AndroidAPS if the raw data becomes too "wobbly", but your curves are way too smooth to trigger that.

Artemdrip:
I'm not looping. I'm only using xdrip+ with Libre2 patched app.
I think everybody has this issue by using xdrip+ with Libre2 patched app .  

Xrip + collects a raw value every minute and if 5 raw values ​​have been collected within the 5 minutes Xdrip + sets the average value on the graph. Xdrip + sets the average value on the graph exactly every 5 minutes.
The small delays of 3 to 6 seconds per 5 minutes while collecting raw values ​​are summarized and they cause a delay of more than 3 minutes after 3-5 hours.
Xdrip + now sets the value on the graph after 8 minutes so that they run synchronously again and the delay grows all over again. This is the hole we see in the top pictures. 

**I think this could be solved as follows.**

Xdrip collects the raw values ​​as usual. As soon as the delay is greater than a minute, xdrip only collects 4 raw values ​​for the next value and only needs 4 minutes. 
So the delay disappears again.

**Is It possible to solve this problem?
  can you or somebody else programm this?**

smuething:
@Artemdrip what version of Xdrip are you on?

Artemdrip:
> @Artemdrip what version of Xdrip are you on?

Nightly build 25th Jul 2020

Artemdrip:
In this picture you can see the lag better. 

Screenshot_2020-08-16-09-01-54-014_com eveningoutpost dexdrip (1)

smuething:
> > @Artemdrip what version of Xdrip are you on?
> 
> Nightly build 25th Jul 2020

Hmm, that's rather current. My code does not touch any timing issues (the code that receives the raw Libre2 values from the app creates a new BG reading if the last reading was more than 4.5 min ago), so I don't think it will help with your problem. Did you open an issue?

smuething:
There's nothing we can do about the lag, the values are processed as soon as they arrive via broadcast from the patched Libre app, and they always get the timestamp of the most recent raw value (the one that triggered the creation of the bg reading entry)

smuething:
Apparently, the sensor does not stick to an exact schedule of one value per minute, but is a little slower than that.

Artemdrip:
The delay can't be avoided, but it is possible to compensate the delay. xdrip + must read only 4 raw values if the delay become  one minute. Sometimes the delay is big, sometimes small but everybody has this issue.

Artemdrip:
> Apparently, the sensor does not stick to an exact schedule of one value per minute, but is a little slower than that.

It is possible to compensate the delay in the code?

smuething:
Maybe we should move this to a separate issue, as this PR is really about the changes to the filtering and the noise estimation.

smuething:
Could you open one, as you can explain the issue better than me I think.

k2s:
I was using now this improved filtering. Without changing defaults provided with this pull request. My experience is that the curve is much more steep which makes me sometimes worry more then I should be.

I appreciate the work in this pull request and specially the simulation tool in Jupyter notebook, but I think more of value would be to let xDrip accept values every minute and read missing values from history, which [seams to be possible](https://github.com/NightscoutFoundation/xDrip/issues/985#issuecomment-686513290), but is not implemented.

linus2100:
I'm trying out the filter now for around 3 weeks and like it a lot. There is some steadiness when raw values are volatile, but still a quick response to changes in BG tendencies. For me it represents true BG values better than pure weighted average results. My current set: smoothing horizon 55 - smoothing lag 1 - weighted average fraction 0.1 - weighted average horizon 14
Screenshot_20200920-190815_xDrip+

linus2100:
Also I found noise calculation quite helpful. At times I've got large raw BG clouds, after a shower for example.  Now noise levels rise and I get messages about blocked BG transmission. Yet it's still received by AAPS, which is a shame, cause those values are always too high. But this might be to some of my settings. Noise blocking is new to me as Libre user.

GuidoMa:
I am using the filter for more than a month now (default parameters) and it is so much better than the original filtering! The delay with the standard filtering is really bad, especially if you have a quick metabolism, like I have.
In my opinion this pull request should be merged!

jamorham:
What is the file size footprint impact of adding kotlin to the project?

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -1260,18 +1264,24 @@ public static synchronized BgReading bgReadingInsertLibre2(double calculated_val
                 bgReading.sensor_uuid = sensor.uuid;
                 bgReading.raw_data = raw_data;
                 bgReading.age_adjusted_raw_value = raw_data;
-                bgReading.filtered_data = raw_data;
```
jamorham, line 1263: What happens to the `raw_data` parameter here?

smuething:
> Also I found noise calculation quite helpful. At times I've got large raw BG clouds, after a shower for example. Now noise levels rise and I get messages about blocked BG transmission. Yet it's still received by AAPS, which is a shame, cause those values are always too high. But this might be to some of my settings. Noise blocking is new to me as Libre user.

Sorry, haven't looked at GitHub in some time. What is the noise value reported in Xdrip when data is noisy? Xdrip will send the data to AAPS as long as the noise value is `< 200`. That number is hardcoded in the Dexcom code path, I actually rescaled the Libre2 noise to fit that threshold (but I only used my data for that). The scaling function is in LibreReceiver.java:209. If you know your way around Java, you could try increasing the scaling factors to make Xdrip more sensitive to the noise.

smuething:
> What is the file size footprint impact of adding kotlin to the project?

I haven't built Xdrip without the patches in quite some time, but my release mode APKs are 10944 kB for the main app and 2972 kB for wear.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
smuething: I'm using the `filtered_data` field to store the filtered (smoothed), but uncalibrated Libre2 values. I need those historical uncalibrated values for future noise estimation, and as `filtered_data` is only used by the Dexcom code path as far as I can tell, that field seemed like a good place to store the data.

Hubiloop:
I'm using the Savitzky-Golay filter for 4 days now and am quite happy about the precision I get from my xdrip+ now. When starting with xdrip+ some months ago, I was a bit disappointed of the data compared to the original LL iOS output. This filter seems to be a good solution for it.

I'm actually testing with the following parameters:
S-G smoothing horizon 35 (+10 to original)
S-G smoothing lag 2
S-G polynom. order 3
S-G weighted average fraction 0,333
S-G weighted average horizon 15 (+15 to original)
noise estimation horizon 30
noise estimation polynom. order 16

I'm quite insulin sensitive and want to get the best combination of fast response but less noise for a stable input for looping. In the meantime I have set up an open loop and will do some more testing with real diabetic life. 

Thanks for this amazing work!

philou78:
What is the file size footprint impact of adding kotlin to the project?

Hereafter, an idea about size impact on debug app apk:

image

Hubiloop:
I'd like to give a short update from user perspective. 
In the meantime, I used the savgol filter together with a loop for some weeks which brought me to analyzing the parameters with the prepared jupyter notebook. The outcome is a parameter set which only differs in the weighted average part:
S-G smoothing horizon 25 
S-G smoothing lag 2
S-G polynom. order 3
S-G weighted average fraction 0,4 (default 0,333)
S-G weighted average horizon 20 (default 15)
noise estimation horizon 15
noise estimation polynom. order 16

This set works fine for my loop. The data transmission to AAPS gets cut when the noise level is above 200 in my xdrip+. This happens with not so good sensors probably 2-5 times a day and is no problem for the loop as it then falls back to the constant basal rates of the active profile. 

Nevertheless I would like to suggest to get users some help with finding their correct parameter set. Probably predefine 2-3 parameter sets for typical scenarios and leave the details to the devoper mode?

tzachi-dar:
Can you please share the Jupyter notebook that directly works on the exported xDrip db with me?
Is there any reason to use kotlin in this pr? seems like the algorithms that are used here are available in java as well.

smuething:
> Can you please share the Jupyter notebook that directly works on the exported xDrip db with me?

It's in the first or second reply by me, August 14. I'm not sure how much I changed in terms of the algorithms since then, I'd have to take a look, but that might take a week or two, I'm pretty busy at the moment.

> Is there any reason to use kotlin in this pr? seems like the algorithms that are used here are available in java as well.

Honestly? Readability, as the kotlin implementations are a lot closer to how I think about the algorithms in terms of the underlying maths. Moreover, Java would be quite a bit more verbose for some parts of it, and Kotlin also saves on a bunch of boilerplate in terms of getters/setters.

tolot27:
> Will your code solve the problem of Xdrp + leaving a small gap (due to the lag by getting raw values) on the graph every 3-5 hours?

No! This will likely be solved by #1760.

smuething:
> No! This will likely be solved by #1760.

Actually, it does, but it's not really what the PR is about. I just rolled it in here as I don't think this PR will be merged and people asked me about an updated version of the PR to build themselves.

The relevant code is in cd0dc54b26bd5258399fb3906292fa4fc06a4f0e. It doesn't bother with the intermediate step taken by #1760, though. I've just encapsulated the required conversion into a function call and fixed all places in the code that muck about with `FUZZER` to call the function instead. If you want to get rid of `FUZZER`, you have to shift the zero point on the time axis somewhat closer to the present, otherwise floating point precision starts to bite you (that's also the reason for the current resolution limit of 2.5 min - if you were to run the code in the 1970s, precision would be much better 😉 ). I've also explained it in the commit message.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -656,7 +680,7 @@ public LineChartData previewLineData(LineChartData hint) {
                 lines.add(subLine); // iob line
             }
 
-            predictive_end_time = simple ? end_time : ((end_time * FUZZER) + (60000 * 10) + (1000 * 60 * 60 * predictivehours)) / FUZZER; // used first in ideal/highline
+            predictive_end_time = simple ? end_time : timeStampToGraphPos(graphPosToTimestamp(end_time) + (60000 * 10) + (1000 * 60 * 60 * predictivehours)); // used first in ideal/highline
```
tolot27, line 683: To get a more readable code, we could use constants for the different millisecond spans.

tolot27:
@smuething Please can you resolve the merge conflicts and check your PR again with the current code?
Afterwards, I'll continue with the review.

app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
```diff
@@ -33,14 +36,27 @@
 
 public class LibreReceiver extends BroadcastReceiver {
 
-    private static final String TAG = "xdrip libre_receiver";
+    private static final String TAG = "xdrip_libre_receiver";
```
tolot27, line 39: Please do not rename the logging tags unless discussed with the developers.

smuething:
The PR builds again for me. Unfortunately, I don't really have time to do much work to get it merged right now, but I still wanted to get an updated version out there, as I'm testing Libre 3 right now, and for that I wanted to have my filtering working. Also, there were a few private requests to get the branch working again.

The main problem is with the Kotlin support, which breaks because of an xdrip-specific token replacement step. I just worked around that by manually applying the token replacement to the actual sources and disabling that step (see [139e928]).

jamorham:
The `token-replace` plugin is a source pre-processor, so you just need to move the `apply plugin: 'kotlin-android'` after the `common-gradle` line and it should be fine. (You might need to add `setExtension('kt')` to `token-replace.gradle` if you want those files processed as well) - The plugin is there for a reason - to preserve source compatibility with forks and PRs. Renaming the packages in source rather than dynamically via the plugin makes merging this PR very difficult.

I still don't completely understand why kotlin is needed for this? Unless I'm missing something I don't see any kotlin specific libraries in use? Math functions like `pow` `abs` etc are standard parts of the java math libraries. I didn't notice any unit tests, but if there were some for these new classes then I would expect it would be quite easy for me to refactor the code in to java as well.

smuething:
> The `token-replace` plugin is a source pre-processor, so you just need to move the `apply plugin: 'kotlin-android'` after the `common-gradle` line and it should be fine. (You might need to add `setExtension('kt')` to `token-replace.gradle` if you want those files processed as well) - The plugin is there for a reason - to preserve source compatibility with forks and PRs. Renaming the packages in source rather than dynamically via the plugin makes merging this PR very difficult.

Thanks for the information, I'll try that!

As I said earlier, I'm using Kotlin because that way, it's a lot easier for me (and others with a math background) to understand what's happening in the code - it reads pretty much like a direct copy of the formulas from a math paper. For a mathematician, the algorithms are pretty straightforward, but they are still quite a bit more complicated than a simple moving weighted-average filter, so I think it's important to keep the implementation readable for people who can reason about the math involved. Switching to Java would really reduce the readability of the code IMHO.

smuething:
Moving the call in `gradle` worked, thanks again! I didn't add Kotlin files to the preprocessor, I guess any Kotlin files should be new enough to stick to the new capitalization rules... :wink:

jamorham:
Can you add unit tests for the new classes and where possible/appropriate provide meaningful variable names within them and also apply standard code reformatting to the new classes. thanks

Issue #1231, opened 2020-02-06
Nightscout treatments are not downloaded


malloy139:
Treatments entered in XDrip are shown on the master and in the Webinterface, but not on the Noghtscout Follower. "Download Treatments" is turned on.
But when a treatment is entered in the Webinterface is shows up on the follower, as can be seen in the attached image.
![nightscout](https://user-images.githubusercontent.com/41596803/73953484-c1e41780-4900-11ea-94ea-27947cb99670.png)
![xdrip follower](https://user-images.githubusercontent.com/41596803/73953488-c27cae00-4900-11ea-9b1a-f78256f35363.png)

misiek2lucky:
I pin up. On stable and nightly the same.

GPLExpert:
Same for me here

> Treatments entered in XDrip are shown on the master and in the Webinterface, but not on the Noghtscout Follower

malloy139:
It is still not working.
This seems to be the only way (if it would work) to distribute treatments for people with phones that have no play services installed and can't use the sync featue.

GPLExpert:
Hello,
Yes if someone can help debugging ?
I can give all informations if needed
Thanks

rocketman10404:
Having the same issue on several phones running XDrip

rocketman10404:
From my brief and uneducated quick scan of Treatments.java, it does indeed appear as though the "Entered By" record is taken from a hard-coded constant (XDRIP_TAG) that currently cannot be changed in the app.

Also, looking at NightscoutTreatments.java (and this is my best interpretation as a hobbyist's first look) - at line 133, it looks like a check on new treatments to only process non-existing (i.e. new) treatments not coming from xdrip (i.e. "Entered By" is not "xdrip").  There's a else condition block to catch changes for pre-existing treatments whether or not they come from xdrip, but there does NOT look to be any code to process new treatments that come from xdrip.  Seems to me like this is the root cause of the issue we're describing.

I'd suggest possibly adding a configurable setting to allow user control over the XDRIP_TAG contents, but have no idea what other negative effects this may have elsewhere.  It may be easier to avoid downloading/uploading duplicates if all XDrip users syncing to a Nightscout instance could have treatments uniquely identified via the XDRIP_TAG.

Alternatively, perhaps non-existing treatments could be process whether or not they came from xdrip?

Lastly - this may be a duplicate of #1196.  Possibly related to #735, but maybe not.

tolot27:
[...]

Issue #735, opened 2019-01-07
Treatments entered by followers are not uploaded to Nightscout


anderser:
Treatments (carbs, insuling) added on follower phone shows up in the graph in the master phone/xdrip and on other followers, but not in Nightscout as they are added. 

Xdrip is set to use Nightscout REST API to upload. Values from sensor and calibrations are successfully synced to Nightscout, but not treatments.

BUT if I use Back-fill data, all the old treatments get synced. So it seems something is wrong with syncing new treatments as they are added.

The Xdrip master is a Samsung Galaxy S4, Android 4.3, Xdrip latest stable version. Nightscout is latest from master. 

I can see no related log messages on the xdrip. Is there some way I can trace these requests to Nightscout in order to figure out where it stops?

anderser:
@jweismann but wouldn't that suggestion mean that all BG-values also are uploaded twice (or more) (from each phone). I suppose that could mean trouble/duplication?

tolot27:
@jamorham Please assign this issue to me and rename it to "Treatments entered by followers are not uploaded to Nightscout".

Anoukmdv:
Is this resolved or worked around? Thanks

rossi32s:
My T1 Son is on iPhone with the dex app.  I have nightscout server setup and has been pulling the data from dexshare fine for months.  My wife and I both have Xdrip on our androids with nightscout as the datasource..  When she enters a treatment on her phone, it shows up on her phone/xdrip and the nightscout site.  If i do a treatment on my phone it shows up on my phone/xdrip and the nightscout site.  I cannot get the treatments she enters to my xdrip and she cannot see the ones I enter.  

If I load the nightscout page and add a treatment,  it shows up correctly/as expected on my phone and hers.

REST-API is setup on both of our phones.

Navid200:
In order for the follower to upload treatments to Nightscout, you will need to enable `Settings` &#8722;> `Cloud Upload` &#8722;>  `Nightscout Sync (REST_API)` &#8722;> `Enabled`.

Then, on the same page, enter Base URL including either an API_SECRET or a token with write privilage.

Then, also enable `Settings` &#8722;> `Cloud Upload` &#8722;>  `Nightscout Sync (REST_API)` &#8722;> `Extra Options` &#8722;> `Upload treatments`.  
  
I'm not sure if the downloaded bg readings will be uploaded again.  I am curious if that happens.  
[...]

PR #1403, opened 2020-08-14
WIP: follow treatments stored in Nightscout

tolot27:
Addresses #1231 and #735 partially.

ToDos:

- [ ] interface hint should not be shown on data import
- [ ] avoid reuploading
- [ ] download treatments only since last received treatment
- [ ] apply removals (from NightScout or other sources) within certain amount of time (i. e. 30 minutes)

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutTreatments.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutTreatments.java
@@ -28,7 +28,7 @@ public class NightscoutTreatments {
     private static final HashSet<String> bad_uuids = new HashSet<>();
     private static final HashSet<String> bad_bloodtest_uuids = new HashSet<>();
 
-    public static boolean processTreatmentResponse(final String response) throws Exception {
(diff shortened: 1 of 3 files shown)


tolot27:
How to avoid reuploading it to the same NS instance?

dubidrubi:
Hi @tolot27,

Will this change fix https://github.com/NightscoutFoundation/xDrip/issues/1231 ?

Running the latest xDrip version still doesn't allow to sync treatments between followers using NS. Any treatment added from xDrip appears in the NS with xdrip tag and doesn't sync to other xdrip devices in the sync group.

tolot27:
@gruoner Do you have any ideas to address the problems mentioned above?

gruoner:
> @gruoner Do you have any ideas to address the problems mentioned above?

i didn't recognized issues with reuploading - i use this feature for quite a year

for the other 3 ToDos i can't imaging the problem; i didn't get any problems with "interface hints" or "download problems". Maybe i'm using it another way.
But as i didn't remove treatments in xdrip master nor follower and i didn't use xdrip's or nightscout's reporting feature (i use nightscout reprter) i did not get any problems with removing treatments (for now)

tolot27:
@gruoner Does this PR work for you as well?

PR #1408, opened 2020-08-16
fix NumberFormatException for invalid noise values

### LINKED ISSUE #1366: Follower receives NS entries download Failed: NumberFormatException - Received Null
Hi,

I have set up my NightScout account, and xDrip can write to it from the master device, but when setting up a following device, I receive the following -

NS entries download Failed:
com.google.gson.JsonSyntaxException:
java.lang.NumberFormatException: For input string: "Null"

The NightScout account was originally populated with entries by Diabox, but this stopped after a few hours, and I can't find a way of getting the details on a Wear OS watch. Hence why I'm trying xDrip. So there are gaps in the data, but I wouldn't expect that to cause an issue.

Thanks,

Brian

tolot27:
Attempt to fix #1366 for new and existing noise values.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollow.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollow.java
@@ -11,7 +11,17 @@ import com.eveningoutpost.dexdrip.cgm.nsfollow.messages.Entry;
 import com.eveningoutpost.dexdrip.cgm.nsfollow.utils.NightscoutUrl;
 import com.eveningoutpost.dexdrip.evaluators.MissedReadingsEstimator;
 import com.eveningoutpost.dexdrip.tidepool.InfoInterceptor;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonSyntaxException;
+import com.google.gson.TypeAdapter;
+import com.google.gson.TypeAdapterFactory;
+import com.google.gson.internal.bind.TypeAdapters;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
 
+import java.io.IOException;
 import java.util.List;
 
 import okhttp3.OkHttpClient;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollow.java
```diff
@@ -150,10 +194,13 @@ public static Retrofit getRetrofitInstance() throws IllegalArgumentException {
                     .addInterceptor(new GzipRequestInterceptor())
                     .build();
 
+            final Gson gson = new GsonBuilder()
+                    .registerTypeAdapterFactory(UNRELIABLE_INTEGER_FACTORY)
+                    .create();
```
tolot27, line 199: We may need to add `.serializeNulls()`.

tolot27:
I'll also add a test case.

app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollow.java
tolot27: `serializeNulls` has an effect on serialization only, not during deserialization. Hence, not needed at that code point.

tolot27:
PR is ready for review.

tolot27:
It looks like the source of the problem was fixed by the diabox developers. Nevertheless, this patch is still useful in case of improperly formatted numbers in json strings from other sources.

jamorham:
Can you briefly describe what the code changes do? I only looked briefly but it looks that you are adding a more tolerant parser for integers to the json decoding?

tolot27:
> Can you briefly describe what the code changes do? I only looked briefly but it looks that you are adding a more tolerant parser for integers to the json decoding?

Yes, this PR adds a special TypeAdapter for native ints and Integer objects which gets called during deserialization of a json stream. Within this TypeAdapter, several string to number convert exceptions are catched and a default value is set.
There are some type differences/inconsistencies between the `Entry` class and the `BgReading` class, for example, noise is declared as `int` in `Entry` whereas it is declared as `String` in `BgReading`. That could lead to different json representations which can cause problems during deserialization. This PR tries to fix this during serialization and deserialization without touching the declarations.

jamorham (approved):
This looks good to me. I assume that there isn't an issue where something intended as a string which contains something that can be parsed as a number gets misrepresented as a result of these changes.

PR #1444, opened 2020-09-22
Use joda-time to parse nightscout date entries.

tzachi-dar:
This allows to read all iso formatted strings.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -294,6 +294,7 @@ dependencies {
     implementation 'com.android.support:multidex:1.0.3'
     //implementation 'com.google.dagger:dagger-android-support:2.x' // if you use the support libraries
     annotationProcessor 'com.google.dagger:dagger-compiler:2.25.4'
+    implementation 'net.danlew:android.joda:2.10.6.1'
 
     testImplementation 'junit:junit:4.12'
     testImplementation 'org.objenesis:objenesis:2.5.1'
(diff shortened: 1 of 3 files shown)


sulkaharo:
This fixes various date parsing issues with Nightscout installations that use multiple uploaders to send data to Nightscout

tolot27:
This is for backward compatibility, isn't it? As long as we have to support Android < 8.0 (SDK level < 26), we need this somehow. Otherwise, we could use java.time as noted in the joda-time readme.

tzachi-dar:
Correct @tolot27 but we will probably have to keep supporting it for a long time.

jamorham:
Can someone just explain the theory of operation of this PR? Uses the existing method and if that generates and exception then try via joda time?

sulkaharo:
The issue with the current xDrip implementation of pulling data from Nightscout is that xDrip is hardcoded to expect a certain subset of ISO-8601 date strings to be returned by the API. This fails if Nightscout syncs with multiple clients, as Nightscout only validates the date strings are compliant with the ISO-8601 specification and allows some date string formats to be used for treatments that are not recognised by xDrip, causing exceptions to be thrown in xDrip. What this attempted to do is replace the date parser with the hardcoded date format string with a ISO-8601 parser that can parse any date that complies with the spec, thus fixing the treatment sync for a large group of Nightscout users.

tzachi-dar:
One comment about the way that this is built.
I could have replaced the 3 methods already used in xDrip with the method from joda-time that *should* always work.
However, always using joda-time would require a bigger test matrix to make sure that indeed everything that worked before continues to work.
So, I choose to leave everything as is, and only once a new format comes in, the new code starts to run.
This makes the PR more safe.

jamorham (approved):

Issue #1469, opened 2020-10-17
Phone Battery Level is sent to AAPS instead of Sensor Battery Level


Philoul:
Hi,

I wanted to add Sensor battery level in AAPS User Interface, but I was a little bit disappointed to find "Phone Battery Level" instead of "Sensor Battery Level" behind key ```com.eveningoutpost.dexdrip.Extras.SensorBattery```...  (currently this value is not used in AAPS, that's why this bug was not detected earlier...)

See screenshot below (Sensor Battery Level 75, Phone Battery Level 97) and extract of Log captured in AAPS...

![Screenshot_20201017-102718_xDrip+ 1](https://user-images.githubusercontent.com/52934600/96332460-5f811080-1064-11eb-950c-a0e3834530a3.jpg)

2020-10-17 10:20:49.259 ? D/BGSOURCE: [IntentService[DataService]]: [XdripPlugin.handleNewData():42]: Received xDrip data: Bundle{ ...; com.eveningoutpost.dexdrip.Extras.SensorBattery => 97; ...; }Bundle

Note: detected with xDrip Nightly build 1st Oct 2020

Philoul:
I'm not a specialist of xDrip code (I'm working on AAPS code...) but after a quick analysis, I suspect line 115 of BroadcastGlucose.java file:

May be 
```bundle.putInt(Intents.EXTRA_SENSOR_BATTERY, PowerStateReceiver.getBatteryLevel());``` 
could be replaced by something like
```bundle.putInt(Intents.EXTRA_SENSOR_BATTERY, BridgeBattery.getBestBridgeBattery());```

(not tested, it's just what I suppose after 1 hour of research into xdrip code)
=> If you think it's correct, I can make a PR if it could help...

tolot27:
@Philoul Please create a PR. You analysis seem to be correct. @jamorham can review the PR.

Philoul:
Thank's for your comment...
PR done.

Philoul:
Do you think this PR could be included in the next nightly build version? I think proposed fix is simple with very few risks...
=> I did my PR on the Master branch two weeks ago, I don't know if it's the right way to do it on the xdrip + project because nothing has changed on this branch for a month, and I didn't find a "dev" branch like on AAPS project ...
Thank's

PR #1471, opened 2020-10-18
Fix Sensor Battery Broadcast

Philoul:
Hi,
Proposal to fix broadcast of Sensor Bat Level instead of Phone Bat Level (fix https://github.com/NightscoutFoundation/xDrip/issues/1469)
PR done on Master Branch (I didn't see any "dev" branch...)

Thank's for all your work...

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BroadcastGlucose.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BroadcastGlucose.java
@@ -112,7 +112,7 @@ public class BroadcastGlucose {
                     }
                 }
 
-                bundle.putInt(Intents.EXTRA_SENSOR_BATTERY, PowerStateReceiver.getBatteryLevel());
+                bundle.putInt(Intents.EXTRA_SENSOR_BATTERY, BridgeBattery.getBestBridgeBattery());
                 bundle.putLong(Intents.EXTRA_TIMESTAMP, bgReading.timestamp);
 
                 //raw value
--- a/wear/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BridgeBattery.java
+++ b/wear/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BridgeBattery.java
@@ -1,0 +1,22 @@
+package com.eveningoutpost.dexdrip.UtilityModels;
+
+import com.eveningoutpost.dexdrip.utils.DexCollectionType;
+
+// jamorham
+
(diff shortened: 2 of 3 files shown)


tolot27 (commented):
`PowerStateReceiver.getBatteryLevel()` always returns a positive value, `50` in case of an error. But `BridgeBattery.getBestBridgeBattery()` could return `-1`. Is this caught upstream?

I recommend broadcasting only valid values.

Philoul:
Currently this information is not used in AAPS...
I prepared a PR in  AAPS to add this information in AAPS UI and -1 managed in this PR (used to hide it in AAPS UI for other CGM plugin or for xDrip plugin if invalid value received)

https://github.com/nightscout/AndroidAPS/pull/41

Philoul:
@tolot27 If you prefer, I can modify PR to not broadcast key "Intents.EXTRA_SENSOR_BATTERY" and associated value (-1) in case of invalid value... (I think it's better than broadcast a "default value" like 50 if information is invalid or not available)

In AAPS, I check if this key exists in received bundle from xDrip, so it works too...

tolot27:
> @tolot27 If you prefer, I can modify PR to not broadcast key "Intents.EXTRA_SENSOR_BATTERY" and associated value (-1) in case of invalid value... (I think it's better than broadcast a "default value" like 50 if information is invalid or not available)

Maybe it is a good idea to always broadcast the sensor battery level and if it is invalid or not available, it could be handled upstream.
A value of 50 (used by `PowerStateReceiver.getBatteryLevel()`) is definitively incorrect for the sensor battery level and could not be distingished from real values. Hence, -1 is okay.

And with this PR recommend not to change the implementation of `PowerStateReceiver.getBatteryLevel()`.

tolot27 (changes requested):
Please also change the wear code (`wear/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BroadcastGlucose.java`).

tolot27 (approved):

Philoul:
Done,
Note, BridgeBattery class was missing in Wear app, so I add a copy of main APP file
"bridge_battery" preference key is updated by cgm modules of wear app, so I think it should work...

tolot27:
> Done,
> Note, BridgeBattery class was missing in Wear app, so I add a copy of main APP file
> "bridge_battery" preference key is updated by cgm modules of wear app, so I think it should work...

Maybe it is not used in wear but anyway, it is compatible now. 👍

jamorham (approved):
looks good to me

PR #1505, opened 2020-11-12
avoids unbonding of Dexcom receiver at every read request

### LINKED ISSUE #1494: Android 10+ asks every 5 minutes for pairing with Dexcom transmitter
After updating in October 2020 from android to samsung a50 every 5 minutes xdrip 2020 asks for dexcom Bluetooth connections. The xdrip app released in 2019 works fine.
How to remove the request to connect dexcom via bluetooth on a50 smartphone?

tolot27:
Fixes #1494 

There is no way around reverting commit b0bb827. During every bond, Android 8.0 onwards with the [Android Security Bulletin—November 2020](https://source.android.com/security/bulletin/2020-11-01?hl=en) installed, shows a Bluetooth pairing request consent dialog (see fixes of CVE-2020-12856, specifically [A-157038281 [3]](https://android.googlesource.com/platform/system/bt/+/b3f12befdc4def7d695b6f1049cd02238eb1e4a8)).

Issue #1061 has to be fixed differently.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -261,7 +261,7 @@ dependencies {
     implementation 'io.reactivex:rxjava:1.3.3'
     implementation 'com.activeandroid:thread-safe-active-android:3.1.1'
     //implementation 'com.github.lecho:hellocharts-android:v1.5.8'
-    implementation "com.polidea.rxandroidble2:rxandroidble:1.11.0"
+    implementation "com.polidea.rxandroidble2:rxandroidble:1.11.1"
     implementation 'com.google.guava:guava:24.1-android'
     implementation 'com.embarkmobile:zxing-android-minimal:2.0.0@aar'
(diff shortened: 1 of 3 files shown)


jamorham:
Thank you for identifying the likely Android framework changes which are the source for this new problem relating to the November security patch. The fix google have implemented looks pretty poor in that it affects `createBond()` calls within the framework when the issue is automatic bonding to encrypted characteristics if I read the proof of concept correctly.

This PR reverts the Android 10 workaround where certain Android 10 devices apparently fail fundamentally to implement certain behaviors in relation to bluetooth bonding and encryption. That is going to restore the issues that those users were having.

We cannot implement this change without at least providing preference switch to control the behavior. All we can hope is that the fundamental bonding/scanning implementation in Android 10 either will get fixed or is already fixed. 

I will implement these changes including a preference switch for the original android 10 unbonding workaround which I will default to off.

tolot27:
Resolved with d90388e.

PR #1508, opened 2020-11-13
Dont unbond if on latest security patch

Emma-Black:
The latest security patch requires user input on every single attempt at pairing with a new device

if we unbond each time, then each time we try to pair again it sees it as a new device

this is to disable that nifty little workaroundn in this case, which unfortunately means that old bug will still exist on some devices if owners of devices with that bug update to the latest security patch

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/Ob1G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/Ob1G5CollectionService.java
@@ -71,6 +71,8 @@ import com.polidea.rxandroidble2.scan.ScanResult;
 import com.polidea.rxandroidble2.scan.ScanSettings;
 
 import java.lang.reflect.Method;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


tolot27:
Thanks for your contribution. I thought about the same solution because it keeps backward compatibility for Android 10+ devices not updated, yet. Hence, it can be merged safely.
The downside is that it reintroduces exactly the same problems as the PR #1505 does.

tolot27:
Solved with commit d90388e merged into nightly [2020.11.15](https://github.com/NightscoutFoundation/xDrip/releases/tag/2020.11.15).

saxainden:
Just downloaded Nov 15th nightly and tried the Android 10 workaround, and I'm still getting the Pairing request every 5 minutes or so.  Will wait until the next nightly to see if it starts working.  The Dexcom app works fine with no pairing requests.. This just started with the Nov 1st security update from Samsung(TMobile US) for the Galaxy S10+ was installed today.   Just a data point in case this helps.. for now. will need to install Dexcom app again since if I don't click OK on the pair request, it stops collecting data. Please let me know if there is more info from me that can help.

tolot27:
@saxainden You may have not enabled/disabled the correct options. Please follow https://github.com/NightscoutFoundation/xDrip/issues/1494#issuecomment-728202441

PR #1515, opened 2020-11-18
consistent trend arrows

### LINKED ISSUE #1439: Bug Report: Inconsistent Arrows in Widget and Notification
The trend arrows in the Widget and the Notification are inconsistent. It is dependent on the arrow direction. 45 degree up and down is different than all the other arrows. The screenshot shows two examples.

Tested on Samsung A40 & Samsung S9, Android 10, xDrip 2020.05.05 & 2020.09.04.

arrows

Edit: The "Number Wall on Lock Screen" is also affected.

tolot27:
fixes #1439 and refactors the usage

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/ImportedLibraries/dexcom/Dex_Constants.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/ImportedLibraries/dexcom/Dex_Constants.java
@@ -85,36 +85,48 @@ public class Dex_Constants {
 
     public enum TREND_ARROW_VALUES {
         NONE(0),
-        DOUBLE_UP(1,"\u21C8", "DoubleUp"),
-        SINGLE_UP(2,"\u2191", "SingleUp"),
-        UP_45(3,"\u2197", "FortyFiveUp"),
-        FLAT(4,"\u2192", "Flat"),
-        DOWN_45(5,"\u2198", "FortyFiveDown"),
-        SINGLE_DOWN(6,"\u2193", "SingleDown"),
-        DOUBLE_DOWN(7,"\u21CA", "DoubleDown"),
+        DOUBLE_UP(1,"\u21C8", "DoubleUp", 40d),
+        SINGLE_UP(2,"\u2191", "SingleUp", 3.5d),
+        UP_45(3,"\u2197", "FortyFiveUp", 2d),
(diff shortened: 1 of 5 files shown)


jamorham:
Can you describe the intended changes of this PR and confirm my understanding of the issue, that on Samsung we have to use a different unicode character to reference the symbol?

Are we sure that this is correctly supported on other devices?

Thanks.

tolot27:
> Can you describe the intended changes of this PR and confirm my understanding of the issue, that on Samsung we have to use a different unicode character to reference the symbol?

We do not use a different character to reference the symbol. We just give the Android system the advice to use the text representation instead of a graphical representation. As Emojis got incorporated into the Unicode set and synchronized with it, no new characters were created. Instead, [variation selectors (VS) ](https://en.wikipedia.org/wiki/Variation_Selectors_(Unicode_block)) were introduced with Unicode version 3.2. Now, one can choose how to show an emoji, either with VS16 (U+FE0F) for emoji-style (with color) or VS15 (U+FE0E) for text style (monochrome).
If no VS is used, it defaults to the system settings which may be differently set by different vendors.

>Are we sure that this is correctly supported on other devices?

Unicode v5 is supported since Android 1.5. Hence, it should be safe to use the text style VS.
We could even try to use the emoji-stale VS to display the graphical version is set in xDrip rather than using images. But this would be a different PR.

jamorham (approved):
Looks good. It could do with more documentation and test cases but I can add those.

tolot27:
> It could do with more documentation and test cases.

You are right. I'll add a detailed description with new PRs and will setup a PR template.

Regarding test cases: Last week I've searched about visual testing and found [Visual tests on every pull request](https://githubuniverse.com/Visual-tests-on-every-pull-request). Maybe, we can find some visual unit testing libs and run them at different screen sizes and Android versions.

jamorham:
The test cases I was thinking of are for different values mapping to trend arrows as the logic of that has been changed, particularly important to test the edge cases where previously it was >= comparison. and that test results exercise all the different trend arrows. 

Also on wear you have `Constants` class instead of `Dex_Constants` in this PR. They should be the same name for sanity.

tolot27:
> Also on wear you have Constants class instead of Dex_Constants in this PR. They should be the same name for sanity.

That's why I added a task at #1522 called "refactor common code of phone and wear app into a library module to avoid code duplication". :smiley:

PR #1516, opened 2020-11-18
Clear Queue button and Stop/Start Sensor options 

### LINKED ISSUE #1118: Feature Request: Deleting Dexcom queue entries
Hello,

sometimes, there are queued entries like „stop sensor“ or „calibrate ...“ in the queue to the Dexcom transmitter (G6 or G5) and it would be nice and helpful, if these entries could be deleted.

Thanks in advance.

Kind regards.

ShusterRooster:
Adds a Clear Queue button to System Status and allows the Stop Sensor and Start Sensor buttons to be shown simultaneously if the user is in Engineering Mode. 

Fixes #1118, the scenario in which a request to start/stop a sensor doesn't go through or fails, thus causing the app to display the opposite of the desired action (i.e. Start Sensor when sensor is already activated or Stop Sensor when the sensor is already inactive). This issue causes users to become stuck in a loop of stopping and starting sensor sessions because the queue still holds the request to either start or stop a sensor even though that's not the desired or intended action.


tolot27 (approved):
The PR does what it claims. I would prefer to not see the Start/Stop Sensor buttons if no sensor-related data source like Nightscout Follower is selected. But that's for a different PR.

ShusterRooster:
> Thx for your effort, @ShusterRooster. The PR does what it claims. I would prefer to not see the Start/Stop Sensor buttons if no sensor-related data source like Nightscout Follower is selected. But that's for a different PR.

I can try messing around with that. Best case scenario you'll see a PR for that in a few days!

ShusterRooster:
Whoops, was trying to get a commit on top of a PR... Clearly very new to this haha

tolot27:
@ShusterRooster Please can you cleanup the commit history? Only 05bb8ac and 421600c are necessary.

ShusterRooster:
I've been having trouble cleaning commit history. Both Git Bash and Android Studio are showing histories with the merges dropped or squashed however nothing's going through. I accidentally deleted my branch thinking it was a different remote which was just dumb on my end. If you could link me to any resources that may help me with this, it would be much appreciated. I have already looked through many guides but they haven't worked. Sorry about this.

tolot27:
@ShusterRooster Your branch [start-stop-fix](https://github.com/ShusterRooster/xDrip/tree/start-stop-fix) is still available. Just fetch it and clean it up with the above commands.

ShusterRooster:
Thank you so much @tolot27 for dealing with my lack of experience and helping me through this. Everything should be cleaned up now. Thanks again!

tolot27 (approved):
Looks good so far. I've tested it for about a week.

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
```diff
@@ -102,15 +108,22 @@ public NavDrawerBuilder(final Context context) {
...
-            else if (!Pref.getBooleanDefaultFalse("engineering_mode") && is_active_sensor) {
-                this.nav_drawer_options.add(context.getString(R.string.stop_sensor));
-                this.nav_drawer_intents.add(new Intent(context, StopSensor.class));
-            }
-            else {
-                this.nav_drawer_options.add(context.getString(R.string.start_sensor));
-                this.nav_drawer_intents.add(new Intent(context, StartNewSensor.class));
+
+            else if (!Pref.getBooleanDefaultFalse("engineering_mode")) {
+                if (!(DexCollectionType.getDexCollectionType().equals(Follower) || DexCollectionType.getDexCollectionType().equals(NSFollow) || DexCollectionType.getDexCollectionType().equals(SHFollow)
```
tolot27, line 113: I recommend to add an additional HashSet to `DexCollectionType` containing devices which support sensor starts/stops as well as a function `hasSensorStartStopSupport`. This function can than called instead of the logical ORs.

ShusterRooster:
Just made that. Initial testing seems good so far. Let me know if you need me to change anything!

ShusterRooster:
There was a simple error I missed, fixed in the latest commit.

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
```diff
@@ -93,12 +101,26 @@ public NavDrawerBuilder(final Context context) {
...
+            }
+            if (Pref.getBooleanDefaultFalse("engineering_mode")) {
                 this.nav_drawer_options.add(context.getString(R.string.stop_sensor));
                 this.nav_drawer_intents.add(new Intent(context, StopSensor.class));
-            } else {
+
                 this.nav_drawer_options.add(context.getString(R.string.start_sensor));
                 this.nav_drawer_intents.add(new Intent(context, StartNewSensor.class));
             }
+
```
tolot27, line 112: Please format the code consistently, i. e. `} else if ...`

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
```diff
@@ -93,12 +101,26 @@ public NavDrawerBuilder(final Context context) {
...
+            else if (!Pref.getBooleanDefaultFalse("engineering_mode")) {
+                if (hasSensorStartStopSupport()) {
```
tolot27, line 114: Please combine this if statement with that of the previous line.

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
```diff
@@ -16,13 +16,21 @@
...
 import com.eveningoutpost.dexdrip.UtilityModels.Experience;
+import com.eveningoutpost.dexdrip.UtilityModels.Pref;
 import com.eveningoutpost.dexdrip.stats.StatsActivity;
 import com.eveningoutpost.dexdrip.utils.DexCollectionType;
 import com.eveningoutpost.dexdrip.utils.Preferences;
 
 import java.util.ArrayList;
 import java.util.List;
 
+import static com.eveningoutpost.dexdrip.utils.DexCollectionType.DexcomShare;
```
tolot27, line 27: Please remove these imports and use `DexCollectionType.hasSensorStartStopSupport` in line 114.

app/src/main/java/com/eveningoutpost/dexdrip/utils/DexCollectionType.java
```diff
@@ -170,6 +170,26 @@ public static boolean isLibreOOPAlgorithm(DexCollectionType collector) {
         return getCollectorServiceClass(getDexCollectionType());
     }
 
+    public static boolean hasSensorStartStopSupport(){
```
tolot27, line 173: Please implement it using a HashSet like in `hasBattery()`.

tolot27 (changes requested):
Sorry to bother you again. I made some comments. IMHO, having a `private static final HashSet<DexCollectionType> supportsSensorStartStop = new HashSet<>();` similar to `usesBattery` is the way do go.

ShusterRooster:
That should be everything from what I've seen. Let me know if you need me to do anything else!

app/src/main/java/com/eveningoutpost/dexdrip/utils/DexCollectionType.java
```diff
@@ -130,6 +132,8 @@ public static boolean hasBattery() {
         return usesBattery.contains(getDexCollectionType());
     }
 
+    public static boolean hasSensorStartStopSupport() { return supportsSensorStartStop.contains(getDexCollectionType()); }
```
tolot27, line 135: Again: please use consistent code formatting! The function should look like the other functions:

```java
public static boolean hasSensorStartStopSupport() {
    return supportsSensorStartStop.contains(getDexCollectionType());
}
```

ShusterRooster:
I don't know what happened there, seems like Android Studio minimized everything.
image

app/src/main/java/com/eveningoutpost/dexdrip/utils/DexCollectionType.java
```diff
@@ -78,6 +79,7 @@
         Collections.addAll(usesBattery, BluetoothWixel, DexbridgeWixel, WifiBlueToothWixel, WifiDexBridgeWixel, Follower, LimiTTer, LibreAlarm, LimiTTerWifi, LibreWifi); // parakeet separate
         Collections.addAll(usesDexcomRaw, BluetoothWixel, DexbridgeWixel, WifiWixel, WifiBlueToothWixel, DexcomG5, WifiDexBridgeWixel, Mock);
         Collections.addAll(usesTransmitterBattery, WifiWixel, BluetoothWixel, DexbridgeWixel, WifiBlueToothWixel, WifiDexBridgeWixel); // G4 transmitter battery
+        Collections.addAll(supportsSensorStartStop, BluetoothWixel, DexbridgeWixel, LimiTTer, LimiTTerWifi, LibreWifi, WifiBlueToothWixel, WifiWixel, DexcomG5, DexcomG6, WifiDexBridgeWixel, LibreAlarm, Medtrum, LibreReceiver);
```
tolot27, line 81: @jamorham Which implementation really supports sensor starts/stops? I believe it is only `DexcomG5`, `DexcomG6` and subset or all entries of `usesLibre`.

tolot27 (approved):
Looks good, now and can be merged.

tolot27:
> I don't know what happened there, seems like Android Studio minimized everything.

Most likely, you have enabled auto-code-reformat. You have to disable it. Otherwise, diffs get much harder to read.

tolot27 (approved):

tolot27:
@ShusterRooster Thx for you effort and patience. Now, we have to wait for @jamorham decision.

jamorham:
I'm not able to reproduce the original problem reported in the linked issue. If you change transmitter the old queue will get sent to the new transmitter but it will ignore/reject a stop sensor session request when a session is not in progress. Items do not get stuck in the queue, rejected stop sensor commands are not retried.

It is true to say that there are some confusing inconsistencies in the user interface regarding stopping and starting sensors which are probably a hang over from when it was possible to run both native and non-native G5/G6 effectively at the same time but those are not really needed now. There are also some potential interactions with automatic restarting of sensors which need to be handled carefully.

What I would suggest is simply clearing the transmitter queue when the transmitter id is changed and probably automatically stopping the sensor session as far as xDrip is concerned when it is in native mode and is informed that the sensor is in a stopped state. This does require careful handling though because sometimes you can send a sensor start and the session will start but you wont get that data till the next reading.

ShusterRooster:
Sorry for the confusion. The linked issue is not exactly what the addressed problem is but this pull request does address some of the problems in that feature request. The issue I, and a few other people, have been experiencing is, for example:

Upon trying to start a sensor, the Start Sensor command doesn't go through or fails and makes it so that the only possible option you have to reverse it is to press Stop Sensor and then Start Sensor. In my experience, what tends to happen is that the Start Sensor command is applied but the Stop Sensor cancels that out leaving you without a current sensor session. This chain of events also leaves you with the Stop Sensor option in the navigation bar thus repeating the cycle. 

I have had days where this problem has prohibited me from using a sensor and there was nothing allowing me to clear the queue. I am not 100% sure about the details and how exactly this problem occurs. For all I know, it could just be the newer Dexcom TX firmwares blocking what it thinks is an attempted restart but the above paragraph states what I assume happens behind the scenes that ultimately causes this issue.

A friend of mine had this issue a few days ago:
image

The showing of both the Sensor Start and Stop options in the navigation bar can be removed but the most important feature needed to fix this issue is the Clear Queue command as it would help stop this issue as it arises. It would also help to fix other issues like canceling undesired actions (i.e. accidental calibrations, unwanted start/stop commands, etc.).

jamorham:
There is a voice command `clear transmitter queue` which was added very recently.

I agree that the xDrip concept of started/stopped sensor state could better track the sensor state but if your transmitter session is stopped and xDrip has a stop session command in the queue it will just send it to the transmitter and then empty it from the queue. I don't know how it can get stuck there because of how it works. The transmitter I would expect to ignore it.

There is the possibility that in this state you need to wait for the queue to empty before trying to start the next session but I think that is a different issue.

From what you're describing I think a better fix would be clear a stop session command from the queue when the state is already listed as stopped and the session start command is issued. That should require no user interaction or ui changes but clean up that scenario. I can implement that.

ShusterRooster:
> From what you're describing I think a better fix would be clear a stop session command from the queue when the state is already listed as stopped and the session start command is issued. That should require no user interaction or ui changes but clean up that scenario. I can implement that.

I agree, that would be a good fix to implement. I can remove [421600c](https://github.com/NightscoutFoundation/xDrip/commit/421600ccea5a2ca657561cf7873cb9e9800dce15) and just keep the part of the commit that shows sensor options depending on the collection method ([ec7f326](https://github.com/NightscoutFoundation/xDrip/pull/1516/commits/ec7f32615e27472abf28a831e3444809dc738d3d)) as I feel it's still a useful UI enhancement to add. 

You'll just need to tell me if the current configuration for the collection type and whether or not to show the sensor options needs fixing or tweaking of any sort. I can implement that fix you mentioned earlier for you if you'd like but the option is yours.  Just let me know if I need to do anything and I'll do it.

ShusterRooster:
I found out another reason why I get stuck in a loop with transmitters recently. When the session ends and the sensor turns off by itself, the Stop Sensor command is still there and is never refreshed and it gets you into that loop. I remembered you talking about the voice commands that allow you to clear the transmitter queue but when I tried to use them, my phone would only pick up "Clear Transmitter **Q**" so the command did not work even when I used a text to speech bot to pronounce it. I doubt my fix for keeping track of the sensor state works and every 10 days or so to test it out is very scarce.

Navid200:
When the transmitter stops on day 10, just tap on stop sensor.  Wait 10 minutes.
Now, you can insert a new sensor, snap in the transmitter and start sensor.
There is no reason to get stuck in a loop.

After the clear queue capability was added, I see no reason for anyone to get stuck.

Navid200:
@ShusterRooster 

If you still get stuck in a loop, would you please provide detail on how I can reproduce the problem?

Otherwise, would you please close this pull request?

Navid200:
Can we close this PR please?

tolot27:
@ShusterRooster Please can you confirm that the problem is fixed with the latest nightly build?

Navid200:
Can we close this PR please?

ShusterRooster:
> @ShusterRooster Please can you confirm that the problem is fixed with the latest nightly build?

Yes, I have been actively using the nightly build and everything works like a charm. Haven't had any aforementioned issues for a while at this point.

PR #1533, opened 2020-12-01
Speak readings/alarms when connected to Android Auto

tim162:
I'm trying to be as precise as possible:

The idea was to have the car read your values/alarms when you are driving and your phone is connected to Android Auto.

In the first implementation I created a messaging style notification when a new value/alarm is detected and send it to the car when connected. The problem here was: 
- I had to migrate the whole project to AndroidX. This should be done in the future nevertheless but I think it would be quite good to have it in a seperate PR.
- To have the notification shown in the car, you would need to activate developer settings in Android Auto-app and then activate unknown sources because xDrip is not installed via Play Store. This may be quite complicated for inexperienced users, and we would have to explain it.
- When the notification is received in the car, you need to press on it on your cars screen to have it read out.

After some ideas, I thought the best way was to simply use the Text-to-Speech feature we already use in the app. I implemented two new Switches in the Settings -> Speak Readings -> `Speak Readings in car` and `Speak  alerts in car`. I also added translation for German. Then everytime it was checked whether Speak is activated I changed to also check whether Speak in car is activated && phone is connected to Android Auto. 

In SpeechUtil I changed the way currently played music is handled. As especially in the car (and I guess this is also way better if not in car than before) when music is currently playing we don't wait 40 sec and then just speak but now request Audiofocus, which means music will be paused for our Speaking Event and resume afterwards.

I think it's now way better and easier to do it with the Text-to-Speech than firing an additional notification that can be read by the car.

For the future it might be interesting to have a look into 'better' texts to be read out. Now for a new value it's "100 flat" and it might be cool to have sth like "Your blood glucose is at 100 and steady, it changed +4 since your last reading". This might also be choosable by the user whether to use long or short form. Before changing this I wanted to have your oppinion on it.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NewDataObserver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NewDataObserver.java
@@ -1,5 +1,10 @@
 package com.eveningoutpost.dexdrip;
 
+import android.app.UiModeManager;
+import android.content.Context;
+import android.content.res.Configuration;
+import android.util.Log;
+
 import com.eveningoutpost.dexdrip.Models.BgReading;
 import com.eveningoutpost.dexdrip.Models.JoH;
 import com.eveningoutpost.dexdrip.Models.LibreBlock;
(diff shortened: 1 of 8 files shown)


tolot27:
I will test this on Monday/Thuesday with my car. 😄

Navid200:
xDrip already has "Speak Readings".
What does this pull request provide that we cannot accomplish already with xDrip as is?

tolot27:
@Navid200 This PR implements Android Auto support as it is mentioned in the title and description. The car computer now "speaks" the BG values and interrupts car music play, etc.

Navid200:
I see.  Sorry I missed that.  Thanks for pointing that out.

Navid200:
Would you please have a look at Motion Tracking menu?  Please see the attached image.
Is this not the capability, which is already available?
Screenshot_20210428-074919

CarleneCram:
At the moment when my phone connects to Bluetooth in the car it silences the speak readings. I have tried to enable all the car connections that I have found but readings do not come through the car audio and my phone is silenced. I don't care what I get the readings through, I just want some readings while phone is connected. A Samsung A32.

Issue #1547, opened 2020-12-09
Android 11 Restore is empty (zip and sqlite is correct)


zehnBE:
I got a Android 11 upgrade on the Pocophone f2 Pro.

I create a xdrip Backup and the zip file with the sqlite was correct created.
A restore with the Backup set all to empty (sgv, treatments etc).

A restore with the same xdrip apk and same Backup (createt on Android 11) on a Android 10 Phone works fine.
I must be comes from Android 11 with the new restriction to read Files.
The import from the preferences Files works.

Xdript Verson f1f44dbb-2020.12.09

nickb24:
> Xdript Verson f1f44dbb-2020.12.09

Where did you get this version? The latest release as of this writing is 2020-11-27

AdrianLxM:
This is the new restriction on file handling in Android 11.
As xDrip is most likely not targeting Android 11 any time soon (target is still 23), we could add a flag to the manifest that most likely should still allow to import backups.

AdrianLxM:
@zehnBE I added a flag in this PR that hopefully solves this for Android 11: https://github.com/NightscoutFoundation/xDrip/pull/1548

Could you please try if that is enough as I have no private Android 11 device at hand right now?

AdrianLxM:
Unfortunately adding the flag did not help.

tolot27:
> This is the new restriction on file handling in Android 11.

Please can you reference it?

jwoglom:
@zehnBE Can you post logcat output? I posted issue #1475 which sounds related.

Navid200:
Closing due to inactivity.
Please update and answer questions if you like to reopen.

Navid200:
@AdrianLxM Would you please explain what could cause some people, using Android 11, to experience this problem and some, also using Android 11, not?

Navid200:
@zehnBE Would you please let us know if you are still experiencing this?
If yes, would you please attempt importing the database file 3-4 times?
There is a suggestion that attempting it 3-4 times will make it work properly.

Navid200:
Closing due to inactivity

UberQueiroz:
> @zehnBE Would you please let us know if you are still experiencing this? If yes, would you please attempt importing the database file 3-4 times? There is a suggestion that attempting it 3-4 times will make it work properly.

[...]

PR #1548, opened 2020-12-10
add requestLegacyExternalStorage to Manifest

AdrianLxM:
Attempt to solve https://github.com/NightscoutFoundation/xDrip/issues/1547
@zehnBE could you please test this?


jamorham:
Doesn't this manifest tag apply only when tagetting Android 10 or above?

zehnBE:
sorry, no. After import, the Data are empty.
I have updated only and i have try it with uninstall and reinstall.

AdrianLxM:
Would have been to easy :/
Thanks for testing

Issue #1573, opened 2020-12-24
xDrip cannot start a G6 sensor in no-code mode with a new (8J) transmitter


Navid200:
xDrip version: July 25, 2020
Dexcom G6
Transmitter serial number: 8J****
Phone (Android 9) runs xDrip as the collector

We were able to run in no-code mode, using xDrip, by entering 0000 as the calibration code.  I did that whenever I restarted my sensors with my 81 transmitters.  

With newer transmitters (I have tried 8J), after entering 0000 as the calibration code, no readings are ever provided after the warm-up period, and the transmitter eventually issues an error.

The Dexcom app can start a sensor (with new transmitter) in no-code mode apparently.  I personally have no experience with that app.  It would be great to be able to do the same using xDrip.

emp-00:
@Navid200 : Is there a good reason for using "0000" as calibration code? As you know - All 8G+ (Firefly) transmitters can only be used in "Native Mode" anyway - therefore the sensor needs the calibration code to be able to show "correct" BG values. 

Manual calibration in xDrip (non-native mode like with 80/81 tx's) does NOT work anymore with these transmitters so you must calibrate them natively in the transmitter... I always use the factory code also for restarts => no problems. Only need to calibrate once after the restart since values are always consistently too high, e.g. 130-150 instead of real 100 mg/dl (due to built-in trauma compensation). One single calibration fixes this issue for me.

Navid200:
@emp-00 
Yes, there is.
In no-code mode, G6 behaves as a G5.  Hence, the factory calibration, which is only valid for the first 10 days becomes disabled in no-code mode.  
The Dexcom app offers that.  So, the fact that the transmitter is limited to Native mode is irrelevant.  The transmitter is in Native mode and in no-code mode, it behaves as a G5 with the Dexcom app.

I would like xDrip code to be modified so that it can offer the same capability to those who use xDrip.

I enter the code also and calibrate to correct the significant, dangerous error because that's all I can do. 
But, there is no reason not to improve things.  That's the purpose of these issues, isn't it?

There is a very good reason to fix this bug.
I always worry someone may inject insulin, not realizing the error, and cause a hypo.

emp-00:
@Navid200 : Sorry, I was not aware this is still possible with the 8G+ transmitters. 

[...]

PR #1583, opened 2020-12-30
Dexcom G6 Calibration Error Codes Update

nickb24:
Based on the discussion here: https://github.com/NightscoutFoundation/xDrip/issues/1573

Will need to determine what the implications are for G5 and non-firefly G6 transmitters.


tolot27 (approved):
Looks good to me. Maybe, we should add a mapping table to the Wiki to know the error codes in the future?

tolot27 (changes requested):
The error message strings should be moved to the translatable strings.
If they should keep an error code, this should be encoded as a variable.

Navid200:
May I ask what the status of correcting the calibration messages is?

Right now, we get logs that suggest xDrip has rejected the calibrations.  Many users find the logs confusing.  There are new users constantly switching from G5 to G6.  They will all be confused.

If there is a reason we cannot change the logs, can we at least stop the erroneous logs as a first step?
No log is better than a log that says calibration was rejected when in reality is has been accepted by the transmitter.

nickb24:
I made this PR as a draft which I assume anyone can edit, right? In any case this PR won't be merged as it stands. We can leave it open for reference or I can close it.

tolot27:
> I made this PR as a draft which I assume anyone can edit, right?

Only you can edit them or developers with a higher permission level.

PR #1589, opened 2021-01-05
Bolus Calculator

ShusterRooster:
This adds a bolus calculator to xdrip. Has customizable settings for bolus increment, max bolus, a correct above value, and some other things. Also added was bolus recommendations in the regular treatments tab where it suggests an amount of insulin depending on current sensor bg and what the user has already put in. There's a carb offset field where you can set how far off you're going to be eating carbs and whatnot. There are probably some minor bugs here and there but I haven't spent an obscene amount of time hunting bugs.

image
image
image

Please let me know if you want anything added or fixed as my code is probably not the best! Thanks for taking the time to review this.


Navid200:
We have to be careful how a capability we add may be used by everyone and the possible consequences.
My worry is someone thinking this could be the answer to their struggles with controlling their blood glucose and overdose on insulin.

It's true that no matter what we do, someone could abuse the tool and hurt themselves.  But, I don't want to feel guilty some day because I overlooked an obvious possibility.

The first thing anyone sees after installing xDrip is that xDrip should not be used to make medical decisions.

xDrip is not a camera app where the worst thing that could happen is bad pictures.  The worst thing that could happen with xDrip is not something any of us likes to think about.  But, we have to.

tolot27:
Personally, I like the "bolus recommendations in the regular treatments tab" because this information is ahead of the current predictive simulation. You do not need to enter the carbs, confirm it and wait for the predictive simulation information, which do not appear immediately because there are no COBs right after entering it. The insulin wich is need for the carbs is displayed immediately. This is the biggest advantage, IMHO.

The bolus calculator view is a common view similar to the FreeStyle Libre Reader and other devices.

Navid200:
If this is an evolution of a feature already existing (simulator), it's best to present it that way.
It would make no sense, at least in my opinion, to leave the existing feature as is and add something else that does the same thing better.

Let me provide an example.  We already have a collector for Dexcom G6.  If someone would like to improve it, they can.
But, if someone writes a G6 collector from scratch, do you think we should leave the existing collector and allow a second one to be added to the list?  If yes, can you explain how that would help the user community?

You say this is how it is in the FreeStyle Libre.  Are you saying FreStyle Libre has two different ways for providing suggestions for how much insulin to take?  Do those two provide exactly thew same suggestions, in other words, do they provide redundant suggestions?  Why?
If not, one of them suggests to inject 1 unit and other one suggests to inject 2 units?!

Wils1907:
> tolot27 mentions "bolus recommendations in the regular treatments tab". Is this in xDrip+? I have " Show Bolus Wizard Preview" and "Always show Bolus Wizard Preview" switched on in the Settings but haven't found where the recommendations are. Sorry if I've posted this in the wrong place; I'm a newbie and have no idea what I'm doing!

tolot27:
> tolot27 mentions "bolus recommendations in the regular treatments tab". Is this in xDrip+? I have " Show Bolus Wizard Preview" and "Always show Bolus Wizard Preview" switched on in the Settings but haven't found where the recommendations are.

They are part of this PR. See 2nd picture in the PR description: https://user-images.githubusercontent.com/66763392/103595776-5288f600-4eca-11eb-8c13-85098e58a1f2.png

Wils1907:
Hi tolot27. I don't see that - it just shows the number of carbs.

tolot27:
You can't see it in xDrip because this PR is not merged. If you have further questions please post it in the support forums.

PR #1590, opened 2021-01-05
Error when dealing with what seems to be a new variant of the BatteryInfoRxMessage

dspnikder:
Guess:

resistance doesn't become useful as a diagnostic tool and/or they are gathering data that is useless prio to a certain time into the length of the sensors active time. However the switch from 10 to 12 means that resistance goes from 0 (or rather nonexistant) to existing. Which wasn't expected here, instead  both runtime and temperature got eaten. As it were. 

Note: I do great big stupid backends and we try to check more edge conditions. It would be very useful with a package dumping console . But that's just me. Anyway, dont't use this particular code (because it doesn't even have a newline at the end but ... inspiration.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/BatteryInfoRxMessage.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/BatteryInfoRxMessage.java
@@ -6,6 +6,7 @@ import com.eveningoutpost.dexdrip.Services.G5CollectionService;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.util.Locale;
+import java.util.function.Predicate;
 
 /**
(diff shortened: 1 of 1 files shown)


tolot27:
@dspnikder Thanks for catching this. Is there an existing issue describing the problem?

tolot27 (changes requested):
To consider this PR at least travis-ci should not fail. It complains about androidx classes. Strange.

dspnikder:
I can tell you why. Because for some reason there is a discrepancy between versions of ByteBuffer used. The one in the build returns an int (IIRC) from the mark() whereas the java.nio.ByteBuffer has been returning ByteBuffer since.... oh a while. The androidx issue is related to a incremental compilation of code generating annotation processors. But I figured out what the problem was. Backwards compatibility and a yay old version of ByteBuffer being used during the build.

Issue #626, opened 2018-10-16
Large database sizes slow down interface?


hardaker:
I have/had an xdrip database on my pixel 2 that's been around since Mar, 2017.  The xdrip interface (eg, when entering treatments) has been getting slower and slower over time to respond to screen presses, etc.    Additionally, I'd get warning messages that 'xdrip was taking too long to respond: close or wait?'.

As a guess that it was the sqlite database getting too big, I exported it, pulled it over to my desktop, ran `delete from BgReadings where timestamp < 1533106800000;` (which killed everything before august 1, 2018, and pushed it back and imported it.  Sure enough, it's now zippy fast again.

Not sure what can be done about this other than to offer a "keep for how long" interface, suggesting at least 3 months of data is wise for the a1c calculations.

MasterPlexus:
Just check the settings. There is are two options which are on Interest for you on this (settings, additive settings, Glucose retention and save database daily. There you can set is as you want (for example 90 days)

Navid200:
@hardaker Would you please close this issue if you are satisfied with the response you got?
Thanks

hardaker:
Thanks for the help.  I've switched to using that and it works well.

PR #1594, opened 2021-01-06
Allow xDrip to show libre trend for libre2 sensors.

tzachi-dar:
Still under testing.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
@@ -82,6 +82,7 @@ import java.net.URLEncoder;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.charset.Charset;
+import java.text.DateFormat;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
 import java.text.SimpleDateFormat;
@@ -188,6 +189,18 @@ public class JoH {
 
     public static long msSince(long when) {
         return (tsl() - when);
+    }
+
+    public static long msSince(long end, long start) {
+        return (end - start);
+    }
+
+    public static String toHumanTime(double t) {
+        return toHumanTime((long)t);
+    }
+
+    public static String toHumanTime(long t) {
+        return DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT).format(new Date(t));
     }
 
     public static long msTill(long when) {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreBlock.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreBlock.java
@@ -1,7 +1,10 @@
 package com.eveningoutpost.dexdrip.Models;
 
+import android.database.Cursor;
+import android.database.sqlite.SQLiteDatabase;
 import android.provider.BaseColumns;
 
+import com.activeandroid.Cache;
 import com.activeandroid.annotation.Column;
 import com.activeandroid.annotation.Table;
 import com.activeandroid.query.Select;
@@ -12,6 +15,9 @@ import com.eveningoutpost.dexdrip.UtilityModels.Pref;
 import com.eveningoutpost.dexdrip.UtilityModels.UploaderQueue;
 import com.google.gson.annotations.Expose;
 import java.text.DecimalFormat;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import java.util.UUID;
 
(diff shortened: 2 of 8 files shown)


keencave:
The PR is working well. Operation is stable so far, no drawbacks. One can see the sensor noise level easy, fast changes could be detected early.

fjpezuela:
Thank you very much for this improvement. I am looking forward to it.

jamorham:
Can this PR be described as to what its purpose is, the nature of the changes and whether it is believed to be ready for merge? Currently it says `still under testing` so I am assuming it isn't ready for review?

tzachi-dar:
Yes, still under testing.
The purpose of existing code is to show the per minute values of libre 2 as well.

Coming next are:
    1) gap closing for libre bluetooth.
    2) data smoothing.
    3) filtering bad points.

jamorham:
Okay, I got tagged to review it but it seems we're not at that stage yet. I will add some inline comments anyway so you can incorporate them now if you like.

fjpezuela:
I don't know how to offer myself to try it, otherwise I would.

app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
```diff
@@ -190,6 +191,18 @@ public static long msSince(long when) {
         return (tsl() - when);
     }
 
+    public static long msSince(long end, long start) {
```
jamorham, line 194: Does this really need a helper function if its just x - y ?

app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
```diff
@@ -190,6 +191,18 @@ public static long msSince(long when) {
...
+        return (end - start);
+    }
+
+    public static String toHumanTime(double t) {
+        return toHumanTime((long)t);
+    }
+
+    public static String toHumanTime(long t) {
```
jamorham, line 202: Does `dateTimeText` not meet your needs? Also is this method even referenced by the PR?

app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreBlock.java
```diff
@@ -129,29 +135,58 @@ public static LibreBlock getLatestForTrend() {
...
-                .where("timestamp <= ?", end_time)
-                .orderBy("timestamp desc")
-                .executeSingle();
+        SQLiteDatabase db = Cache.openDatabase();
+
+        // Using this syntex since there is no way to tell the DB which index to use.
+        // Using ActiveAndroid method would take up to 8 seconds to complete.
+        Cursor cursor= db.rawQuery("select * from libreblock  INDEXED BY  index_LibreBlock_timestamp "+
+                "where bytestart == 0 AND byteend == 344 OR byteend == 44 " +
+                "AND timestamp >= " + start_time + " AND timestamp <= " + end_time +
```
jamorham, line 162: Does the SQL query need brackets to encapsulate the `OR` clauses? 

Do you need to make sure you close the cursor in the event of exception or other unexpected code path? I'm not familiar enough with sqlite optimization to know why you need to specify the index manually. Does the order of the clauses have any impact? 

I am concerned overall about the performance impact of the per minute Libre trend stuff so it would be very good to get this as efficient as possible.

app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreOOPAlgorithm.java
```diff
@@ -311,13 +314,39 @@ public static void handleDecodedBleResult(long timestamp, byte[] ble_data, byte
         glucoseData.glucoseLevelRaw = raw;
         
         libreAlarmObject.data.trend.add(glucoseData);
+
+        //========= add code here for smmothing and gap closing.
```
jonaschn, line 318: What is the current progress of gap closing?
There is a PR open for smoothing #1401

app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
tzachi-dar: Well, I can remove it if this is a problem.

app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
tzachi-dar: This was used in debugging this PR. In any case I'll use dateTimeText.

app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreBlock.java
tzachi-dar: Yes, the SQL needs brackets for the OR, I have missed this.

I think that the code as written should be fine, but moved this using a try block to make sure that it gets closed.
I have tried changing the order of clauses but it did not help.

As for performance impact, there is nothing new here. Just adding the libre bluetooth. This changes will make things much faster for old code as well. In any case this feature is optional.

By the way, I believe that we are having the same issue on many of our DB queries. See for example: https://github.com/NightscoutFoundation/xDrip/issues/626

app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreOOPAlgorithm.java
tzachi-dar: gap closing is implemented in https://github.com/NightscoutFoundation/xDrip/pull/1640
Looking at 1401 to see if I can get ideas from it.

tzachi-dar:
This pr is replaced by https://github.com/NightscoutFoundation/xDrip/pull/1640 (fixes for comments are there).

PR #1596, opened 2021-01-08
securing Google Maps key

tolot27:
This PR can read the Google Maps key from the Gradle project properties or an environment variable.

**Rationale**: Currently, the Google Maps keys are stored as resource string variables in two files in the debug and release output directory `res/values/google_maps_api.xml `, redundantly.
Creating a new build type results in a build error until `google_maps_api.xml` is duplicated again. Since the key is identical for all builds, it can be stored centrally. 

In general, it is a good practice to exclude keys from public repositories. This is not the case with xDrip. With this PR, it is easy to externalize the key from `app/gradle.properties` and store a newly generated key into an environment variable or a project or local property (`local.properties` file). It is also possible to set it using `gradle -Pgoogle_maps_key='<key>'` or specify it in `.travis.yml` in the `env` section, possibly as [encrypted variable](https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml).


tolot27:
rebased on current master (35c786c1)

jamorham (approved):

PR #1598, opened 2021-01-09
enable incremental debug builds and apply changes

tolot27:
This PR enables incremental debug builds and minimizes build time if "Apply changes" is executed by introducing an additional build type and a [productFlavor](https://developer.android.com/studio/build/build-variants#product-flavors) `fast`.

A cleaner way would be to disable `minifyEnabled` and `shrinkResources` in the debug buildType, which is the default and not really necessary. This would avoid the introduction of the new buildType `dev`.

The main reason that the incremental compilation did not work is the usage of timestamps and random UUIDs in the `buildConfigField`. These are not necessary for debug or development builds.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -5,11 +5,20 @@ buildscript {
     }
 
     dependencies {
-        classpath "com.newrelic.agent.android:agent-gradle-plugin:5.27.1"
+        if (getGradle().getStartParameter().getTaskRequests().toString().endsWith("Release")) {
+            println("Enabling New Relic")
+            classpath "com.newrelic.agent.android:agent-gradle-plugin:5.27.1"
+        }
     }
 }
 apply plugin: 'com.android.application'
-apply plugin: 'newrelic'
+if (getGradle().getStartParameter().getTaskRequests().toString().endsWith("Release")) {
+    apply plugin: 'newrelic'
+    newrelic {
+        variantMapsEnabled false
+        uploadMapsForVariant ''
+    }
+}
 //apply plugin: 'me.tatarka.retrolambda'
 //apply plugin: 'io.fabric'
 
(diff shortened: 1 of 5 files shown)


jamorham:
Can you provide a little more detail regarding the implementation of these changes? Will there be any changes to the standard build process by applying these gradle changes or does it just add a variant which is unused by default?

tolot27:
This PR does not influence the current build process. It just adds another build type (dev) to the app and flavors (fast and prod) to app and wear. To get incremental compilation (apply changes) working, one has to switch the build type of app to `fastDev' and of wear to 'fastDebug'. The first build takes as long as usual, but any further build process just takes a few seconds (2 to 15 seconds, depending on the compute power) until something changes in the manifest or major code changes. Then it takes slightly longer.

image

I still recommend to reconfigure the debug build type because it is the recommend build type for incremental changes. The release build type is completely untouched.

I also recommend to merge #1596. Then, line 153 and 154 of apps build.gradle can be removed.

jamorham:
Thanks for the description. Can you pull master and push those changes to your PR as well to confirm that it still works with the unit tests?

tolot27:
The build-time unit tests running fine locally at my computer. I assume, they will run smoothly at GitHub as well.

BTW: Incremental compilation is still not fully enabled, as long as the annotation processors are not incremental because of the old databinding compiler.

tolot27:
@jamorham Why do you need minify, shrink and Proguard in the debug build together with dynamic buildConfigFields?

jamorham:
Without minify I expect the build to be greater than >65k public methods allowed by Android 4.x and the shrinking seems the best way to deal with that issue. Because proguard can result in behavior changes I have preferred testing using the the shrunken form of the apk.

tolot27:
I've further improved compilation time (reduced from 1 min, 12 s to 47 s) by switching to Gradle 6.8. Documentation is added to the new gradle build properties.

tolot27:
BTW: Parallel execution (`--parallel`) and build cache (`--build-cache`) should be used with `gradlew` during the build unit test, as well.

jamorham:
Hi thanks, this seems to work and it looks like the unit tests are being applied to each variant flavor which is probably what we want to catch any differences between shrunk vs not shrunk code.

jamorham (approved):

PR #1601, opened 2021-01-10
AuthRequestTxMessage.java was inherently broken

dspnikder:
It was generating values which pretty much destroyed any expected randomness by converting them into a hex stirng and then getting the bytes from the chars which meant that the token size was effectively >> 4, from 8 bits per byte to 4 which was.... painful. The use of double ByteBuffers didn't make it any better either. So here, with tests, and again the Log.java to help ensure that we can write quick unit tests. There's even a test for randomness but since I can't be bothered to write two sample Kolmogorov-Smirnov tests this shows a more.... reasonable distribution ;)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,9 @@ build/
 app/release
 localeapi/gradle/
 localeapi/lint.xml
-
+gradle/native
+gradle/wrapper/dists
+gradle/caches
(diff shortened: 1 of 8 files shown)


dspnikder:
Bah broke it....

PR #1605, opened 2021-01-11
Dspnikder auth request tx message

dspnikder:
My note: the first version actually ended up having a failing unit test.

So it goes. Now it has been fixed. Always something. 

------------ The problem with AuthRequestRxMessage---------------

It was generating values which pretty much destroyed any expected randomness by converting them into a hex stirng and then getting the bytes from the chars which meant that the token size was effectively >> 4, from 8 bits per byte to 4 which was.... painful. The use of double ByteBuffers didn't make it any better either. So here, with tests, and again the Log.java to help ensure that we can write quick unit tests. There's even a test for randomness but since I can't be bothered to write two sample Kolmogorov-Smirnov tests this shows a more.... reasonable distribution ;)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,9 @@ build/
 app/release
 localeapi/gradle/
 localeapi/lint.xml
-
+gradle/native
+gradle/wrapper/dists
+gradle/caches
 
 # Local configuration file (sdk path, etc)
 local.properties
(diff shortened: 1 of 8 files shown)


jamorham:
There is already a version that uses more entropy in the packet construction which I will merge. 

We use Robolectric to mock android framework dependencies such as logging.

tolot27:
Just a small note: please write descriptive commit messages. This makes code review and cherry-picking much easier.

tolot27:
@dspnikder Please can you narrow down your PR to the documentation and UnitTest, only?

If it is worth changing the new implementation (commit d7837eb) again, please discuss this here.

PR #1609, opened 2021-01-12
replace "yes" and "no" with resource strings

### LINKED ISSUE #1608: 'no' string not translated in initial calibration dialog box
Version of xDrip: 2101081940

### Expected behavior
Code should use translated string instead of native english string.

### Actual behavior
'NO' is displayed instead of translatef string 'NON' in case of french language for current screenshots exemple

### Steps to reproduce the behavior:
1. Start New sensor menu
2. Wait for readings...
3. Dialog box will be prompted asking to calibrate

### Screenshots

Screenshot_2021-01-12-09-03-55-189_com eveningoutpost dexdrip

tolot27:
All UI-related hard-coded single word strings of "Yes" and "No" were replaced by their resource strings.

fixes #1608


jamorham (approved):
This looks sensible and like it should work.

PR #1610, opened 2021-01-15
render BG graph at small widget sizes

### LINKED ISSUE #1560: BG graph is missing in small widget
With the latest xdrip+ ("xDrip-plus-20201218-056b822.apk") I´ve got a problem.
In my xdrip+ widget:
- the white dots and 
- the two lines (for high and low glucose level) dissapeared.

Please look at the picture.
android

tolot27:
PR #1024 introduced rendering the BG graph only if the widget has a certain size (https://github.com/NightscoutFoundation/xDrip/pull/1024/files#diff-ddde3c059ce7449a35352211887a9567b61470117b683eb914daf7011366962aR115). This value was chosen, arbitrary and is too large.
This PR just reduces the value more reasonable value and fixes #1560 . It is still unclear why this cutoff was introduced.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/xDripWidget.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/xDripWidget.java
@@ -74,7 +74,7 @@ public class xDripWidget extends AppWidgetProvider {
             appWidgetManager.updateAppWidget(appWidgetId, views);
             // needed to catch RuntimeException and DeadObjectException
         } catch (Exception e) {
-            Log.e(TAG, "Got Rexception in widget update: " + e);
+            Log.e(TAG, "Got exception in widget update: " + e);
         }
     }
 
@@ -91,7 +91,7 @@ public class xDripWidget extends AppWidgetProvider {
             appWidgetManager.updateAppWidget(appWidgetId, views);
             // needed to catch RuntimeException and DeadObjectException
         } catch (Exception e) {
-            Log.e(TAG, "Got Rexception in widget update: " + e);
+            Log.e(TAG, "Got exception in widget update: " + e);
         }
     }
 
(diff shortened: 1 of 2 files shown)


jamorham:
I've been testing a different resolution to this issue and one other related to the widget. I will push these later and we can see if that covers all the bases. I agree that the UI experience has been negatively affected by the PR.

tolot27:
I've also tried to use [autoSizeText](https://developer.android.com/guide/topics/ui/look-and-feel/autosizing-textview) in the xml and programmatically to adjust the font size of the extra status line but it does not work after resizing the widget. It should work according to the docs, even down to API level 14. Autosizing in general, is a nice feature.

jamorham:
I provided a preference switch for the option so that if anyone does want to hide the graph on the widget they can do so and also adjusted where it makes the font too small with mmols. See: https://github.com/NightscoutFoundation/xDrip/commit/1d757edab8d1d47737faa82ede4d6765c308bc15

Regarding autosizing, this feels like it should be the best solution instead of where the font size is adjusted programatically, are you saying that it doesn't work properly? It would be nice to use it if we can but if its not reliable then perhaps not.

tolot27:
> I provided a preference switch for the option so that if anyone does want to hide the graph on the widget they can do so and also adjusted where it makes the font too small with mmols. See: 1d757ed

A good idea in general, but we have an overwhelming number of preferences, already. Nevertheless, it is already merged without discussion. I believe, this preference will never be used. It is just additional work for translators.

Regarding autosizing: this will work on androidx migrated builds but it did not work with the rather old support libs currently in use. I spend two ours to get it working but gave up.

tolot27:
@tolot27 Please create another PR for the typo fixes.

PR #1614, opened 2021-01-18
Dspnikder weneedmoretests

dspnikder:
We need more tests.

I am being VERY nice here. For reasons that should be obvious there are issues in the code that are bad that don't need to point out although I thing @tolot27  that we should have an open disclosure policy in due time. Right now that would not be a good idea.

So here's just some examples.


dspnikder:
Let me be fair and remove something that caused extra issues

PR #1621, opened 2021-01-25
Package name pre-processor to fix issues with databinding library

jamorham:
This PR uses a gradle pre-processor plugin to adjust package names within the source tree during compilation to use lower case.

The reason for doing this is that later versions of the Android databinding library cannot handle mixed case package names (despite this being part of previous versions test suites). Cosmetic refactoring of the code creates additional work for many unmerged feature branches and so this is intended as a pragmatic alternative.

In tests, the plugin takes less <1 second to run. Indeed tests showed it at around 80-330ms execution time. This may vary depending on the machine hardware, ssd etc.  The plugin is intelligent in terms of only rewriting changed code and so is significantly faster after the first build. 

To demonstrate that it works, this PR also raises the gradle android build tools version to `3.5.4` and sets `android.databinding.enableV2=true` although in fact only V2 is possible with this newer build tools version anyway. 

Only very limited testing at this stage has been done on the resultant app which has been tested only in as far as specimen databinding screens within the app appear to function correctly. Further testing is needed and also with the Wear microapk.

An additional question is whether this in any way negatively affects the build process or development cycle/consistency. Summed up really as, does everything else still work?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,28 +12,7 @@ buildscript {
     }
 }
 apply plugin: 'com.android.application'
-apply plugin: 'com.github.jamorham.android.replace.token.preprocessor'
-
-replaceAndroidTokenPreprocessorSettings {
-
-    // Adjust package names
-    replace 'com.eveningoutpost.dexdrip.Services': "com.eveningoutpost.dexdrip.services"
-    replace 'com.eveningoutpost.dexdrip.UtilityModels': "com.eveningoutpost.dexdrip.utilitymodels"
-    replace 'com.eveningoutpost.dexdrip.Models': "com.eveningoutpost.dexdrip.models"
-    replace 'com.eveningoutpost.dexdrip.Tables': "com.eveningoutpost.dexdrip.tables"
-    replace 'com.eveningoutpost.dexdrip.G5Model': "com.eveningoutpost.dexdrip.g5model"
-    replace 'com.eveningoutpost.dexdrip.GlucoseMeter': "com.eveningoutpost.dexdrip.glucosemeter"
-    replace 'com.eveningoutpost.dexdrip.localeTasker': "com.eveningoutpost.dexdrip.localetasker"
-    replace 'com.eveningoutpost.dexdrip.ImportedLibraries': "com.eveningoutpost.dexdrip.importedlibraries"
-
-    // For AndroidManifest shorthand format
-    replace '".UtilityModels.': '".utilitymodels.'
-    replace '".localeTasker.': '".localetasker.'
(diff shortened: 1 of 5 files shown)


app/token-replace.gradle
```diff
@@ -0,0 +1,24 @@
...
+    // Adjust package names
+    replace 'com.eveningoutpost.dexdrip.Services': "com.eveningoutpost.dexdrip.services"
+    replace 'com.eveningoutpost.dexdrip.UtilityModels': "com.eveningoutpost.dexdrip.utilitymodels"
+    replace 'com.eveningoutpost.dexdrip.Models': "com.eveningoutpost.dexdrip.models"
+    replace 'com.eveningoutpost.dexdrip.Tables': "com.eveningoutpost.dexdrip.tables"
+    replace 'com.eveningoutpost.dexdrip.G5Model': "com.eveningoutpost.dexdrip.g5model"
+    replace 'com.eveningoutpost.dexdrip.GlucoseMeter': "com.eveningoutpost.dexdrip.glucosemeter"
+    replace 'com.eveningoutpost.dexdrip.localeTasker': "com.eveningoutpost.dexdrip.localetasker"
+    replace 'com.eveningoutpost.dexdrip.ImportedLibraries': "com.eveningoutpost.dexdrip.importedlibraries"
+
```
tolot27, line 15: What about `com.eveningoutpost.dexdrip.ShareModels`?

app/build.gradle
```diff
@@ -276,11 +276,11 @@ dependencies {
 
     // use specific version of data binding until google resolves the bugs with mixed case package names
     //noinspection GradleDependency
-    implementation 'com.android.databinding:adapters:3.1.4'
+    //implementation 'com.android.databinding:adapters:3.1.4'
```
tolot27, line 279: Please remove unused code. If necessary, commits can be reverted or partially restored from git history of a file.

tolot27 (commented):
Looks good for me and compiled apks did not produce any error during the last few weeks.

build.gradle
```diff
@@ -5,12 +5,14 @@ buildscript {
...
         jcenter()
         mavenCentral()
+        maven { url "https://jitpack.io" }
 
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.3'
+        classpath 'com.android.tools.build:gradle:3.5.4'
         classpath 'com.google.gms:google-services:4.3.3'
  //       classpath 'me.tatarka:gradle-retrolambda:3.7.0'
```
tolot27, line 14: Please remove unused code.

tolot27:
@jamorham Any progress on this PR?

Issue #1647, opened 2021-02-17
xDrip confuses G5 with a G6 and performs a preemptive restart on day 9 as well as on day 6


Navid200:
### Expected behavior
Dexcom G5 can be used in Native mode.  But, since G5 sends raw data, you can still access the calibration graph.  And xDrip removes "Initial calibration" from the menu after initial calibrations have been submitted and replaces it with "Add calibration".

### Your environment
The problem was observed with the February 11, 2021 release, and February 3, 2021.

### Actual behavior
Calibrations are sent to the transmitter and are accepted by the G5 transmitter.  However, xDrip logs show they are rejected by xDrip.
![tempFileForShare_20210216-234941](https://user-images.githubusercontent.com/51497406/108206590-96d7fa80-70f4-11eb-93a3-64454447f949.jpg)

Logs show calibrations have been rejected.
No access to adding calibration in the top left menu as only initial calibration is shown there.
No access to calibration graph.
Empty calibration data table.

The symptoms match the behavior of xDrip with a device that does not send raw data like a Firefly G6,

After installing the January 29, 2021 version, the problem was resolved. 

### Steps to reproduce the behavior:
Install xDrip release February 11, 2021
Start a G5 sensor in native mode
Apply initial calibrations
Check calibration data table, or check the logs.

@viq:hackerspace.pl reported this in the Gitter room.
Another individual reported it on facebook where the calibration graph being missing and the calibration data table being empty were reported.
The calibration rejection logs were reported by both users.

jamorham:
The screenshot implies that a bloodtest was delivered from a bluetooth connected glucose meter which it believes was ~9000 seconds old. If the finger stick test was just done recently then we need to determine if the clock on the meter is correct or if there is an issue with the user's local timezone vs the clock on the meter and if that is the source of the >1 hour rejection message. It may also be (more likely?) that there were multiple tests synced from the meter and that only one of those was used and the other rejected which is why there is a report of success and also rejection in the screenshot.

We also want to know the transmitter firmware version to determine the issue regarding raw.

Navid200:
You are talking to him in the Gitter room.
The firmware version is 1.0.4.15.

jamorham:
Please try the 17th Feb nightly to see if it resolves the issue.

viq:
[...]

PR #1660, opened 2021-02-28
Update FirmwareCapability.java

Navid200:
108576647-5cd14900-72ec-11eb-8ed1-3f2a483af37d
Added 1.0.4.14 to the G5 firmware list.


tolot27 (approved):
Looks good so far. This firmware was first seen at solving #1647.

@jamorham Wouldn't it be better to store the lowest and highest firmware version number for a particular sensor/feature set and accept all other versions in between instead of adding every version that we see at sometime?

jamorham (approved):

PR #1662, opened 2021-03-01
Add a new command 0x08

bubbledevteam:
Add a new command 0x08(implemented in Bubble FW v2.6 and upper) which request decrypted Libre2/US 14days raw data from Bubble, the data Structure is the same as Libre1,make xdrip+ algorithm work with all Libre type sensors(Libre Pro is not included yet).

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/Bubble.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/Bubble.java
@@ -31,7 +31,7 @@ public class Bubble {
         if (activeBluetoothDevice == null || activeBluetoothDevice.name == null) {
             return false;
         }
-        return activeBluetoothDevice.name.contains("Bubble");
+        return activeBluetoothDevice.name.startsWith("Bubble");
     }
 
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Models/Bubble.java
```diff
@@ -31,14 +31,25 @@ public static boolean isBubble() {
         if (activeBluetoothDevice == null || activeBluetoothDevice.name == null) {
             return false;
         }
-        return activeBluetoothDevice.name.contentEquals("Bubble");
+        return activeBluetoothDevice.name.contains("Bubble");
```
tolot27, line 34: Maybe, `startsWith` is more appropriate here.

tolot27 (approved):
Looks good to me.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Bubble.java
parapenT1sta: Thank you!

jamorham (approved):
This looks okay to me, but I have no knowledge of the hardware changes it addresses. I can only assume that it has been properly tested.

PR #1668, opened 2021-03-11
fix: use correct speech language strings if different from display

k2s:
This fixes situation if you for example use english language on system, also for display and then set in preferences `Speak Readings / Custom Locale for Spoken Readings` another value. For example `sk`.

You still need to disable/enable speaking to activate the custom language.
From this moment the whole sentence will be in `sk` language.
Without this fix the numbers are spoken correctly, but for other strings (delta name and some others) translation of the display language is used.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/SpeechUtil.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/SpeechUtil.java
@@ -3,6 +3,8 @@ package com.eveningoutpost.dexdrip.UtilityModels;
 import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
+import android.content.res.Configuration;
+import android.content.res.Resources;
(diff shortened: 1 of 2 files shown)


Navid200:
What happens, in what we have now, if the user just restarts the phone?
Is that not enough to sort everything out?
There are several settings that require a restart of xDrip to take effect.  I don't think we need to fix all of those.

Navid200:
Why is there a need for choosing a language other than the language you have chosen as your main language?
I think I understand what this PR intends to allow.  But, I don't understand why it would ever be needed.
Can you provide an example?

Let's say someone chooses to change the display language from the default English to say Spanish as an example.  I assume that happens because that's the language the user prefers to use.
Now, why would the same user choose Russian as the spoken language?

My understanding of why we have the different choices for the spoken language has been the accent or the tone of voice.  But, all of the very same chosen language.
Shouldn't that be the only intent of the option?
In that case, the issue this PR is trying to solve does not exist, does it?

PR #1671, opened 2021-03-13
change: git ignore signing.properties

k2s:
it is very dangerous to commit this file


tolot27:
Where does this file come? I don't have it on my system. Adding all untracked files is not a good idea, IMHO. Hence, how would it be added to the repository?

Navid200:
@k2s 
Would you please explain why we need this PR?
The only thing you have posted is "it is very dangerous to commit this file".
That is not enough for explaining why we need to make a change.

PR #1674, opened 2021-03-14
fix: make it possible to translate all xdrip preferences

k2s:


app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -2792,7 +2792,7 @@ private void displayCurrentInfo() {
         if (sensor_age > 0 && (DexCollectionType.hasLibre() || hasLibreblock())) {
             final String age_problem = (Pref.getBooleanDefaultFalse("nfc_age_problem") ? " \u26A0\u26A0\u26A0" : "");
             if (Pref.getBoolean("nfc_show_age", true)) {
-                sensorAge.setText(getResources().getQuantityString(R.plurals.sensor_age, sensor_age, JoH.qs(((double) sensor_age) / 1440, 1)) + age_problem);
+                sensorAge.setText(getResources().getQuantityString(R.plurals.sensor_age, sensor_age / 1440, JoH.qs(((double) sensor_age) / 1440, 1)) + age_problem);
```
tolot27, line 2795: Did you test this expression?

tolot27:
@k2s Please can you add a description to this PR?

app/src/main/res/values/strings.xml
```diff
@@ -1649,11 +1649,11 @@
...
     <plurals name="sensor_expires">
-        <item quantity="one">Expires: %dd</item>
-        <item quantity="other">Expires: %dd</item>
+        <item quantity="one">Expires: %sd</item>
+        <item quantity="other">Expires: %sd</item>
     </plurals>
     <plurals name="sensor_age">
-        <item quantity="one">Age: %dd</item>
-        <item quantity="other">Age: %dd</item>
+        <item quantity="one">Age: %sd</item>
```
tolot27, line 1656: Please can you link this format specifiers? What's the difference between `%dd` and `%sd`? I've never seen these double char specfiers.

Navid200:
@k2s Would you please answer the questions?

Navid200:
Can we close this PR please?  If/when the developer has time, we can always reopen it.

tolot27:
> Can we close this PR please? If/when the developer has time, we can always reopen it.

I took this PR over and will finalize it. It is a reasonable improvement.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
tolot27: The int division is necessary for `getQuantityString` to select the correct plural.

tolot27 (approved):
LGTM and is one step further to get more strings translated.

jamorham (approved):
No description provided and this does more than change preferences but it looks okay from what I can see and tolot27 has had a good look at it also. The sensor age should be verified in practice to see if it works correctly.

PR #1679, opened 2021-03-20
add xDrip+ Sync Follower to source setup wizard

### LINKED ISSUE #1027: Feature Request: add xDrip+ Sync Follower to SourceWizard on first page
It is not possible to select `xDrip+ Sync Follower` in data source wizard. It is not on first page and it is not nested in other.

I suggest to add it to the fist wizard page, under `Others`.

Selecting this item should set xDrip hardware source to `xDrip+ Sync Follower` and start `ActivityDisplayQrcodeBinding`.

k2s:
fixes #1027

It adds `xDrip+ Sync Follower` source to source setup wizard. It starts qrcode scanner same way as `Auto configure` does.


tolot27:
Please can you move it below "Other"? I don't see any reason why it should be displayed at the first view but "Nightscout Follower" not. Hence, below "Other" is more suitable.

Navid200:
What does this address?
We can use xDrip as a Sync follower already.  Can we not?

jwoglom:
Just wondering, do we have any telemetry around which data sources are used most often in xDrip? Ideally this setup screen would be built out for all of the most common data sources. Perhaps @jamorham could give an assessment, as there is "send feedback to developer" functionality in the app which might include this data.

I could imagine this PR, as simple as it is, potentially making it significantly easier for a new, follower user of xDrip to quickly get set up following a primary xDrip user, especially if they're not familiar with xDrip.

Navid200:
There are already too many settings in xDrip.
Every new setting that is added has the danger of being impossible to remove later because of the resistance of the development team against making changes that could confuse users.

I will object to every new settings that is NOT absolutely necessary for that reason.

We need to reduce the number of settings rather than increasing it.

tolot27:
@Navid200 That's not a new setting. It is just made visible at the right place and should improve users first experience with xDrip because the most common data sources are shown on setup. @jwoglom asked the right question, IMHO.

tolot27:
I suspect either "Nightscout Follower" or "xDrip+ Sync Follower" is the most common data source beside Dexcom and Libre. Both of them should be listed at least below "Other" and the most frequent one at the main source list.

Issue #1351, opened 2020-05-25
About alerts


Air69:
Can the rising（Falling） threshold be set to be adjustable?  Fixed 0.1mmol / l and 0.16mmol / l, too few options.
Or add 0.2, 0.25, 0.3, etc.  Thank you for your development, this project has benefited me a lot.@bigdigital

bigdigital:
The alert thresholds can be adjusted directly in the xdrip settings. Just go to Settings-> Alarm and alerts -> Glucose level alert list and configure thresholds for low and high alerts.

Air69:
> 警报阈值可以直接在xdrip设置中进行调整。只需转到“设置”->“警报和警报”->“葡萄糖水平警报”列表，然后为低警报和高警报配置阈值。

Hello there.  I may not be able to express clearly. What I am saying is that in other alarms, the rate of rise of blood glucose and the rate of decrease of blood glucose.

Navid200:
Duplicate:  https://github.com/NightscoutFoundation/xDrip/issues/907#issuecomment-778728153

Navid200:
A lower threshold, of 1, has been added. You can test it by updating to the latest Nightly build.
https://github.com/NightscoutFoundation/xDrip/releases/tag/2021.04.06

Please report here:(https://github.com/NightscoutFoundation/xDrip/issues/907) if you discover any unexpected behavior.
Thanks

Issue #907, opened 2019-06-11
Falling/Raising BG Falling Threshold Adjustment


Seithx:
Hello,

Current available threshold for Xdrip to start alerting, is either a raise or fall of 2 or 3 mgdl per reading.

Personally, I don't get any of the alerts, because the way my blood sugar raises after a meal, if I missed a shot or did not inject enough insulin for example, is lower than the threshold and I have no way of changing it.

So what happens is that my blood sugar raises steadily (for example), and I don't get any alerts until it reaches to what I defined as a high blood sugar level. 

I'm a low carber, so I don't have huge jumps, but steady raises in blood sugar, or steady decreases in blood sugar OVERTIME.

So if we could set it ourselves, or at least get a lower threshold, it will make Xdrip much much more usable for us low carbers. either 0.5 or 0.7 mgdl threshold, will improve a lot of users experience I reckon.

Seithx:
Any comments regarding this?

Navid200:
You can specify as many alerts for as many different levels as you wish.
You are correct that you cannot change the fast falling or fast rising alarm values other than choosing one of the two available values for each.

I am looking at the scenario you have described and and I don't understand why you cannot use additional levels to address your concern.

Let's say your high alarm is set at 7.0.  I am just picking a number here.  It can be any other value.  But, I am going to try to make a point.
You say that you get no alarm until your glucose reaches 7.0.  Why?
You can add another alarm set at 6.0 and another at 6.5.   Now, you will get 2 alarms before your blood glucose reaches 7.0.  You can add as many high alarms as you want at any value you like.

Can you please explain why you cannot do that?

Seithx:
In theory, I can.

I also recognize its a good workaround, instead of adding extra work, though, without of course undermining what youve done so far, it sounds like it should be easy to adjust to accommodate for low carb users, needing this, since this feature is already existent. 

But, its not the same and that feature is completely unusable for me.

Also, think about the amount of false alerts I would go, since sometimes my sugar is just comfy and stable at 80 mg, so, on a day like that, I`d have to shut my alerts off for the whole day.

Or at all, since you know, blood sugar is dynamic to an extent, even for a strict low carber.

So lets say I set 80 as my HIGH alert. You say, alert when 70, alert when 65.... I`d have constant alerts. Sometimes its like, 75, 72, 80, 79.... within an hour or two....

Enabling the feature would make a whole big diff.

Navid200:
[...]

PR #1691, opened 2021-04-01
Update arrays.xml to add a lower threshold for the fast rising and fast falling alarms

Navid200:
Screenshot_20210331-223319
Issues addressed: https://github.com/NightscoutFoundation/xDrip/issues/907
https://github.com/NightscoutFoundation/xDrip/issues/1351

This makes it easier for those who use a low-carb diet (not me) to use the fast rising (and fast falling) alarm for their needs.
It is a very old issue.  And seems like a legitimate request to me.

It is quite possible that I have overlooked requirements for a pull request.  Thanks for any feedback in advance.

### DIFF
--- a/app/src/main/res/values/arrays.xml
+++ b/app/src/main/res/values/arrays.xml
@@ -152,11 +152,13 @@
     </string-array>
 
     <string-array name="risingEntries">
-        <item>2 mgdl (0.1 mmol)</item>
-        <item>3 mgdl (0.16 mmol)</item>
+        <item>1 mgdl (0.06 mmol)</item>
+        <item>2 mgdl (0.11 mmol)</item>
+        <item>3 mgdl (0.17 mmol)</item>
     </string-array>
 
     <string-array name="risingValues">
+        <item>1</item>
         <item>2</item>
         <item>3</item>
     </string-array>
--- a/app/src/main/res/values/arrays.xml
+++ b/app/src/main/res/values/arrays.xml
@@ -152,9 +152,9 @@
     </string-array>
 
     <string-array name="risingEntries">
-        <item>1 mgdl (0.06 mmol)</item>
-        <item>2 mgdl (0.11 mmol)</item>
-        <item>3 mgdl (0.17 mmol)</item>
+        <item>1 mg/dl (0.06 mmol/l)</item>
+        <item>2 mg/dl (0.11 mmol/l)</item>
+        <item>3 mg/dl (0.17 mmol/l)</item>
     </string-array>
 
     <string-array name="risingValues">


tolot27 (approved):
Looks good and is ready to merge.

Navid200:
Thanks.
I wanted to change the units too.  But, thought I should minimize the number of changes, to increase the chances of being merged.
Thanks

tzachi-dar (approved):
LGTM,
If you did not do it, please make sure that for people with existing alert nothing changes.

jamorham (approved):
Looks good, thanks

Navid200:
> LGTM,
> If you did not do it, please make sure that for people with existing alert nothing changes.

I don't have a simulator.
I'm really new as a contributor.
I set up my old phone as a Nightscout follower and set it to the new low threshold.  I set my main phone (the master) to the next level up.  I confirmed that the two phones go off at different times at different slopes.  
This way I confirmed that the new setting was not just imitating the next level up threshold.

But, I did something really dumb.  I only had my branch installed on my old phone!
So, I am not going to repeat this experiment this time installing my branch on both phones.
I will report back soon.

Navid200:
I decided not to install my branch on my main phone because I don't know how to do it other than uninstalling first, which would lose all my settings.

I have now set both phones (main phone with March 30, 2021) to the same setting of 2.  This will confirm that my branch matches the old behavior at setting 2.

After that is done, I will set both phones to 3.  I guess then I will have to have something really sweet to confirm that my branch also behaves as it did before on 3.

Navid200:
Set phone on March 30, 2021 to 2.  Set the phone on my branch to 1.  Pass
2 triggered one BG reading behind 1 (a higher rate since the rate was increasing).

Set both phones (one on my branch and the other on March 30, 2021) to 2.  Pass  
Both triggered on the same BG point.

Set the phone with March 30, 2021 to 2.  Set the phone on my branch to 3.  Pass
3 triggered one BG point behind 2 (a higher rate as the rate was increasing).

Navid200:
@tzachi-dar If this is merged, I will mention it in facebook making sure that everyone knows about the change.  So, we will find out right away if I have missed anything in the testing.

Assuming everything goes smoothly, the next on my plan is to add override silent mode and sound file choice to the missed reading alert.  That would close two old open issues.

Would you have any objections to that?
If not, would you have any advice?

tzachi-dar:
Adding override silent mode and sound file choice to the missed reading alert is a good thing.

I'm not sure how simple it is but if that can be made generic it would be nice to add this to all 'other' alerts. If it is complicated, only add them to the missed reading alert.

Overriding silent mode might be hard and depend on the specific android maker and version. Try to use the same method that is used for the general alerts in order to have a consistent solution working.

jamorham:
With engineering mode enabled it is possible to use the voice command `enable fake data source` which create a sine wave of raw values which can be useful for testing in certain situations.

Issue #1309, opened 2020-04-17
Feature Request: UI Move "Stop Sensor" to something more inconvenient


wgcj:
The "stop sensor" button is prominently placed on the main menu. User error can cause the "stop sensor" function and "snooze alert" function to be confused -- especially by someone with low blood sugar at 3am who just wants to make the noise stop and clicks through.

This is definitely a PEBKAC issue. But, the risk of clicking incorrectly is ruining an expensive sensor early and that the target audience sometimes experiences mental difficulties (low blood sugar). Additionally, if someone is trying to temporarily take over treatment or just mute the alarm (like a pregnant T1 in labor, or a T1 being hospitalized), the urge to "just 'stop' the alert" seems to pull people towards "stop sensor" more than "snooze".

I think it would be helpful for "stop sensor" to be more inconvenient than "snooze alert" (which currently has the same amount of clicks, even with the "are you sure?" screen).

Possible solutions would be to move "snooze" higher up on the menu, to hide "stop" behind the "settings" or "system status" menu, or to add a more obnoxious "are you sure?" screen that might require more steps (for example, having to type "OFF" in a text box).

thehappyhippo:
I vote for a relocation of Stop Sensor to a Systemstatus tab.

adanvdo:
you have to confirm a stop.

Navid200:
You should be able to snooze an xDrip alert from the corresponding Android notification.  You don't even need to open xDrip to do that.

Navid200:
Closing due to inactivity

PR #1696, opened 2021-04-15
Change Stop Sensor page

Navid200:
There are cases where a user accidentally stops their sensor.
#1309
This comes up occasionally on facebook as well.
The user intends to go to the system status page or another page, but, accidentally taps on "Stop Sensor" and end up on the stop sensor page and stop sensor.
I can see how annoying it can be if I made such a mistake.

There are requests for the stop sensor button, on the main menu, to be relocated.
I believe that is an extreme change.

Instead of moving the button, I have edited the stop sensor page to make it stand out so that the user hopefully can realize that they are on a page other than the one they intended.
Currently the stop sensor page looks like this:

I have changed it to look this way:

I have intentionally picked ugly colors, different from the xDrip norm colors.
I have intentionally written a long story and I have added warning at the top.
The intent is to give visual cues to a user who may be half sleep or hypoglycemic or riding a bicycle etc. that they are on the wrong page.


tzachi-dar:
I think this is too frightening for the normal user.
A sensor can be stopped and started again. This solves some problem.
The wording might need to reflect that this is not that dangerous.

Navid200:
@tzachi-dar Thanks for the feedback.  Understood.

Would it be better if I add the following paragraph after?

However, if you indeed intend to stop sensor or clear calibrations, you are on the right page.

Navid200:
@tzachi-dar Is there a way to stop and start a sensor without losing all the existing calibrations?

Navid200:
I think we can even go back to the original text if you prefer as long as we keep the colors.  Please see this image.




Again, I hope the colors will provide a visual cue to someone who is not reading the text.

Would this be acceptable?

tzachi-dar:
This makes more sense now.
Please note that stopping a sensor is meant to clear all calibrations, in order to have a new start. No reason to keep them.

tzachi-dar:
By the way, this PR only changes two files, any reason why it has 96 commits?
Maybe recreate it with relevant files only?

Navid200:
It's because I am trying to find my way around Github!  I am sorry abut that.
I will close this and open a new pull request.

PR #1697, opened 2021-04-15
Change Stop Sensor page (second attempt)

Navid200:
I closed my previous pull request because I am still learning pull requests and everything else with github:
https://github.com/NightscoutFoundation/xDrip/pull/1696

There are cases where a user accidentally stops their sensor.
#1309
This comes up occasionally on facebook as well.
The user intends to go to the system status page or another page, but, accidentally taps on "Stop Sensor" and ends up on the stop sensor page and stops sensor.
I can see how annoying it can be if I made such a mistake.

There are requests for the stop sensor button, on the main menu, to be relocated.
I believe that is an extreme change.

Instead of moving the button, I have edited the stop sensor page to make it stand out so that the user hopefully can realize that they are on a page other than the one they intended.
Currently the stop sensor page looks like this:

I have changed it to look this way:

I have intentionally picked ugly colors, different from the xDrip norm colors.
I have not changed anything else.
The intent is to give visual cues to a user who may be half sleep or hypoglycemic or riding a bicycle etc. that they are on the wrong page.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/StopSensor.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/StopSensor.java
@@ -48,12 +48,33 @@ public class StopSensor extends ActivityWithMenu {
 
         button.setOnClickListener(new View.OnClickListener() {
             public void onClick(View v) {
-                stop();
-                JoH.startActivity(Home.class);
-                finish();
+                confirm_stop();
             }
 
         });
+    }
+
+    public void confirm_stop() {
+        final AlertDialog.Builder build_stop = new AlertDialog.Builder(this);
+        build_stop.setTitle(gs(R.string.are_you_sure));
+        build_stop.setMessage(gs(R.string.do_you_want_to_end_session));
+        build_stop.setNegativeButton(gs(R.string.no), new DialogInterface.OnClickListener() {
+            @Override
+            public void onClick(DialogInterface dialog, int which) {
+                dialog.dismiss();
+                JoH.startActivity(Home.class);
+            }
+        });
+        build_stop.setPositiveButton(gs(R.string.yes), new DialogInterface.OnClickListener() {
+            @Override
+            public void onClick(DialogInterface dialog, int which) {
+                stop();
+                dialog.dismiss();
+                JoH.startActivity(Home.class);
+                finish();
+            }
+        });
+        build_stop.create().show();
     }
 
     public synchronized static void stop() {
--- a/app/src/main/res/layout/activity_stop_sensor.xml
+++ b/app/src/main/res/layout/activity_stop_sensor.xml
@@ -8,7 +8,6 @@
     android:orientation="vertical" android:layout_width="match_parent"
     android:layout_height="match_parent"
     android:weightSum="1"
-    android:background="#e8e800"
     android:id="@+id/stop_sensor_container">
     <ScrollView
         android:layout_width="fill_parent"
@@ -25,7 +24,6 @@
                 android:textAppearance="?android:attr/textAppearanceLarge"
                 android:text="@string/stop_sensor"
                 android:id="@+id/textView3"
-                android:textColor="#0000ff"
                 android:layout_gravity="center_horizontal" />
 
             <TextView
@@ -40,7 +38,6 @@
                 android:layout_gravity="right"
                 android:paddingLeft="20dp"
                 android:paddingRight="20dp"
-                android:textColor="#0000ff"
                 android:gravity="center_vertical|center_horizontal" />
 
             <Button
--- a/app/src/main/res/layout/activity_stop_sensor.xml
+++ b/app/src/main/res/layout/activity_stop_sensor.xml
@@ -40,7 +40,7 @@
                 android:layout_gravity="right"
                 android:paddingLeft="20dp"
                 android:paddingRight="20dp"
-                android:textColor="#0000ee"
+                android:textColor="#0000ff"
                 android:gravity="center_vertical|center_horizontal" />
 
             <Button
--- a/app/src/main/res/layout/activity_stop_sensor.xml
+++ b/app/src/main/res/layout/activity_stop_sensor.xml
@@ -8,6 +8,7 @@
     android:orientation="vertical" android:layout_width="match_parent"
     android:layout_height="match_parent"
     android:weightSum="1"
+    android:background="#e8e800"
     android:id="@+id/stop_sensor_container">
     <ScrollView
         android:layout_width="fill_parent"
@@ -17,7 +18,6 @@
             android:orientation="vertical"
             android:layout_width="fill_parent"
             android:layout_height="fill_parent"
-            android:background="#e8e800"
             android:gravity="center_horizontal">
             <TextView
                 android:layout_width="wrap_content"
@@ -25,7 +25,7 @@
                 android:textAppearance="?android:attr/textAppearanceLarge"
                 android:text="@string/stop_sensor"
                 android:id="@+id/textView3"
-                android:textColor="#dd0066"
+                android:textColor="#0000ff"
                 android:layout_gravity="center_horizontal" />
 
             <TextView
@@ -35,12 +35,12 @@
                 android:text="@string/only_stop_your_sensor_when_you_actually_plan"
                 android:layout_below="@+id/textView"
                 android:layout_alignParentStart="true"
-                android:layout_marginTop="60dp"
+                android:layout_marginTop="25dp"
                 android:id="@+id/stop_sensor_instructions"
                 android:layout_gravity="right"
                 android:paddingLeft="20dp"
                 android:paddingRight="20dp"
-                android:textColor="#dd0066"
+                android:textColor="#0000ee"
                 android:gravity="center_vertical|center_horizontal" />
 
             <Button
@@ -52,7 +52,7 @@
                 android:layout_alignParentBottom="true"
                 android:layout_alignEnd="@+id/textView3"
                 android:layout_gravity="center_horizontal"
-                android:layout_marginTop="150dp" />
+                android:layout_marginTop="30dp" />
 
             <Button
                 android:text="@string/dont_stop_just_reset_all_calibrations"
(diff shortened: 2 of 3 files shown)


jamorham:
Why don't I add a confirmation dialog to confirm the button press for stop sensor instead? Would that be better than uglifying the user interface?

Navid200:
@jamorham There is already a confirmation page that comes up to ask user if they really want to stop.

People who have complained about this have either been experiencing low sugar resulting in not paying attention to the text, or were riding a bicycle, not paying attention to the text.

My intent was to provide a unique set of colors that never show when pressing the adjacent buttons.  So, the unexpected colors, I hope, will be noticed even if the person is not paying attention to the text.

If you prefer to add an additional confirmation dialog, I say it will be OK as long as the colors on it are different than the usual xDrip colors and stand out.

tolot27:
There is no need for a second confirmation dialog. The different color should be sufficient.
IMHO, I don’t like the Sensor stop entry in the main menu. It is visible even if the data source has no sensor at all (NS). But this is discussed elsewhere.

Navid200:
@jamorham I don't intend make it ugly.  I intend to make it look different than the rest of the xDrip pages.  
Would you please tell me what you mean by ugly?
I can come up with a slew of options for you to choose from.  
As long as it provides a visual cue to someone who is not capable of seeing the text, it does the job.  It doesn't have to be ugly.  It just has to be obviously different.

tolot27 (changes requested):

Navid200:
@tolot27 Thanks for the review.
Thanks for the suggestion.  Quite a valid concern.

I moved the buttons to exactly where they were originally (as is in the current xDrip).  So, the buttons will be exactly where they are on a small screen as well now.
I changed the red color to blue as blue on yellow provides max contrast for visibility.
Screenshot_20210512-180016

app/src/main/res/layout/activity_stop_sensor.xml
```diff
@@ -49,7 +52,7 @@
                 android:layout_alignParentBottom="true"
                 android:layout_alignEnd="@+id/textView3"
                 android:layout_gravity="center_horizontal"
-                android:layout_marginTop="30dp" />
+                android:layout_marginTop="150dp" />
```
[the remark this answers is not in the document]
Navid200: The buttons are now back to where they were originally.

tolot27:
The color yellow should be save even for most of the color-blindness cases. Just one additional thought: What happens if you switch to the night/dark mode?

tolot27 (dismissed):

Navid200:
Night/dark mode?  
Is that an xDrip setting?!
Or, do you mean Android?  Thanks for the wonderful questions.

tolot27:
I mean the Android Darke Theme (display settings). Likely, it is not affected but we should keep an eye on it if we ever move to API level 29 and support the OS theme.

jamorham:
I don't think there is already a confirmation dialog on `Stop Sensor` only on `Reset All Calibrations` - I would propose to add one there.

tolot27:
> I don't think there is already a confirmation dialog on `Stop Sensor` only on `Reset All Calibrations` - I would propose to add one there.

You are right, there is no confirmation dialog. @Navid200 Please can you implement it instead of the yellow colored background?

Navid200:
Undid the color change by reverting back to the original file.

Added a confirmation dialog to  stop sensor.


Screenshot_20210515-120958

app/src/main/java/com/eveningoutpost/dexdrip/StopSensor.java
```diff
@@ -48,12 +48,33 @@ public void addListenerOnButton() {
...
+                confirm_stop();
+            }
+
+        });
+    }
+
+    public void confirm_stop() {
+        final AlertDialog.Builder build_stop = new AlertDialog.Builder(this);
+        build_stop.setTitle(gs(R.string.are_you_sure));
+        build_stop.setMessage(gs(R.string.do_you_want_to_end_session));
```
tolot27, line 60: I suggest naming the resource string `confirm_sensor_stop`.

tolot27: Or `sensor_confirm_stop`. A naming scheme for resource strings would be helpful for translators to understand the context.

tolot27 (changes requested):
Thanks @Navid200 !

tolot27:
> Added a confirmation dialog to stop sensor.

Maybe we can show the confirmation dialog in a different color? @jamorham ?

jamorham:
I don't think it is necessary to use anything other than the standard confirmation dialog. I did already implement that, but the exact wording of the dialog can perhaps be improved to explain that the user should really only use the option when they are going to remove the sensor.

Navid200:
@jamorham The user may not want to remove the sensor.  In case of restarting a G6, we use the same sensor.
Should I change the resource string as suggested?  Or, do you want to use what you have done?

Navid200:
@tolot27 The only new resource string in this PR is the following.
do_you_want_to_end_session

confirm_stop is not a resource string.  Am I misunderstanding you?  Please explain.

tolot27:
The confirmation dialog was implemented today by jamorham with 84e4376. Unfortunately, no resource string was introduced for the confirmation text: https://github.com/NightscoutFoundation/xDrip/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/StopSensor.java#L52

app/src/main/java/com/eveningoutpost/dexdrip/StopSensor.java
tolot27: > @tolot27 The only new resource string in this PR is the following.
> do_you_want_to_end_session
> 
> confirm_stop is not a resource string. Am I misunderstanding you? Please explain.

I thought it was obvious which resource string I meant because I created the comment at line 60 and only one new resource string was introduced. Try to improve this in the future.

Yes, I suggest to rename `do_you_want_to_end_session` into `confirm_sensor_stop` or better into `sensor_confirm_stop` because the previous name is out of context. Meanwhile, this PR should just address the missing resource string in 84e4376.

Navid200: Your suggestion reads: I suggest naming the resource string confirm_sensor_stop.  And you have linked to the existing confirm_stop() in StopSensor.java.

Is confirm_stop() is a resource string?
I thought resource strings are in strings.xml.

I want to understand what you are suggesting so that I can do it right in the future.  Asking me to do it right in the future is pointless and humiliating for no reason.

Navid200:
Was implemented mush nicer by jamorham.  There is no need for this PR.

app/src/main/java/com/eveningoutpost/dexdrip/StopSensor.java
Navid200: The resource string for confirming clear calibration is:
do_you_want_to_delete_and_reset_the_calibrations_for_this_sensor

I set the one for asking for confirmation to stop to the following based on that.
do_you_want_to_end_session

Because I thought similar to the calibration, it reflects the sentence that should be translated.

Do you think that the existing resource string for calibration is also bad?
If not, I don't see how confirm_sensor_stop represents the sentence that should be translated.

You can point me to documentation that explains what you need.
Or, you can tell me what you need.  In that case, you should be open to questions coming from me.
If you don't have time, there is no rush.  Your feedbcak is always welcome.

Saying do better next time is not useful.

tolot27: `confirm_stop()` is a Java function you have added. But you also added the line `build_stop.setMessage(gs(R.string.do_you_want_to_end_session));`. With `gs(R.string.do_you_want_to_end_session)` the translated resource string with name `do_you_want_to_end_session` is retrieved from the AppContext. As you can see, every resource string is prefixed with `R.string.`.

> I thought resource strings are in strings.xml.

You have to specify it in strings.xml and you reference it where you need it. It does not matter where you rename the string (in Java or xml file) as long as you are using code refactoring. This ensures, it is renamed at every place of use.

tolot27:
> Was implemented mush nicer by jamorham. There is no need for this PR.

But he missed the resource string for translation. 😞

app/src/main/java/com/eveningoutpost/dexdrip/StopSensor.java
tolot27: > Saying do better next time is not useful.

Where did I say this? I tried to explain my suggestion with "A naming scheme for resource strings would be helpful for translators to understand the context."

Navid200: I followed the existing resource strings, as examples, to figure out what should be acceptable.  
If what I have done is not good, I welcome directions.  
For example, you can say, the resource string should match the entire sentence.  Or, you can say, the resource string must have exactly 12 words, or, the resource string must ...

I still don't know what is wrong with my resource string.
And, I have no choice but to follow the existing examples unless you give me clear directions.

Navid200: Screenshot 2021-05-16 155410

Navid200: Screenshot 2021-05-16 155410

tolot27: The resource string should not repeat the sentence which is shown (without cutting some words). A resource string should contain the context, i. e. my suggestion `confirm_sensor_stop` or `sensor_confirm_stop`. It is short, it is like a label, it does not repeat the text to translate, it contains the context "sensor", a verb "confirm" and an "action". The suggested `sensor_confirm_stop` is according some (currently un-) defined nomenclature containing the class name (`Sensor`, at least partially) and method of use (`confirm_stop()`). Hence, the context (of use) is obvious and the string is meaningful for translators, as the Java function is.
My suggestion is to be precise and provide context.

Navid200: I follow that.  
So, the existing resource string for the calibration confirmation also does not follow your requirement, right?
If that's correct, I understand.

Issue #553, opened 2018-08-18
Speak Readings>"Or per number of minutes" text is too large


dave1812:
doesn't match other text on that page and therefore I can't see the current setting (number of minutes).  That line and the line of text above it is out of proportion to all other text.

If you fix THIS issue, but don't fix the issue I posted about 5-6 days ago, I won't be able to use the fixed version as I won't get data on my Gear 2.

jamorham:
Can you provide a screenshot to illustrate the problem? Thanks.

dave1812:
Here u go...

dave1812:
I sent the screenshot via email but I don't see it here on get hub. I hope you have it somewhere. Please let me know if you don't. Thank you

jamorham:
You need to drop it in on the github website

dave1812:
![screenshot_20180819-062740_xdrip](https://user-images.githubusercontent.com/37376075/44314492-b7615580-a3ce-11e8-952c-0f3191b4655e.jpg)

dave1812:
See how large the font is for those first few items?  Way larger than the following items, and therefore I can't see the current number of minutes that I've selected.

dave1812:
16 days and not a peep from anyone on the dev team??   Hello??

dave1812:
ah, thanks for the explanation.  if the text size is fixed in a daily release, will anyone be updating this thread?

MasterPlexus:
Normally yes.

yosepito:
@jamorham & colleagues
An additional feature of this section: I use xdrip while biking. To me, is quite important to know how much it raises/decreases, and not only the label of the trend (slight up/down...). The possibility that one can select between xdrip speaking aloud the number of the delta or the text label would be a nice improvement. I don't know how to do this in the code. Anyone else see this feature as interesting for having a "normal" life?
Sorry, but I don't know how to use the "pull request" feature of this site
thanks!

increddibelly:
this looks like something I could to to get into the xDrip code. I can confirm the symptoms on my phone, that's something. *forked*

MasterPlexus:
[...]

PR #1702, opened 2021-04-21
Shorter lines on the "Speak Readings" page

Navid200:
| Subject                      | Detail |
|------------------------|----------------------------------------------------------------------------------------------------------|
| Why do we need this? | So that small devices can show the number of minutes <br/> See the images after the table.  <br/>  fixes: https://github.com/NightscoutFoundation/xDrip/issues/553 |
| Is there a work-around? | No |
| Are there any side-effects to this fix? | The title text now will appear in the summary with a smaller font | 
|Tests | This has been tested, on a small device, using the Android studio virtual devices. |

The following shows the current release on a screen with a resolution of 240X400.
Before

After the fix, this is what the same page will look like on the same device:
115640297-9de1db80-a2e4-11eb-9fce-e054a914202f


tolot27:
This PR likely fixes it for English only. I suggest to find a general solution, i.e. by allowing text-wrapping.

Navid200:
Even if we fix the problem for English only first, that is a step forward as opposed to the current situation.   For a 3-year old issue, I would consider that a step forward.

I agree with you that there may be other solutions.  I strongly disagree that we should wait for perfection.  I prefer to offer users a solution even if it is not perfect.

I don't see why other languages cannot find alternative phrases, like I did for English.

tolot27:
> Even if we fix the problem for English only first, that is a step forward as opposed to the current situation. For a 3-year old issue, I would consider that a step forward.

Problem here is that the change of the English source strings invalidates all corresponding translations and none of the translators know why and that the should sorten the translation.

> I agree with you that there may be other solutions. I strongly disagree that we should wait for perfection. I prefer to offer users a solution even if it is not perfect.

IMHO, it has nothing todo with perfection. You provided a workaround, not a solution. A solution would be to either wrap the text automatically or shrink the size automatically. Both are simple solutions. Please can you propose either of these solutions? In Android Studio you can easily switch the display language and screen sizes for testing.
 
> I don't see why other languages cannot find alternative phrases, like I did for English.

That's sometimes impossible because English text is about 15% shorter than German (for instance) in terms of string length. Other translations are much longer than English even if the the shortest translations are used.
Also, nobody can test it because a screen size of 240X400 is very rare.

Navid200:
I spent a few days on this trying to figure out a way to reduce the font size.  I was not able to.
 jamorham stated, in the original issue, that he ran into trouble when he attempted to solve the problem.

The suggested work-around solves the problem for English.  It does not solve the problem for all languages.
Therefore, it is an improvement.  The original issue was raised by someone who used the English language.

If this is merged, no one will be stopping anyone from finding a way to reduce the font size or make it wrap.  So, a better solution can be found later.  
But, we have waited 3 years for that.  I am happy for it whenever it comes.  But, please let's not object to a work-around.

Navid200:
> Also, nobody can test it because a screen size of 240X400 is very rare.

I just used the virtual device in Android Studio to test a size of 240X400.    Why wouldn't everyone be able to do the same?

tolot27:
> I just used the virtual device in Android Studio to test a size of 240X400. Why wouldn't everyone be able to do the same?

Because translation happens at Crowdin and not in Android Studio.

tolot27:
Most of the problems can be solved easily if a higher API level would be used. We will see if jamorham will merge this PR.

Navid200:
I moved the text, from title to summary.  With no change to the text itself, the following is the result.
Screenshot_1619053121


Is this better?  Please note that the font size is now quite small (summary).
If this is better, how can I change an open pull request without closing it and opening a new one?

Or, is the text too small now?

nightscout-ricsi:
I prefer:
If changes by ()
Frequency in minutes ()

Navid200:
@nightscout-ricsi To be correct, we cannot call it frequency.  We can call it period, cycle, how often, ...  But, not frequency.  Frequency, by definition, is the number of occurrences per unit time.  While what this is is how often something happens.
The two are related, but, not the same.   f = 1/ Period

nightscout-ricsi:
> @nightscout-ricsi To be correct, we cannot call it frequency. 
Agree, but it is much easier to understand, with this name. Maybe we should change the unit, ex. speak per hour.
But maybe "period of speak" is better. I am not a native english speaker.

tolot27:
> If this is better, how can I change an open pull request without closing it and opening a new one?

Meanwhile, you should know how to update a PR. 😄

Navid200:
> > If this is better, how can I change an open pull request without closing it and opening a new one?
> 
> Meanwhile, you should know how to update a PR. 😄

Yes, that was a while ago.
I can spend time and figure out how to change the slider.  It is used in several places in xDrip and all have larger text.  So, that would be a good solution.  
But, it will take quite some time for me to figure it out.

tolot27:
I've assigned this PR to me and will have a look during this weekend. Hopefully, I can suggest a solution.

Navid200:
I have reverted the wordings to the original.
But, I have moved the text to the summary, which has a smaller font and does wrap.  
The numerical value being adjusted is still shown in the title.
For what it looks like, please view my last message of April 21 above.

Navid200:
@jamorham That's exactly what I have done.
However, that's not what I had originally.  That's why there are so many commits even though the change now is only very simple.
I can close this PR and open a new one only containing moving the content from the title to the summary, which is what this is  right now.

Navid200:
https://github.com/NightscoutFoundation/xDrip/pull/2076

PR #1706, opened 2021-04-25
Preemptive restart settings

Navid200:
Every now and then, someone on facebook asks why their G6 sensor has stopped on day 9 instead of day 10.
After checking, sure enough, they have preemptive restarts enabled for a firefly G6.

Looking at the G5/G6 Debug Settings, you can see that every single item on that page, except preemptive restarts, is a checkbox.
It's very easy for someone to look at that page and conclude that they don't have preemptive restarts enabled just because they don't see a checkmark on it.  
But, they cannot make that conclusion because that is not a checkbox.  Since every other entry on that page is a checkbox, it is a common confusion.

I have changed that label from "Preemptive restarts" to "Preemptive restart Settings".  And, I have changed the summary also accordingly.
Screenshot_1619325715

This should reduce the confusion and hopefully the number of errors resulting in sensors stopping on day 9.
There was a person who posted and said that her sensors stopped on day 9, instead of day 10, and she believed it was because of the low quality of the sensors.  So, it's not even clear exactly how many people are affected really.  Some people are not even aware that they have a problem that can easily be resolved.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1586,7 +1586,8 @@
     <string name="summary_ob1_g5_preemptive_restart_extended_time_travel">Restart the session to start at day 3 to avoid causing jumps</string>
     <string name="title_ob1_g5_preemptive_restart_extended_time_travel_all_firmwares">Always allow extended time-travel</string>
     <string name="summary_ob1_g5_preemptive_restart_extended_time_travel_all_firmwares">Allow extended time-travel with all firmware versions (engineering mode only)</string>
-    <string name="collection_summary_ob1_g5_preemptive_restart">Automatically restart a sensor before it stops</string>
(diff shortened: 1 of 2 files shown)


jamorham:
I think it is better to fix the logic problem, which is: pre-emptive restarts being applied where they wont work. 
I've made a fix here which hopefully resolves the issue:
https://github.com/NightscoutFoundation/xDrip/commit/0109ed4ab438d04f9825918626e06b6d5e0b2cf0

Navid200:
@jamorham If you mean that you will be changing the behavior of xDrip such that even if the user has preemptive restart enabled, it will not attempt a restart on day 9 if the transmitter is a firefly, that will be awesome.

But, still, it makes sensor to have different headings for the page containing the settings and the command itself.  So, merging this will still be a good thing even after you make your change.

And what happens when we get the new transmitters with modified firmware?
They are firefly but can they not run preemptive restart?

This pull request only presents a valid representation of what lies underneath, the settings rather than enabling the function.

PR #1709, opened 2021-05-01
Optionally store g5/g6 transmitter battery information in nightscout

jwoglom:
This is an updated version of https://github.com/NightscoutFoundation/xDrip/pull/584 (which was originally submitted as a PR in 2018).

When enabled, the changes in this CR allow for uploading battery and status data for Dexcom G5/G6 transmitters to Nightscout. 

The following appears at /api/v1/devicestatus in Nightscout:

```
[
  {
    "_id": "608d030f695b8526acd3fabc",
    "device": "G6 Transmitter",
    "uploader": {
      "days": 85,
      "daysEstimate": "85",
      "status": "OK",
      "voltagea": 309,
      "voltageb": 290,
      "resistance": 0,
      "temperature": 32,
      "battery": "85 days (voltage: 309/290)",
      "type": "DEXCOM_TRANSMITTER"
    },
    "created_at": "2021-05-01T07:28:15.656Z"
  },
  ...
]
```

Other devicestatus uploads now contain a type field, e.g.:

```
{
  "_id": "608d0561695b8526acd3fabe",
  "device": "Google Pixel 2 XL",
  "uploader": {
    "battery": 100,
    "type": "PHONE"
  }
}
```

The transmitter details are now visible in the Nightscout UI:

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1DexTransmitterBattery.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1DexTransmitterBattery.java
@@ -1,0 +1,135 @@
+package com.eveningoutpost.dexdrip.G5Model;
+
+
+import com.eveningoutpost.dexdrip.Models.JoH;
+import com.eveningoutpost.dexdrip.Services.G5BaseService;
+import com.eveningoutpost.dexdrip.Services.Ob1G5CollectionService;
+import com.eveningoutpost.dexdrip.UtilityModels.Pref;
+import com.eveningoutpost.dexdrip.UtilityModels.StatusItem;
+
+
+/**
+ * External public-facing accessor for getting transmitter battery and related status information
+ * for Ob1G5/G6, which abstracts away the G5/G6 internals.
+ *
+ * Used in the Ob1G5/G6 status page as well as for Nightscout transmitter battery upload.
+ *
+ * @author James Woglom (j@wogloms.net)
+ */
+public class Ob1DexTransmitterBattery {
+
+    private final String tx_id;
+    private final BatteryInfoRxMessage battery;
+    private final VersionRequestRxMessage firmware;
+
+    public Ob1DexTransmitterBattery(String tx_id, BatteryInfoRxMessage battery, VersionRequestRxMessage firmware) {
+        this.tx_id = tx_id;
+        this.battery = battery;
+        this.firmware = firmware;
+    }
+
+    private Ob1DexTransmitterBattery(String tx_id) {
+        this.tx_id = tx_id;
+        this.battery = Ob1G5StateMachine.getBatteryDetails(tx_id);
+        this.firmware = (VersionRequestRxMessage) Ob1G5StateMachine.getFirmwareXDetails(tx_id, 0);
+    }
+
+    public Ob1DexTransmitterBattery() {
+        this(Pref.getStringDefaultBlank("dex_txid"));
+    }
+
+    public boolean isPresent() {
+        return battery != null && firmware != null;
+    }
+
+    public TransmitterStatus status() {
+        return TransmitterStatus.getBatteryLevel(firmware.status);
+    }
+
+    public int days() {
+        if (battery.runtime > -1) {
+            return battery.runtime;
+        }
+
+        return DexTimeKeeper.getTransmitterAgeInDays(tx_id);
+    }
+
+    public String daysEstimate() {
+        final int timekeeperDays = DexTimeKeeper.getTransmitterAgeInDays(tx_id);
+        long last_transmitter_timestamp = Ob1G5CollectionService.getLast_transmitter_timestamp();
+
+        StringBuilder b = new StringBuilder();
+
+        if (battery.runtime > -1) {
+            b.append(battery.runtime);
+        }
+
+        if (timekeeperDays > -1) {
+            if (b.length() > 0) {
+                b.append(FirmwareCapability.isTransmitterG6Rev2(tx_id) ? " " : " / ");
+            }
+            b.append(timekeeperDays);
+        }
+
+        if (last_transmitter_timestamp > 0) {
+            b.append(" / ");
+            b.append(JoH.qs((double) last_transmitter_timestamp / 86400, 1));
+        }
+
+        return b.toString();
+    }
+
+    public int voltageA() {
+        return battery.voltagea;
+    };
+
+    public int voltageB() {
+        return battery.voltageb;
+    }
+
+    public boolean voltageAWarning() {
+        return voltageA() < G5BaseService.LOW_BATTERY_WARNING_LEVEL;
+    }
+
+    public boolean voltageBWarning() {
+        return voltageB() < (G5BaseService.LOW_BATTERY_WARNING_LEVEL - 10);
+    };
+
+    public int resistance() {
+        return battery.resist;
+    }
+
+    public enum ResistanceStatus {
+        UNKNOWN(null),
+        GOOD(StatusItem.Highlight.GOOD),
+        NORMAL(StatusItem.Highlight.NORMAL),
+        WARNING(StatusItem.Highlight.NOTICE),
+        BAD(StatusItem.Highlight.BAD)
+        ;
+
+        // Passes-through the highlight used on the status page
+        public final StatusItem.Highlight highlight;
+        ResistanceStatus(StatusItem.Highlight highlight) {
+            this.highlight = highlight;
+        }
+    }
+    public ResistanceStatus resistanceStatus() {
+        if (!FirmwareCapability.isFirmwareTemperatureCapable(firmware.firmware_version_string)) {
+            return ResistanceStatus.UNKNOWN;
+        }
+
+        if (resistance() > 1400) {
+            return ResistanceStatus.BAD;
+        } else if (resistance() > 1000) {
+            return ResistanceStatus.WARNING;
+        } else if (resistance() > 750) {
+            return ResistanceStatus.NORMAL;
+        } else {
+            return ResistanceStatus.GOOD;
+        }
+    };
+
+    public int temperature() {
+        return battery.temperature;
+    }
+}
--- a/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1DexTransmitterBattery.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1DexTransmitterBattery.java
@@ -4,6 +4,7 @@ package com.eveningoutpost.dexdrip.G5Model;
 import com.eveningoutpost.dexdrip.Models.JoH;
 import com.eveningoutpost.dexdrip.Services.G5BaseService;
 import com.eveningoutpost.dexdrip.Services.Ob1G5CollectionService;
+import com.eveningoutpost.dexdrip.UtilityModels.PersistentStore;
 import com.eveningoutpost.dexdrip.UtilityModels.Pref;
 import com.eveningoutpost.dexdrip.UtilityModels.StatusItem;
 
@@ -128,6 +129,10 @@ public class Ob1DexTransmitterBattery {
             return ResistanceStatus.GOOD;
         }
     };
+
+    public long lastQueried() {
+        return PersistentStore.getLong(G5BaseService.G5_BATTERY_FROM_MARKER + tx_id);
+    }
 
     public int temperature() {
         return battery.temperature;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutBatteryDevice.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutBatteryDevice.java
@@ -1,0 +1,103 @@
+package com.eveningoutpost.dexdrip.UtilityModels;
+
+
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.os.BatteryManager;
+import android.os.Build;
+
+import com.eveningoutpost.dexdrip.G5Model.Ob1G5StateMachine;
+import com.eveningoutpost.dexdrip.G5Model.Ob1DexTransmitterBattery;
+import com.eveningoutpost.dexdrip.Services.DexCollectionService;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Represents a device type tracked by xDrip which can be uploaded to Nightscout devicestatus.
+ *
+ * Contains methods for identifying the current battery level and details for how the device's
+ * status will be sent to Nightscout.
+ *
+ * @author James Woglom (j@wogloms.net)
+ */
+public enum NightscoutBatteryDevice {
+    PHONE {
+        @Override
+        int getBatteryLevel(Context mContext) {
+            Intent batteryIntent = mContext.registerReceiver(null, new IntentFilter(Intent.ACTION_BATTERY_CHANGED));
+            if (batteryIntent != null) {
+                int level = batteryIntent.getIntExtra(BatteryManager.EXTRA_LEVEL, -1);
+                int scale = batteryIntent.getIntExtra(BatteryManager.EXTRA_SCALE, -1);
+                if (level == -1 || scale == -1) {
+                    return 50;
+                }
+                return (int) (((float) level / (float) scale) * 100.0f);
+            } else return 50;
+        }
+
+        @Override
+        String getDeviceName() {
+            return Build.MANUFACTURER + " " + Build.MODEL;
+        }
+    },
+    BRIDGE {
+        @Override
+        int getBatteryLevel(Context mContext) {
+            return Pref.getInt("bridge_battery", -1);
+        }
+
+        @Override
+        String getDeviceName() {
+            return DexCollectionService.getBestLimitterHardwareName();
+        }
+    },
+    PARAKEET {
+        @Override
+        int getBatteryLevel(Context mContext) {
+            return Pref.getInt("parakeet_battery", -1);
+        }
+
+        @Override
+        String getDeviceName() {
+            return "Parakeet";
+        }
+    };
+
+
+    /**
+     * Returns the battery level of the device.
+     *
+     * @param mContext is supplied so that the phone battery can be retrieved.
+     */
+    abstract int getBatteryLevel(Context mContext);
+
+    /**
+     * Returns the JSON object containing the battery details.
+     * Defaults to returning a JSONObject with integer key battery.
+     *
+     * @param mContext is supplied so that the phone battery can be retrieved.
+     */
+    public JSONObject getUploaderJson(Context mContext) throws JSONException {
+        JSONObject uploader = new JSONObject();
+        uploader.put("battery", getBatteryLevel(mContext));
+        return uploader;
+    }
+
+    /**
+     * Returns the device name sent to Nightscout for the device.
+     */
+    abstract String getDeviceName();
+
+    /**
+     * If true, the battery status is always uploaded to Nightscout.
+     * If false, the battery status is only uploaded when it changes.
+     *
+     * Nightscout doesn't currently display a device's status if it thinks
+     * that it is stale, so this has always been set to TRUE.
+     */
+    boolean alwaysSendBattery() {
+        return true;
+    }
+}
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutBatteryDevice.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutBatteryDevice.java
@@ -63,6 +63,83 @@ public enum NightscoutBatteryDevice {
         String getDeviceName() {
             return "Parakeet";
         }
+    },
+    DEXCOM_TRANSMITTER {
+        /**
+         * This is used for checking if the battery value is stale.
+         */
+        @Override
+        int getBatteryLevel(Context mContext) {
+            Ob1DexTransmitterBattery b = new Ob1DexTransmitterBattery();
+            if (b.isPresent()) {
+                return b.voltageA();
+            }
+            return -1;
+        }
+
+        /**
+         * Returns a JSON blob such as:
+         * {
+         *     "days": 50,
+         *     "daysEstimate": "50" or "50 / 100" or "50 100",
+         *     "status": "UNKNOWN" or "BRICKED" or "LOW" or "OK",
+         *     "voltagea": 285,
+         *     "voltagea_warning": true,
+         *     "voltageb": 275,
+         *     "voltageb_warning": true,
+         *     "resistance": 100 or null,
+         *     "resistance_status": "GOOD" "NORMAL" "WARNING" "BAD" or "UNKNOWN",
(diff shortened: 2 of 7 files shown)


Navid200:
Thanks for this.  
I have a few questions, and a comment.   I am not very knowledgeable.  So, please be patient.  
My questions are not criticisms.

Currently, there is no resistance measurement for new G6 transmitters (as far as I know).  Will "0" be uploaded for resistance in that case?  Or, will there be an error?

After this is enabled, will these values be available, in Nightscout, versus time?  Or, will there only be one current value present?

How practical/easy/reasonable will it be to add the (additional) capability to store Va and Vb only in a text file in the xdrip folder on the phone?   Or, they can be saved in any other format as long as they can be viewed.  This will be useful when helping someone who has not enabled the setting to upload the values to Nightscout.  The scenario I have in mind is a device that has run into trouble because of the battery being old.  Being able to see the voltage trend is much more informative (about the condition of the battery) than just being able to see the voltage for the past 2 days.

I suggest leaving the corresponding issue open.  Regardless of how you implement the idea, the idea is still the original precious idea.  The open issue shows how old the request has been (maintain history).  It has evolved and has changed.  That's not a problem.  We still have open issues that were opened in 2017.  If someone opens a new issue about it, I will just link them to your issue and close their new issue anyway.  So, I prefer to have the issue open until it is either merged or decided that will never be done.

Fixes: https://github.com/NightscoutFoundation/xDrip/pull/584

tolot27:
> How practical/easy/reasonable will it be to add the (additional) capability to store Va and Vb only in a text file in the xdrip folder on the phone? Or, they can be saved in any other format as long as they can be viewed.

I recommend storing these values in the SQLite database because that is our central data store. Please can you open a separate issue for it?

jwoglom:
> Currently, there is no resistance measurement for new G6 transmitters (as far as I know). Will "0" be uploaded for resistance in that case? Or, will there be an error?

It will get uploaded as the value 0.

> After this is enabled, will these values be available, in Nightscout, versus time? Or, will there only be one current value present?

Nightscout stores all of the values. The main nightscout UI only shows the most recent reading, I believe, but via the API or manually looking at the sqlite database you can view the complete history. If there isn't a UI in nightscout for viewing battery status over time, it would just be a question of adding an interface to let you view it since the data is already stored in the nightscout database. 

> How practical/easy/reasonable will it be to add the (additional) capability to store Va and Vb only in a text file in the xdrip folder on the phone? Or, they can be saved in any other format as long as they can be viewed. This will be useful when helping someone who has not enabled the setting to upload the values to Nightscout. The scenario I have in mind is a device that has run into trouble because of the battery being old. Being able to see the voltage trend is much more informative (about the condition of the battery) than just being able to see the voltage for the past 2 days.

Yeah, this would be a separate task that would be best served in a separate issue/PR. You can use the Ob1DexTransmitterBattery helper class I included to make querying that information easier if this ever got implemented.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutBatteryDevice.java
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * Returns the device name sent to Nightscout for the device.
+     */
+    abstract String getDeviceName();
+
+    /**
+     * If true, the battery status is always uploaded to Nightscout.
+     * If false, the battery status is only uploaded when it changes.
+     *
+     * Nightscout doesn't currently display device device status if it thinks its stale,
```
tolot27, line 175: remove the second `device` and replace `its` with `it's`.

app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1DexTransmitterBattery.java
```diff
@@ -0,0 +1,135 @@
...
+            return ResistanceStatus.NORMAL;
+        } else {
+            return ResistanceStatus.GOOD;
+        }
+    };
+
+    public int temperature() {
+        return battery.temperature;
+    }
+}
```
tolot27, line 135: There should be a newline at EOL.

tolot27 (changes requested):
Nice abstraction. Many thanks for this PR. A real test starts soon.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutBatteryDevice.java
tolot27: That's comment was copied from somewhere else, I know.

jwoglom: Indeed it was!

app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1DexTransmitterBattery.java
jwoglom: Done

jwoglom:
@tolot27 I forced push and edited the existing commits to fix the typos

tolot27:
I've tested it but no tooltip appears if I hover over the battery icon. I've verified that the data is uploaded to NS (present in devicestatus, including the new type attribute). I have to enable the [devicestatus plugin](https://github.com/nightscout/cgm-remote-monitor#devicestatus-device-status), somehow.

Navid200:
I don't understand the priority of this.

I ask myself, if I was not able to see my blood glucose in xDrip, and I could spend time on enabling to see it in xDrip or in Nightscout, which task would I prioritize first?  
My thought is xDrip first.

Why not implement the capability in xDrip first?

tolot27:
> I don't understand the priority of this.

Currently, there is no priority of any issue/PR defined, unfortunately.

> I ask myself, if I was not able to see my blood glucose in xDrip, and I could spend time on enabling to see it in xDrip or in Nightscout, which task would I prioritize first?
> My thought is xDrip first.
> 
> Why not implement the capability in xDrip first?

Feel free to create a PR. I think it is much easier to extend the NS uploader service than changing/extending the xDrip database objects and enhance the GUI or write to files.

tolot27:
@jwoglom Does the Nightscout site also need some updates? I've the head version 14.2.3 installed but still do not see the tooltip. I've checked that the devicestatus plugin is enabled and DEVICESTATUS_ADVANCED is true. Any hints?

jwoglom:
@tolot27 I am running a very old version of Nightscout in my current setup (from circa 2018, I haven't gotten around to updating it), so I don't think Nightscout version is a concern. Instead, I think that Nightscout doesn't show devicestatus data which is "old", for some definition of old. There's a comment in the xdrip source about this in fact, which is why for all other battery data sources, xDrip publishes new devicestatus data to Nightscout on every upload instead of whenever the battery percent changes.

I was a little weary of setting the dexcom transmitter battery stats to upload at such a high frequency to Nightscout, given that we're only getting the source data about the battery every 6 hours or so and we would be polluting a lot of extra data to the Nightscout database. If someone explicitly opts-in on the xDrip side to uploading this battery data then I suppose it is okay, albeit a little misleading given that we're recording battery details in the database that could be as much as 6 hours out of date.

If we think the primary consideration here is to make sure this is easily visible in the nightscout UI then I presume we'll just have to enable it to upload the battery data on every nightscout upload, short of adding specialized code for this on the Nightscout end I'm unaware of a better solution.

Navid200:
I believe, by default, the battery info is updated once every 12 hours.
Of course, the user can request battery info any time.  
If there could be a choice in how often to upload, I would choose once a day.

jwoglom:
This code as written will only send new battery info to Nightscout whenever the voltage A changes (checking whenever the normal Nightscout sync occurs, and not changing the xDrip default battery check interval)

Navid200:
I hope you mean whenever Voltage A is updated.  If you really mean that it only uploads only when voltage A changes (for example from 315 to 314) and will not upload if Voltage A is the same as it was last time, that would be really bad.

jwoglom:
> I hope you mean whenever Voltage A is updated. If you really mean that it only uploads only when voltage A changes (for example from 315 to 314) and will not upload if Voltage A is the same as it was last time, that would be really bad.

I believe that is what is happening based on this check:

https://github.com/NightscoutFoundation/xDrip/pull/1709/files#diff-a9f0fc69fd72f2078337cc3cf814930111fbf45f6007c952898ec13cec1e8b80R1087

For DEXCOM_TRANSMITTER, getBatteryLevel returns the voltage a value.

Is this of significant concern? What's so bad about it?

For what it's worth, like I elaborated on above, I think we may need to enable uploading the battery on every single nightscout upload if we want to make this visible in the Nightscout UI, which negates the significance of this specific behavior.

Navid200:
Just because Voltage A has not changed does not mean that Voltage B has not changed.
Voltage B changes faster than Voltage A does.
That's why it is a better representation of how healthy the battery is.  The absolute values of those voltages is useless.  It's the trend that matters.  The trend has no meaning for Voltage A as it hardly changes with respect to voltage B.

tolot27:
@sulkaharo Please can you comment on this PR, especially on the update frequency and visualization?

sulkaharo:
Update frequency: the Nightscout client only loads last few device statuses per device into memory at any time, so any field needed for visualisation should be included in _every_ device status update.

Visualization: it'd be great if xDrip would provide the data on how to interpret the values. We're on G6 and I have no idea what the voltage ranges are really and I'm used to xDrip showing the values in red (which I take means a problem) almost every time I look at the voltages. For visualisation, what could work is something like xDrip also uploading the reference range in the device status, so the visualisation could actually show if the voltage value is in range or not. I bet most users have no idea what the dual voltage values mean either - I most definitely don't given G6 has a single battery, vs the two in G5. :)

Navid200:
As far as the G5 transmitter is concerned, there is only one power source, the regulated voltage source.
The two batteries of G5 are in series forming a single voltage to the transmitter.   An electronic circuit, almost always, has a voltage regulator, which takes the battery voltage as an input and generates an output voltage, which is lower than the battery voltage but more stable with respect to the load and temperature and production parameter variations.

Voltage A and Voltage B are the regulator output voltages unloaded and loaded respectively.  There is no ideal (perfect) voltage regulator.  Any voltage regulator has an output impedance.  That's why the loaded output voltage (B) is lower than the unloaded voltage (A).  

The unloaded voltage (A) only provides info about the regulator itself and the battery.  It is the loaded voltage (B) that is more valuable because it also shows the impact of the load, the transmitter circuitry. 

So, if the output voltage of the regulator is stable, how can we tell the age of the battery from it?  Again, the regulator is not perfect.  The output voltage of the regulator is affected gradually as its input voltage (the battery voltage) drops as it ages.  

Have a look at this image.  
Screenshot 2021-05-04 172200

It shows Voltage B for my current and last 2 G6 transmitters.  Why are they different?  
There are two reasons.  No 2 transmitters are identical.  But, more importantly, the measurements are made by the transmitters themselves.  They don't have perfect accuracy.  

It's true that we pay a lot of money for a single transmitter.  But, a transmitter costs no where near what we pay for it for the manufacturer.  There is nothing in the transmitter that's not absolutely required.  And a very accurate volt meter is not one of those.

So, it's not surprising that the voltages are different.  But, one thing is consistent.  When the time comes that the battery starts dying, Voltage B will consistently decrease.  That's why the trend is more important than the absolute values of the voltages.

If your Voltage B, for G6, is greater than 280, you should be OK.  But, if you ask for how long, that's when I would say, you need to create a spreadsheet.
I hope to add the capability to xDrip to store and show an image of Voltage B.  Then, the user will not need to create a spreadsheet.
Storing it on Nightscout would be great for the same reason storing our blood glucose readings, on Nightscout, is great.

jwoglom:
@sulkaharo it seems a little redundant if we were to publish the ranges for the transmitter battery inside every single item uploaded to devicestatus -- however we are publishing the "status" field as either OK, LOW, BRICKED, or UNKNOWN (that is self reported by the dexcom transmitter) in addition to voltagea_warning and voltageb_warning as booleans (not publishing them if they're false, just to save a little bit of space). There's also a self-reported transmitter resistance_status as either GOOD, NORMAL, WARNING, or BAD -- but I believe we can't read that info on newer transmitters at the moment.


Perhaps as a start if any one of those fields is present and contains a warning status, nightscout could show a warning? I'm not sure in what specific cases xDrip alerts on each of those properties' state -- I think voltagea/voltageb warning is the primary one.

jwoglom:
I have updated the PR to always upload the dexcom battery on every nightscout upload cycle, and to also include a lastQueried field which contains a timestamp of the last time xdrip queried the tx battery.

jwoglom:
Unrelated to these specific changes, something seems wrong with the build (see CI details tagged on the PR). It's happening locally to me as well, even building from the master branch. Did these packages become unavailable?

```
* What went wrong:
Execution failed for task ':app:preFastDebugBuild'.
> Could not resolve all files for configuration ':app:fastDebugCompileClasspath'.
   > Could not resolve com.github.nightscout:android-uploader:CORE_FOR_XDRIP.
     Required by:
         project :app
      > Skipped due to earlier error
   > Could not resolve com.github.jamorham:amazfitcommunication:master-SNAPSHOT.
     Required by:
         project :app
      > Skipped due to earlier error
   > Could not resolve com.github.jamorham:QR-Code-generator:d0454644af.
     Required by:
         project :app
      > Skipped due to earlier error
```

sulkaharo:
Sorry, didn't explain the full why reason I mentioned uploading the ranges for every status: the idea was that ideally we'd have a generic way to visualise the state of a battery inside Nightscout, which would work based on the uploaded data and not need battery type specific logic inside Nightscout. An uploader knows the specifics of a battery type (hopefully) and could upload a status message that allows simple logic to compare if a battery is in range or not and decide to act if so. This would then allow an implementation where any device's battery could trigger alarms etc without NS needing to have device specific logic. Uploading the OK/LOW information would also allow this, if we add that same status to other devices. So as an example,  I'm also thinking of OpenAPS uploading battery state of the rigs in voltages and some of the related configuration to tune the values being pretty hard on users / not existing, while the system could upload data that allowed for generic logic to visualise the battery state.

tolot27:
After updating to commit f2b13da, it works as shown with the screenshot. But I don't really like sending unchanged data repeatedly. This pollutes the NS database and consumes traffic, which drains the battery, especially if this feature is integrated into the wear app.
@sulkaharo Do you have any idea how to avoid sending these repeated unchanged data? Can we implement something in NS to show the last device status entry if not older than x days?

Navid200:
I don't understand!  Are we polluting the Nightscout database if our blood glucose is always the same?

jwoglom:
@Navid200 There's a significant difference in that in regular operation we would be sending the exact same battery value for 6-12 hours (however long the default xdrip battery query interval is) in to nightscout. It makes it look like there is more precision than there actually is, and it wastes storage space and bandwidth. If someone's glucose level is the same, that is at least discrete readings from the sensor which are being uploaded.

Navid200:
Of course, data takes space.  
If you can write the code to only send the battery voltage when it changes, it would do the job.  But, that has to be done for the two different voltages separately.  I personally see no value in Voltage A.  But, I don't want to start a debate.  
If you are really worried about saving space, why do you want to upload the temperature?  What value does that provide?
Same question for resistance.

If someone asks me "how much longer will my battery last", the only way I can answer that question is if I have Voltage B at least once every 48 hours. 

If we have the info inside xDrip, I personally see no reason for uploading voltages, resistance or temperature to Nightscout.

Would there be a switch to allow the user to only upload glucose and not any of this?

Navid200:
Sorry, I may have completely misunderstood you guys.
Are you saying, we will be uploading the battery voltage once every 5 minutes?  If yes, that's not good.  Can we not upload the battery info only once a day?

tolot27:
> Are you saying, we will be uploading the battery voltage once every 5 minutes? If yes, that's not good. Can we not upload the battery info only once a day?

Yes, now it is sent every five minutes. Before commit f2b13da it was only send if the battery status was queried from the transmitter. But this occurs not often and you will only see the status in NS if it is not older than a few minutes (@sulkaharo how long)?
That's why I've posted https://github.com/NightscoutFoundation/xDrip/pull/1709#issuecomment-833054107.

sulkaharo:
Some comments on above questions:

The Nightscout database query layer is already quite complicated and getting the queries to scale with the free MongoDB solutions was challenging, so we effectively can't change the Mongo access to load data selectively and have the latest update from each uploader be available, with arbitrary data upload intervals. Note also the only way Nightscout can know an uploader is alive is through the periodic uploads, where if a device would only send a device status message once a day, it'd be impossible know if this is due to the device being offline. The expected behavior is for an active data uploader to post a device status upload fairly frequently.

There's no hard rule for the intervals, but most uploaders are uploading the data when they have other data to upload, so in case of a CGM data source, the behavior has been to upload the battery level every 5 minutes. The web client is sent a few of the latest updates from each of the devices sending the device status information and if an uploader is sending the data more infrequently than other uploaders, there is no guarantee data from it's devivcestatus messages is passed to the client. Also there is no generic mechanism to aggregate data fields of the uploaders, so if the set update messages contain varied data, the client needs implement logic to try to find the wanted information from those messages and this is not always 100% consistent, so for an uploader the only way to guarantee a piece of data like batter state is available for the client and over the API, the data needs to be included in every status message.

Note a lot of users use Nightscout as the primary view into device information, so including any data that might be relevant for an xDrip user into the messages is fine. Don't worry about the data consumption here - the messages from 24 hours is tiny compared to the user loading even a single animated GIF over mobile data and we have mechanisms to clear old device status data from Mongo. FWIW OpenAPS device status messages are several kilobytes each and it sends those twice as frequently as xDrip, so there's plenty of room to grow here. :)

jwoglom:
Thanks for the info @sulkaharo! With that in mind, I have updated the code so that dexcom tx battery info uploads just as frequently as phone battery info/other device batteries uploaded by xDrip. There'll still be a lastQueried field which tracks when the battery detail was last queried from the transmitter -- perhaps that could be integrated into the Nightscout UI.

tolot27:
> 

What does the "% 32" mean at the end of the last line of the tooltip?

tolot27:
@sulkaharo 

> we have mechanisms to clear old device status data from Mongo.

Are these mechanisms enabled by default? How do they work? Do they aggregate older data just ignoring the `created_at` timestamp or do they simply delete old data?

And how does the query for displaying the status works? Does it simply fetch the last x records or is it based on the `created_at` field?

Navid200:
xDrip now shows the temperature for Firefly correctly, in the logs.  However no one has bothered adding it to the system status page.  But, it is there.  And, it is useless.  So, I will not add it to the system status page and don't care for it.
And it's not always 32.

Navid200:
I may be wrong.  I'm going to add the temperature back to the system status page.

tolot27:
> > What does the "% 32" mean at the end of the last line of the tooltip?
> 
> @tolot27 I think that is the temperature reported by the sensor. xDrip isn't placing that inside the panel info, Nightscout is. I suspect it is reading the value from uploader["temperature"]
> 
> I'm actually not sure if newer G6 transmitters provide a real temperature value that is parsed by xDrip or whether it's always 32. Perhaps someone who is more familiar with the g5/g6 native mode inside xDrip could inform.

@sulkaharo Please can you link the corresponding code in NS? It would be great if the unit of measurement (°C) could be added to the temperature value.

Also, the percent sign is irritating. Can this be avoided? In theory, it should be calculated based on days/daysEstimate. But likely this is to specific to the Dexcom transmitter battery. @jwoglom Since the "battery" inner string is shown in NS, maybe we could add the battery state in percent right after the closing bracket to get a more readable status string.

jwoglom:
If someone was confident enough after looking at trends of voltage values to make an estimated battery percent, I suppose that could be an option? Alternatively this could be a percent of days used / 100 (or, if xDrip is able to calculate it, the day that the battery will expire). I doubt any calculated value would ever be fully accurate, though. So maybe we just leave the hanging % and focus on better displaying this in Nightscout. 

FYI, I haven't tried a new release of Nightscout in a while so maybe it looks a bit better than my screenshot. If not, seems simple enough to only have NS show the % if the battery field is a number, and maybe affix a unit to the temp of one is available.

Navid200:
@sulkaharo Thanks for having answered our questions.
My database size is 23%.
Annotation 2021-06-13 112343

I've always wondered what happens if/when that reaches 100% and what my options will be.
What will be the impact of adding these to how fast the database will be full?

Someone requested that we also upload the calibration code and transmitter serial number to Nightscout.  The calibration code changes once every 10 days.  The transmitter serial number changes approximately once every 3 months.
What would be the impact, on database size, if we also uploaded those?

Where do we draw the line, or do we?

jwoglom:
@Navid200 I think it would be useful to store tx serials in Nightscout/xDrip however I think we should track that as a separate GH issue.

I just force-pushed to resolve a merge conflict on the branch. @tolot27, are there any remaining important things we feel are necessary to work out before we could merge this? (Keeping in mind that this adds a feature not enabled by default)

Navid200:
@jwoglom I did not mention that because I thought we should upload the serial numbers to Nightscout.  In fact, I hope we never will if my understanding of Nightscout storage is correct.
The reason for my question was to understand how Nightscout data storage capacity is used.
I'd like to know if uploading the voltages to Nightscout takes space that could otherwise be used for storing blood glucose.

Unfortunately, it is still not clear for me.

If the storage capacity is limited, we should not waste it.  We should not even upload the voltages because I don't know why it is needed to be there.
On the other hand, if it makes no differences, then, we can plan using it differently.

jwoglom:
@Navid200 There is only one storage capacity in play here, and it's however much disk space a Nightscout user has on their machine. I believe NS has a feature to delete old device status entries after some period of time, which could ensure that years worth of devicestatuses didn't result in less storage space being available for blood glucose data.

Navid200:
@jwoglom What if a user likes to give priority to blood glucose over voltages?  How would a user do that?
Hearing that there is only one storage capacity is not something that gives me confidence.  It actually worries me why we are wasting the limited space for storing voltages, which are of less importance than blood glucose.

jwoglom:
@Navid200 In that case a user would need to use Nightscout's existing mechanisms for regularly deleting old device statuses, such that they didn't take up as much space. There's no way to "give priority" since everything is stored in one database. Nightscout never automatically deletes data unless the user asks it to.

Also, it isn't the case that every user of Nightscout has the same amount of limited space. For some users they might be using NS on a server with plentiful storage space, and thus never want to delete anything, while others might only have a few hundred megabytes of storage available.

Navid200:
If that's true and there is no way to give priority for deletion, then, we need to only upload content to Nightscout that is absolutely required to be uploaded.

If we ever add a capability to upload something else to Nightscout, for that reason, we should allow the user to opt in.  If not, by default, the new content should not be uploaded.

If this is merged, is it going to upload the voltages for everyone by default?

jwoglom:
@Navid200 This PR does not enable anything by default. You must enable this preference which defaults to false: https://github.com/NightscoutFoundation/xDrip/pull/1709/files#diff-ea6c08af36bba55c22c04153f1cf14f714345e8e442f3cc63c7693c959c0cc88R76

Navid200:
Great

jamorham:
This looks like it should work. I assume from the age of the PR and comments that significant real world testing has occurred.

Issue #684, opened 2018-11-14
G6 uploaded as G5


quizzmaster:
The uploaded device String using a Dexcom G6 is "xDrip-Dexcom G5 Native" but should be "xDrip-Dexcom-G6 Native".

jwoglom:
I haven't tried the patch with newer versions of xDrip but #559 should fix this

quizzmaster:
With the latest nightly the device string is "xDrip-DexcomG5 G5 Native". Now the G5 is doubled.

Navid200:
@quizzmaster Has this been resolved?

quizzmaster:
@Navid200 no, it's still "xDrip-DexcomG5 G5 Native" for the Dexcom G6

Navid200:
Where is this uploaded to?
How can I see this?  I also use a G6.

quizzmaster:
It is uploaded to nightscout. Just look at the endpoint for the BG values `/api/v1/entries/sgv.json`

Navid200:
Sorry, I am a novice.  Can you be more specific?  When I go to Nightscout, I see my blood glucose versus time in a graph.

quizzmaster:
Well, just append `/api/v1/entries/sgv.json` to your Nightscour URL and you get the latest blood glucose informations from the API (in json format).

Navid200:
Thanks.

Navid200:
@quizzmaster Would you please try the latest Nightly (May 10, 2021) and see if this is fixed or not?

Navid200:
@quizzmaster 
Please reopen if it still has not been fixed.

PR #1714, opened 2021-05-04
BgReading: store "G6 Native" label instead of "G5 Native" when using g6.

jwoglom:
This will cause the correct G6 label to also be stored in Nightscout.

As far as I can tell, this codepath only affects the internal sourceinfo stored in the xDrip database which is passed to Nightscout, and does not affect AndroidAPS or local broadcasts.

If this somehow does affect those, per the commits linked by @AdrianLxM in the original PR, AndroidAPS code has been aware of the "G6 Native" label since 2018. There also appears to be code that checks for backwards-compatibility at https://github.com/NightscoutFoundation/xDrip/blob/38fee03be56da6ce9cbe90e08dbe8e4b651c6edc/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BroadcastGlucose.java#L38

See original discussion in https://github.com/NightscoutFoundation/xDrip/pull/559

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -1089,7 +1089,11 @@ public class BgReading extends Model implements ShareUploadableBg {
             bgr.uuid = UUID.randomUUID().toString();
             bgr.calculated_value = calculated_value;
             bgr.raw_data = SPECIAL_G5_PLACEHOLDER; // placeholder
-            bgr.appendSourceInfo("G5 Native");
+            if (Ob1G5CollectionService.usingG6()) {
+                bgr.appendSourceInfo("G6 Native");
+            } else {
+                bgr.appendSourceInfo("G5 Native");
+            }
             if (sourceInfoAppend != null && sourceInfoAppend.length() > 0) {
                 bgr.appendSourceInfo(sourceInfoAppend);
             }


tolot27 (approved):
Since current AAPS supports both G5 and G6 native, everything is fine.

This tiny PR fixes a misleading label which avoids confusion.

tolot27:
> https://github.com/NightscoutFoundation/xDrip/blob/38fee03be56da6ce9cbe90e08dbe8e4b651c6edc/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BroadcastGlucose.java#L38

I don't really like this long-lasting backward compatibility. New changes are in the nightlies/alpha versions first. Why should one update xDrip to a nightly build but stick with a very old AAPS version, especially if also G6 sensors are used?
G7 sensors are on the horizon (second half of 2021) which further increases mislabeling risks.

Navid200:
Is there a reason to close an open PR and open another one to address the exact same issue?
The possible concern I have with this approach is that there may be some posts in the original PR that are relevant but are left out of the new PR discussions.

If this is an acceptable approach, we should have a clear instruction as to when/how it can be performed.
Otherwise, I suggest we never do it.

tolot27:
@Navid200 Rebasing of PRs combined with `git push --force-with-lease` is the way to go, but most developers do not know about it. Should be added to the [dev docs](https://xdrip.github.io).

jamorham (approved):
For info I still use AAPS from 2017 so appreciate the backwards compatibility but I don't think this causes any problems for it.

Navid200:
@jwoglom Would you please tell me how I can test this?

This is what I get.  I am using a G6.
Screenshot 2021-05-11 080057

Why does it still show G5?

jwoglom:
@Navid200 are you using native mode? I believe this change will only take effect if you are using native mode.

e.g., in my case before this change my device string showed in Nightscout was:
```
"device":"xDrip-DexcomG5 G5 Native"
```

This PR only changed the "G5 Native" to "G6 Native", not the "DexcomG5" part.

Navid200:
I have a Firefly.  I can only use it in Native mode.

jwoglom:
@Navid200 Ah, it looks like this info is only appended when the "[nightscout_device_append_source_info](https://github.com/NightscoutFoundation/xDrip/blob/38fee03be56da6ce9cbe90e08dbe8e4b651c6edc/app/src/main/res/xml/pref_data_sync.xml#L86)" preference is set to true:

https://github.com/NightscoutFoundation/xDrip/blob/38fee03be56da6ce9cbe90e08dbe8e4b651c6edc/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java#L598

I know this is a discussion for another place, but there are just so many obscure preference settings in xDrip that should really just be enabled all of the time. Is there a reasonable use case for _not_ wanting this source info to be appended? Setting this preference to default to on is one solution, I suppose, but that doesn't help existing xDrip users who might already have it disabled.

Navid200:
The reason I asked is that there is an open issue for this.  I would like to close it.  But, I wanted to ensure that it is fixed before closing it.
https://github.com/NightscoutFoundation/xDrip/issues/684

I take it it is fixed.  I will close it and ask the individual opening to reopen if they still have problems.

Thanks

PR #1715, opened 2021-05-04
Mediaplayer bugfix - closes the player correctly and releases resources

gruoner:
JoH plays media files using MediaPlayer but it doesn't release ressources and closes the player correctly
after a while this results in silent alarm without any sound

fixed by defining a onCompleteListener as used in AlertPlayer

tested on by productivity used samsung S3 compact with lineage os


Navid200:
When the problem happens, what is shown in xDrip event logs?

Would you mind listing the alerts that this issue affects?  If I understand you correctly, it does not affect any alarms under "Glucose Level Alerts List".   Is that correct?
But, would you please clarify exactly which alerts have the problem?

What is the total number of files changed by this pull request?  It shows 10 commits.

Can you suggest an easy way to reproduce/test the problem?

gruoner:
i encountered the issue after synchronizing a bloody BG check from my accu check guide via Bluetooth.
right after the BG was synchronized (and the sound was played) ALL glucose alarms were silent (regardless whether there as still an active alert while synchronizing or not) - restarting the phone or restarting the app solved this issue until next synchronization took place. i couldn't find any logentries belonging to this issue - i already tried to use debug logs of ADB logcat but there was way to much information to extract the root cause.

i encountered the issue with ALL alerts (too high or too low and any alarm between; with system sounds or custom sounds) - the only way to work around this issue was switching off any sound messages belonging to BG meters via Bluetooth.

That way i found the place where the issue might be - and found out, that in most of the cases where JoH.playSoundUri was called, the used MediaPlayer wasn't closed correctly. So fixing the bug was as easy as assigning a listener to MediaPlayer to release all resources and closing the player right after sound was played......

i'm quite sure, my lineage os patch for samsung s3 compact will cause this issue - because my galaxy star doesn't has this issue some months ago - but nether the less, the used MediaPlayer has to release all acquired resources.

the first 9 commits are just downstream syncs from "global master" to my forked master - the only relevant commit is the last one. U can see it via the files changed.

app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
```diff
@@ -1063,6 +1063,19 @@ public static synchronized MediaPlayer playSoundUri(String soundUri) {
             JoH.getWakeLock("joh-playsound", 10000);
             final MediaPlayer player = MediaPlayer.create(xdrip.getAppContext(), Uri.parse(soundUri));
             player.setLooping(false);
+            player.setOnCompletionListener(mp -> {
+                UserError.Log.i(TAG, "playSoundUri: onCompletion called (finished playing) ");
```
tolot27, line 1067: Please log it to debug and remove the trailing whitespace.

app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
```diff
@@ -1063,6 +1063,19 @@ public static synchronized MediaPlayer playSoundUri(String soundUri) {
...
+                try {
```
tolot27, line 1068:
```java
                player.release();
```

Should be sufficient. According to the [docs](https://developer.android.com/reference/android/media/MediaPlayer#release()), release() does not throw any exception.

tolot27 (changes requested):
Thanks for catching this. It is a simple and recommend solution to solve this general issue.

app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
gruoner: as this code is copied from AlertPlayer, i have to change it there too....

gruoner:
how do i rebase a PR? Github published a guide (https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) but this seems very complicated and i'm afraid it will lead to mistakes.....

app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
```diff
@@ -1063,6 +1063,15 @@ public static synchronized MediaPlayer playSoundUri(String soundUri) {
...
+                    player.stop();
```
tolot27, line 1069: You don't have to stop the player before. Just release it.

tolot27:
> how do i rebase a PR?

It is very simple if you did not push any commits or merge other PRs to your master branch:

```
git fetch origin master
git checkout <your branch name>
git rebase origin/master
git push --force-with-lease origin <your branch name>
```

tolot27:
@gruoner Please can you address the latest change request and rebase this PR? Afterwards, I'll mark it as ready to merge for the next build. Again, only jamorham can merge it.

Thanks!

tolot27:
@gruoner Please can you address the review comments? We will be happy to merge this PR, if ready. Thanks!

gruoner:
@tolot27 rebase didn't work - maybe because my master has also upstream fetches.... any thoughts?

Navid200:
> @tolot27 rebase didn't work - maybe because my master has also upstream fetches.... any thoughts?

I'm not sure why you think that. 
<br/>
<br/>

Untitled

gruoner:
@Navid200 because there are 14 commits in the PR but just two of them are containing bugfix-changes. The rest are upstream fetches. i tried to rebase using tolot27's process but there wasn't any rebase.
tolot27 wrote "It is very simple if you did not push any commits or merge other PRs to your master branch:" and my master branch contains PRs resulting from upstream fetches.....

PR #1717, opened 2021-05-05
Double-Fork: Add Afrezza, Apidra, Levemir, Basaglar and Tresiba profiles, and correct Lantus, Toujeo and Novo Rapid profiles based on more recent data. #1388

ckidow:
This pull requests should fix all remaining tasks of the [original pull request](https://github.com/NightscoutFoundation/xDrip/pull/1388) from the [J141 / xDrip Fork](https://github.com/J141/xDrip).

The tasks of the original pull request cannot be fixed because nobody but J141 itself has access and permissions to change the code.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/insulin/InsulinProfileEditor.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/insulin/InsulinProfileEditor.java
@@ -1,6 +1,7 @@
 package com.eveningoutpost.dexdrip.insulin;
 
 import android.os.Bundle;
+import android.view.MotionEvent;
 import android.view.View;
 import android.widget.AdapterView;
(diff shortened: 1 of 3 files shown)


tolot27:
@ckidow Thanks for your effort. But that's not the way to go. It is not worth to create a new PR and loosing all the conversations just because of a line-break fix. 
IMHO we should give J141 some time to respond. But meanwhile, I've fixed this small issue in-place.

ckidow:
> @ckidow Thanks for your effort. But that's not the way to go. It is not worth to create a new PR and loosing all the conversations just because of a line-break fix.

Yes you are absolutely right.

> IMHO we should give J141 some time to respond. But meanwhile, I've fixed this small issue in-place.

Fine for me and thx for fixing it.

PR #1718, opened 2021-05-06
Improve Tasker snooze functionality

jwoglom:
This PR allows for snoozing a specified number of minutes via Tasker, as well as a specified number of minutes for snoozing low/high alerts. It includes some cleanup in SnoozeActivity to allow for this to be queried via tasker.

Added tasker commands:
* **SNOOZE [mins]** - snoozes for the given number of minutes (Existing no-argument SNOOZE behavior is preserved -- snoozes the current event length or default snooze length)
* **SNOOZE_HIGH** - snoozes high alerts for the current alert length or default snooze length
* **SNOOZE_HIGH [mins]** - snoozes high alerts for the given number of minutes
* **SNOOZE_LOW** - snoozes low alerts for the current alert length or default snooze length
* **SNOOZE_LOW [mins]** - snoozes low alerts for the given number of minutes

These commands can be executed via the WebServiceTasker interface (e.g. http://localhost:17580/tasker/SNOOZE_LOW+30 to SNOOZE_LOW for 30 minutes). To test this functionality you can run `adb forward tcp:17580 tcp:17580` to access the web interface.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
@@ -34,38 +34,50 @@ public class SnoozeActivity extends ActivityWithMenu {
     private static String status;
 
     // Snooze types supported by this class
-    public static final String ALERTS_DISABLED_UNTIL = "alerts_disabled_until";
-    public static final String LOW_ALERTS_DISABLED_UNTIL = "low_alerts_disabled_until";
-    public static final String HIGH_ALERTS_DISABLED_UNTIL = "high_alerts_disabled_until";
+    public enum SnoozeType {
+        ALL_ALERTS("alerts_disabled_until"),
+        LOW_ALERTS("low_alerts_disabled_until"),
+        HIGH_ALERTS("high_alerts_disabled_until")
+        ;
+
+        private final String prefKey;
+        SnoozeType(String prefKey) {
+            this.prefKey = prefKey;
+        }
+
+        public String getPrefKey() {
+            return prefKey;
+        }
+    }
 
     /**
      * Snoozes a given type for the specified number of minutes.
      */
-    public static void snoozeForType(long minutes, String disableType, SharedPreferences prefs) {
+    public static void snoozeForType(long minutes, SnoozeType disableType, SharedPreferences prefs) {
         if (minutes == -1) {
             minutes = infiniteSnoozeValueInMinutes;
         }
         long disableUntil = new Date().getTime() + minutes * 1000 * 60;
 
-        prefs.edit().putLong(disableType, disableUntil).apply();
+        prefs.edit().putLong(disableType.getPrefKey(), disableUntil).apply();
 
         //check if active bg alert exists and delete it depending on type of alert
         ActiveBgAlert aba = ActiveBgAlert.getOnly();
         if (aba != null) {
             AlertType activeBgAlert = ActiveBgAlert.alertTypegetOnly();
-            if (disableType.equalsIgnoreCase(ALERTS_DISABLED_UNTIL)
-                    || (activeBgAlert.above && disableType.equalsIgnoreCase(HIGH_ALERTS_DISABLED_UNTIL))
-                    || (!activeBgAlert.above && disableType.equalsIgnoreCase(LOW_ALERTS_DISABLED_UNTIL))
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/localeTasker/receiver/FireReceiver.java
```diff
@@ -162,8 +167,59 @@ public void onReceive(final Context context, final Intent intent) {
...
+                            if (message_array.length > 1) {
+                                minutes = Integer.valueOf(message_array[1]);
+                                JoH.static_toast_long("SNOOZE from Tasker for "+minutes+" min");
+                            } else {
+                                JoH.static_toast_long("SNOOZE from Tasker");
+                            }
+                            AlertPlayer.getPlayer().Snooze(xdrip.getAppContext(), minutes);
+                            break;
+
+                        case "SNOOZE_LOW":
```
tolot27, line 181: Here, an enum would also be appropriate.

tolot27 (changes requested):
Thanks for your contribution. Please apply code formatting in general. Should be consistent with the existing code (AS default code style).
For the future, it would be nice to commit code refactoring changes separately.

app/src/main/java/com/eveningoutpost/dexdrip/localeTasker/receiver/FireReceiver.java
jwoglom: Would you mind clarifying what you think should be an enum here? An enum representing all of the supported top-level messages inside tasker?

tolot27: > Would you mind clarifying what you think should be an enum here? An enum representing all of the supported top-level messages inside tasker?

Yes, that would be the correct way to implement. But keep it as is, for now.

app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
```diff
@@ -35,6 +33,43 @@
     //public static String menu_name = "Snooze Alert";
     private static String status;
 
+    // Snooze types supported by this class
+    public static final String ALERTS_DISABLED_UNTIL = "alerts_disabled_until";
```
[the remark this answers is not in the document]
tolot27: Thanks. Later on, we should use it anywhere.

tolot27 (approved):
LGTM

jwoglom:
@jamorham I don't think it should affect backwards compatibility. The only existing tasker command changed is "SNOOZE", which when invoked without an argument runs `AlertPlayer.getPlayer().Snooze(xdrip.getAppContext(), -1)`, which is equivalent to the previous `AlertPlayer.defaultSnooze()` invocation which runs that explicitly.

I did just notice that it looks like the [tasker configuration edit layout](https://github.com/NightscoutFoundation/xDrip/blob/38fee03be56da6ce9cbe90e08dbe8e4b651c6edc/app/src/main/res/layout/activity_tasker_edit.xml#L124) (which appears when you add xdrip inside tasker to select the command) references a "SNOOZE ALL" command, although from looking at the old code this was never actually implemented to do anything separate from the existing "SNOOZE". Now that we're checking the second parameter this won't crash the app or anything, but will raise a log message since the string "ALL" can't be parsed as an int. Might also want to update this screen to include the new commands from this section as well. 

I can tack that change on to this PR or make a new one, since it's just a UI and not related to the explicit handling code.

tolot27:
> I can tack that change on to this PR or make a new one, since it's just a UI and not related to the explicit handling code.

Please create a new PR for this. Maybe we can also implement a simple enumeration test case which checks for the existence of all tasker command implementations.

jamorham:
Can you briefly describe the changes that have been made regarding snooze and whether there is any risk to existing functionality given that this is a critically important part of the system for alerts etc? Thanks.

jwoglom:
@jamorham the only changes here involve adding additional arguments to the existing snooze tasker command, and adding new snooze high/snooze low tasker commands. This should not affect any existing usages of tasker, it only adds new functionality. 

The way in which the snooze is called from within the xDrip tasker plugin has changed very slightly, but it performs the same functionality and ends up calling the same codepath to actually perform the snooze.

PR #1722, opened 2021-05-12
G6 Temperature back

Navid200:
| Subject | Detail |
| ---------|-------|
|Why do we need this? | A G6 Firefly does not show the sensor temperature on the system status page now. <br /> After this fix, it will. |
| Is there a work-around? | The temperature is shown in the logs. |
| Are there any side-effects to this fix? | No |
| Tests | Has been tested with 2.27.2.98 and 2.27.2.103 firmwares. |

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/FirmwareCapability.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/FirmwareCapability.java
@@ -44,7 +44,11 @@ public class FirmwareCapability {
     }
 
     public static boolean isFirmwareTemperatureCapable(final String version) {
-        return !isG6Rev2(version) && !isG6Plus(version);
+        return !isG6Plus(version);
+    }
+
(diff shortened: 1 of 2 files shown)


tolot27 (approved):
Quite obvious. 👍

jamorham:
Can you let me know which firmware versions this has been tested with? Thanks

Navid200:
@jamorham I personally have tested this with the following two firmware versions.
2.27.2.98
2.27.2.103

Navid200:
Screenshot_20210521-090910

jamorham (approved):
This looks logically correct but beware that temperature is only updated every 12 hours so its not really that useful to the end user and is more of a curiosity I would think. Maybe it should be hidden behind engineering mode to avoid being misleading.

Navid200:
@jamorham Thanks
How much battery energy does the transmitter use as a result of an enquiry?
We currently submit an enquiry once every 12 hours.  What if we submit it once every half hour?  What will be the impact on the battery life of the transmitter?

Issue #1720, opened 2021-05-09
Dexcom G5: Scan Error: Unable to Start scanning (Pixel 4a 5G with Android 12 / S Preview)


parapenT1sta:
**xDrip Nightly build 5th May 2021**
**Dexcom G5 receiver**
**Mobile:** Google Pixel 4a 5G
**Android Version**: S (12 Preview) build bramble-spp3.210325.010-factory-b282d767

**Issue:** Scan Error: Unable to start scanning
**Error log:**
```
Unable to Start Scanning
com.polidea.rxandroidble2.exceptions.BleScanException: Bluetooth Cannot Start (code 0)
No transmitter mac address!
```

Dexcom G5 works perfectly fine with Dexcom Official App and also working fine with MiaoMiao 2 / Bubble 1 & Mini.
With xDrip+ the scanning is not even starting/working.  

![image](https://user-images.githubusercontent.com/608650/117578480-6ea1db80-b0e6-11eb-9eec-6921043802c7.png)

![image](https://user-images.githubusercontent.com/608650/117578552-d5bf9000-b0e6-11eb-827a-8a5cf80dde0a.png)

![image](https://user-images.githubusercontent.com/608650/117580239-00154b80-b0ef-11eb-9169-48306bd62dd5.png)

tolot27:
Please enable verbose logging for Ob1G5CollectionService.

parapenT1sta:
@tolot27 I will try it on the next time I update to Android 12.

Smip:
@parapenT1sta did you solve the problem? I have pixel 4 and after I updated the system have a problem like you.

parapenT1sta:
@Smip I downgraded from Android 12 to 11 and problem solved.

Smip:
@tolot27 
![photo_2021-05-19_22-59-07](https://user-images.githubusercontent.com/11481452/118883944-22c91080-b8f6-11eb-84eb-60002e985625.jpg)

tolot27:
Looks like either a permission problem or a new security layer. Did xDrip ever ask to bond the transmitter? I assume it is not bonded, yet.

Smip:
Probably xDrip didn't ask to bond. I used this transmitter with Android 11 and everything was good. After updating, xDrip worked for a while and after some time I got this error. I tried to change a lot of different settings and nothing helps. The last thing I tried is to reinstall the application. After installation of the app, I didn't see a bond request.

tolot27:
[...]

PR #1723, opened 2021-05-13
update rxandroidble2 to 1.12.1

tolot27:
This PR bumps RxAndroidBle to version 1.12.1 to fix a memory leak in [scanBleDevices](https://github.com/Polidea/RxAndroidBle/issues/607) which can cause the RxAndroidBle stack to hang.

It also add support to parse service solicitation UUIDs, which is required by the updated library but currently not used in xDrip.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -344,7 +344,7 @@ dependencies {
     implementation 'io.reactivex:rxjava:1.3.3'
     implementation 'com.activeandroid:thread-safe-active-android:3.1.1'
     //implementation 'com.github.lecho:hellocharts-android:v1.5.8'
-    implementation "com.polidea.rxandroidble2:rxandroidble:1.11.1"
+    implementation "com.polidea.rxandroidble2:rxandroidble:1.12.1"
     implementation 'com.google.guava:guava:24.1-android'
     implementation 'com.embarkmobile:zxing-android-minimal:2.0.0@aar'
     implementation 'com.embarkmobile:zxing-android-integration:2.0.0@aar'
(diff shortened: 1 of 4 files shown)


parapenT1sta:
@tolot27 Do you think will help fixing this issue with latest Android 12 Preview version? https://github.com/NightscoutFoundation/xDrip/issues/1720

tolot27:
> @tolot27 Do you think will help fixing this issue with latest Android 12 Preview version? #1720

No, I don't think so. I cannot test it because Bluetooth/Dexcom is not available in the simulator and I don't have a device which is capable of running Android 12.

BTW: It is running fine on my S4 (Android 10) for three days, so far.

parapenT1sta:
> > @tolot27 Do you think will help fixing this issue with latest Android 12 Preview version? #1720
> 
> No, I don't think so. I cannot test it because Bluetooth/Dexcom is not available in the simulator and I don't have a device which is capable of running Android 12.
> 
> BTW: It is running fine on my S4 (Android 10) for three days, so far.

Thank you for your reply. In fact only stopped working on Pixel 4a 5G after the Preview 3 version of Android 12, as soon I reverted to Android 11 started working right away.

tolot27:
@jwoglom Please can you review this PR as well?

jwoglom (approved):
@tolot27 Built an APK from this PR and tested it out on my old Pixel 2XL running Android 10, and it seems to function OK connecting to a G6 transmitter. I did run into some awkward behavior when initially pairing a transmitter, but I've run into that with stock xDrip before so I doubt it's related.

tolot27:
@jwoglom thanks for testing. It runs well on my phone for several days. During transmitter change it took a while to bond. But that also happens with the nightly build and no extra message is logged. Hence, it is unrelated to rxble.

jamorham:
I'll try this out here and see if it looks like everything still works okay.

jamorham (approved):

Issue #1239, opened 2020-02-15
G6 8G transmitter always shows Initial Calibration


jamespaulley:
Restarting a G6 sensor with an 8G transmitter using the sensor's code causes the fly-out menu to always display "Initial Calibration". Calibrations do seem to take, however, but I'm not totally sure it's working correctly.

https://photos.app.goo.gl/aMdnUtsfd8rdGaht9

jamespaulley:
Has this been addressed?

Navid200:
You are correct that it always show the initial calibration option even after you have done calibrations already.
Yes, the calibrations are sent to the transmitter.

Navid200:
You can ignore the messages.  Not, they still show up and have not been addressed yet.

Navid200:
I am closing this issue.  
There is already work in progress to update xDrip to provide correct messages.
You can use xDrip as is knowing what messages you can ignore.  This is not elegant and hopefully will be addressed soon.

PR #1726, opened 2021-05-16
Add resource string for the newly added Stop Sensor Confirmation

Navid200:
Adding a resource string for the stop sensor confirmation that was added with build [2021.05.16](https://github.com/NightscoutFoundation/xDrip/releases/tag/2021.05.16) (https://github.com/NightscoutFoundation/xDrip/commit/84e4376c601e52c561ab17a34f691136b62ef27d).

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1496,6 +1496,7 @@
     <string name="__do_you_want_to_change_settings_or_start_sensor">\n\nDo you want to change settings or start sensor?</string>
     <string name="change_settings">Change settings</string>
     <string name="do_you_want_to_delete_and_reset_the_calibrations_for_this_sensor">Do you want to delete and reset the calibrations for this sensor?</string>
+    <string name="sensor_stop_confirm">Do you want to Stop sensor and end the current session?</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1496,6 +1496,7 @@
     <string name="__do_you_want_to_change_settings_or_start_sensor">\n\nDo you want to change settings or start sensor?</string>
     <string name="change_settings">Change settings</string>
     <string name="do_you_want_to_delete_and_reset_the_calibrations_for_this_sensor">Do you want to delete and reset the calibrations for this sensor?</string>
+    <string name="sensor_stop_confirm">Do you want to Stop sensor and end the current session?</string>
```
tolot27, line 1499: Please change it into "Do you want to stop this sensor." IMHO, many users do not understand what "session" means. The term in not introduced, AFAIK.

tolot27 (changes requested):
Please always provide a description and refer the original commit (84e4376 in this case), if appropriate. Tanks!

app/src/main/res/values/strings.xml
Navid200: Done

Navid200:
This is an issue that I closed because it is an expected behavior.
https://github.com/NightscoutFoundation/xDrip/issues/1239

There are behaviors that we tell users to accept and put up with until we fix them.
Many such issues have already been addressed.

tolot27 (approved):
Since it is evident for non-English users to fully understand the consequences of the stop sensor confirmation dialog, it is important to provide a translatable resource string.

jamorham:
Since this string probably is unlikely to change much then this makes sense, in future we can use the `translatable="false"` suggestion for strings where the wording may change during the early stages of a feature.

jamorham (approved):

Issue #1415, opened 2020-08-21
Unable to fully perform calibration on 8K series G6 transmitter with latest nightly (67014b0-20202.07.25)


PedanticAvenger:
All attempts to calibrate result in "Got invalid impossible slope calibration!".
The calibration is accepted by the transmitter but xdrip doesn't register as such and stays in "initial calibration" stage.
Was working on previous transmitter (81 series)

Logs uploaded, calibration attempts start at 06:54, there are more than one.

PedanticAvenger:
> @PedanticAvenger : Actually I wanted to refer [to this thread](https://github.com/NightscoutFoundation/xDrip/issues/1310). Do you use the "syringe icon" for entering calibrations? If I remember correctly, I also had the issue of immediate rejections for some time. Ever since using only the syringe (and the settings as per thread in issue #1310) this issue never occurred again for me. Hoping this helps.

I'm digesting this thread now. Different but I see similarities. I'll continue to document things here for my case.  Further note, this is the first sensor for the 8k transmitter so maybe after a stop and start in a few days this will go away.

PedanticAvenger:
> > The calibration is accepted by the transmitter but xdrip doesn't register as such and stays in "initial calibration" stage.
> 
> Considering that you can only run 8K in native mode, which makes the transmitter responsible for calibration, if the calibration is accepted by the transmitter, what is the problem then?

The issue is that while I can accept this personally the vast majority of users would totally panic and xdrip should handle this. Either this is an edge case and we document it and learn or this is a bug and we figure it out and a fix is discovered and rolled out. The process itself is important to continue to build comfort with the medical community in solutions like this. That's all.

PedanticAvenger:
Ok, so the message on calibration attempt (every calibration, this example was exactly the same as G6 thought BG was at.)
![Screenshot_20200824-205349](https://user-images.githubusercontent.com/15637097/91352481-284e5b80-e7a7-11ea-9b15-de79b24f1ee2.jpg)

Log entries to correlate.
![Screenshot_20200824-205403](https://user-images.githubusercontent.com/15637097/91352503-2e443c80-e7a7-11ea-8dbf-99e884009d33.jpg)

Logs showing calibration was sent to transmitter.
![Screenshot_20200824-205805](https://user-images.githubusercontent.com/15637097/91352512-313f2d00-e7a7-11ea-9e09-f8faafd2fb1e.jpg)

Further log entries, not sure if this is correct behavior or not so included.
[...]

PR #1727, opened 2021-05-17
"Add Calibration" menu for a Firefly G6 instead of "Initial Calibration"

### LINKED ISSUE #1239: G6 8G transmitter always shows Initial Calibration

Navid200:
**Why do we need this?**
This is not a critical issue.
However, this PR makes the behavior of xDrip more intuitive for a G5 user who has switched to a Firefly G6 or a Dexcom app user who has switched to xDrip.  So, this makes life easier for new comers to Firefly on xDrip.

**What changes**
Only a Firefly user experiences a difference.  A G5, or old G6 (80XXXX or 81XXXX) user will experience no change.
The calibration option becomes "Add Calibration" if you are using a Firefly after you have received the third reading.
Screenshot_20210531-172141  

The following shows a calibration.
Screenshot_20210531-174935

**Tests**
Tested with a G6 Firefly
2.27.2.103

Fixes #1239
Fixes https://github.com/NightscoutFoundation/xDrip/issues/1415


app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
```diff
@@ -77,18 +77,23 @@ public NavDrawerBuilder(final Context context) {
...
                                 this.nav_drawer_intents.add(new Intent(context, Home.class));
                             }
                         } else {
                             if (BgReading.isDataSuitableForDoubleCalibration() || Ob1G5CollectionService.isG5WantingInitialCalibration()) {
-                                this.nav_drawer_options.add(context.getString(R.string.initial_calibration));
-                                this.nav_drawer_intents.add(new Intent(context, DoubleCalibrationActivity.class));
+                                if (last_two_bgReadings.size() > 1) {  // Firefly after second reading, now, gets add calibration.
+                                    this.nav_drawer_options.add(context.getString(R.string.add_calibration));
+                                    this.nav_drawer_intents.add(new Intent(context, AddCalibration.class));
+                                } else {    // G5 or non-native G6 after warm-up still get the option to perform initial calibration
```
tolot27, line 93: Please apply code formatting to get a consistent style, also for the spacing before comments.

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
```diff
@@ -77,18 +77,23 @@ public NavDrawerBuilder(final Context context) {
...
-                            } else {
+                            } else {  // G5 or old G6 not long after a calibration
                                 this.nav_drawer_options.add(context.getString(R.string.cannot_calibrate_right_now));
                                 this.nav_drawer_intents.add(new Intent(context, Home.class));
                             }
                         } else {
                             if (BgReading.isDataSuitableForDoubleCalibration() || Ob1G5CollectionService.isG5WantingInitialCalibration()) {
-                                this.nav_drawer_options.add(context.getString(R.string.initial_calibration));
-                                this.nav_drawer_intents.add(new Intent(context, DoubleCalibrationActivity.class));
+                                if (last_two_bgReadings.size() > 1) {  // Firefly after second reading, now, gets add calibration.
```
tolot27, line 90: I'm unsure whether this is the correct check, especially after `isG5WantingInitialCalibration()`. There are so many if/else here. Is there a state machine for these checks?

tolot27 (changes requested):
Thanks for this PR. Maybe, it is a good idea to create a state machine, first. The tests and alternatives are not easy to understand.

tolot27:
@Navid200 At the right pane, you can link this PR to an open issue, if any. You can also put Fixes #<number> in the description to achieve the same.

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
Navid200: Do you mean a flowchart?
How do I add a diagram to a code?
Or, do you mean something else?

Navid200: There is an OR statement between `isDataSuitableForDoubleCalibration()` and `isDataSuitableForDoubleCalibration`

I may have confused people by adding my comments to existing code.

I looked at the existing code and found where initial calibration is located.  That's where a Firefly triggers after its  third reading after a warm-up.   The only other device that ever gets to the same spot (initial calibration) is either a G5 or an old G6 in non-native mode.  Those are the only two devices that get initial calibration option under the top left menu.

However, none of those two have readings before initial calibration.
So, by adding that test, I can create a fork that sends G5 and old G6 to one branch of the fork, and Firefly to the other branch.

Navid200: >  I'm unsure whether this is the correct check, especially after isG5WantingInitialCalibration(). 

The test is not after `isG5WantingInitialCalibration()` logically.
There is an `else` on line 88.

Navid200:
This is an issue that I closed because it is an expected behavior.
#1239

There are behaviors that we tell users to accept and put up with until we fix them.
Many such issues have already been addressed.

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
Navid200: I did it.  Is it OK now?

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
tolot27: > Do you mean a flowchart?
> How do I add a diagram to a code?
> Or, do you mean something else?

No, I don't mean adding a diagram (i. e. as an image) to the code. I'd like to have a technical documentation which shows the Dexcom state machine and all things around it. This can be done without any graphics in the Wiki, i. e. using [mermaid](https://mermaid-js.github.io). It is just to understand the code much better and reduces the error probability.

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
tolot27: Sorry, to be pedantic. Everything in the code should have the same formatting and style, including comments. You still have two different comment styles: ` // comment` and ` //comment`.  Please be consistent. 

BTW: I really like your comment additions because it increases the logical understanding a lot.

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
Navid200: We can do that.  That will benefit many PRs.  
But, why are we talking about it here in this PR?

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
Navid200: Thanks for catching this.

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
tolot27: Because it is hard to review a nested and incomplete if/then/else statement were a further condition is added and no unit test is available. It is just a general suggestion and does not influence this PR. If we would have this kind of documentation, we would see immediately that the surrounding if clause is incomplete.

tolot27 (approved):
This PR changes the calibration behavior for the most common G6 Firefly transmitters to be more intuitive. It just overcomes the confusing initial calibration which is not necessary and not possible with G6 Firefly.

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
Navid200: I am going to learn mermaid and put this together.  What do you mean by doing it in the wiki?

tolot27: We need some place to store the tech doc. The wiki might be a good place because it does not require pull requests.

app/src/main/java/com/eveningoutpost/dexdrip/AddCalibration.java
```diff
@@ -195,16 +203,31 @@ public void onClick(final View v) {
                             final double calValue = JoH.tolerantParseDouble(string_value);
 
                             if (!Home.get_follower()) {
-                                Calibration calibration = Calibration.create(calValue, getApplicationContext());
-                                if (calibration != null) {
-                                    UndoRedo.addUndoCalibration(calibration.uuid);
-                                    //startWatchUpdaterService(v.getContext(), WatchUpdaterService.ACTION_SYNC_CALIBRATION, TAG);
-                                    //Ob1G5StateMachine.addCalibration((int)calibration.bg, calibration.timestamp);
-                                    NativeCalibrationPipe.addCalibration((int)calibration.bg, calibration.timestamp);
+                                if (FirmwareCapability.isTransmitterRawIncapable(getTransmitterID())) { // Firefly only
```
jamorham, line 206: Can you explain the changes in `AddCalibration` is this because creating a calibration results in a null if the sensor sanity tests fail on the raw value? So you're creating a blood test entry instead?

Navid200: My first step was to add the add calibration option in the menu only for a Firefly.  This was done by the added fork to NavDrawerBuilder.java.
After I did that, I saw that the calibrations still did not go through for a Firefly.
I realized that Calibration.java had a fork that checked for raw sanity and just rejected a Firefly.

My first attempt to fix that was to modify Calibration.java.  It became too complex.  So, I decided to take a different approach.  I created a fork in AddCalibration.java.  This fork side is only taken for a Firefly.  A G5 or an old G6 just goes through as before.

The Firefly side of the fork, in AddCalibration,java needs to do everything that is done in AddCalibration.java as well as what is done in Calibration.java that is not executed for a Firefly.  Those are creating a blood glucose reading point on the plot as well as sending the calibration to the transmitter.

If there is a more graceful way of accomplishing this, please tell me.  I am so sorry if this is messed up.  
I can also add more comments to the code to explain all of this better if you like.

Thanks for looking.

jamorham:
@Navid200 Do you think this is ready to go? Have you tested it okay?

Navid200:
@jamorham I have tested this with my Firefly only.  I have only tested it in the original code mode.  My plan is to test it in the no-code mode tomorrow when the time comes for another restart.

I have a few old G5 transmitters.  But, I don't have a workshop where I can replace the batteries.  I also don't have simulators so that I can test without having to get sensors and insert the sensor on my skin.
I don't have any old G6 that can be hard reset.  Therefore, I have not way of testing this with G5 or old G6.

A Nightly release would be good if we used the Nightlies for testing by the user community.  I would announce it on facebook right before release if you tell me and ask G5 and old G6 users to update knowing that it has not been tested.

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
```diff
@@ -69,26 +69,31 @@ public NavDrawerBuilder(final Context context) {
...
+                            } else {  //G5, old G6 or Firefly, in no-code mode, not long after a calibration
                                 this.nav_drawer_options.add(context.getString(R.string.cannot_calibrate_right_now));
                                 this.nav_drawer_intents.add(new Intent(context, Home.class));
                             }
-                        } else {
+                        } else { //If the has been no two initial calibrations
                             if (BgReading.isDataSuitableForDoubleCalibration() || Ob1G5CollectionService.isG5WantingInitialCalibration()) {
-                                this.nav_drawer_options.add(context.getString(R.string.initial_calibration));
-                                this.nav_drawer_intents.add(new Intent(context, DoubleCalibrationActivity.class));
+                                if (last_two_bgReadings.size() > 1) { //Firefly G6, in factory calibration mode, after third reading
```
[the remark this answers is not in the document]
Navid200: I have made changes since this question was asked.  And the question was being asked from someone else.
Is there still a question, for me, after the changes?

jamorham:
Ok, I'm planning to merge this for a nightly.

Navid200:
@jamorham Sounds good.
I will announce on facebook now.

Navid200:
@jamorham Done:
Screenshot 2021-08-04 141059

PR #1728, opened 2021-05-17
Stop sensor on master disable switch

Navid200:
The follower can stop sensor on master only if the existing switch on master is enabled.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
@@ -56,6 +56,8 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 
+import static com.eveningoutpost.dexdrip.xdrip.gs;
+
 import static android.support.v4.content.WakefulBroadcastReceiver.completeWakefulIntent;
 import static com.eveningoutpost.dexdrip.Models.JoH.isAnyNetworkConnected;
 import static com.eveningoutpost.dexdrip.Models.JoH.showNotification;
@@ -542,8 +544,12 @@ public class GcmListenerSvc extends JamListenerSvc {
                 } else if (action.equals("ssom")) {
                     if (Home.get_master()) {
                         if (payload.equals("challenge string")) {
-                            UserError.Log.e(TAG, "Stopping sensor by remote");
-                            StopSensor.stop();
+                            if (Pref.getBoolean("plus_accept_follower_actions", true)) {
+                                UserError.Log.e(TAG, "Stopping sensor by remote");
+                                StopSensor.stop();
+                            } else {
+                                UserError.Log.wtf(TAG, gs(R.string.not_follower_switch));
+                            }
                         } else {
                             UserError.Log.wtf(TAG, "Challenge string failed in ssom");
                         }
@@ -581,14 +587,14 @@ public class GcmListenerSvc extends JamListenerSvc {
             return;
         }
         LibreBlock.Save(lb);
-        
+
         PersistentStore.setString("LibreSN", lb.reference);
-        
+
         if(Home.get_master()) {
             if (SensorSanity.checkLibreSensorChangeIfEnabled(lb.reference)) {
                 Log.e(TAG, "Problem with Libre Serial Number - not processing");
             }
-            
+
             NFCReaderX.HandleGoodReading(lb.reference, lb.blockbytes, lb.timestamp, false, lb.patchUid,  lb.patchInfo);
         }
     }
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1387,8 +1387,9 @@
     <string name="summary_show_home_on_boot">Bring up xDrip screen automatically after device boots</string>
     <string name="title_show_home_on_boot">Show xDrip on Boot</string>
     <string name="eula">EULA</string>
-    <string name="summary_plus_accept_follower_actions">Treatments and Calibrations from followers will be accepted</string>
+    <string name="summary_plus_accept_follower_actions">Treatments, Calibrations and remote sensor stop from followers will be accepted</string>
     <string name="title_plus_accept_follower_actions">Accept follower actions</string>
+    <string name="not_follower_switch">Follower actions switch is disabled</string>
     <string name="summary_plus_whole_house">Participate in a Whole House Network</string>
     <string name="title_plus_whole_house">Whole House</string>
     <string name="summary_libre_whole_house_collector">This phone will be a collector in a whole house network</string>


app/src/main/res/values/strings.xml
```diff
@@ -1387,8 +1387,9 @@
     <string name="summary_show_home_on_boot">Bring up xDrip screen automatically after device boots</string>
     <string name="title_show_home_on_boot">Show xDrip on Boot</string>
     <string name="eula">EULA</string>
-    <string name="summary_plus_accept_follower_actions">Treatments and Calibrations from followers will be accepted</string>
+    <string name="summary_plus_accept_follower_actions">Treatments, Calibrations and remote sensor stop from followers will be accepted</string>
```
jwoglom, line 1390: I think this might be cleaner as "Treatments, Calibrations and Sensor Stop from followers will be accepted"

app/src/main/res/values/strings.xml
```diff
@@ -1387,8 +1387,9 @@
...
     <string name="title_plus_accept_follower_actions">Accept follower actions</string>
+    <string name="not_follower_switch">Follower actions switch is disabled</string>
```
jwoglom, line 1392: This message could probably be clearer as well. Also, this is only being used when a follower attempts to stop a sensor, right? Not any of the other locations where a follower performs a disallowed action?

Perhaps "Rejected action from follower because follower actions are disabled", in case you want to future-proof this message to use it in the other contexts as well.

app/src/main/res/values/strings.xml
Navid200: Thanks.  Updated

app/src/main/res/values/strings.xml
Navid200: Yes, this is a log on the master only when stop sensor is rejected.  Updated.  Thanks

tolot27 (commented):
Hmm, do we really need log translations which needs much more work to the translators? Currently, non of the log messages are translated.
If we keep the possibility, we could tag the resource strings as [non-translatable](http://tools.android.com/recent/non-translatablestrings).

Navid200:
> Hmm, do we really need log translations which needs much more work to the translators? Currently, non of the log messages are translated.
> If we keep the possibility, we could tag the resource strings as [non-translatable](http://tools.android.com/recent/non-translatablestrings).

Sorry, I didn't know that.
I can remove that and make it English only.
Should I do that?  No problem.  I just didn't know.

tolot27:
> Should I do that? No problem. I just didn't know.

Not so fast. Looks like jwoglom is supporting as further and we can discuss this. 😄

@jwoglom What is your opinion about log translations or at least creating non-translatable resource strings for it?

Navid200:
When I help on facebook, often I need to see the logs.  if the logs were translated, it would make the support task harder.
I wasn't thinking when I added this resource string.
I'd like to remove it and make it consistent with the rest of the logs.

app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
```diff
@@ -542,8 +544,12 @@ public void run() {
                 } else if (action.equals("ssom")) {
                     if (Home.get_master()) {
                         if (payload.equals("challenge string")) {
-                            UserError.Log.e(TAG, "Stopping sensor by remote");
-                            StopSensor.stop();
+                            if (Pref.getBoolean("plus_accept_follower_actions", true)) {
+                                UserError.Log.e(TAG, "Stopping sensor by remote");
+                                StopSensor.stop();
+                            } else {
+                                UserError.Log.wtf(TAG, "Stop sensor by follower rejected because follower actions are disabled");
```
tolot27, line 551: Is this really a WTF? WTFH (What a terrible failure) is more severe than error log. The error which should never ever happen. It may force the app to hold for writing the logs before termination.

Navid200: What are the alternatives?  
I just copied the format from another adjacent log, "Challenge string failed for ssom".

jwoglom:
> @jwoglom What is your opinion about log translations or at least creating non-translatable resource strings for it?

Thinking about it more broadly, it'd be impossible to translate _every_ log message, and it would be weird to have some translated log messages and some non-translated, so we probably shouldn't do it. That being said, for WTF-level specific logs maybe it would be useful in some contexts. If there are flows that currently WTF log that we think might be better suited as notifications or popup messages of some kind, then there could be value in translating them. 

But, for now, probably not worth the effort to translate these strings and we should probably remove this one.

app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
Navid200: I thought more about this.  This log is generated only when there is a disagreement between a follower and the master.    Either the master is a kid who needs to enable the switch to allow the parent to control, or the master is a grown-up who needs to change the security code to disable an idiot from following.
Either way, this is serious situation that requires immediate attention.  So, I would leave it as wtf.

tolot27: But there is no notification about this. It is simply a log entry at the collectors side. Neither the sensor-wearing user nor the remote controller knows about the stop failure.

log.wtf is not a categorization at the user level. It is a categorization from the point of program flow and in this case it is not a terrible failure. Please change it to `UserError.Log.e()` or better to `UserError.Log.w()`.

app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
```diff
@@ -542,8 +544,12 @@ public void run() {
                 } else if (action.equals("ssom")) {
                     if (Home.get_master()) {
                         if (payload.equals("challenge string")) {
-                            UserError.Log.e(TAG, "Stopping sensor by remote");
-                            StopSensor.stop();
+                            if (Pref.getBoolean("plus_accept_follower_actions", true)) {
+                                UserError.Log.e(TAG, "Stopping sensor by remote");
```
tolot27, line 548: Logging sensor stop by remote as an error is also not correct because the stop is intentional. I would categorize it as info or at maximum as a warning.

Navid200: Is info "i"?

Navid200: Wait a minute!
That's not my code.  That was there already.  You are asking me to change someone else's code.  You know that, right?

app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
Navid200: Done

app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
Navid200: Done

tolot27 (approved):

app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
tolot27: > That's not my code. That was there already. You are asking me to change someone else's code. You know that, right?

Indeed, yes. The whole file is not your code. That's open source. :smirk:

tolot27 (approved):
@jamorham As discussed in the dev group, this PR implements the workaround.

app/src/main/res/values/strings.xml
```diff
@@ -1387,7 +1387,7 @@
     <string name="summary_show_home_on_boot">Bring up xDrip screen automatically after device boots</string>
     <string name="title_show_home_on_boot">Show xDrip on Boot</string>
     <string name="eula">EULA</string>
-    <string name="summary_plus_accept_follower_actions">Treatments and Calibrations from followers will be accepted</string>
+    <string name="summary_plus_accept_follower_actions">Treatments, Calibrations and Sensor Stop from followers will be accepted</string>
```
jamorham, line 1390: I think this should just say "other actions" instead of "Sensor Stop" as otherwise it just has to be re-translated if that functionality is ever extended - for example if starting sensor remotely was added.

app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
```diff
@@ -56,6 +56,8 @@
 import java.util.List;
 import java.util.Map;
 
+import static com.eveningoutpost.dexdrip.xdrip.gs;
```
jamorham, line 59: I don't think this import is used is it?

app/src/main/res/values/strings.xml
Navid200: Done

app/src/main/java/com/eveningoutpost/dexdrip/GcmListenerSvc.java
Navid200: Thanks for catching this;  sorry about that; removed.

Issue #650, opened 2018-10-29
25 hour Main Screen


DunePlodder:
The main xDrip screen currently covers 24 hours, but I often want to look back at what I did at this time (or a little before) yesterday, to see what I did & how effective it was, or what I did wrong so I don't repeat it. Currently I use the history screen but a 25 hour screen might be a little more convenient?

jamorham:
It is an interesting idea.

Navid200:
I have been there.  Interesting idea.

emp-00:
+1 --- enhancement highly appreciated --- 
I also posted this in #1273 (was closed due to duplicate with this issue)

Navid200:
I'm going to close this for now.  We can always reopen a closed issue.  This is my thinking.

In order to decide what to do, a better approach may be to look at the percentile chart that shows an overlap of possibly the last 90 days.
I understand you may have changed a practice so dramatically recently that the past 90 days are not a valid representation.
You are not limited in using the percentile chart to 90 days.  You can also set it to the last 30 days or the past 7 days.

Please note that implementing this, which I almost di will have a negative consequence.
We currently have an issue with a very high reading causing the y axis to extend making the slopes to look wrong.  This lasts 24 hours.  Meaning you will have to put up with the wrong (visual) slopes for 24 hours before the y axis goes back to normal.
If we implement this feature, we will then have to wait 25 hours instead of 24 hours.

Issue #1273, opened 2020-03-15
Implement 36h Trend (>24h) as quick alternative to History


emp-00:
Currently the max timespan in the xDrip+ main window is 24 h. Very often I would like to see my carbs and bolus just slightly before 24h. For example: When I start eating my breakfast only 5 mins later than the day before I have to use the History function to check which dose I used the day before and how the trend evolved. Looking up "what happened yesterday" is very important in daily life (for me).

It would therefore be great to add functionality extending the timespan in the main window e.g. to 36 h (3, 6, 12, 24, 36h).  Anybody else like this idea?

ANDREAPIPPI:
I absolutely agree with emp-00. 
This could be very useful on some occasions.
Pippo

Navid200:
This is a very good idea.  Would you please add your voice to the existing issue? 
https://gitter.im/jamorham/xDrip-plus
I am going to close this.  But, if you believe this should have a different issue, you can always reopen.

PR #1729, opened 2021-05-18
Change the length of time covered on screen from 24 hours

Navid200:
**Status**
The span of time shown on screen, by default, is 2.5 hours.  However, this expands if low prediction is predicting a value.  These two spans are affected by the changes this PR makes.  So, the 2.5-hour default increases if we increase the total available span from 24 hours to 36.   
Work is in progress to figure out why this happens and disable it.  The behavior of xDrip should not change by this PR other than the intended change.  

**What is the change?**
If you scroll back on screen, you can only go back to 24 hours ago.  After this, you can go back more or less based on a setting.

Fixes https://github.com/NightscoutFoundation/xDrip/issues/650

**Why do we need this?**
We don't!  It is a feature request.  It can be handy.

**Why each file changed**
BgGraphBuilder is where the chart boundaries are defined.

Home had 24 as the length of time covered on screen.  It had to be changed to the variable so that the time buttons would function properly.  The default chart width of 2.5 hours would not reset correctly until the last changes.

arrays includes the available choices for the new setting.

strings allows the title of the new setting to be translated.

xdrip_plus_prefs includes the new setting implementation

**Detail**
Please see the post below:  https://github.com/NightscoutFoundation/xDrip/pull/1729#issuecomment-855171796


jwoglom:
Just wanted to ask -- any reason not to do something more round such as 36 hours? Obviously the decision here is fairly arbitrary, and I imagine it might impact performance slightly. But 36 hours would give a bit more of a range to see what happened yesterday around the same time.

As much as I hate to suggest adding yet another preference field, this could also be a decent candidate for a user-changeable setting.

Navid200:
I will strongly object to yet another new setting.
Every time we add a setting, we should make sure it is absolutely necessary.

There are already too many settings in xDrip.  As soon as xDrip stops readings, there are users who start flipping settings hoping to fix the problem.
Then, they post on facebook asking for help.
When I ask them if they have changed any settings, they say yes they did and they don't remember which ones!

This issue has been sitting there since 2018.  I wish you made a comment there about your preference of 25 versus 24, not now that there is a PR.

Unless there is a very very good reason for 36 instead of 25, I see no reason 25 is not adequate.

tolot27:
> Unless there is a very very good reason for 36 instead of 25, I see no reason 25 is not adequate.

Argue the other way around is valid, too. 😏 Hence, we need decision criteria. @jwoglom and feedback from #1273 implies a longer time.

Navid200:
No, it's not valid.  the only proposal was 25 hours for years.  Everyone could have viewed the issue and post their opinion.

I am doing my best to reduce the number of open issues.  However, not everyone helps and many of the issues that re pointless remain open because of the constant disagreements.
So as an example, the same person who complains that there are too many open issues and that's why developers have no time looking at them also objects when I want to close a redundant issue.

I value all feedback here.
So far, I haven't seen any reason why going from 24 to 25, which is the only thing on the table, is a bad idea.

jwoglom:
I don't think anyone would disagree that upping the displayed length from 24 to 25 hours is a bad thing. I was just hoping to raise a suggestion that might better solve the original use case of "want[ing] to look back at what I did at this time (or a little before) yesterday to see what I did."

Like you mentioned there is no point wasting time on relatively pointless arguments when either option would be a user-facing improvement from the status quo. I just think that perhaps going "big and bold" here and showing 1.5 days might end up being a better use of our time relative to the end-user impact than only increasing the view by 1 hour.

Navid200:
I don't know the impact on performance.  I am happy with 36 hours as long as we know there is no negative impact on performance.

tolot27:
I'll do some tests over the weekend. I don't expect any performance decrease but there is a visual impact on smaller screens if the chart preview is enabled.
@Navid200 Does your PR also extends the chart preview?

Navid200:
@tolot27 This PR does not affect the chart preview.
I only intended to allow the user to scroll back more than 24 hours.
I thought of testing 25 versus 36 myself.  But, I didn't know how.

jamorham:
The more data that is processed the more overhead there will be, particularly with some of the libre trend stuff where I think there can be data rendered every minute. 

Some people are using on really tiny devices so any change would need to be mindful of that and I would prefer to introduce a setting that allows the default of 24 hours to be changed for the preview graph and then users can enable that if they want and their device is large enough to handle it properly.

We also need to ensure that there are not any other unintended side effects. I can't immediately think of any but it would need to be considered.

tolot27:
> I thought of testing 25 versus 36 myself. But, I didn't know how.

The easiest would be to create a mock database containing raw BG values from Libre (every minute) and the resulting calculated BG values as well as a lot of treatments. Maybe some basal info as well. With this database, we can test the performance and whether there are some side effects.

Having a setting for this feature would be appropriate in this case. But it should be limited to the range 24 - 36 hours. More than one 25 hours are especially useful on traveling to the east into a time zone which is some hours ahead.

tolot27:
We could also add the `<profilable android:shell="true" android:enable="true" />` to the `AndroidManifest.xml` and use a profiling tool like [perfetto](https://developer.android.com/studio/command-line/perfetto) to determine the resource utilization.

Navid200:
What I really wanted to do was to only extend how far back in time the user can scroll the main screen.
I didn't want to also extend the width of the preview chart.  But, after I opened this PR, and read the reviews, I realized that I had done that.

So, it is embarrassing that I didn't test this enough before opening a PR.
Anyway, I have been looking at it for a while to see how I can do exactly what I wanted.  I need more time.  So, I changed this to a draft.

I have no issue with changing this from 25 to 36.
My hope is that if the width of the preview chart remains at 24 hours, the performance should not be affected.   Neither should a smaller device behave any differently than it does now.
Or, is there a reason still for the performance to drop?

Navid200:
I have thought more about this.  I now think this is not a good idea.

First, we can already see what we did yesterday or last week or any time using History.  It's not like this change is providing access to something that was unattainable before.  it only provides a slight convenience.

But, this is done at a price.  If your glucose goes over 250 or below 40, the max or min shown on screen extends to be able to show your readings.  This change of the vertical axis scale lasts for 24 hours.  So, as long as the readings that have gone beyond the default min and max are still included in the 24-hour period that is being shown, the vertical axis remains modified.
If we extend the period to 25 hours (or 36 hours), you will have to wait even longer for the vertical axis to reset.

It just seems to be too much trouble to offer something that is not really lacking.  Again, we can use history to see what we did a little bit more than 24 hours ago.

I am curious of anyone has a serious objection to my new thinking.
I would like to close this PR as well as the issue.
May I have your opinion please?

jwoglom:
@Navid200 I hadn't considered the axis change this would result in until you mentioned it. In the past, I've definitely found it annoying when 12-24 hours after having an extremely high BG, I'm still more-or-less reminded about it by the adjusted axis on the graph. So I think a 36 hour timeline would in fact get a bit annoying in that situation, at least.

I think that behavior suggests the ideal paths forward here are either going up to 25 hours or not changing this at all. Or, what if we could add a button to jump to the history of the previous day that only appears when you scroll back all the way to the left edge of the graph, to make it easier to switch between the main screen and the history when looking at yesterday?

Navid200:
There is now a setting that let's you choose the time span available on screen with 24 hours as default:
Screenshot_20210604-221050

Screenshot_20210604-224446

After a change, the user needs to restart for it to take effect.  I'm not sure how to make it take effect with no need to a restart.

I have tested it on an Android 8 and an Android 9 phone and Android 11 virtual device.

If you choose a time span of 12 hours, the 24-hour (optional) time button will behave the same as the 12-hour (optional) time button.

jamorham:
Thanks for adding the configuration option and defaulting it to the existing standard so that there are no surprising user interfaces changes. I think that is important.

I want to approve this PR but because there are a lot of lines changed which look like whitespace formatting only, I need a statement of intent that those lines are just whitespace only changes and there is nothing else happening in this PR other than the functional changes already mentioned. Without this I'd need to review every line changed manually to confirm that those are purely whitespace changes. Thanks.

Navid200:
@jamorham Sorry about that.  I use auto formatting and for some files, that makes lots of changes.
The following lists the only changes, that re not formatting for the first 2 files:

Home.java: 2016, 2029, 2034, 2196, 
BgGraphBuilder.java: 98, 108, 112, 172-182, 190, 2104, 

I have to say that I am not quite happy with the behavior of the default time span.  I have made it really to be the default.  Meaning, it will stay at 2.5 hours now.  But, before this change, the behavior is weird.  Sometimes the view is 2.5 hours; sometimes it's 6 hours.  And I am not sure why it changes.  Some users may notice this change.  I'm not sure if everyone is going to be happy with it.

Considering this is a nightly, I wouldn't stop because of this.  But, I know you want to be 100% sure.  So, I had to be transparent.
I can work on this more if you think needed.

Navid200:
Let me explain in more detail what I mean.
If I long press the xDrip icon and make the time buttons visible and long press on 6 hours, for example, the screen will zoom in to 6 hours.   If I then pinch the screen to zoom in, after a few seconds, the zoom size will automatically go back to 6 hours.  So, long pressing a time button makes the screen remember it, which is very nice.

But, If I remove the time buttons from the screen, they have no effect.  I wish the long pressed button would stay active even when the buttons were not showing.  But, that's a different discussion.

Anyway, with no time buttons on screen, the length of time the screen zooms in varies from time to time and I cannot put a logic into it.  I cannot explain when it shows what value.

When I first started working on this PR, I noticed that the effective zoom value was being affected by changing the total length from 24 hours to something else.  
I did not like that.  I wanted the zoom behavior to be independent of the total length of time included.
I made things predictable, and independent of the selected time_span, by changes I made to lines 2029 and 2034.  
But, this has removed the randomness, which may have been intentional even though I don't understand it.

Navid200:
https://github.com/NightscoutFoundation/xDrip/pull/2081

Issue #403, opened 2018-04-22
Rest-Api - Nightscout data download not working


muehle28:
"Also try to download data from Nightscout" is activated, but no data from nightscout is beeing downloaded.

Splitfox:
Is there already a solution, because I have the same Issue

bobhavlin:
I too have this problem.  Treatments are downloaded from Nightscout and can be seen in Xdrip, but not BG data.  Thus, it does not appear to be a Nightscout configuration problem, but an Xdrip BG data processing/acceptance issue.  "Nightscout Sync (REST-API)" -> "Download data" is ON.  This is version d1d4887-2018.09.08.

anfichtn:
I'll have a similar problem. Treatments get backfilled when I set the download to on, but cgm data keeps missing.

ikniwn:
Did this get resolved?  I have the same issue - I can see treatments but no BG data.  I am using a Huawei p8 lite as a follower.  Base URL data is correct and Night Scout is receiving the uploads from the master.

Navid200:
@psonnera

psonnera:
I believe the menu item "Download data" should be renamed "Download treatments" or xDrip+ should accept incoming BG as it is and not bloodtest.

Navid200:
xDrip cannot download BG readings from Nightscout.
You will need to transfer database from the old phone to the new phone.
I have opened a pull request to change the menu to avoid this confusion as suggested by Patrick.

I am closing this issue as the capability does not exist.

jwoglom:
It strikes me as odd that xDrip, to my knowledge, doesn't support automatically downloading treatments (or CGM readings, although I don't have a direct need for that) on some periodic schedule from Nightscout. In my own workflow, I have pump treatments which upload directly to Nightscout via tconnectsync but unless I manually go into the xDrip settings and hit this download treatments button, they aren't ingested into xDrip. Ideally, whenever xDrip already uploads CGM data to Nightscout it could query to see whether there were new treatments to download and automatically ingest them to xDrip, if enabled by the user.

Automatically downloading treatments is a feature I was planning to build out for my own personal use, if it was implemented would others make use of it?

tolot27:
I do not fully understand the use case. The Nightscout Follower data source is designed to fetch the data from NS, both BG readings and treatments as well as notes.
Now, you are requesting the same feature from Cloud Upload Agent "Nightscout Sync (REST-API)", aren't you?

[...]

PR #1731, opened 2021-05-18
Change Nightscout "Download data" to "Download treatments"

Navid200:
Related: https://github.com/NightscoutFoundation/xDrip/issues/403

xDrip cannot download CGM readings from Nightscout.
There is an option, on the Nightscout upload page, that reads "Download data".

When I saw that the first time, I thought I could download all my CGM BG readings from Nightscout into xDrip on a new phone.  But, xDrip cannot do that.  The title of the option can be misleading as you can see from the linked issue.

I have changed the menu to look as shown below following Patrick's suggestion (please see [his post in the issue](https://github.com/NightscoutFoundation/xDrip/issues/403#issuecomment-782720963)).
Screenshot_1621366736

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1194,8 +1194,8 @@
     <string name="manual_overwrite_not_possible">Manual overwrite not possible!</string>
     <string name="upload_even_when_using_mobile_data">Upload even when using mobile data</string>
     <string name="use_mobile_data">Use mobile data</string>
-    <string name="summary_cloud_storage_api_download_enable">Also try to download data from Nightscout</string>
-    <string name="title_cloud_storage_api_download_enable">Download data</string>
+    <string name="summary_cloud_storage_api_download_enable">Also try to download treatments from Nightscout</string>
+    <string name="title_cloud_storage_api_download_enable">Download treatments</string>
     <string name="summary_bluetooth_meter_for_calibrations_auto">Calibrate using new blood glucose readings if the conditions appear right to do so without asking confirmation (experimental)</string>
     <string name="title_bluetooth_meter_for_calibrations_auto">Automatic Calibration</string>
     <string name="title_rest_api_extra_options">Extra Options</string>


jwoglom:
This comment isn't so much a reply to this PR as it is a comment on the download treatment functionality this PR references itself, but hopefully it's an okay place to raise it.

It strikes me as odd that xDrip, to my knowledge, doesn't support automatically downloading treatments on some periodic schedule from Nightscout. In my own workflow, I have pump treatments which upload directly to Nightscout via [tconnectsync](https://github.com/jwoglom/tconnectsync) but unless I manually go into the xDrip settings and hit this download treatments button, they aren't ingested into xDrip. Ideally, whenever xDrip already uploads CGM data to Nightscout it could query to see whether there were new treatments to download and automatically ingest them to xDrip, if enabled by the user.

This is a feature I was planning to build out for my own personal use, do you think it's worth opening a new GitHub issue to track this or are you aware of it being suggested or asked for before?

tolot27:
@jwoglom Your question is better placed in the issue. Please move it to 403.

tolot27:
@Navid200 @jamorham Maybe, we should rename the "Nightscout Sync" into "Nightscout Uploader" as well.
The "Sync" is misleading and implies a bidirectional communication.

Navid200:
@tolot27 I believe you can download treatments.  It's just that you cannot download BG readings.  So, it is not just upload.

tolot27:
> @tolot27 I believe you can download treatments. It's just that you cannot download BG readings. So, it is not just upload.

Indeed, but it is listed in the "Cloud Upload" section and "Nightscout Follower" is for following data. If we enhance "Nightscout Sync" to download BG readings as well, we would have two redundant agents, which is IMHO not necessary.

Navid200:
>Maybe, we should rename the "Nightscout Sync" into "Nightscout Uploader" as well.
The "Sync" is misleading and implies a bidirectional communication.

Why?  
It is bidirectional.

tolot27:
> > Maybe, we should rename the "Nightscout Sync" into "Nightscout Uploader" as well.
> > The "Sync" is misleading and implies a bidirectional communication.
> 
> Why? It is bidirectional.

Because it is listed below Uploader and is not fully bidirectional, indeed, partially but not full. Maybe, we should discuss merging Nightscout Sync with Nightscout Follower instead of wording. 😏

tolot27 (approved):
This PR improves wording and improves the understanding what the option really provides (if enabled).

jamorham:
It is also possible to use NSClient which I think syncs in both directions but I believe the xDrip implementation is more power efficient.

Issue #874, opened 2019-05-12
Bg Data Table


doc77:
When a bad reading is flagged in the Bg Data Table, would it be possible to turn the dots on the main screen graph a different color?  This would make bad readings be differentiated from valid ones.

tolot27:
BTW: How do you flag a BG reading as bad? Manually, by changing the entry in the SQlite DB?

doc77:
You mark it as bad in the Bg Data  Table (readings). This is still a feature request, and one that I don't imagine would be too difficult to institute.

ARkaine2:
If leaving post here can speed up realizing this FR, then I leave it here.

Navid200:
@ARkaine2 This feature was added more than a year ago.  
Do you have a question about it or does it not work for you or what?

ARkaine2:
**@Navid200**

Sometimes I get "jumping" BG Data (Reading) and it makes main view vertically narrow.
I'd like to fix it.
The most logic way I see is to move certain BG Data (Reading).
But is(n't) it possible?

![photo_2022-12-29_23-00-52 с](https://user-images.githubusercontent.com/31447095/210011017-60f908b4-585a-4764-bdfd-fd5f25985cae.jpg)

Navid200:
@ARkaine2 That was never the request.  It was to change the color.
But, your idea is interesting.
If I were you, I would open a discussions (not an issue), and we can talk about it.
I will already go ahead and ask to see what the others think about it.  Thanks

Navid200:
@ARkaine2 There is already an open issue for this.  You can add your suggestions there:  https://github.com/NightscoutFoundation/xDrip/issues/2173

ARkaine2:
[Navid200](https://github.com/Navid200)
:) So, should I use time buttons to force the x axis range?
Of course I'll try, but ATM it's difficult (not possible) to see/get jump BG Data (Reading), but I'm all interested.
Or NVM, then I see it later.

Navid200:
@ARkaine2 Unless I am misunderstanding you, this is not related to the x axis.  It is related to the y axis.  And there is nothing you can do about it as is.

ARkaine2:
[Navid200](https://github.com/Navid200), oh I got that. That's sad. :)

ARkaine2:
[Navid200](https://github.com/Navid200)
I got that. That's sad.
If it can be realized, know, that I need that.
Thanks.

PR #1732, opened 2021-05-19
Different color for marked bad readings

Navid200:
Fixes https://github.com/NightscoutFoundation/xDrip/issues/874

If you use the data table to mark a reading as bad, the reading is not used for statistical analysis.  However, there is no visual distinction for such readings on the main screen.
This update changes the color of such readings to a different unique color.  The user can also tweak the color the same as low, in range and high.  

The following image shows an example where low, in range, high and bad colors can be seen side by side.
  

I changed the color to white to make it distinctly different from the existing 4 colors. 
Screenshot_20210529-154434

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
@@ -146,6 +146,7 @@ public class BgGraphBuilder {
     private final List<PointValue> remoteValues = new ArrayList<>();
     private final List<PointValue> highValues = new ArrayList<>();
     private final List<PointValue> lowValues = new ArrayList<>();
+    private final List<PointValue> badValues = new ArrayList<>();
     private final List<PointValue> pluginValues = new ArrayList<PointValue>();
     private final List<PointValue> rawInterpretedValues = new ArrayList<PointValue>();
     private final List<PointValue> filteredValues = new ArrayList<PointValue>();
@@ -701,6 +702,7 @@ public class BgGraphBuilder {
 
             lines.add(remoteValuesLine()); // TODO conditional ?
             lines.add(backFillValuesLine()); // TODO conditional ?
+            lines.add(badValuesLine());
             lines.add(inRangeValuesLine());
             lines.add(lowValuesLine());
             lines.add(highValuesLine());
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -744,6 +746,15 @@ public Line highValuesLine() {
         return highValuesLine;
     }
 
+    public Line badValuesLine() {
+        Line badValuesLine = new Line (badValues);
```
tolot27, line 750: CF: Please remove the space before `(`.

tolot27 (commented):
This PR introduces a new setting which is in contrast to your objections reducing the number of settings.

Nevertheless, a visual distinct representation of unused BG values is a good hint, IMHO.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
Navid200: Thanks for catching that.  I will never forget to do code formatting again.

Navid200:
My objection is to adding settings that are unnecessary.
If this is unnecessary, I am happy to remove it.  It will then mean the color will be dictated and cannot be adjusted by the user.  The setting is buried several levels down and no were near any setting that could affect Bluetooth or any other connectivity-related settings.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
tolot27: Uhh, now a lot more changes were introduced which distracts a little from the main changes. Nevertheless, one more consistently formatted file.

tolot27 (approved):
This PR improves the visual differentiation of flagged bg readings as bad. The main logic can be seen in 8de959c. The last commit just applies consistent code formatting.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
Navid200: Are there still issues?

jamorham:
There is nothing in particular wrong with this implementation, it doesn't look like it would introduce much overhead other than the additional setting but I've never personally flagged a reading as bad and so I'm not sure exactly the use-case or more simply why this is needed?

Navid200:
@jamorham I personally have no need for it either.
But, we have an open issue for it.  I could be wrong.  But, the person asking for it could be someone who has helped many over the years on a different platform.   
I'm looking at old open issues and if I don't see any seriously wrong with the request, and if it is easy to implement without adding to many lines and complexity to the code, I open a PR.
I hope that's OK.

Navid200:
@jamorham The capability to mark a reading as bad already exists.  
But, after doing that, there is no visual distinction.  The only way to find such a reading is by searching the data table.  I don't consider this issue an urgent issue.  I consider it a good to have.  We have more serious issues open.  It's just that it was easy enough for me to implement.

Issue #768, opened 2019-02-05
Build troubles.  (xDrip should have a reliable documented automated build procedure)


afcady:
I'm trying to build xDrip.  I have never before built any Android program or any java program.  Here are the troubles I'm running into.

----

EDIT: Nevermind actually.  At last, I successfully built it.  Here is the working build procedure:

```
wget -c https://dl.google.com/android/repository/sdk-tools-linux-4333796.zip -O ../sdk-tools-linux-4333796.zip.partial
mv ../sdk-tools-linux-4333796.zip.partial ../sdk-tools-linux-4333796.zip
unzip -d ${ANDROID_HOME} ../sdk-tools-linux-4333796.zip
mkdir ${ANDROID_HOME}
yes | ${ANDROID_HOME}/tools/bin/sdkmanager --licenses
yes | ${ANDROID_HOME}/tools/bin/sdkmanager "platforms;android-26" "build-tools;27.0.3" "extras;google;m2repository"
bash gradlew assembleRelease
ln -sf app/build/outputs/apk/release/app-release-unsigned.apk .
```

-----

I have a `Makefile` that generated these commands.  It should make it possible simply to run `git clone`, `cd xDrip` and then download and build the `.apk` just by typing `make`.  I will try it out on a fresh directory later to make sure.  Then I'll put it in a PR.

Navid200:
@afcady Sounds like your problem has been resolved.
Would you please close the issue then?
Thanks

afcady:
Not exactly resolved, as #903 is not merged.

Navid200:
But, isn't 903 self-contained?  Or, are you saying 903 and 768 are duplicates?

afcady:
903 is a pull request, not an issue.

Navid200:
I see thanks.

tolot27:
@afcady As commented at #903, the PR can be merged only after removing the hardcoded version numbers. When do you have time to implement this?

Navid200:
@tolot27 Considering there is already a PR, shouldn't this be closed?  What if someone else sees this open issue and starts spending time on this developing a solution?  Wouldn't that be a waste of time?

tolot27:
@Navid200 Issues will be closed automatically, as soon as the linked PR gets merged. It is recommend that people should read the complete discussion at the issues and if they are willing to contribute, discuss this with the developers.

Anyway, this issue can be closed since there is a tiny HowTo at https://xdrip.github.io and there is no need for a Makefile.

PR #1735, opened 2021-05-21
add Makefile

afcady:
This allows the project to be built in a single command (`make`).  The `Makefile` handles downloading and installing the SDKs in an essentially identical way to the `.travis.yml`, then builds with `gradle`.

This is an updated version of PR #903 which incorporates the requested change by @tolot27 (obtain version strings from `.travis.yml`) and uses a new .zip for `sdkmanager`.

Also related #768.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,5 @@ patches/
 *.dll
 *.exe
 *.iml
+installed-android-sdk-stamp
+unzipped-android-sdk-stamp
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,9 @@
 ANDROID_HOME=$(PWD)/../android
 export ANDROID_HOME
+
+GIT_DISCOVERY_ACROSS_FILESYSTEM=y
+export GIT_DISCOVERY_ACROSS_FILESYSTEM
+
 SDKS = "platforms;android-26" "build-tools;27.0.3" "extras;google;m2repository"
 ANDROID_SDK_ZIP=sdk-tools-linux-4333796.zip
 
@@ -7,7 +11,15 @@ DEBUG_APK = app/build/outputs/apk/debug/app-debug.apk
 RELEASE_APK = app/build/outputs/apk/release/app-release-unsigned.apk
 
 .DEFAULT_GOAL = app-release-unsigned.apk
-.PHONY: assembleRelease clean clean-global-caches install
+.PHONY: assembleRelease clean clean-global-caches install tasks
+
+gradle := $(shell [ "$(which gradle >/dev/null)" ] && echo gradle || echo bash gradlew)
+
+exec:
+	@if [ "$(sh)" ]; then sh -c "$(sh)"; else exec bash; fi
+
+tasks:
+	$(gradle) tasks
 
 install: app-release-unsigned.apk
 	adb install $^
@@ -18,7 +30,7 @@ app-release-unsigned.apk: $(RELEASE_APK)
 $(RELEASE_APK): assembleRelease
 
 assembleRelease: | installed-android-sdk-stamp
-	bash gradlew assembleRelease
+	$(gradle) assembleRelease
 
 BUILD_DIRS = app/build/ build/ localeapi/build/ wear/build/
 clean:
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,14 @@
 ANDROID_HOME=$(PWD)/../android
 export ANDROID_HOME
 
+BUILD_TOOLS_VERSION := $(shell sed -n 's/^ *- build-tools-//p' .travis.yml)
+ANDROID_PLATFORM_VERSION := $(shell sed -n 's/^ *- yes | sdkmanager "platforms;android-\([0-9]*\)"/\1/p' .travis.yml || echo 26)
+
+
 GIT_DISCOVERY_ACROSS_FILESYSTEM=y
 export GIT_DISCOVERY_ACROSS_FILESYSTEM
 
-SDKS = "platforms;android-26" "build-tools;27.0.3" "extras;google;m2repository"
+SDKS = "platforms;android-$(ANDROID_PLATFORM_VERSION)" "build-tools;$(BUILD_TOOLS_VERSION)" "extras;google;m2repository"
 ANDROID_SDK_ZIP=sdk-tools-linux-4333796.zip
 
 DEBUG_APK = app/build/outputs/apk/debug/app-debug.apk
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,14 @@
 ANDROID_HOME=$(PWD)/../android
 export ANDROID_HOME
 
-BUILD_TOOLS_VERSION := $(shell sed -n 's/^ *- build-tools-//p' .travis.yml)
-ANDROID_PLATFORM_VERSION := $(shell sed -n 's/^ *- yes | sdkmanager "platforms;android-\([0-9]*\)"/\1/p' .travis.yml || echo 26)
-
+BUILD_TOOLS_VERSION != sed -n 's/^ *- build-tools-//p' .travis.yml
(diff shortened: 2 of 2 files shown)


Navid200:
It's more efficient if you work with the person who has reviewed your PR instead of opening another PR.
Is there a reason you couldn't use the same PR to make the necessary changes?

afcady:
I am absolutely working with @tolot27: I wrote this code per their request!

I made a new PR because the new commits that I made didn't show up in Github UI on the other PR.

I guess it was because the PR was closed.

Navid200:
This was in his last message:

>I suggest, you keep the make file in your master branch and use it on your own. There is no need to integrate it into xDrip master because xDrip builds fine and your on branches will build, too.

Which means, he is suggesting we don't need this.

afcady:
There is no need to remind me of that.

Nevertheless, I made the requested changes.

Makefile
```diff
@@ -0,0 +1,71 @@
+ANDROID_HOME=$(PWD)/../android
+export ANDROID_HOME
+
+BUILD_TOOLS_VERSION != sed -n 's/^ *- build-tools-//p' .travis.yml
+ANDROID_PLATFORM_VERSION != (sed -n 's/^ *- yes | sdkmanager "platforms;android-\([0-9]*\)"/\1/p' .travis.yml; echo 26) | head -n1
```
tolot27, line 5: What is the fixed `echo 26` for? Isn't this a hard-coded version number which should be fetched from `.travis.yml` as well?

Makefile
```diff
@@ -0,0 +1,71 @@
...
+	mv $@.partial $@
+
+$(ANDROID_HOME):
+	mkdir $@
+
+unzipped-tools-stamp: ../$(ANDROID_CLI_TOOLS_ZIP)
+	unzip -d $(ANDROID_HOME) $^
+	touch $@
+
+sdkmanager = $(ANDROID_HOME)/tools/bin/sdkmanager
```
tolot27, line 54: One common code style is upper-casing of variables. Please change it to `SDKMANAGER` or better to `SDK_MANAGER`.

Makefile
```diff
@@ -0,0 +1,71 @@
...
+export ANDROID_HOME
+
+BUILD_TOOLS_VERSION != sed -n 's/^ *- build-tools-//p' .travis.yml
+ANDROID_PLATFORM_VERSION != (sed -n 's/^ *- yes | sdkmanager "platforms;android-\([0-9]*\)"/\1/p' .travis.yml; echo 26) | head -n1
+
+GIT_DISCOVERY_ACROSS_FILESYSTEM=y
+export GIT_DISCOVERY_ACROSS_FILESYSTEM
+
+SDKS = "platforms;android-$(ANDROID_PLATFORM_VERSION)" "build-tools;$(BUILD_TOOLS_VERSION)" "extras;google;m2repository"
+ANDROID_CLI_TOOLS_ZIP=commandlinetools-linux-7302050_latest.zip
```
tolot27, line 11: I'm still not very happy with the fixed version in the `commandlinetools`. As far as I know, they can be simply installed using the OS common package managers like `sudo apt install android-tools` (see [How To Set Up Android Development Tools On Linux](https://www.addictivetips.com/ubuntu-linux-tips/set-up-android-development-tools-on-linux)).

Hence, why not their installation status?

.gitignore
```diff
@@ -57,3 +57,5 @@ patches/
 *.dll
 *.exe
 *.iml
+installed-android-sdk-stamp
```
tolot27, line 60: Are these two files still be created or can it be avoided, somehow?

tolot27 (changes requested):
Thanks for addressing my requests. It looks much better than before. I still have some suggestions.

Since you have claimed not using xDrip, what is your intention to open this PR? Do you develop some automated deployment pipelines for xDrip or do you plan to address existing issues?

afcady:
>what is your intention to open this PR?

To share the code that I wrote to build xDrip.

Can I ask you a question: what is your intention to ask me all of these questions?  Why do you care about the "code style" of the `Makefile`?  I made the changes you requested.

I wrote the entire file and it is in my typical style.  It is the only file in that language in the repository.  So it should be in my style, as the sole author in that language in the project.  In my style, variables that hold the names of commands are lowercase for consistency with other commands.

afcady:
1. The stamp files are good.  That's the best way to do these things.  Forget about it Jake, it's Unix.

2. Using the package manager requires root.  Also, supporting all the various package management systems is a lot of work.  Many hours of setup just to do any kind of testing.  If it is really important, better to merge first, then create  a new issue for such complicated feature.

15characterlimi:
My personal view is that this PR adds negative value and should not be merged.

The reason it adds negative value is that it duplicates some assumptions such as Android SDK version and etc. with which the project can be build, or gradle target names. Every time those change in future, one would then have to make a corresponding change in the makefile, or else the makefile would undergo bitrot.

Since we already have a build system (gradle) and since not even the author of this PR needs this since they don't use xDrip, this seems to not solve any problem that anyone has, while creating new problems as per my previous paragraph.

tolot27:
> My personal view is that this PR adds negative value and should not be merged.

I would not call it a negative value as long as it has no negative impact.

tolot27:
> > what is your intention to open this PR?
> 
> To share the code that I wrote to build xDrip.

But what are the benefits for us? We already have an established build system and if we like to build it using Linxu, I would simply install travis and using the existing `.travis.yml` file.

> Can I ask you a question: what is your intention to ask me all of these questions?

Because we appreciate contributions but have to identify the benefits for us and the impact of a PR rejection.

> Why do you care about the "code style" of the `Makefile`? I made the changes you requested.
> 
> I wrote the entire file and it is in my typical style. It is the only file in that language in the repository. So it should be in my style, as the sole author in that language in the project. In my style, variables that hold the names of commands are lowercase for consistency with other commands.

You are free to use your code style in your fork. But we like to have a common code style in the main repository. That are our rules and we recommend to follow common code styles and apply code formatting.

15characterlimi:
The negative impact that I was referring to was the fact that it adds duplication of the names of gradle build targets and version numbers of Android SDK and build tools, which already exist in gradle files. This only adds a little bit of work towards changing these (because now the makefile needs to be changed and tested as well), but it sounded like no one is actually using the makefile, not even the original author?

Note that I don't feel strongly. There are much, much bigger problems in this project's code health than the potential duplication added by this PR.

afcady:
>The reason it adds negative value is that it duplicates some assumptions such as Android SDK version with which the project can be build, and gradle target names. Every time those change in future, one would then have to make a corresponding change in the makefile, or else the makefile would undergo bitrot.

How is the non-working makefile of less value than a non-existent makefile?

Suppose we assign the values like this:

    Makefile works | Makefile does not work | Makefile does not exist
      1            |  x                     |  0               

Why is `x` negative?  I believe that `x` is positive and very close to `1` because the makefile in that condition can be more easily updated to work than in the non-existent condition.

afcady:
I totally accept a 0 for the sake of argument though.  But negative?  How???

afcady:
Let me just throw on another advantage.  If a developer wanted to do automated builds for tests but not on Microsoft servers, the Makefile would enable that.  It reduces dependency on Github and Microsoft testing servers.

You've got code that's incomplete, non-portable, runs only on one system, Travis.  Now I add code that does the same thing but works on MY system, and is sure to work on MORE systems than the build system you guys had.  Somehow my more portable solution is the problem, and especially the casing of its variables?  Maybe it is a problem in YOUR mind.  Syndrome called NIHS?

afcady:
I got xdrip to build on a system it could not build on until I added those capabilities.

Why?  That was the system I was using.  

Why submit it here?  Because it is more useful to have a more portable build system.  Not just to me.  More portable is objectively better.

Many other people have systems that will not run .  I share my work with them for their benefit.  You people here stand in the way of me sharing my free work with other users.  To no benefit to you or me or anyone whatsoever to block my makefile nor to have any of these arguments.  What are you trying to do?  Gatekeep?  Drive me away?  This is not helping xdrip or anyone else in the world.  Unlike the makefile.  The makefile can help someone with a system like mine accomplish the build that I accomplished but withot performing the work I performed.

Issue #1733, opened 2021-05-19
Initiative: Better Support for Android Wear 


jwoglom:
Some news coming out of Google I/O this week is that Google seems to be suggesting they will be putting effort into Android Wear/Wear OS again. There's not a ton of details yet, but the [Android Developer page on Wear](https://developer.android.com/wear) mentions "big changes to the platform" and "major improvements to the core experience" are coming. With this will likely come a new API release, new smartwatches that are more heavily promoted by Google/Samsung, and potentially more users of xDrip who wish to make use of our existing Wear support.

Given this, I think it's worth discussing how xDrip can be best prepared for upcoming changes in the Wear ecosystem. My understanding is that we currently are in violation of several best-practices and soon-to-be or even currently-enforced guidelines around building an Android app that functions on phone and watch. One of these is https://github.com/NightscoutFoundation/xDrip/issues/1648 which affects how we sync the app to wearables.

We also have some extreme code duplication that goes on between the app/src and wear/src folders which is not kept in sync very well -- essentially all development occurs in app/src and is then very infrequently backported to wear/src on an irregular basis. This seems like an area of significant concern for getting xDrip's codebase in to a more maintainable state. I wonder whether it would be possible to avoid this code duplication entirely and get everything important into shared code that can be used by both modules.

The ability for xDrip running on Wear to function as a collector is also an area that is potentially worth re-investigating. My understanding, at least back from mid-2018 when I used xDrip and Android Wear, was that only one or two specific watches functioned properly as collectors due to some Bluetooth stack related bug affecting watches. I have no clue whether a significant chunk of the xDrip community still uses that specific ~4+ year old watch today, or whether there's been any recent checks done to see whether newer watches function okay as collectors. Needless to say, it would be amazing if we could test whether new Wear devices could be workable as collectors for G5/G6.

While I don't intend for this GitHub issue to serve as explicit tracking for some of the specific issues mentioned here, I hope this can spark a conversation around how we can improve xDrip to build a truly great wearable experience for Android Wear users in the near future.

Navid200:
@jwoglom 
Would you please see the following?
https://github.com/NightscoutFoundation/xDrip/issues/1648

Is there anything in this new issue that is not already covered by 1648?
If yes, would you please add it to the existing issue rather than opening a new one?

Or, is there a reason this needs a dedicated new issue?

tolot27:
[...]

Issue #1522, opened 2020-11-24
Code Review: refactor code


tolot27:
Over the last couple of days I spent some time to understand the xDrip code. After some reviewing I found the following issues which I like to address soon:

- [ ] cleanup Gradle files
- [ ] remove local Android Archive (AAR) files and use the libraries from public repositories like maven, instead
  - [ ] update ColorPicker
  - [ ] remove hellocharts
  - [ ] search-preference.aar v1.0.4 requires casts to AppCompatActivity
  - [ ] remove usb-serial-for-android-v010 (implementation 'org.clojars.nakkaya:usb-serial-for-android-v010:010') becaue it is imported into ImportedLibraries\usbserial; it looks like it is a much older version 0.2.0-pre
- [ ] ImportedLibraries.dexcom is imported from https://github.com/nightscout/android-uploaderand requires ImportedLibraries\usbserial. Remove unused code.
- [ ] upgrade all dependencies to the newest one which still supports the minSdkVersion we require
  - [ ] setup [Dependabot](https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot) at the main repository; did this on my [fork](https://github.com/tolot27/xDrip/pulls) and it works so far
- [ ] avoid calling new methods on older versions: minSdkVersion is set to 18/19 but methods from level 21 or 26 are called
- [ ] migrate to AndroidX (mandatory as of Android SDK 28.0), previous versions support `com.android.support:support-annotations` which is currently in use
- [ ] migrate rxjava (1.3.3) to rxjava2 
- [ ] enable incremental compilation
- [ ] enable parallel and incremental annotation (`android.enableSeparateAnnotationProcessing=true`)
- [ ] upgrade Android Gradle Plugin from 3.4.1 to at least 3.5.x (https://developer.android.com/jetpack/androidx/releases/databinding)
  - [ ] requires package names written in all lowercase (see [Naming a Package](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html)) because old data binding (`android.databinding.enableV2=true`) is removed)
- [ ] upgrade Android Gradle Plugin to 3.6.x or higher requires migration to AndroidX or `android.useAndroidX=true` and `android.enableJetifier=true` in `gradle.properties`
- [ ] migrate from Fabric Crashlytics (io.fabric) to Firebase Crashlytics
[...]

PR #1738, opened 2021-05-25
Ensure wakelocks are released on more exception paths.

15characterlimi:
There are a lot of places around xDrip+ where wakelocks are acquired but not released until the (generally inconsistently picked) timeout runs out. Many of those are deliberate (although probably not a good idea), but others occur by accident, on exception paths. This commit fixes a subset of the latter.

 * There were a few cases where no try/finally block was used but there were multiple return statements that were each accompanied by a release of the wakelock. This is needlessly complex and error prone. I've replaced these with a try/finally block; note that even in case of a return statement being encountered in the try block, the finally block will still run.

 * Many of these occur on background threads; when uncaught exceptions on background threads kill the app, this is okay - but these many threads should not hard code assumptions about the UncaughtExceptionHandler, which is a global setting. This commits introduces a helper class WakeLockThread that takes care of the acquisition/release of the wakelock, and ensures that the thread has a name (which can help during debugging); in some cases, this also allows dropping a try/finally block to avoid indentation in the source code. The presence of the new class should hopefully encourage more mindful handling of wakelocks in future code.

 * There were also a handful of cases where there was a try/finally block, but the wakelock acquisition occurred a few statements before the try block; I've moved the wakelock acquisition to just before try block in order to make sure that it is always released.

Unrelatedly, I've also fixed a handful of needless constructions of Runnable objects for invocations of the form new Thread(new Runnable() { public void run() { ... } } which can instead just override Thread.run() instead (same behavior, but one less object).

I ran all Robolectric unit tests and confirmed that they still pass.

Potential work for follow-up commits:
 - Potentially address the two TODOs and one FIXME that I've left in the code.
 - Add convenience API to specify (TimeUnit timeUnit, long duration) instead
   of (long millis) for more readable / less error-prone code.
 - Add an AbstractWakeLockService class that takes care of acquiring/releasing
   a wakelock during onStartCommand(), since xDrip+ has multiple independent
   implementations of this behavior, some of which were buggy. That said,
   the app will probably be killed during onStartCommand() only in extreme
   cases, so acquiring the wakelock might be unnecessary in the first place.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/AddCalibration.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/AddCalibration.java
@@ -20,6 +20,7 @@ import com.eveningoutpost.dexdrip.UtilityModels.Constants;
 import com.eveningoutpost.dexdrip.UtilityModels.PersistentStore;
 import com.eveningoutpost.dexdrip.UtilityModels.UndoRedo;
 import com.eveningoutpost.dexdrip.calibrations.NativeCalibrationPipe;
+import com.eveningoutpost.dexdrip.utils.framework.WakeLockThread;
 
 import java.util.UUID;
 
@@ -87,12 +88,9 @@ public class AddCalibration extends AppCompatActivity implements NavigationDrawe
                     if (!TextUtils.isEmpty(string_value)) {
                         if (!TextUtils.isEmpty(bg_age)) {
                             final double calValue = Double.parseDouble(string_value);
-                            new Thread() {
+                            new WakeLockThread("xdrip-autocalibt", 60000) {
                                 @Override
-                                public void run() {
-
-                                    final PowerManager.WakeLock wlt = JoH.getWakeLock("xdrip-autocalibt", 60000);
-
+                                public void runWithWakeLock() {
                                     long bgAgeNumber = Long.parseLong(bg_age);
 
                                     if ((bgAgeNumber >= 0) && (bgAgeNumber < 86400)) {
@@ -156,8 +154,6 @@ public class AddCalibration extends AppCompatActivity implements NavigationDrawe
                                     } else {
                                         Log.wtf("CALERROR", "bg age either in future or older than 1 day: " + bgAgeNumber);
                                     }
-
-                                    JoH.releaseWakeLock(wlt);
                                 }
                             }.start();
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
@@ -34,6 +34,7 @@ import com.eveningoutpost.dexdrip.UtilityModels.Pref;
 import com.eveningoutpost.dexdrip.utils.CipherUtils;
 import com.eveningoutpost.dexdrip.utils.DisplayQRCode;
 import com.eveningoutpost.dexdrip.utils.SdcardImportExport;
+import com.eveningoutpost.dexdrip.utils.framework.WakeLockThread;
 import com.google.android.gms.common.ConnectionResult;
 import com.google.android.gms.common.GoogleApiAvailability;
 import com.google.android.gms.gcm.GoogleCloudMessaging;
@@ -544,10 +545,9 @@ public class GcmActivity extends FauxActivity {
 
     static synchronized void syncBGTable2() {
         if (!Sensor.isActive()) return;
-        new Thread() {
+        new WakeLockThread("syncBGTable", 300000) {
             @Override
-            public void run() {
-                final PowerManager.WakeLock wl = JoH.getWakeLock("syncBGTable", 300000);
+            public void runWithWakeLock() {
                 //if ((JoH.ts() - last_sync_fill) > (60 * 1000 * (5 + bg_sync_backoff))) {
                 if (JoH.pratelimit("last-sync-fill", 60 * (5 + bg_sync_backoff))) {
                     last_sync_fill = JoH.tsl();
@@ -576,7 +576,6 @@ public class GcmActivity extends FauxActivity {
                 } else {
                     Log.d(TAG, "Ignoring recent sync request, backoff: " + bg_sync_backoff);
                 }
-                JoH.releaseWakeLock(wl);
             }
         }.start();
     }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreAlarmReceiver.java
@@ -24,6 +24,7 @@ import com.eveningoutpost.dexdrip.UtilityModels.Intents;
 import com.eveningoutpost.dexdrip.UtilityModels.Pref;
 import com.eveningoutpost.dexdrip.utils.CheckBridgeBattery;
 import com.eveningoutpost.dexdrip.utils.DexCollectionType;
+import com.eveningoutpost.dexdrip.utils.framework.WakeLockThread;
 import com.google.gson.Gson;
 
 import org.apache.commons.math3.analysis.interpolation.SplineInterpolator;
@@ -104,10 +105,9 @@ public class LibreAlarmReceiver extends BroadcastReceiver {
 
     @Override
     public void onReceive(final Context context, final Intent intent) {
-        new Thread() {
+        new WakeLockThread("librealarm-receiver", 60000) {
             @Override
-            public void run() {
-                final PowerManager.WakeLock wl = JoH.getWakeLock("librealarm-receiver", 60000);
+            public void runWithWakeLock() {
                 synchronized (lock) {
                     try {
 
@@ -172,7 +172,6 @@ public class LibreAlarmReceiver extends BroadcastReceiver {
                         }
                     } finally {
                         JoH.benchmark("LibreReceiver process");
-                        JoH.releaseWakeLock(wl);
                     }
                 } // lock
             }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
@@ -18,6 +18,7 @@ import com.eveningoutpost.dexdrip.UtilityModels.Intents;
 import com.eveningoutpost.dexdrip.UtilityModels.Pref;
 import com.eveningoutpost.dexdrip.UtilityModels.StatusItem;
 import com.eveningoutpost.dexdrip.utils.DexCollectionType;
+import com.eveningoutpost.dexdrip.utils.framework.WakeLockThread;
 
 import java.text.DecimalFormat;
 import java.util.ArrayList;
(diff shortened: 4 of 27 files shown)


tolot27:
@15characterlimi Thanks for catching this topic. It will take a wile to review it. What it makes harder is that there is just one commit with many changed files. But it looks like the same change is applied to several files.
I know we have a huge code duplication between app and wear. Shouldn't the same changes be applied to wear sources as well?

15characterlimi:
> Thanks for catching this topic. It will take a wile to review it. What it makes harder is that there is just one commit with many changed files. 

Would you prefer if I split this PR up into different "kinds" of changes, ie.

 - Introduce and use WakeLockThread
 - Replace blocks that release the WakeLock separately on multiple return statements -> a single try/finally block
 - ...

?

> I know we have a huge code duplication between app and wear. Shouldn't the same changes be applied to wear sources as well?

I only looked through usages of JoH.getWakeLock(). If that class/method is also duplicated for Wear, then that would explain it. Let me look into this some time outside of work time (I will have limited time to engage with this during the week, please bear with me).

If this is all duplicated in Wear then the #1 goal should be to reduce/remove this duplication so that all future work doesn't need to be doubled-up. Would you agree?

tolot27:
@15characterlimi No hurry. We all have limited time. BTW: Issue #1522 catches many of the code review aspects and the app/wear code duplication is also mentioned in #1733. The problem is that we cannot touch to many files at once.

As more and more developers contribute to xDrip we have to get more organized. I'll propose structure, soon.

jamorham:
This PR isn't unreasonable in itself and looks like it has good practice behind it, but the main issues I see go like this:
- Is this solving a particular problem? Does an audit of operational wakelocks of the app running show that they are not being released in good time?
- Have the modified code paths been tested operationally, for example if NFC reading is modified has that been tested? Wakelock debugging can be tricky because the device can behave differently depending on whether they want to doze or not and how strictly things are being enforced. Some items are tune-able at runtime and different manufacturers tweak these settings in potentially annoying ways. A quick read of the PR doesn't *look* like it changes any behavior, but actual testing is important to validate that.

15characterlimi:
> Is this solving a particular problem? Does an audit of operational wakelocks of the app running show that they are not being released in good time?

The main goal of this commit is to set precedence for good wakelock hygiene and to make it easy to ensure wakelocks are released in a timely and deterministic manner. Specifically:

  - Releasing wakelocks in a deterministic manner (after completion of some action rather than after a set number of seconds) makes the code behave more deterministically, which makes it more likely that any bugs reproduce reliably and their presence and root causes identified. While the current practice of holding on to wakelocks longer than needed is unlikely to trigger a serious bug, it's hard to know this for sure because any bug triggered in this manner would be hard to reproduce and therefore would have a high chance of remaining undetected, even if its impact was severe.
  - Releasing wakelocks once they are no longer needed saves battery. Again the impact per user is probably small, but (a) there are lots of users, and (b) if there are cases where the impact is not small then we'd be unlikely to find out because of (again) the nondeterministic and hard-to-reproduce causal chain.

In summary, the impact of each of the places toughed by this commit is probably small, but (a) this CL fixes a lot of them, and (b) we don't know for sure that it isn't large because it's so difficult to track battery use back to a particular piece of sloppy coding.

The idioms introduced by this commit aim to make the code more maintainable in the long run by setting precedent and by making it easier to write correct code. For example, 
 - Subclasses of WakeLockThread can't forget to give up the wakelock.
 - Code snippets inside a try/finally block no longer need to remember to release the wakelock on every single return statement, which is error prone (example: DexCollectionService.onServicesDiscoveredWithWakeLockHeld())

> Have the modified code paths been tested operationally, for example if NFC reading is modified has that been tested?

Not yet.

---------

I'd like to take a step back and ask you about the general engineering philosophy, because I think this will determine whether me trying to contribute to this project will deliver positive value vs. just be annoying for both of us.

My personal impression is that xDrip+ has a **lot** of features, but the code supporting those features has become quite brittle and hard to maintain. Examples:

 - There's lots of code duplication, e.g. between Wear and non-Wear code, which doubles the effort of any commits that touch this code.
 - The app still targets Android Marshmallow, which means that:
    - The app would not be accepted on Play Store (I gather that this is a non-goal, for liability reasons)
    - Modern tooling such as the error-prone plugin (static analysis that detects bugs) cannot be used.
    - We're stuck on the JDK 8 toolchain from 2014 - Java 17 LTS is coming out this September
    - Android exhibits a large number of compatibility behavior; the behavior changes since then have generally been made in the interest of the user (battery saving, privacy, ....) so the user will not get those benefits.
 - There is lots of dead code. For example, CipherUtils.java contains a hard-coded key (which seems like an insecure practice?), but also appears to be nearly completely unused.
 - Sloppy code that is *almost* correct is all over the place - such as wakelocks that aren't deterministically released or that are even kept around for a seemingly arbitrary amount of time. 
 - Test coverage is nowhere near enough to give confidence in code changes (examples: neither this PR nor https://github.com/NightscoutFoundation/xDrip/pull/1737 touches any existing tests, doesn't break any existing tests, yet still people are very worried about it breaking stuff)
  - There are lots of "haunted graveyards" that no one dares to touch. For example
    - The comment on the targetSdkVersion vaguely says "increasing target SDK version can cause compatibility issues with Android 7+", rather than being specific about what needs to change in order to target a modern version of Android.
    - No one knows where the constant value MMOLL_TO_MGDL comes from yet still people are afraid to change it to a very similar value that breaks no tests and has a justification.
  - Most commits deal with adding new features, rather improving the maintainability of the existing code. PRs that try to (such as this one here) don't feel very welcomed).

My personal perception is that xDrip+ is stuck in a local optimum that it can't get out of because the code is so brittle that no one dares to make any changes, beyond adding even more features. My personal perception is that xDrip+ has for years been stuck in a place where most effort was invested in more and more features, and very little effort was invested in keeping the existing code healthy (for example, by deleting dead code, refactoring sloppy or poorly abstracted code, or targeting recent versions of Android). My personal perception is that what xDrip+ needs most is not new features, but a higher code quality through static analysis (error-prone), more comprehensive unit tests, deduplication of code that was previously duplicated (Wear), ... Getting there will likely result in *less* stability in the short term as some things will invariably break as things get refactored, but it will result in much *more* stability and code quality in the long term. 

Does this match your philosophy of what you would like xDrip+s future to be? If not, then no problem, it's your project. But then I think it'd be good to know, because I think that my attempts to provide value would then just result in frustration for both of us. I used to maintain a codebase that had a **lot** more users than xDrip+ (actually, xDrip+ was one of those users...) and I'm definitely not a cowboy when it comes to making code changes. At the same time, I'm a strong believer in "No haunted graveyards" ([example](https://www.usenix.org/sites/default/files/conference/protected-files/srecon17americas_slides_reese.pdf) out of many excellent talks/articles). From your and others' reactions to my bait in form of two pull requests, it seems to me that this approach to engineering might not be welcome here. If that's the case, then so be it. But it'd be good to know so we can each go our separate ways.

WDYT?

jamorham:
The main issue is integration testing. I don't believe there is really any automated way to unit test the interactions with the framework that occur over the (fragmented) Android ecosystem and xDrip commonly is interacting with remote devices which have their own quirks as well. 

This means that complete testing needs to be done with real handsets and real devices in my opinion. The existing code has been tested like this and so if its going to be changed then I have to weigh up the pros and cons of doing so and factor in the level of testing that has occurred.  

The changes proposed in this PR are relatively minor and I'm not suggesting that there are errors in it (although I haven't meticulously reviewed it either) but taking a strict engineering view when asking the question "does this break anything" the answer is that it hasn't been tested so we don't know for sure. This is why I asked that question.

When someone produces changes for a particular part of the system which they are using every day, it is usually to address a specific need, like something is broken or a new feature is added and there is a greater level of confidence that it has been operationally tested (at least on the handsets they are using) - In my case I'm typically testing over a range of handsets with various chipsets and android versions and often live with multiple different users before merging the change.

Last time I did an audit of wakelocks, what I found was that the only dangling one consuming anything significant was inside of google play services. This is why xDrip patches this at runtime to resolve that issue. The fact that this is necessary when it is google who are hounding developers to reduce background cpu usage tells you a lot about the chaotic nature of working with android and their engineering practices.

One of the reasons for not targeting higher sdk versions is because of restrictions put in place by doing so that can break app behavior. In some situations this can be worked around with code changes and in others there is no replacement functionality (eg calling hidden api's to resolve framework bugs) 

Often google will introduce a replacement for something they are removing from a targetsdk level, provide documentation for it that is wrong (or manufacturers will integrate it incorrectly - hard to tell which) only to abandon it on the next major release. Such instability in the framework doesn't give so much incentive to adopt the new mechanisms while the old ones are still working. Eventually changes will be required as backward compatibility wanes in future android versions, a good example of this is package targeted broadcasts which are mandatory whichever android version is targeted.  

I do want to raise the minimum sdk level to android 4.4 (and in time 5.0) but last time I increased it there were complaints from users still using those handsets which is why that was reverted. I'm thinking enough time has passed now that such a change should be much safer to make.

15characterlimi:
Thanks, that makes sense. There may be some common ground around better test coverage and/or refactoring code to be more testable that we can agree on. Let me think about this over the weekend and get back to you; perhaps I can think of some test coverage improvements that will make up for any error budget spent in the short term on refactorings/fixes such as in this PR here. Obviously a full integration test will not be possible without having all the other respective systems to integrate with, but it is possible in principle to define an API boundary and then have unit tests that operate at that boundary. I haven't checked, but I wonder for example whether there are sufficient tests at the layer of the HTTP server; I've also been thinking of improving the UI of the appwidget (some extra lines could make it more readable IMO), and while I'm at it perhaps adding some screenshot-based integration tests. We'll see - I'll look into it.

I'll also have to address @tolot27 's comments - I'll have to check for duplicate code that also needs to change, and might want to split this commit up into multiple that separately tackle different parts of the issue.

Please hold on and I'll get back to you, hopefully by the end of the upcoming weekend (I also have a bunch of other things to deal with this weekend so can't 100% promise). I'll get back to you eventually.

tolot27:
I'm not sure whether this PR would fix the NPEs which can be seen during unit tests, i. e. https://github.com/jamorham/xDrip-plus/pull/115/checks?check_run_id=2908921032#step:5:793:
```
    Exception in thread "Thread-75" java.lang.NullPointerException
    	at org.robolectric.shadows.ShadowPowerManager.newWakeLock(ShadowPowerManager.java:32)
    	at android.os.PowerManager.newWakeLock(PowerManager.java)
    	at com.eveningoutpost.dexdrip.Models.JoH.getWakeLock(JoH.java:831)
    	at com.eveningoutpost.dexdrip.UtilityModels.BgSendQueue.handleNewBgReading(BgSendQueue.java:115)
    	at com.eveningoutpost.dexdrip.Models.BgReading.notifyAndSync(BgReading.java:1240)
    	at com.eveningoutpost.dexdrip.Models.BgReading.lambda$bgReadingInsertFromG5$0(BgReading.java:1104)
    	at com.eveningoutpost.dexdrip.UtilityModels.Inevitable$Task.poll(Inevitable.java:114)
    	at com.eveningoutpost.dexdrip.UtilityModels.Inevitable.lambda$task$0(Inevitable.java:60)
    	at java.lang.Thread.run(Thread.java:748)
```

15characterlimi:
> I'm not sure whether this PR would fix the NPEs which can be seen during unit tests

Probably not, the code still builds on top of JoH.getWakeLock()

Issue #1216, opened 2020-01-30
Feature Request: Default plot maximum


dkumor:
Currently, I have not been able to find a way to set the default range for the main glucose plot - the default plot goes to 250mg/dl. If possible, I'd love to be able to set a custom maximum value.

LanFeusT23:
To add to that I would love to also set a custom minimum value. I'd rather have a 50 to 230mg/dl range than a 0 to 250!

Navid200:
Those (min and max) are automatically adjusted based on the values present during the current 24-hour period.

dkumor:
Yes, the max can increase based on the data, however, I have not noticed the max adjusting *down* below 250, even with no data.

PR #1742, opened 2021-05-28
Changing min and max glucose shown on screen

Navid200:
**Why do we need this?**
The main screen vertical axis has a default minimum of 40 (2.22) and default maximum of 250 (13.9).
If your readings are far from those extremes, the screen still shows those as min and max and most of your screen is blank.

After the change, you can change those defaults.  The minimum and maximum will continue to change according to your readings as they did before.    But, if you have a very good day, with minimal fluctuations, the axis scale will shrink to the new defaults.

fixes https://github.com/NightscoutFoundation/xDrip/issues/1216

**Is there a workaround?**  
No

Unless the user uses the settings to change the defaults, nothing changes, from the user perspective, after this PR is merged.

**Tests**
This has been tested with Android 9.


Navid200:
If the user is unaware of this change, it could result in a visual impression that the glucose variation has increased.
We should inform users to avoid unintended consequences.

One solution is to add a setting that the user would need to set in order to enable this.  But, I'd rather not add yet one more setting.

Any suggestions?

jamorham:
The mostly fixed Y axis with the wider range is to make it easier to at-a-glance see what is going on. In the lower range it should be preferable to see a lot of empty space so you know its in the low range and you don't have to read the scale of the graph to evaluate mentally what the range is likely to be.

If users would like differing behavior then it should be introduced as an advanced display preference option defaulting to the currently used values.

Navid200:
@jamorham Done

app/src/main/res/xml/xdrip_plus_prefs.xml
```diff
@@ -424,6 +424,14 @@
                     android:key="illustrate_remote_data"
                     android:summary="Show data received remotely (diagnostic only)"
                     android:title="Illustrate remote data" />
+                <CheckBoxPreference
```
jwoglom, line 427: Can I suggest instead of having these as checkbox preferences, having a numeric value for these settings fields that defaults to 40 for the low and 250 for the high? (the current defaults)

As-is these are sort of confusing, since it's a checkbox for changing a number between two presets.

jwoglom (commented):
TBH, in my personal opinion I would find changing the default glucose window to 70-180 to be pretty annoying. After enough time I've definitely gotten used to how the current range displays on my phone, and I'd actually rather the screen stay constant nearly all of the time (except if I go super high) than have the y axis changing from day to day.

Navid200:
> TBH, in my personal opinion I would find changing the default glucose window to 70-180 to be pretty annoying. After enough time I've definitely gotten used to how the current range displays on my phone, and I'd actually rather the screen stay constant nearly all of the time (except if I go super high) than have the y axis changing from day to day.

That's why there is a setting.  So, you don't have to change the defaults if you don't like to.  The default, if you don't use these new settings, leaves everything as it was.

app/src/main/res/xml/xdrip_plus_prefs.xml
Navid200: The defaults are 40 and 250.  

I can change it to what you suggest still keeping the defaults at 40 and 250.  
But, I'm not sure someone else is not going to prefer with checkboxes.

Navid200: I changed the settings to lists.  There are now multiple options for each with 40 and 250 as defaults.

tolot27:
Once again, the number of preferences increases with this PR. I thought we would avoid this and even try to reduce the number of settings.

Navid200:
Reducing the number of settings to avoid making it complicated for users.  The way I always thought about that was to reduce, if possible, the number of settings.  Not to reduce the number of options for each setting.
If we have a new setting, I don't see why it would make it more complicated if the number of options for a single setting goes up.

The main issue, in my opinion at least, are caused by settings that re right there in Bluetooth and G5/G6 Debug settings.  Those are the settings that affect connectivity.  Users start flipping those like it was lottery hoping that it would fix their connectivity issue.
But, I didn't think adding a setting deep under the graphics settings would cause complexity for connectivity.

Navid200:
> Once again, the number of preferences increases with this PR. I thought we would avoid this and even try to reduce the number of settings.

Will increasing the number of options reduce xDrip performance?

PR #1746, opened 2021-06-06
Enable Add calibration menu option in case Session is in state NeedsD…

LoriLori:
…ifferentCalibration


tolot27:
@LoriLori Thanks for your contribution. Is your PR related to an existing issue? If yes, please link it. It would be very helpful if you add a detailed description of the problem this PR addresses rather than an a wrapped PR title.

I also wonder whether is is the right place to add the new condition because the containing function is named `isG5WantingInitialCalibration()`. It is counterintuitive to me to add a `NeedsDifferentCalibration` condition to it. Maybe we should rename the function if there is no better place to catch this state.

tolot27:
@LoriLori Please can you explain why you have closed this PR? Is the problem solved, somehow?

Issue #1741, opened 2021-05-26
Please don't show user ID and/or password in event logs


Navid200:
**Problem**
Helping people on facebook, we often ask them to show their logs.  Sometimes, their Nightscout ID and password shows in the logs.   We ask them to delete it.  But, those could already have been exposed to eyes that should have seen them.

**Solution**
Please identify all instances where a user ID or password is included in event logs.  Please replace the actual ID and password with words ID and Password.

PR #1747, opened 2021-06-06
Exclude Nightscout ID and password from logs

Navid200:
Users who ask for support are often asked to show event logs.  
They may not be aware that the logs may include their Nightscout ID and password.
This PR removes base URL, which includes the ID and password, from the logs.

Fixes https://github.com/NightscoutFoundation/xDrip/issues/1741

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
@@ -299,7 +299,7 @@ public class NightscoutUploader {
         try {
             uri = new URI(baseURI);
         } catch (URISyntaxException e) {
-            Log.e(TAG, "Error URISyntaxException for " + baseURI, e);
+            Log.e(TAG, "Error URISyntaxException for the base URL", e);
             return baseURI;
         }
         String host = uri.getHost();
@@ -419,7 +419,7 @@ public class NightscoutUploader {
                             PersistentStore.setString(LAST_MODIFIED_KEY, last_modified_string);
                             checkGzipSupport(r);
                         } else {
-                            Log.d(TAG, "Failed to get treatments from: " + baseURI);
+                            Log.d(TAG, "Failed to get treatments from the base URL");
                         }
 
                     } else {
@@ -429,7 +429,7 @@ public class NightscoutUploader {
 
 
             } catch (Exception e) {
-                String msg = "Unable to do REST API Download " + e + " " + e.getMessage() + " url: " + baseURI;
+                String msg = "Unable to do REST API Download " + e + " " + e.getMessage() + " url: ";
                 handleRestFailure(msg);
             }
         }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -489,7 +489,7 @@ private boolean doRESTUpload(SharedPreferences prefs, List<BgReading> glucoseDat
                     last_success_time = JoH.tsl();
                     last_exception_count = 0;
                 } catch (Exception e) {
-                    String msg = "Unable to do REST API Upload: " + e.getMessage() + " url: " + baseURI + " marking record: " + (any_successes ? "succeeded" : "failed");
+                    String msg = "Unable to do REST API Upload: " + e.getMessage() + " url: " + " marking record: " + (any_successes ? "succeeded" : "failed");
```
jwoglom, line 492: Can you remove the "url:" section in the raw string here?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
```diff
@@ -429,7 +429,7 @@ private synchronized boolean doRESTtreatmentDownload(SharedPreferences prefs) {
 
 
             } catch (Exception e) {
-                String msg = "Unable to do REST API Download " + e + " " + e.getMessage() + " url: " + baseURI;
+                String msg = "Unable to do REST API Download " + e + e.getMessage() + " url: ";
```
jwoglom, line 432: Also remove the trailing "url: "

jwoglom (changes requested):

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
Navid200: Thanks for the review.
I intentionally left it.  I just wanted to remove the url itself and nothing else.   I thought leaving the url acronym may serve as a hint that's where the problem is.

What would be the problem with leaving it?

jwoglom: As-is it looks pretty confusing, because the syntax of how the message is worded implies that it's going to put something there. Reading this out of context I'd probably assume the URL was blank, versus that the URL was omitted intentionally.

I would recommend changing it to something like

```
"Unable to do REST API Upload. Check the configured Nightscout URL in the xDrip settings: " + e.getMessage() + " marking record: " + (any_successes ? "succeeded" : "failed");
```

Navid200: Removed "url:"
Thanks

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/NightscoutUploader.java
Navid200: Done

jwoglom (approved):
LGTM

jamorham:
What is the situation for a user who needs to look in the event logs to discover for example a typo in their url? How should we handle that?

Navid200:
@jamorham Can they not just go to the Nightscout Sync (REST-API) page to check the spelling?

jamorham:
Ok fair enough. LGTM

jamorham (approved):

Issue #1586, opened 2021-01-03
update translation credits


tolot27:
With the ongoing merge of the recent translations, the [credits.xml](../blob/master/app/src/main/res/values/credits.xml) should be updated as well.

Navid200:
Is this not done already?
Can you please explain what this is?
Isn't every contributor's login name added to the code when their contribution is merged?

tolot27:
The latest update of the file was in January, 2018. But many people contributed the the translations, recently. If you visit the "End User License Agreement" (in Settings), you can see the rather old list of translators obtained from [credits.xml](../blob/master/app/src/main/res/values/credits.xml). Translators at Crowdin are invisible to GitHub.

Navid200:
Why is a translator different than anyone else who is contributing to the code, like any other developer?
Aren't they all developers?

tolot27:
A translator is not a developer. But that is a different discussion. As translators are not seen by GitHub or because @jamorham wanted to give them credits, he created the credits.xml.
Either credits.xml gets removed or updated.

jamorham:
Before crowdin was used there was a different system for translators which is where the previous credits document comes from. I'd consider keeping this file in the source tree for historical record and recognition but removing its visibility from the EULA screen as it doesn't reflect the current state of translator contributions.

Navid200:
I would like to open a PR for this.

I know how to remove the visibility from the EULA screen.
But, I'm not sure what "keeping this file in the source tree" means.

PR #1748, opened 2021-06-07
Hidden original translator credits

Navid200:
Fixes https://github.com/NightscoutFoundation/xDrip/issues/1586

Am I supposed to move the original credits to some specific file?


jamorham:
I will add some commentary documentation about this to `credits.xml`

jamorham (approved):

Issue #1537, opened 2020-12-04
The glucose readings aren't added in 5 minute intervals anymore


flango:
Hi,
I use Dexcom G6 with xDrip+, it works very well but suddenly it started to add readings from 2 minutes to 8 minute intervals instead of 5 and its really visible and bothering me, i can still keep a good eye on my bloodsugar, but its really annoying.

Does anyone know why it happens?
Screenshot below to show what i mean, can easily see the dots arent added in 5 minute intervals anymore.

<img width="350" src="https://user-images.githubusercontent.com/17218692/101171253-ca4fd080-363f-11eb-9375-840f86b7b61b.png" />

Does anyone know why this suddenly started to happen?

Edit:
It suddenly fixed itself. Only thing i know did earlier tonight was to go to system status, forget device, force close app, clear cache, start xdrip+ again, and start source setup wizard to connect to my G6 again.

tolot27:
There are some missing data points. Please click on some adjacent data points and check the time stamps. Are they differ by 5 minutes? What does the log shows?

flango:
Its the correct amount of dots, just doesnt look right with them uneven spaced.

It just started to add readings from 2 minutes to 8 minute intervals.

All the data points are there but not added in even 5 minute intervals, the shortest interval is 2 minutes in between and the longest 8 minutes.

They can say like 10:00am and next can be 10:02am next 10:07am and next can be at 10:15am, 10:20am, so that's the correct amount of reading just look weird visually.

It suddenly started to add the dots on uneven times even if the reading is every whole 5 minute from the logs.
Even the notification reading tell the same there, a reading every whole 5 minute.
So correct number with readings, just added in weird intervals, and the log says it reads every whole 5 minute.

IntoxRRR:
I notice this behaviour too, once in a while. I get it fixed again by just restarting the phone.

Navid200:
@AdrianLxM 
Any ideas what could cause this?

Navid200:
@nickb24

nickb24:
Possibly related to https://github.com/NightscoutFoundation/xDrip/issues/1612

nickb24:
Possibly need to enable debug logs for 'BgGraphBuilder' to get more information to diagnose. If someone with the issue can do that and then post a screenshot of the logs that might help.

Navid200:
[...]

PR #1750, opened 2021-06-10
Lower the FUZZER value to 37.5 seconds

Navid200:
**Why do we need this?**
This reduces the delta between what is shown on screen and what is in the data table ( what is sent to Nightscout).

A consequence of the delta being great is that occasionally, when the reading arrival time is very close to the boundary of the grid buckets, the readings on the xDrip main screen are not equidistant.  Lowering the value of the FUZZER to 37.5 seconds (from 2.5 minutes) will make the points look equidistant even if they are not.

My hope is that, if everyone is happy with this new setting, some day, we will remove the FUZZER altogether.  So, this is a first step.

Fixes https://github.com/NightscoutFoundation/xDrip/issues/1537  

**Is there a workaround?**
No  

Unless the user enables the setting, nothing changes from the user perspective.  

**Tests**  
This has been tested on an Android 9 handset.


jamorham:
As there are quite a few lines here, could you summarize the changes that you understand to be in this PR? thanks

Navid200:
The following is a list of lines that contain real changes (in the BgGraphBuilder file).  The rest are auto formatting.

BgGraphBuilder.java:  85, 458, 1431

PR #1751, opened 2021-06-10
G5/G6: Automatically create Sensor Start and Stop treatments when they occur

jwoglom:
Currently, xDrip is aware of when a G5/G6 sensor reports that the sensor stopped or has been started, but this does not tie in to xDrip's own reporting of manual treatment events.

After this change, xDrip will create a "Sensor Stopped" treatment (which can also sync to Nightscout using existing mechanisms) whenever a G5/G6 sensor reports itself as being stopped.

xDrip will also create a "Sensor Started" treatment whenever a G5/G6 sensor reports itself as being in the warm-up period. Starting a sensor from within xDrip already creates a sensor start treatment, so that behavior is not modified. However, if a sensor start treatment was within the xdrip database within the past 15 minutes then an additional treatment is not created, to avoid double-creating a treatment if you manually hit "start" within xDrip which then triggers xDrip to detect that the transmitter reports itself as started soon after.

With these changes applied, when my G6 sensor automatically stopped at the end of the 10-day window, a sensor stop treatment was created with "Stopped by transmitter" in the notes field. When I started a G6 sensor from a non-xDrip device, a new start treatment was created with "Started by transmitter" in the notes field. 

image image

Additionally, some minor code cleanup is done in StartNewSensor.java in a separate commit.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/Treatments.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/Treatments.java
@@ -8,6 +8,8 @@ import android.content.Context;
 import android.provider.BaseColumns;
 import android.util.Pair;
 
+import androidx.annotation.Nullable;
+
 import com.activeandroid.Model;
 import com.activeandroid.annotation.Column;
 import com.activeandroid.annotation.Table;
@@ -59,7 +61,11 @@ import static com.eveningoutpost.dexdrip.Models.JoH.emptyString;
 @Table(name = "Treatments", id = BaseColumns._ID)
 public class Treatments extends Model {
     private static final String TAG = "jamorham " + Treatments.class.getSimpleName();
+
+    public static final String SENSOR_START_EVENT_TYPE = "Sensor Start";
+    public static final String SENSOR_STOP_EVENT_TYPE = "Sensor Stop";
     private static final String DEFAULT_EVENT_TYPE = "<none>";
+
     public final static String XDRIP_TAG = "xdrip";
 
     //public static double activityMultipler = 8.4; // somewhere between 8.2 and 8.8
@@ -356,20 +362,54 @@ public class Treatments extends Model {
         return treatment;
     }
 
-    public static synchronized Treatments SensorStart(long timestamp) {
-        if (timestamp == 0) {
+    /**
+     * Returns a newly created treatment entry in the database for Sensor Start,
+     * and pushes the new treatment to followers.
+     * @param timestamp is optional, defaults to right now
+     * @param notes is optional
+     */
+    public static synchronized Treatments SensorStart(@Nullable Long timestamp, @Nullable String notes) {
+        if (timestamp == null || timestamp == 0) {
             timestamp = new Date().getTime();
         }
 
-        final Treatments Treatment = new Treatments();
-        Treatment.enteredBy = XDRIP_TAG;
-        Treatment.eventType = "Sensor Start";
-        Treatment.created_at = DateUtil.toISOString(timestamp);
-        Treatment.timestamp = timestamp;
-        Treatment.uuid = UUID.randomUUID().toString();
-        Treatment.save();
-        pushTreatmentSync(Treatment);
-        return Treatment;
+        final Treatments treatment = new Treatments();
+        treatment.enteredBy = XDRIP_TAG;
+        treatment.eventType = SENSOR_START_EVENT_TYPE;
+        treatment.created_at = DateUtil.toISOString(timestamp);
+        treatment.timestamp = timestamp;
+        treatment.uuid = UUID.randomUUID().toString();
+        if (notes != null && notes.length() > 0) {
+            treatment.notes = notes;
+        }
+        treatment.save();
+        pushTreatmentSync(treatment);
+        return treatment;
+    }
+
+    /**
+     * Returns a newly created treatment entry in the database for Sensor Stop,
+     * and pushes the new treatment to followers.
+     * @param timestamp is optional, defaults to right now
+     * @param notes is optional
+     */
+    public static synchronized Treatments SensorStop(@Nullable Long timestamp, @Nullable String notes) {
+        if (timestamp == null || timestamp == 0) {
+            timestamp = new Date().getTime();
+        }
+
+        final Treatments treatment = new Treatments();
+        treatment.enteredBy = XDRIP_TAG;
+        treatment.eventType = SENSOR_STOP_EVENT_TYPE;
+        treatment.created_at = DateUtil.toISOString(timestamp);
+        treatment.timestamp = timestamp;
+        treatment.uuid = UUID.randomUUID().toString();
+        if (notes != null && notes.length() > 0) {
+            treatment.notes = notes;
+        }
+        treatment.save();
+        pushTreatmentSync(treatment);
+        return treatment;
     }
 
     private static void pushTreatmentSync(Treatments treatment) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
```diff
@@ -160,21 +160,45 @@ public void onTimeUpdated(int newmins) {
...
 
-                realStartSensor();
+                startSensorOrAskForG6Code();
             }
         });
         timePickerFragment.show(activity.getFragmentManager(), "TimePicker");
     }
 
-    private void realStartSensor() {
+    private void startSensorOrAskForG6Code() {
```
tolot27, line 169: Isn't `realStartSensor()` also used by other sensors like Libre?

app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
```diff
@@ -160,21 +160,45 @@ public void onTimeUpdated(int newmins) {
...
+            G6CalibrationCodeDialog.ask(this, this::startSensorAndSetIntent);
         } else {
-            realRealStartSensor();
+            startSensorAndSetIntent();
         }
     }
 
+    private void startSensorAndSetIntent() {
+        long startTime = ucalendar.getTime().getTime();
+        Log.d(TAG, "Starting sensor time: " + JoH.dateTimeText(ucalendar.getTime().getTime()));
```
tolot27, line 179: I still don't understand why `UserError.Log.x` isn't used all the time.

tolot27 (commented):
Thanks for this good catch. Looks good so far. Please respond to the minor comments.

app/src/main/java/com/eveningoutpost/dexdrip/Services/Ob1G5CollectionService.java
```diff
@@ -1717,9 +1719,28 @@ public static void processCalibrationState(final CalibrationState state) {
             final PendingIntent pi = PendingIntent.getActivity(xdrip.getAppContext(), G5_SENSOR_STARTED, JoH.getStartActivityIntent(Home.class), PendingIntent.FLAG_UPDATE_CURRENT);
             JoH.showNotification(state.getText(), "Sensor Stopped", pi, G5_SENSOR_STARTED, true, true, false);
             UserError.Log.ueh(TAG, "Native Sensor is now Stopped: " + state.getExtendedText());
+
+            Treatments.SensorStop(null, "Stopped by transmitter");
+
         } else if (is_started && !was_started) {
             JoH.cancelNotification(G5_SENSOR_STARTED);
             UserError.Log.ueh(TAG, "Native Sensor is now Started: " + state.getExtendedText());
+
```
jamorham, line 1728: I would prefer if we can move this logic in to Treatments so that there is just a simple call from the collection service that specifies the event occurring.

app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
```diff
@@ -160,21 +160,45 @@ public void onTimeUpdated(int newmins) {
...
         if (Ob1G5CollectionService.usingCollector() && Ob1G5StateMachine.usingG6()) {
-            G6CalibrationCodeDialog.ask(this, this::realRealStartSensor);
+            G6CalibrationCodeDialog.ask(this, this::startSensorAndSetIntent);
         } else {
-            realRealStartSensor();
+            startSensorAndSetIntent();
         }
     }
 
+    private void startSensorAndSetIntent() {
```
jamorham, line 177: Is this just a rename or are there deeper changes going on here?

app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
jwoglom: This is just a method rename -- clarified below

app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
jwoglom: The cleanup in this file was just a rename from:
* realStartSensor => startSensorOrAskForG6Code
* realRealStartSensor => startSensorAndSetIntent

The former realStartSensor only asks for a G6 code if necessary and then calls realRealStartSensor, and the ambiguity in the names was bothering me a bit

app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
jwoglom: I don't know, Log.d is a standard Android debugging method, right? versus the custom UserError.Log stuff

app/src/main/java/com/eveningoutpost/dexdrip/Services/Ob1G5CollectionService.java
jwoglom: I can do this -- will just have to name the methods such that it's obvious that they sometimes don't do anything (e.g., in a state where an event was recently created), since most of the methods in Treatments are direct database row creators/accessors.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Treatments.java
```diff
@@ -448,6 +488,15 @@ public static Treatments lastNotFromXdrip() {
                 .executeSingle();
     }
 
+    public static Treatments lastEventTypeFromXdrip(String eventType) {
+        fixUpTable();
+        return new Select()
+                .from(Treatments.class)
+                .where("enteredBy LIKE '" + XDRIP_TAG + "%' and eventType = ?", eventType)
```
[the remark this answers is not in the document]
jwoglom: enteredBy is never set differently from XDRIP_TAG, right? The only reason why there's a "enteredBy LIKE .." here is that the other getter methods in this file used the same convention.

app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
tolot27: Currently, I don't know whether the standard logging `Log` is intercepted to store the data also in the `UserError` database table. If yes, we don't have to care about writing `Log.x` or `UserError.Log.x` otherwise we have to write `UserError.Log.x` to get it logged to the database as well.

app/src/main/java/com/eveningoutpost/dexdrip/Models/Treatments.java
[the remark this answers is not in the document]
tolot27: It looks like there is sometimes and additional text: https://github.com/NightscoutFoundation/xDrip/blob/38fee03be56da6ce9cbe90e08dbe8e4b651c6edc/app/src/main/java/com/eveningoutpost/dexdrip/Models/Treatments.java#L272

app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
jamorham: I think its possible with the correct import to have `Log.x` use `UserError` class.

app/src/main/java/com/eveningoutpost/dexdrip/Services/Ob1G5CollectionService.java
jamorham: You can create an additional helper class if you think that makes sense for this. With the logic moved outside of the collector so its just a simple call out on the event occurring then I'd expect to approve the PR.

jamorham:
I'm planning to adopt this PR myself and make the logic changes that I requested so that we can get it merged quickly.

jamorham:
Testing now. Will merge soon if successful.

jamorham:
Thanks for this, simple and useful.. see the modified version I committed here: https://github.com/NightscoutFoundation/xDrip/commit/fab3e2bff411efc32d18fe4741703715e9ee3974

Main differences are that the logic is moved more in to `Treatments` and it handles sensor start failures as well.

jamorham:
Closing as all changes now merged via a different branch. Thanks

Issue #541, opened 2018-08-09
Wiki - process description for translation


Kywalh:
Hello
I started to do translation in French and I already submitted a couple of them.
I gave my email address but I don't know what is the next step ?

Thanks

themotoden:
I can translate to Russian, but how can I use my translated version? Please, add option for this

tolot27:
@Kywalh I did not find you at [crowdin](https://crowdin.com/project/xdrip) and do not understand what you mean by "I already submitted a couple of them."

@themotoden I accept your join request at crowdin. You are now able to suggest translations.

Navid200:
Is this still an issue?
Or, can we close it?

tolot27:
We still don't have a documentation describing the translation process.

Navid200:
@tolot27 With respect to documentation, can we not have a link to crowdin in Wiki?
And then, you can add any detail documentation under discussions as a post pinned to the top, no?

Can we close this issue?

tolot27:
Better, someone creates a PR against [CONTRIBUTING.md](https://github.com/NightscoutFoundation/xDrip/blob/master/CONTRIBUTING.md). Mayb you can do this?

We close this issue as soon as it is fixed/implemented. Linking it to a PR will close it automatically, if the PR gets merged.

tolot27:
> Do I understand you correctly that you are suggesting that I submit a pull request to change https://github.com/NightscoutFoundation/xDrip/blob/master/CONTRIBUTING.md to include how translators can submit their translations?

Yes, you or someone else. This is a simple task which does not need developer skills. The PR should briefly describe the translation process (Crowdin account, translation, approval, automatic merging to https://github.com/jamorham/xDrip-plus/tree/l10n_master7 and final merging to xDrip by jamorham) and link to the Crowdin project and Crowdin manual.

PR #1755, opened 2021-06-13
Crowdin info

Navid200:
I think it is better for the users to first ask before opening an issue.  Many issues are just user error and it is easier to resolve it and not open an issue.

This also adds a translation heading.
I am not a member of Crowding project.  But, as no one was making this change, I am doing it.  So, please let me know if you want any detail added or any changes to be made.  Or, go ahead and make the changes yourself.

Fixes https://github.com/NightscoutFoundation/xDrip/issues/541

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ Try not to bundle different patches together and keep the change set to the mini
 Maintainers are not always immediately aware of pull request submissions, which is another reason why it helps to have discussed your work prior to submission. Feel free to ask about the progress of processing your PR as maintainers may not realise a PR is awaiting action.
 
 
-#Translations
+# Translations
 Please use [xDrip+ on Crowdin](https://crowdin.com/project/xdrip) if you have any questions about translations, or like to add a language or contribute to an existing language.
 
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,6 +10,7 @@ We will encourage you to open an issue if it is not already known and there is n
 * If it is a bug, please try to explain how to reproduce it, roughly which settings you are using, what you are seeing or not seeing. Screenshots welcomed.
 * If it relates to a new feature, please explain the rationale for the feature, use cases and detail your suggestion.
 
+<br/>
 
 # Contributing patches to xDrip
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,10 +1,16 @@
 # Reporting Issues with xDrip
+Before opening an issue, please connect with us by posting in the support channels:
```
tolot27, line 2: There should be an empty line after every heading and there is no need for double-empty lines in most cases.

"Before you open an issue, please contact us by posting in the support channels:"

CONTRIBUTING.md
```diff
@@ -1,10 +1,16 @@
...
+https://gitter.im/jamorham/xDrip-plus  
+We will encourage you to open an issue if it is not already known and there is no work-around.
+
 
 ### Please try to provide as much information as possible when raising an issue:
 
 * If it is a bug, please try to explain how to reproduce it, roughly which settings you are using, what you are seeing or not seeing. Screenshots welcomed.
 * If it relates to a new feature, please explain the rationale for the feature, use cases and detail your suggestion.
 
+<br/>
```
tolot27, line 13: Please remove all HTML tags unless they are really necessary (i. e. tables).

tolot27 (changes requested):

CONTRIBUTING.md
Navid200: Annotation 2021-06-14 123545
The image shows what it looks like now.  The red line is the distance between two consecutive paragraphs.  The blue line is the distance between the heading and the content of the same paragraph.    The red line being shorter than the blue line is not very nice.  
I have used the html tag to insert an extra line between paragraphs to make it easier to read and more pleasing to the eye.

Is there another way to add a space?

CONTRIBUTING.md
Navid200: Why should there be an empty line after each heading?

CONTRIBUTING.md
tolot27: Markdown is a text format an can be read very well as plain text (the main goal). It is not used to style anything. For styling, it is converted to some other formats which supports styles like HTML with CSS. If you look at https://github.com/NightscoutFoundation/xDrip/blob/master/CONTRIBUTING.md it can be read very well. I don't see any reason why the space between the headings and paragraphs should be changed. Nevertheless, it is the GitHub Markdown CSS which is applied to view Markdown files.

CONTRIBUTING.md
tolot27: The [GFM spec](https://github.github.com/gfm/#atx-headings) does not require an empty line after a heading bit it increases the readability in plain text mode and is recommend (see [Empty Lines Before And After](https://arcticicestudio.github.io/styleguide-markdown/rules/headings.html#empty-lines-before-and-after). The reason is likely historical because it allows to change between ATX and Settext headings. The later require a blank line because it supports multiline headings. But that's off-topic.

Navid200: I have placed a line after every heading now.

CONTRIBUTING.md
Navid200: The reason I added those lines was to make our writings look more professional.  The heading goes with the paragraph it is heading not the previous paragraph.  Look at any article or book.

But, my objective is to close the issue.  So, I have removed the lines as you desire.

tolot27 (approved):

jamorham (approved):

Issue #193, opened 2017-10-04
Missed Readings override silent


ruenightscout:
Hello, I have missed the missed readings alert a few times recently. I have discovered that if notification volume is on then missed reading alarm sounds but if my phone is on silent it automatically switchs off notification volume so missed reading alert just vibrates.   Is there a setting to make missed reading alert override silent mode? It seems to get its sound from whatever alert sound I have 'Other Alerts' set as but doesn't seem to recognised the override silent mode being ticked in the 'Other Alerts' settings page.  I can't see anything in the missed readings settings for choosing the sound or overriding silent.  It's a good sign that the system has worked well for so long that I haven't needed the missed readings alert! Many thanks, Esther

Navid200:
The missed readings alarm uses the same sound file as the fast rising and fast falling alerts.

There is a claim that on Android 9 and below, the override silent mode also followed the setting of fast  rising and fast falling alerts.
Can anyone confirm that?

Navid200:
@raduiliescu Do you still experience this issue?
What is the version of Android on your phone?

kasapis:
A friend is having the same issue, she has android 10.

Navid200:
If your Android is level 8 or higher, you can use notification channels to accomplish this.

Enable:
Settings -> Alarms and Alerts -> Glucose Alerts Settings -> Notification Channels

Now, you need to trigger the missed readings alert.  To artificially cause that, just change the hardware data source to something non-existent.

After the alert is triggered, use Android settings to go to Settings -> Apps & Notifications -> xDrip+ -> Notifications 
You should see a channel titled: bgMissedAlertChannel.
Tap on the following line (Missed reading Alert).
Now, you can customize your preferences.  You can choose the sound clip, if it should vibrate or not, and if it should override do not disturb or not.

The following may vary by phone.
If you are overriding do not disturb, carefully read the description:
![Screenshot_20210406-151801](https://user-images.githubusercontent.com/51497406/113767033-2ad54400-96ec-11eb-90db-3ff34a8ce64a.png)

So, in the case of the phone shown, "... when Do not disturb is set to Priority only, you need to ensure Android do not disturb is set to Priority only.

Navid200:
I don't believe this is an issue.

In the phone Android settings, please go to:
Settings Apps & notifications -> Notifications -> xDrip+

Make sure "Override Do Not Disturb" is enabled.
[...]

PR #1756, opened 2021-06-17
Override silent mode - other alerts Android 10 & 11

Navid200:
**Why do we need this?**
Override silent mode does not work for "Other Alerts" on Android 10 or 11.

This PR fixes that for missed signal alert only.
If it is merged and no issues are experienced, we can gradually move other members of the alert group to use the alert player (with future PRs).  But, I am doing it like this (one step at a time) to minimize impact on users.

Fixes: https://github.com/NightscoutFoundation/xDrip/issues/193  
This PR is one small step towards goals described here:  https://github.com/NightscoutFoundation/xDrip/issues/1762

**Is there a workaround?**
The only workaround, if you could call it that, is to disable the silent mode on the phone.  

* Unfortunately, I have changed two functions in AlertPlayer from private to public.  AlertPlayer is what all the main original alerts use.  So, I need you all developers to be as brutal as you can be in reviewing this PR.    Please point out anything you don't like.  Please let me know if there is a more graceful way to do any of it.  
* The alert volume is equivalent to the medium volume profile.  But, the user cannot change that.  The capability to change the volume profile, for this alert, is on my roadmap.  Please see issue 1762 linked above.  I have left it out of this pull request to simplify the review process by keeping the change small.

**Tests**  
I have tested the impact on Android 10 and 11, the objective of this PR, on with Android Studio virtual devices.  
I have tested this with an Android 9 handset.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
@@ -319,7 +319,7 @@ public class AlertPlayer {
         return false;
     }
 
-    private synchronized void playFile(final Context ctx, final String fileName, final float volumeFrac, final boolean forceSpeaker, final boolean overrideSilentMode) {
+    public synchronized void playFile(final Context ctx, final String fileName, final float volumeFrac, final boolean forceSpeaker, final boolean overrideSilentMode) {
         Log.i(TAG, "playFile: called fileName = " + fileName);
         if (volumeFrac <= 0) {
             UserError.Log.e(TAG, "Not playing file " + fileName + " as requested volume is " + volumeFrac);
@@ -609,7 +609,7 @@ public class AlertPlayer {
     }
 
     // True means play the file false means only vibrate.
-    private boolean isLoudPhone(Context ctx) {
+    public boolean isLoudPhone(Context ctx) {
         final AudioManager am = (AudioManager)ctx.getSystemService(Context.AUDIO_SERVICE);
         try {
             switch (am.getRingerMode()) {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/Notifications.java
@@ -82,6 +82,7 @@ public class Notifications extends IntentService {
     private static final boolean use_best_glucose = true;
 
     private static String last_noise_string = "Startup";
+    private static boolean AlertPlayerInstead = false; // Set this for alerts that must use AlertPlayer instead of the default
 
 
     Context mContext;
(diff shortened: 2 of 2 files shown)


Tigerroad:
Anything stopping you from merging this pull request. I'm a Pixel 4 with Android 11 and I am not getting any sound on my "missed readings" alerts when my phone is on silent mode at night time

Navid200:
@Tigerroad Thanks for your interest.
It needs to be reviewed to verify that it will not break anything;  and that it is necessary.
The developers are volunteers and we need to be patient.
The feedback you have provided is valuable.  Thanks

jamorham:
Okay so first off I don't think setting a static flag is the way to handle this as it potentially creates a race condition as well as not being a very clean way to do it. Without significant refactoring I think it could be better to extend the signature of `OtherAlert` to accommodate the flag and wrap all existing calls with that set to false.

But there is a deeper question as to whether this changes existing behavior in terms of that we ask the user whether to bypass silent mode for regular alarms and also for these type of alerts but isn't the issue that this doesn't work on android 10+ so instead of this type of workaround shouldn't you instead be targeting the conditional code that attempts to do that already and going to alert player if the sdk version is android 10+ ? Wouldn't that be a simpler solution needing only some small changes in that method? Then you don't need to change any signatures or flags? Please let me know if I've misunderstood the code flow here.

Navid200:
@jamorham I hope I am not misunderstanding the question.

The silent mode effect on the alert is accomplished for alerts under "Glucose Level Alerts List" (the original alerts) by lowering the volume to 0.  So, if the user looks at the logs, they will see that the alarm has been triggered but has not played because the volume has been reduced to 0.
This mechanism works on Android 10 and 11.

But, the silent mode for the other alerts is accomplished using a different mechanism:
The alert is only categorized as an alarm if the override silent mode is set.  If not, the alert is categorized as a media.  This, for some reason, does not work on Android 10 and 11.
I don't know how to fix it for Android 10 and 11.

So, I asked myself, if the main glucose level alerts are working fine and we are happy with them, why not use the same mechanism they use for other alerts?
That's all I am doing.  I am using the same mechanism already used for all the glucose level alerts.
  
May I ask what is wrong with the way override silent mode is accomplished with the existing glucose level alerts?

Navid200:
@jamorham Any alert under "Glucose Level Alerts List" allows the user to choose any of the following for the alert:
specific audio file
enable or disable the vibrator
override silent mode or not
Also, such alerts follow the selected volume profile.

The other alerts don't consistently offer these selections.  None of the other alerts follow the selected volume profile.
I would like all the other alerts to offer the same capabilities.
This PR is only a first step.  Please see:  https://github.com/NightscoutFoundation/xDrip/issues/1762

Navid200:
I am closing this pull request because I believe xDrip, as is, can handle override silent mode fine.  It's just that Android needs to be set up correctly.
Please see this:
https://github.com/NightscoutFoundation/xDrip/issues/193#issuecomment-903731657
Please try that suggestion.  If your setup looks different and you cannot get operation, please post in that issue and we can go over the details.

Navid200:
The issue that the missed readings alert does not have a dedicated override silent mode setting and inherits it from the other alerts remains.
I will open another PR for addressing that.  The reason I have closed this PR is that I was under the impression that the functionality did not work for Android 10 or 11.  So, the changes I had made in this PR were drastic, and unnecessary, just to add a new dedicated setting.

PR #1760, opened 2021-06-19
Lower the FUZZER value to 37.5 seconds

### LINKED ISSUE #1537: The glucose readings aren't added in 5 minute intervals anymore

Navid200:
**Why do we need this?**
This reduces the delta between what is shown on screen and what is in the data table ( what is sent to Nightscout).

A consequence of the delta being great is that occasionally, when the reading arrival time is very close to the boundary of the grid buckets, the readings on the xDrip main screen are not equidistant. Lowering the value of the FUZZER to 37.5 seconds (from 2.5 minutes) will make the points look equidistant even if they are not.

My hope is that, if everyone is happy with this new setting, some day, we will remove the FUZZER altogether. So, this is a first step.

Fixes #1537

**Is there a workaround?**
No

No.  Nothing changes from the user perspective if this is merged unless the user enables engineering mode and enables the setting.

**Tests**
This has been tested on an Android 9 handset.

**Changed files**
Code formatting in BgGraphBuilder.java has caused many changes. The only real changes are the following line numbers:
85, 457, 458, 1430, and 1431  

**Changed Interface**
This image shows the new page at: Settings -> xDrip+ Display Settings -> Graph Settings  
Screenshot_20210711-231826
The option at the very bottom of the page, Lower FUZZER, is a new setting.  It cannot be enabled except in engineering mode.  After enabling engineering mode and enabling the new setting, a phone restart is required to see the impact of lower FUZZER on the main screen.  If engineering mode is disabled, after a phone restart, the setting will be automatically disabled.   

**Previous closed PR**
https://github.com/NightscoutFoundation/xDrip/pull/1750


jamorham (approved):
Okay lets try this

PR #1743, opened 2021-05-30
Tzachi libre improvements 4

tzachi-dar:
This replaces: https://github.com/NightscoutFoundation/xDrip/pull/1640

It adds the following for libre2:
    Per minute values.
    gap closing.
    Filtering of bad points.
    Data smoothing.

For libre 1 filtering of bad points is added.

This has passed all my testing and is ready for review.

PR #1761, opened 2021-06-19
interprete only 344 bytes for Libre2 FRAM NFC data

tzachi-dar:
Fix a NPE introduced with #1743

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreBlock.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreBlock.java
@@ -101,7 +101,7 @@ public class LibreBlock  extends PlusModel {
         final LibreBlock lb = create(reference, timestamp, blocks, byte_start, patchUid, patchInfo);
         if (lb != null) {
             lb.save();
-            if(byte_start == 0 && blocks.length == 344 && allowUpload) {
+            if(byte_start == 0 && blocks.length == Constants.LIBRE_1_2_FRAM_SIZE && allowUpload) {
                 Log.d(TAG, "sending new item to queue");
                 UploaderQueue.newTransmitterDataEntry("create" ,lb);
             }
@@ -161,7 +161,7 @@ public class LibreBlock  extends PlusModel {
         // Using this syntax since there is no way to tell the DB which index to use.
         // Using ActiveAndroid method would take up to 8 seconds to complete.
         try (Cursor cursor = db.rawQuery("select * from libreblock  INDEXED BY  index_LibreBlock_timestamp " +
-                "WHERE bytestart == 0 AND (byteend == 344 OR byteend == 44) " +
+                "WHERE bytestart == 0 AND (byteend == " + Constants.LIBRE_1_2_FRAM_SIZE + " OR byteend == 44) " +
                 "AND timestamp BETWEEN " + start_time + " AND " + end_time +
                 " ORDER BY timestamp DESC LIMIT 1", null)) {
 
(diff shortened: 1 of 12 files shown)


jamorham (approved):

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -277,7 +277,11 @@ public static boolean HandleGoodReading(final String tagId, byte[] data1, final
     
     // returns true if checksum passed.
     public static boolean HandleGoodReading(final String tagId, byte[] data1, final long CaptureDateTime, final boolean allowUpload, byte []patchUid,  byte []patchInfo, boolean decripted_data ) {
-        Log.e(TAG, "HandleGoodReading called");
+        Log.e(TAG, "HandleGoodReading called dat1 len = " + data1.length);
+        if(data1.length > 0x158) {
+            // It seems that some times we read a buffer that is bigger than 0x158, but we should only use the first 0x158 bytes.
+            data1 = java.util.Arrays.copyOfRange(data1, 0, 0x158);
```
tolot27, line 283: Is this save for all Libre NFC readings? Maybe we should declare a constant for it and use this constant to iterate over the array rather than shrinking it.

tolot27 (commented):
Why do we read sometimes more then 0x158 bytes? Is this caused by read errors? In such cases, the checksum should be wrong.

tolot27:
@tzachi-dar Please address my questions. Hopefully, we can get it merged into this weekend release.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
tzachi-dar: Right now it is the biggest libre buffer that we are aware of, and many places it is used to check for a legal buffer.
I'll change it to a constant later, and you will see all it's hidden usages.

tzachi-dar:
> Why do we read sometimes more then 0x158 bytes? Is this caused by read errors? In such cases, the checksum should be wrong.

We have a single block read and multi block reads. See the code in https://github.com/NightscoutFoundation/xDrip/blob/cc967428b40259d43890380de10174e16d7d5e2d/app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java#L577
Each block is 8 bytes long. so 0x158==344 or 43 blocks.
When doing multiblock reads, we read 3 block each time so we end up with reading 45 blocks for the data, although we only need 43. (so we have 2 invalid data.)

app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreBlock.java
```diff
@@ -161,7 +161,7 @@ public static LibreBlock getLatestForTrend(long start_time, long end_time) {
         // Using this syntax since there is no way to tell the DB which index to use.
         // Using ActiveAndroid method would take up to 8 seconds to complete.
         try (Cursor cursor = db.rawQuery("select * from libreblock  INDEXED BY  index_LibreBlock_timestamp " +
-                "WHERE bytestart == 0 AND (byteend == 344 OR byteend == 44) " +
+                "WHERE bytestart == 0 AND (byteend == " + Constants.LIBRE_1_2_FRAM_SIZE + " OR byteend == 44) " +
```
tolot27, line 164: Please can you replace the `44` by a constant, too?
What size is this?

tolot27: I know, this is a bit off-topic but just have seen this.

tolot27 (approved):
Thanks for this fix. If you have time, please turn the number `44` into a constant and provide java comments at the declarations of the constants.

PR #1764, opened 2021-06-22
hide password of Dexcom Share Follower

### LINKED ISSUE #1187: Dexcom Share Following Password visibility
The login password for Dex Share Following is always visible. It should be invisible except the last later during input.

tolot27:
This PR hides the password in the Dexcom Share Follower settings. This is consistent with the other password fields, i. e. Dexcom Account Password of the Dexcom Share Server Upload.

The two additional commits cleaning up the xml file similar to 6d3f629.

fixes #1187

### DIFF
--- a/app/src/main/res/xml/pref_data_source.xml
+++ b/app/src/main/res/xml/pref_data_source.xml
@@ -12,7 +12,7 @@
             android:key="dex_collection_method"
             android:summary="@string/how_receive_data"
             android:title="@string/hardware_data_source" />
-        <PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android"
+        <PreferenceScreen
             android:icon="@drawable/ic_nfc_grey600_48dp"
             android:key="xdrip_plus_nfc_settings"
             android:summary="@string/nfc_options"
@@ -140,10 +140,7 @@
             android:summary="Login password for Dex Share Following"
             android:title="Share Password" />
 
-
-
-
-        <PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android"
+        <PreferenceScreen
             android:key="xdrip_plus_g5_extra_settings"
             android:summary="@string/advanced_g5_settings"
             android:title="@string/g5_debug_settings">
--- a/app/src/main/res/xml/pref_data_source.xml
+++ b/app/src/main/res/xml/pref_data_source.xml
@@ -288,7 +288,7 @@
                 android:key="g5-battery-warning-level"
                 android:title="@string/title_g5_battery_warning_level"
                 android:defaultValue="300"
-                android:numeric="integer"
+                android:inputType="number"
                 />
             </PreferenceCategory>
 
--- a/app/src/main/res/xml/pref_data_source.xml
+++ b/app/src/main/res/xml/pref_data_source.xml
@@ -133,7 +133,7 @@
             android:title="Share Username" />
         <EditTextPreference
             android:defaultValue=""
-            android:inputType="textNoSuggestions|textVisiblePassword"
+            android:inputType="textNoSuggestions|textPassword"
             android:key="shfollow_pass"
             android:maxLines="1"
             android:singleLine="true"


Navid200:
I personally have no problem with this as a user.
But, I can predict posts on facebook by people who have difficulty verifying their password questioning why they cannot see the password so that they could verify it.

For an app that let's you log into your bank account, I have no doubt this is a requirement.  And, there are many elderly individuals who don't use mobile devices because they cannot deal with all of its requirements.  So, they walk into a bank instead.
There is a justification for that.  We cannot allow someone to look over someone else's shoulder and find out their login password to their bank account.

But, who is going to look over someone's shoulder to access their Dexshare account?  What for?  Do we really want to make it harder for the elderly to use xDrip?  And what would be the justification?

tolot27:
The share follow password (`shfollow_pass`) can be extracted from an exported preferences file, if necessary. The password is not visible in the Dexcom app, as well. It is just a matter of privacy and consistency. xDrip should not be used to remember passwords. A password manager is more suitable for that purpose. And no one expects that a password is visible. The opposite is the case.

Navid200:
If there is a malfunction, the first thing one would check is the password.  If the password is invisible, the only option is to delete it and reenter it.  

Imagine a parent wondering why the child's phone not uploading any longer, and wondering if the password has been accidentally and unintentionally changed by the child.
It's not the end of the world.  But, it will be an added nuisance, which doesn't exist currently.  

So, the question is what's the point and is it worth the added new hassle?

Navid200:
If this is absolutely necessary (I personally don't think it is), can we have a new button added to the page, which would make the password visible?  And we can have the password invisible by default and we can make it invisible again after a restart.

tolot27:
> So, the question is what's the point and is it worth the added new hassle?

This PR does not introduce any hassle, it fixes a security and privacy problem. 
You expect some hassle somewhere that in the future which is coursed by a potentially loss of a password from someone. 
That are completely different things. 

Every password field in the modern world is hidden except this one in xDrip.

Also, the following statements are contrary:

> And, there are many elderly individuals who don't use mobile devices because they cannot deal with all of its requirements.

and

> Do we really want to make it harder for the elderly to use xDrip?

It is very likely that elderly people don't use xDrip at all because they either don't have a smartphone (first statement) or they are using the much simpler Dexcom Follow app which does not require any password, just an invitation.

Summary of the Dexcom Share requirements and description of the security problem:

- A Dexcom account is necessary and the password for it isn't stored at all at the Dexcom servers, just a salted hash.
- During account creation, no one sees the password, neither during typing nor during SSL transport. 
- Maybe or very likely the user stores the password in the browser password manager or some external password manager. There, the password is stored in encrypted form.
- With the Dexcom account it is possible to configure followers, view patient privacy information and order new sensors and transmitters. That's all possible, because most of the users do not create two different accounts at the Dexcom servers.
- In xDrip it is possible to configure Upload to Dexcoms Share servers. The account name and password is necessary to configure it AND the password input field is hidden.
- Inviting followers is also possible at the same config page.
- In the rare case a follower is not using Dexcom Follow, but xDrip without Nightscout (which syncs all kind of data and not just sgvs), it can use Dex Share Follower.
- The owner of the account has to share the password with the follower which is risky as explained above.
- The follower enters the password into a visible text field and xDrip is storing the password in plain text.

If the follower loses its phone, the finder can read the password from the field or the config and can login to Dexcom servers. What a nightmare even if not likely.

What would an owner of the Dexcom account would do if it cannot remember the password? 1. Looking into its password managers if any. 2. Resetting the password at the Dexcom server.
Even in the rare case the owner would look at the Dexcom Share upload settings in xDrip, no password is visible. Guess, no one expects this. And lastly, the owner could ask some of the followers for the password. But that is very unlikely because both account owner and follower should know that there is at least one app in the world - the medical app xDrip storing bg values and treatments - which also stores the Dexcom account password unencrypted and shows this in the follower configuration. How unlikely is that!

Indeed, follower and account holder could be the same person or from the same family. But this does not matter because password reset is much easier and a known practice.

I will not discuss this further. The developer community should decide.

Navid200:
>Also, the following statements are contrary:
>>And, there are many elderly individuals who don't use mobile devices because they cannot deal with all of its requirements.

>and
>>Do we really want to make it harder for the elderly to use xDrip?

Since you say my statements are contrary, I will explain here.  Hopefully, my point will be clear then.
An elderly could be using xDrip because their son or daughter has set it up for them.  
If it stops working, the son/daughter may like to resolve it over the phone instead of having to go to them.  If we make xDrip harder to use in a scenario like that, what we gain from the proposed change is a legitimate question to answer.  

Every change we make to xDrip very likely has consequences.  It's only fair to consider them.

jamorham:
I'm not that familiar with this data source so please educate me, but am I right in thinking that xDrip can both upload and download to this data source?

The upload password isn't obscured and neither are others in the configuration. Indeed xDrip allows complete backup of the settings and I would think the ability to view passwords is very useful. 

What type of attack scenario are we trying to protect against here? Is this just if someone is looking over someone's shoulder while they are editing settings and they would see the password? Or are we trying to protect against an attack where someone has control over a user's phone? Or that one user would set up a follower password for another user but doesn't want that user to easily be able to pass that password on to another third party? What is the risk profile that is being thought of?

If this is just to protect against someone casually seeing the password then I would suggest just wrapping it inside another screen so its only visible when clicked if that is needed? If it isn't already obscured like this.

If it is when someone is in control of a user's phone then I'm thinking the whole settings feature would want a password lock so that it can't be accessed.

PR #1767, opened 2021-06-25
add unit gram (g) to carb input activities

### LINKED ISSUE #1339: Ambiguous Carb Unit
I have the problem that I usually calculate my Insulin based on KEs (german meaning: Kohlenhydrat-Einheiten, english translation: Carbohydrate units), which equal 10g of carbohydrates. So then I multiply by my factor (for instance 1.25), and I directly get the amount of Insulin I need to inject. However, xDrip+ is very unambiguous when it comes to those units.

tolot27:
Add gram (g) to the GUI labels/texts were carbohydrates are shown (see screenshots). TextToSpeech should not be affected.

Fixes #1339

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -1469,7 +1469,7 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
             case "carbs":
                 if (!carbsset && (thisnumber > 0)) {
                     thiscarbsnumber = thisnumber;
-                    textCarbohydrates.setText((int) thisnumber + " carbs");
+                    textCarbohydrates.setText((int)thisnumber + " g carbs");
                     carbsset = true;
                     Log.d(TAG, "Carbs eaten: " + thisnumber);
                     btnCarbohydrates.setVisibility(View.VISIBLE);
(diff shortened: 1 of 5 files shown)


tolot27:
One translation/abbreviation problem remains. The term "carbs" in the second screenshot is not translated. In theory, the source string `carbs` could be used but it is not abbreviated and in some languages a long word. That would not fit in line. Either we introduce a new source string like `carbs_abbreviated` or one called `carbohydrates`.

jamorham:
What is the situation if the user wants to use for example carbohydrate portions as their input units. I know the ratios are likely set vs grams but I wonder if that is the only place and so by specifying grams it might make it too restrictive for users who might like to work like that.

tolot27:
> What is the situation if the user wants to use for example carbohydrate portions (10g) as their input units? 

@ jamorham @vguttmann Were can a user change the input unit? Currently, xDrip assumes g carbohydrates and not portions. Here in Germany it is called carb units like insulin units and mostly 1 carb portion/unit equals 10 g carbohydrates.

As far as I know, we do not have such preference setting.

tolot27:
Any thoughts on this PR, especially about carb portions?

tolot27:
There is currently no way to switch between "g carbohydrates" and "carbohydrate portions (10g)". We can open an issue on request.

Hence, I propose to merge this PR because it is simple fix for #1339.

jamorham (approved):

PR #1792, opened 2021-07-28
Update FirmwareCapability.java

Navid200:
There seems to be a new G5 firmware released.
I don't have access to the transmitter.  It is from a post on a forum.

225229444_4036987536350326_6897882416693704483_n


tolot27 (approved):

Navid200:
After updating xDrip, this is what is shown on the system status page, a different firmware!
218960171_4039399212775825_5631109065694584644_n

I am closing this PR as there doesn't seem to be a new firmware.  I will not jump the gun next time I see something like that by opening a PR.

PR #1800, opened 2021-08-03
XdripWebService: send HTTP 500 when exception occurs during processing

jwoglom:
Previously, if an error occurred while processing a web service
endpoint then the connection would be terminated with an EOF.

This adds a test which verifies that instead a HTTP 500 is sent
with MIME type text/plain, containing the traceback of the error
to ease debugging.


tolot27 (approved):
LGTM with the exception of the renaming of `RouteFinderTest`. But since it is a test class it should be okay.

jwoglom:
The rename of the test file was intentional -- the source file was RouteFinder.java but the test file was named RouterFinderTest.java, so it made sense imo to fix the mistake. Thanks for the LG. @tolot27

jamorham (approved):

PR #1805, opened 2021-08-08
add Bulgarian language support

tolot27:
The translation to Bulgarian has started: https://crowdin.com/project/xdrip/bg.
This PR enables this language in xDrip and helps the translation process further.

### DIFF
--- a/app/src/test/java/com/eveningoutpost/dexdrip/ui/TranslationTest.java
+++ b/app/src/test/java/com/eveningoutpost/dexdrip/ui/TranslationTest.java
@@ -7,11 +7,13 @@ import com.eveningoutpost.dexdrip.R;
 import com.eveningoutpost.dexdrip.RobolectricTestWithConfig;
 import com.eveningoutpost.dexdrip.xdrip;
 
-import org.apache.commons.lang3.ArrayUtils;
 import org.junit.Test;
 
 import java.text.MessageFormat;
+import java.util.Arrays;
 import java.util.Locale;
+import java.util.Set;
+import java.util.TreeSet;
 
 import lombok.val;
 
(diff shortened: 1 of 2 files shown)


Navid200:
I am not very familiar with the translation process.

Why does this PR change the TranslationTest file?  That seems like a change that is going affect all languages and not just Bulgarian.

app/src/test/java/com/eveningoutpost/dexdrip/ui/TranslationTest.java
```diff
@@ -30,7 +32,8 @@ public void testFormatStrings() {
         final Configuration config = xdrip.getAppContext().getResources().getConfiguration();
         val internal = xdrip.getAppContext().getResources().getStringArray(R.array.LocaleChoicesValues);
         val extra = new String[]{"ar", "cs", "de", "el", "en", "es", "fi", "fr", "he", "hr", "it", "iw", "ja", "ko", "nb", "nl", "no", "pl", "pt", "ro", "ru", "sk", "sl", "sv", "tr", "zh"};
-        val locales = ArrayUtils.addAll(internal, extra);
+        Set<String> locales = new TreeSet<>(Arrays.asList(internal));
```
jamorham, line 35: is this change just to ensure entries are not duplicated?

jamorham:
Can you confirm that the translation test still works okay? Otherwise looks fine to me assuming the Bulgarian word for Bulgarian is correct also.

Navid200:
According to Google translate, the text translates to "Bulgarian language" rather than "Bulgarian".

tzachi-dar (approved):
LGTM

tolot27:
> I am not very familiar with the translation process.

Thanks for looking at this PR.

> Why does this PR change the TranslationTest file? That seems like a change that is going affect all languages and not just Bulgarian.

It is mentioned in the commit message of 2f27dc9 and it change does not affect any language. It avoids only duplicated tests of the translations.
The call `xdrip.getAppContext().getResources().getStringArray(R.array.LocaleChoicesValues);` reads all language codes from the xml resources file. Line 'val extra = new String[]{"ar", "cs", "de", "el", "en", "es", "fi", "fr", "he", "hr", "it", "iw", "ja", "ko", "nb", "nl", "no", "pl", "pt", "ro", "ru", "sk", "sl", "sv", "tr", "zh"};' declares a constant array with almost of the same codes present in the xml file.
With 'val locales = ArrayUtils.addAll(internal, extra);' they will be concatenated into a new array and duplicated. Hence, they will be tested twice which takes unnecessary long time. I've just replaced the array with a `TreeSet` which avoids duplication and also orders the codes.

tolot27:
> Can you confirm that the translation test still works okay?

Yes, it does. Unfortunately, I cannot link to the CI unit test anymore, because it was run too long ago.

> Otherwise looks fine to me assuming the Bulgarian word for Bulgarian is correct also.

I got it from the official [ISO 639-1 list](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes).

PR #1807, opened 2021-08-09
support TLSv1.2 for NightscoutUploader on pre-Lollipop devices 

tolot27:
Earlier, it was [announced](https://help.heroku.com/SXWA00YI/tls-v1-0-v1-1-end-of-life-schedule) that Heroku will drop TLS v1.0 and v1.1 by July 31, 2021 (as major browser vendors already did). Since August, xDrip on older Android versions below 4.4 but maybe also on some 4.4 devices cannot connect to Nightscout instances running on Heroku because of this TLS drop. This is a breaking bug of xDrip which is addressed by this PR.
It implements the same workaround for NS uploader which was already implemented for Nightscout Follower. It also logs the TLS version used to connect to the NS instance, if extra logging is enabled for NS uploader (`NightscoutUploader:v`), but without intercepting the communication if the extra logging is not enabled.
Also, OkHttp library and logger is updated to [v3.2.13](https://square.github.io/okhttp/changelog_3x/#version-31213), which is just a bug fixed version without new features.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -327,9 +327,8 @@ dependencies {
     implementation 'com.google.android.gms:play-services-location:10.2.1'
     implementation "com.google.android.gms:play-services-gcm:10.2.1"
     implementation 'com.squareup.wire:wire-runtime:2.2.0'
-    implementation 'com.squareup.okhttp:okhttp:2.7.5'
-    implementation 'com.squareup.okhttp3:okhttp:3.12.1'
-    implementation 'com.squareup.okhttp3:logging-interceptor:3.10.0'
+    implementation 'com.squareup.okhttp3:okhttp:3.12.13'
+    implementation 'com.squareup.okhttp3:logging-interceptor:3.12.13'
     implementation "com.newrelic.agent.android:android-agent:5.27.1"
     //implementation 'com.jakewharton.retrofit:retrofit1-okhttp3-client:1.1.0'
     implementation 'com.squareup.retrofit2:retrofit:2.4.0'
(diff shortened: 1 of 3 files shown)


tolot27:
Here is an unsigned debug build which requires reinstalling xDrip from untrusted source. Backup of DB and settings are highly recommend beforehand, otherwise you will loose your data.

https://www.dropbox.com/s/ktjsibkvmxwykvq/xDrip_20210804_e4f7fa3_NSUploadTLSv12_debug.apk?dl=0

tzachi-dar:
This has worked well with the following phone:
Samsung Galaxy Young 2.
According to https://www.gsmarena.com/samsung_galaxy_young_2-6480.php this is android 4.4.2
Many thanks for working on this.

tolot27:
Good to know. It would be great if someone reviews this PR.

app/build.gradle
```diff
@@ -327,9 +327,8 @@ dependencies {
     implementation 'com.google.android.gms:play-services-location:10.2.1'
     implementation "com.google.android.gms:play-services-gcm:10.2.1"
     implementation 'com.squareup.wire:wire-runtime:2.2.0'
-    implementation 'com.squareup.okhttp:okhttp:2.7.5'
```
jamorham, line 330: I think this line should be restored otherwise okhttp 2.5 will be used in the dependency instead of 2.7.5

tolot27: I don't know why it should use v2.5 if no dependency for v2.x is declared at all. I got no compilation errors. Anyway, I'll revert this change in favour of #1812.

jamorham (approved):

Issue #565, opened 2018-08-28
Bug: Manually Entering U to 3+ DP miscalculates Treatment


old-square-eyes:
Entering 0.299U results in 99U being registered.

Navid200:
Seems the problem has been resolved for units of insulin.
But, the problem remains for carbs.

Navid200:
@old-square-eyes Thanks for reporting this.
I am working on it.  Hopefully it will be fixed soon.

PR #1808, opened 2021-08-11
Carb entry error

Navid200:
Updated current description

Why we need this
When entering carb values, if the user enters 3 decimal points, what is entered is an integer value equal to the last two digits of the fractional part of the number. For example, if the user enters 1.298, what will be entered is 98, which is a significant error.

Fixes https://github.com/NightscoutFoundation/xDrip/issues/565

Is there a workaround?
The user can avoid entering more than 2 decimal points.
The problem is that anyone could accidentally enter too many digits.

Are there any side effects?
No

Tests
I have tested this on an Android 8 follower

Credits
Thanks to Mathias Walter for suggesting the alternative regex

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -1359,7 +1359,6 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
         allWords = allWords.trim();
         allWords = allWords.replaceAll(":", "."); // fix real times
         allWords = allWords.replaceAll("(\\d)([a-zA-Z])", "$1 $2"); // fix like 22mm
-
         allWords = allWords.replaceAll("([a-zA-Z]+ \\d{1,3}(?:\\.\\d)*)(\\d+(?:\\.\\d)* [a-zA-Z]+)", "$1 $2"); // fix multi number order like blood 3.622 grams
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
             allWords = allWords.replaceAll(new String(RTL_BYTES, StandardCharsets.UTF_8), "");
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -1360,13 +1360,7 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
         allWords = allWords.replaceAll(":", "."); // fix real times
         allWords = allWords.replaceAll("(\\d)([a-zA-Z])", "$1 $2"); // fix like 22mm
 
-/** The following line converts 1.234 carbs to 1.2 34 carbs.  It inserts a space between 1.2 and 34.  This results in an array of two words
- *  to be converted into an array of 3 words.  This then results in carb entry with more than 2 decimal points to be unintentionally
- *  converted into an integer containing the last 2 decimal points of the original fractional number.
- *  I am commenting it out and will ask the developers if it is really needed.
- *  Navid  August 10, 2021
-*/
-//        allWords = allWords.replaceAll("([0-9]\\.[0-9])([0-9][0-9])", "$1 $2"); // fix multi number order like blood 3.622 grams
+        allWords = allWords.replaceAll("([a-zA-Z]+ \\d{1,3}(?:\\.\\d)*)(\\d+(?:\\.\\d)* [a-zA-Z]+)", "$1 $2"); // fix multi number order like blood 3.622 grams
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
             allWords = allWords.replaceAll(new String(RTL_BYTES, StandardCharsets.UTF_8), "");
         }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -1359,7 +1359,14 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
         allWords = allWords.trim();
         allWords = allWords.replaceAll(":", "."); // fix real times
         allWords = allWords.replaceAll("(\\d)([a-zA-Z])", "$1 $2"); // fix like 22mm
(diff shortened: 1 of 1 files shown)


Navid200:
Why would the user enter "Blood 3.622 grams"?  Regardless of if they are using speech or not.
Perhaps, originally the menu allowed such an entry.
But right now, the user has to first select blood or carbs or insulin; then, enter a value.
After all the entries are entered, xDrip combines then with space between.

I don't understand the need for this splitting.   Unless this applies to a different keypad that has to do with simulation or some other function that I have never used.

Is there any entry in xDrip that accepts blood and carbs entered in one shot by the user?  If not, I don't see the need for this splitting.

tolot27 (changes requested):
Entering values manually using the keypad is totally different from speech input. If you use voice control you would not speak extra voice commands to switch between the different treatment input pages. You just say: "blood 3.6 and 10 carbs and 2 insulin units" or something like that.
The speech-to-text processing is completely separated from manual input.
To resolve #565, just the manual input route should be fixed and not the speech processing. Sometimes, the speech recognition returns strange text which is filtered/adapted by the function `naturalLanguageRecognition` you have changed.

Navid200:
There is only one route.  The splitting should only be applied to speech input.  But, it also is applied to text entry resulting in the reported bug.

After 1.234 is split into two entries of 1.2 and  34, the damage is done.  
I can place a conditional before this and for carb entry only convert that to 1 as the carbs should be entered as integer anyway.
But, what happens if someone adds another speech command tomorrow that requires another manipulation added, which screws up text entry again?

The best solution, in my opinion, is to only have speech go through naturalLanguageRecognition, but, not text from the keypad.  Currently, keypad entries also go through it.

I am happy to do the work.  But, can we please wait for input from jamprham as to his preferred approach?  I will be happy to do the work.  I am absolutely not trying to avoid to do the work.  But, I'd like to do it after we get more input.

Navid200:
The speech recognition scheme has limitations.
Someone posted on facebook claiming that when he attempted to use the speech command to clear the transmitter queue, the speech recognition typed "clear transmitter cue", which did not clear the queue.

I told them that they should long press on the microphone symbol, instead of tapping on it, so that they can enter the command using the keyboard.
In fact, all my guides instruct people to use the keyboard instead of using speech.

Now, I hope no one adds a replacement of cue with queue in xDrip.  Please don't do that.  

I admire the intent of speech input in xDrip.  But, it has serious limitations.

tolot27:
> > To resolve #565, just the manual input route should be fixed and not the speech processing.
> 
> Would you please highlight where the "manual input route" is?

I've debugged this. Unfortunately, input using the keypad/touchscreen is also passed to the voice input processing. I'm not particularly happy about that.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -1359,7 +1359,14 @@ private void naturalLanguageRecognition(String allWords) {
         allWords = allWords.trim();
         allWords = allWords.replaceAll(":", "."); // fix real times
         allWords = allWords.replaceAll("(\\d)([a-zA-Z])", "$1 $2"); // fix like 22mm
-        allWords = allWords.replaceAll("([0-9]\\.[0-9])([0-9][0-9])", "$1 $2"); // fix multi number order like blood 3.622 grams
```
tolot27, line 1362: Commenting this line out removes the implemented fix. It looks like sometimes people speak two voice treatment commands right after another, maybe in different word/number ordering like "blood 3.6" followed by "22 grams". The voice recognition engine concatenates the recognized text to a single string without a separator between the two commands.
The regex tries to fix this but is incomplete. Here is an improved regex: `([a-zA-Z]+ \d{1,3}(?:\.\d)*)(\d+(?:\.\d)* [a-zA-Z]+)`. This handles the following cases correctly:
```
blood 3.622 grams
grams 223.6 blood
grams 1802.9 blood
grams 180114 blood
1.298 carbs
```

They will be changed to:
```
blood 3.6 22 grams
grams 22 3.6 blood
grams 180 2.9 blood
grams 180 114 blood
1.298 carbs
```

Now, there is a separator between the numbers but only if the input string starts with a word and ends with a word. The latest example is unaffected.

All examples will be processed correctly with this regex.

tolot27, line 1362: Commenting this line out removes the implemented fix. It looks like sometimes people speak two voice treatment commands right after another, maybe in different word/number ordering like "blood 3.6" followed by "22 grams". The voice recognition engine concatenates the recognized text to a single string without a separator between the two commands.
The regex tries to fix this but is incomplete. Here is an improved regex: `([a-zA-Z]+ \d{1,3}(?:\.\d)*)(\d+(?:\.\d)* [a-zA-Z]+)`. This handles the following cases correctly:
```
blood 3.622 grams
grams 223.6 blood
grams 1802.9 blood
grams 180114 blood
1.298 carbs
```

They will be changed to:
```
blood 3.6 22 grams
grams 22 3.6 blood
grams 180 2.9 blood
grams 180 114 blood
1.298 carbs
```

Now, there is a separator between the numbers but only if the input string starts with a word and ends with a word. The latest example is unaffected.

All examples will be processed correctly with this regex.

Navid200:
@tolot27 Are you saying I replace 
("([0-9]\\.[0-9])([0-9][0-9])", "$1 $2")
with
([a-zA-Z]+ \d{1,3}(?:\.\d)*)(\d+(?:\.\d)* [a-zA-Z]+)
on line 1362?

tolot27:
BTW: We should consider implementing a unit test for the voice command fix. But this requires separating the multiple regex from the remaining function code.

tolot27:
@Navid200 I dug into it more and this PR indeed solves a bug in xDrip - even if we are not happy about the common handling of manual and language input by the natural language processing code and even if language recognition is not perfect.

> I don't believe we should be telling people how much insulin to take.

We don't do that. We just improve the code to handle the input. The processed values should match as close to the input as possible.

> Even more so if they are visually impaired such that they need to use voice to tell xDrip how much they have eaten.

That's way it is important to fix the processing because otherwise such people does not have any chance to get their voice input processed correctly.

Hence, please open this PR again and add the [comment](https://github.com/NightscoutFoundation/xDrip/pull/1808#discussion_r735097769), otherwise I'll open a new PR.

Issue #1736, opened 2021-05-21
Issues experienced by G6 Firefly users


Navid200:
#### Your environment
xDrip: May 21, 2021
Android: any
<br/>  

#### Expected behavior
No confusing, misleading, or unexpected logs, messages, menu options, or behavior
<br/>  

#### The following is a list of issues that could be observed by a G6 Firefly user.
1- Temperature is not shown on the system status page.  
2- Add Calibration is not an option.  There is only an initial calibration option.  
3- The following message is observed after a calibration.  Got invalid impossible slope calibration!  
4- Multiple instances of the following log are observed after a calibration.  Received backfill packet out of sequence  
5- Pop-up option to hard reset transmitter  
6- Failure to start sensor
7- The following message after calibrating using the treatment menu.  Cannot send calibration older than 1 hour.  

#### Summary
None of these issues is critical.
They can all be circumvented, or safely ignored.  However, for an inexperienced user, these can be very confusing or time consuming.
<br/>  

#### Steps to reproduce the behavior:
Use a Firefly G6, meaning a G6 transmitter with a firmware starting with 2.18 or 2.27.
<br/>  

#### There are open or closed issues related to some of these:  
https://github.com/NightscoutFoundation/xDrip/issues/1239
https://github.com/NightscoutFoundation/xDrip/issues/1415
https://github.com/NightscoutFoundation/xDrip/issues/1310
<br/>  

#### Plan  
I would like to address all of these.  However, it may take time.  We have to ensure nothing is affected negatively before any fix is accepted.   There are already 2 open pull requests for the first two items on the list.

tolot27:
@Navid200 Please can you reformat your issue text to use a [task list](https://github.github.com/gfm/#task-list-items-extension-) rather than the numbered list and append the related PR/issue to each list item? Then, we can simply click on the checkbox if a task has been completed. Thanks!

If you need help to address the remaining issues, please inform me at Gitter.

robertrub:
Strangely, in the logs, after G6/Firefly restart (code 0) the calibration was _accepted_ and _rejected..._
![Screenshot_20210814-160516_xDrip+](https://user-images.githubusercontent.com/76626474/129449385-382f704e-d9d0-4812-9a36-0fd5dc42bd58.jpg)
Very confusing but 5 min later, without having done anything, BGs appeared.

PR #1810, opened 2021-08-13
Don't offer to hard reset Firefly

Navid200:
**Why we need this**
It's confusing if the user of a Firefly G6 sees an offer to hard reset transmitter.
This PR removes the offer, if the transmitter is a Firefly, and replaces it with a log.
Over the period of 3 years, I have only seen may be 5 cases of this on facebook.  The last time was less than a month or so ago.  But, I'm not sure what xDrip version they were using.

This is one of the actions listed in this issue:  https://github.com/NightscoutFoundation/xDrip/issues/1736

**Is there a workaround?**
The user can ignore the message if they are knowledgeable that xDrip has not been updated after the G6 firmware change.

No

**Tests**
I am running this build on my main device now.  But, I don't know how to trigger the error that would execute the new conditional.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
@@ -610,7 +610,9 @@ public class Ob1G5StateMachine {
                                 UserError.Log.ueh(TAG, "Session Start failed info: " + JoH.dateTimeText(session_start.getSessionStart()) + " " + JoH.dateTimeText(session_start.getRequestedStart()) + " " + JoH.dateTimeText(session_start.getTransmitterTime()));
                                 if (session_start.isFubar()) {
                                     final long tk = DexTimeKeeper.getDexTime(getTransmitterID(), tsl());
-                                    if (tk > 0) {
+                                    if (tk > 0 & FirmwareCapability.isTransmitterRawIncapable(getTransmitterID())) { // Firefly, which cannot be hard reset
+                                        UserError.Log.e(TAG, "Unusual session start failure");
+                                    } else if (tk > 0) {
                                         DexResetHelper.offer("Unusual session start failure, is transmitter crashed? Try Hard Reset?");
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
```diff
@@ -610,7 +610,9 @@ public static boolean doGetData(Ob1G5CollectionService parent, RxBleConnection c
                                 UserError.Log.ueh(TAG, "Session Start failed info: " + JoH.dateTimeText(session_start.getSessionStart()) + " " + JoH.dateTimeText(session_start.getRequestedStart()) + " " + JoH.dateTimeText(session_start.getTransmitterTime()));
                                 if (session_start.isFubar()) {
                                     final long tk = DexTimeKeeper.getDexTime(getTransmitterID(), tsl());
-                                    if (tk > 0) {
+                                    if (tk > 0 & FirmwareCapability.isTransmitterRawIncapable(getTransmitterID())) { // Firefly, which cannot be hard reset
```
tolot27, line 613: I suggest changing this to: 
```java
                                    if (tk > 0) {
                                        if (FirmwareCapability.isTransmitterRawIncapable(getTransmitterID())) { // Firefly, which cannot be hard reset
                                            UserError.Log.e(TAG, "Unusual session start failure");
                                        else {
                                            DexResetHelper.offer("Unusual session start failure, is transmitter crashed? Try Hard Reset?");
                                        }
                                    } else {
```

But this is just a minor suggestion improving code readability.

tolot27 (approved):
Thanks for this PR. This improves usability and avoids confusion.

app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
Navid200: Makes sense.  Thanks.
Done

jamorham (approved):

Issue #1299, opened 2020-04-13
xDrip follower sees compressed bg scale if on mmol/l while master on mg/dl


steve8x8:
![photo_2020-04-13 20 57 44](https://user-images.githubusercontent.com/11962538/79150667-74d16580-7dc9-11ea-863a-ce161c20cc4d.jpeg)

This happens if the master records mg/dl data (and treatments are "scaled" into the range), and a follower runs mmol/l (for whichever reason). Apparently treatment y values aren't scaled.
Since there are entries with a bg value in the 100-150 range, the real bg range is shown, with correct values, in a compressed (by a factor 18) manner.

Navid200:
@steve8x8 What is the follower scheme?
Does it happen if you use Nightscout to follow?  Does it happen if you use xDrip sync to follow?

tolot27:
Also, which version of xDrip has the follower running?

I've tried to reproduce this with the current nightly but couldn't.

Navid200:
Please provide more info if you experience it again.
Please help us recreate the problem if you do and we can reopen the issue.

Navid200:
I can recreate the problem after seeing the report here:
https://github.com/Navid200/xDrip/discussions/58

It requires adding a note to the entered carb value before changing the unit on the same device.  No follower involved.

mikel0101:
I have exactly the same bug.
Latest stable release (Nov 16th i believe) working on a Pixel 5.
I use mg/dl and I've never changed it.
No idea how I can fix this.

PR #1811, opened 2021-08-16
Don't save a vertical position for notes

Navid200:
**Why we need this**
This is not critical.  
Unfortunately, the person opening the issue has not provided detail.  I cannot recreate it even though I have tried:
https://github.com/NightscoutFoundation/xDrip/issues/1299
fixes https://github.com/NightscoutFoundation/xDrip/issues/1299

But, you can recreate another version of the problem:
Set your unit to mg/dL.  Use the treatment menu to enter some carbs.  Then, tap on it and add a note.  Then, change your unit to mmol/L.  And, you will see this:
Screenshot_20210815-213911

I don't know why a (vertical ) position is given to the note.  I have just changed it to 0.  All needed is the horizontal position (time).  The vertical position should just be  the reading at that same time, which is restored at the time of creating the graph.    

**Is there a work around**
Assuming this is experienced with a follower, the workaround is for the follower to always use the same unit as the master.  To avoid the issue described here (second version), the user should not keep changing their unit back and forth.

I don't know!
I don't know why a position is given to the note.  If there is an answer for that, that could explain what the side-effect may be.  

**Tests**
Tested with an Android 8


Navid200:
I just tested this again.  It actually creates a new problem!
I set the unit to mg/dL and entered a value and note.  it changed the vertical axis range.  I suspect the reason is that now with this PR the position is 0, which is very low for mg/dL.

Therefore, I am closing this PR until we have a solution that solves the problem without creating a new one.

PR #1814, opened 2021-08-17
Fix "Never use them"

Navid200:
**Why we need this**
Having automatic calibration disabled and selecting "Never use them" on this form sends entered blood test value for calibration.
Screenshot_20210816-121504
There are individuals, facebook group, who have stopped entering their blood test results because of the unexpected behavior of xDrip.

**Is there a work-around?**
User can select "Automatic mode" and answer no.  That's a work-around for the user who needs to enter their blood test results.  However, it does not address the incorrect behavior of "Never use them" setting.

No

**Tests**
Tested with an Android 9 master collecting from G6.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -809,6 +809,10 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
                     alert.show();
                 }
                 // offer choice to enable auto-calibration mode if not already enabled on pratelimit
+            } else if (calibration_type.equals("never")) {
+                Log.d(TAG, "Creating bloodtest record");
+                BloodTest.createFromCal(glucosenumber, timeoffset, "Manual Entry");
+                GcmActivity.syncBloodTests();
             } else {
                 // if use for calibration == "no" then this is a "note_only" type, otherwise it isn't
                 calintent.putExtra("note_only", calibration_type.equals("never") ? "true" : "false");


app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -809,6 +809,10 @@ private void processCalibrationNoUI(final double glucosenumber, final double tim
                     alert.show();
                 }
                 // offer choice to enable auto-calibration mode if not already enabled on pratelimit
+            } else if (calibration_type.equals("never")) {
```
tolot27, line 812: Now, the code of the "never" branch is identical with the first three lines of the "auto" branch (except the log message). Is this intentional?

Code duplication can be avoided with:

```java
            } else if (calibration_type.equals("auto") || calibration_type.equals("never")) {
                Log.d(TAG, "Creating bloodtest record from cal input data");
                BloodTest.createFromCal(glucosenumber, timeoffset, "Manual Entry");
                GcmActivity.syncBloodTests();
                if ((calibration_type.equals("auto") && (!Pref.getBooleanDefaultFalse("bluetooth_meter_for_calibrations_auto"))
                        && (DexCollectionType.getDexCollectionType() != DexCollectionType.Follower)
                        && (JoH.pratelimit("ask_about_auto_calibration", 86400 * 30))) {
                    ...
                }
            } else {
```

Navid200: Yes, it's intentional.

Functionally, most of this makes no sense to me.  I don't even like automatic calibration.  
If someone wants to always use their entered test results for calibration, they can use automatic calibration.  So, I don't know why this whole thing has been organized this way.

When I use always, it does not add the reading on the main screen.  Why?  

Choosing auto, will ask you if you want to enable automatic calibration.  It's like we are trying to convince people to use automatic calibration.  Why?

If I answer no to enabling automatic calibration, after having auto enabled, I know that it adds the reading to the main screen.  That's what many people like to be able to do.  They can also tap on it and use it as calibration if they want.
Because I knew that did what was missing as an independent option, I created a new fork that just did that.
 
If this is merged, I will open a second step to remove the redundant (always use) fork.
If that is also merged, I would like to also remove the auto fork leaving only ask me always and never use.

Navid200: Thinking more about this, I don't think there should be any conditionals here at all.
Either the user enables automatic calibration or not.
If they haven't any readings through Bluetooth or manual should just be entered in the database and shown on screen as a BG test reading.

But, let's keep this as a small step.  In the process, someone may explain why this was done this way and clear up my confusion.

tolot27:
This PR needs some more discussion. Please convert it into draft.

Navid200:
No, this PR is a step forward as is.  It will allow the users to be able to enter their readings, as they did in the past, without it being used for calibration.  
It does that.  I have no doubt as I have tested it.  I don't believe this is a draft.

jamorham:
As far as I can tell this isn't correct. I haven't had time to test this, but as far as I know the entering of a `BloodTest` to the `NativeCalibrationPipe` is gated by the preference boolean `bluetooth_meter_for_calibrations_auto` - Where is the data path / code flow where it gets sent as a native calibration other than this?

jamorham:
Oh I think I see it now, so I'm thinking the `NativeCalibrationPipe` needs to be gated by the `note_only` flag in `AddCalibration` `automatedCalibration` but that flow change would need to be confirmed that it wont impact anything else.

jamorham:
Reviewing the code path again I think the PR as it is may be the best solution because even with `note_only` there will still be a calibration created regardless of what happens with the `NativeCalibrationPipe` - The question remains though what the effect of excluding the else clause which uses the "never" option to determine whether to set `note_only` or not.

Navid200:
@jamorham I would love to just remove the whole conditional:

Either the user wants to use automatic calibration or not.  There is already a place to make that choice by enabling or disabling automatic calibration.
If the user has disabled automatic calibration (or has not enabled it), there is no reason to ask them if they want to enable it, there is no reason to ask them if they always want to submit their blood tests entered to be used for calibration, and there is no reason for them to say they never want their entered values to be submitted for calibration.  All of those are already answered by either enabling or disabling automatic calibration.

But, I didn't want to do that here and create a PR that makes too much change.

If you approve what I have explained, I am happy to do it.

Navid200:
If a user needs to just add a note, not a blood test result, not insulin injection, and not carb intake, they can tap on a blood glucose reading, at the time they like the note to appear, and add a note to it.

jamorham:
There are two different ways that a calibration can be used and when we receive a blood test. 

With raw data there is code to look at whether the trend is flat enough to use the blood test data received previously at a point in the future when we have the lagged cgm data. With native calibrations we don't need to do this and can (in theory) use every one and let the native algorithm decide what to do with it.

Likewise blood test data can come in at a point other than the test time, for example when synced from a glucose meter.

Navid200:
The automatic calibration scheme is a genius concept in non-native mode for G5.
For Dexcom G6, native, I have always strongly advised people to disable it.
The fact that the user can enable or disable automatic calibration and the fact that the default setting is disabled make me happy.
I am not advocating removing automatic calibration here.

Navid200:
@jamorham 
The user has 4 options on the first line and 2 options on the second line:
Screenshot_20210817-205931

Since those two lines are independent, that results in a total of 2 x 4= 8 different combinations (8 different possibilities).
If a user asks me to describe the 8 different conditions, and explain when one would use each, I will not be able to answer.

These are the 4 possibilities of the first line:
129652260-79fc4542-abb8-4971-bac8-866c1fc243e6

Isn't "Never use them" identical to disabling automatic calibration?
Isn't "Always use them" identical to enabling automatic calibration?
So, I don't think the two lines are really independent.
What's the point of "Ask me every time"?  If the user has a Bluetooth meter linked, is xDrip going to ask them every time a Bluetooth reading arrives?  Wouldn't that drive them mad?

I would just eliminate the form with 4 possibilities.  
If automatic calibration is enabled, every blood test result, Bluetooth or manual, will be used for calibration if the flat trend condition is satisfied for raw data and always if there is no raw data.
If automatic calibration is disabled, every blood test result, Bluetooth or manual, will be saved as a blood test (square red point on screen) but not used for calibration.

Navid200:
@jamorham The "Never use them" option uses them!
This PR fixes that and makes xDrip behave as expected when that option is selected. 

Is there anything I can do to make this move forward?

jamorham (approved):

PR #1827, opened 2021-09-15
Removed broken sound file selection option

Navid200:
**Why we need this**
If you tap on "Choose File" on the edit alert page in order to change the sound file for one of your main glucose level alerts, you will see this:
Screenshot_20210915-140641

If you then tap on custom sound and navigate to an mp3 file, you will see that it is grayed out.  So, you cannot choose a sound file.
The solution is to add the custom sound file to your Notifications folder (as explained here: https://navid200.github.io/xDrip/docs/Custom-sound-grayed-out.html) and use the "System Sound/Alarm" option.

The custom sound option on the form is not functional and confusing to the users.  There was a question posted on facebook and made me realize the problem.
There is an identical problem if you go to xDrip reminders and attempt to change the sound file for one of the reminders.

This PR removes the "Custom sound/alarm" option and makes tapping on the choose file behave as shown below.  
Screenshot_20211117-225530

The same for reminders:
Screenshot_20211117-230119

**Is there a workaround?**
The only workaround is to avoid using the custom sound option.  

No

**Tests**
Tested with Android 8 and virtual Android 11


tolot27 (commented):
If we remove this option, we should remove the unnecessary code as well (`AlertPlayer.java`, `arrays.xml`, etc.). Otherwise, we should fix the AlertPlayer:https://github.com/NightscoutFoundation/xDrip/blob/5c0debba376648ef2a94ef94363cc94b430d5994/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java#L549

Navid200:
@tolot27 I don't follow.
https://github.com/NightscoutFoundation/xDrip/blob/49817a5e788316ad5efaa23fc0a79e3c8fb98e33/app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java#L594-L611  

In the code I am showing in this post, I am making the case (which == 0) to be the default.  I have eliminated which == 1 and what comes after the last else.

The Alert player is still being used.  So, we don't need to edit alert player.

I agree with you about the array.  
But, I wonder if I should leave the default xDrip sound alone, in which case, the array will remain and only needs to be modified instead of deleted.

Should I leave the Default xDrip Sound alone?

Navid200:
**Why we need this**
If you tap on "Choose File" on the edit alert page in order to change the sound file for one of your main glucose level alerts, you will see this:
Screenshot_20210915-140641

If you then tap on custom sound and navigate to an mp3 file, you will see that it is grayed out.  So, you cannot choose a sound file.
The solution is to add the custom sound file to your Notifications folder (as explained here: https://github.com/Navid200/xDrip/wiki/Custom-Alert-Sound) and use the "System Sound/Alarm" option.

The custom sound option on the form is not functional and confusing to the users.  There was a question posted on facebook and made me realize the problem.

This PR removes the "Custom sound/alarm" option and makes tapping on the choose file behave as shown below.  
Screenshot_20211005-191507


**Is there a workaround?**
The only workaround is to avoid using the custom sound option.  

No

**Tests**
Tested with Android 8.

app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
```diff
@@ -589,7 +589,7 @@ public void onClick(View v) {
         buttonalertMp3.setOnClickListener(new View.OnClickListener() {
             public void onClick(View v) {
                 AlertDialog.Builder builder = new AlertDialog.Builder(mContext);
-                builder.setTitle("What type of Alert?")
+                builder.setTitle("")
```
tolot27, line 592: What is the idea of removing the title?
I suggest keeping the title but creating a translatable resource string of it. Default English title could be "Select type of sound" or "Select sound source" or something similar.

tolot27 (changes requested):

app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
Navid200: Thanks for the review.
I have followed your advice.
In the process discovered what "Reminders" were.  Made the same change to reminders.

Navid200:
As I kept modifying alertType values, I noticed that virtual test machines would reflect the changes but real phones wouldn't!  The real phones kept showing the original values no longer in the code.  I don't know why this happened.

I changed the name "alertType" itself to "alertSound".  That solved the problem.
But now, I'm not sure if there are any associations that I might have messed up by changing the name of this array element.

tzachi-dar:
I'm not sure what this PR is about.
This seems to remove a functionality that is working.

https://user-images.githubusercontent.com/7942372/152702709-27b48b56-4d91-4e22-a97a-7b712e87700d.mp4

In any case, if you think that this does not work, here is (for example) a complete solution that should work.
https://stackoverflow.com/questions/3592717/choose-file-dialog

It should allow to keep the functionality.

Navid200:
@tzachi-dar It's about a problem that has been reported in the facebook group over and over again and I can see it on my phones as well.
Using the custom option and finding an mp3 file show it grayed out.
 
I thought all phones had the problem.  From now on, I will ask anyone else who posts the question, to post in the discussions so that you can see.

tzachi-dar:
Do they have some kind of file manger that is installed?

Navid200:
I don't have any file manager installed.  
I have older phones on which, I have uninstalled everything other than xDrip and I still see this behavior.

tzachi-dar:
Can you try installing it from here and see if it will solve your problem
https://play.google.com/store/apps/details?id=com.alphainventor.filemanager

Navid200:
@tzachi-dar I have recreated a problem in virtual device.  I have shown all the steps in this discussion:  https://github.com/NightscoutFoundation/xDrip/discussions/1983
Would you please have a look?
Thanks for your time.

Navid200:
I don't know how to resolve conflicts through GitHub.
I'm so sorry the only way I know is to close the PR and open a new one.

Issue #1740, opened 2021-05-25
Please move infrequently used settings to engineering mode


Navid200:
**Problem**
When a beginner runs into trouble using xDrip, they can easily change a lot of settings they should not touch.

There are documents that describe the recommended settings.  Yet, everyday, someone posts, in facebook, asking how they can get xDrip going again.
<br/>

**Suggested Solution**
It would be better if only settings were available to the user, out of engineering mode, that would be related to user preferences, not settings that are supposed to be always set or always disabled.

Is any of the following settings needed to be different than shown for your device or for any reason other than testing?  Can we move them to engineering mode? 
<br/>
 
**Bluetooth Settings**
Turn Bluetooth on    Enabled
Bluetooth Watchdog    Enabled
Use Background Scans   Enabled
Constantly reset Bluetooth   Disabled
Transmitter (PL) support  Disabled
RFduino support   Disabled
xBridge+ Polling Mode    Disabled

**Other misc. options**
Run Collector in foreground   Enabled
Battery Optimization prompt   Disabled
Samsung Workarounds   Enabled

**G5/G6 Debug Settings**
Use OB1 Collector   Enabled
Fallback to xDrip   Disabled
Allow OB1 initiate bonding   Enabled
Authenticate G5 before each read   Enabled
Unbod G5 before each read   Disabled

tolot27:
Beside showing extended settings only in engineering mode a lot of settings should not be visible if the data source/sync agent is not active. This will reduce the number of visible settings significantly. I'm unsure whether we should treat this as a separate issue. At least the PRs should address this independently.

jamorham:
If we put too much behind engineering mode then it will end up getting enabled too frequently. Some of the options above yes engineering mode is appropriate as we really don't expect users to need to modify the defaults, but for some others we could have an `advanced mode` and add that as a dependency to the setting and then iterate the settings and automatically hide those when that mode is not enabled.

Navid200:
@jamorham I hope I understand correctly:
I will create a new setting deep under other misc options.  I will call it something like "Advanced" for English.
Then, I will make the settings I have identified here to be accessible only if the new advanced setting is enabled.   If the new advanced setting is disabled, the identified settings automatically go back to their default settings.

jamorham:
[...]

PR #1829, opened 2021-09-18
Gray out unnecessary and apply recommended settings

Navid200:
**Why we need this**
To make xDrip less confusing for newcomers.
See details here: https://github.com/NightscoutFoundation/xDrip/issues/1740
Fixes https://github.com/NightscoutFoundation/xDrip/issues/1740

The new setting (Experiment) is shown below.
Screenshot_20210918-003537
It is disabled by default, which forces the recommended settings.

**Is there a workaround?**
No, there are too many settings available for anyone to change hoping for a solution.  But, changing the numerous settings usually doesn't solve the problem for inexperienced people and they ask for help in the facebook group.

The idea is that none of these recommended settings should cause a problem for anyone.  Therefore, there should be no side effect.  But, if you notify me before merging, I will announce on facbook and suggest that they save their settings.  

**Tests**
Tested with a Firefly G6 on Android 8 and Android 11.

**Notes**
Please have a look at the list of settings (Preferences.java) and let me know if:
1- we should add any other setting if you believe there are any other settings that we can recommend to everyone.
2- we should remove any setting if you believe there is a setting that could cause a problem for a particular device or Android level.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
@@ -434,6 +434,28 @@ public class Preferences extends BasePreferenceActivity implements SearchPrefere
                     //
                 }
             }
+            if (!prefs.getBoolean("enable_experiment", false)) { // If Experiment is disabled
+                try { // Apply recommended settings
+                    prefs.edit().putBoolean("automatically_turn_bluetooth_on", true).apply(); // Enable auto Bluetooth
+                    prefs.edit().putBoolean("bluetooth_watchdog", true).apply(); // Enable Bluetooth watchdog
+                    prefs.edit().putBoolean("bluetooth_allow_background_scans", true).apply(); // Enable background scan
+                    prefs.edit().putBoolean("bluetooth_frequent_reset", false).apply(); // Disable constant Bluetooth reset
+                    prefs.edit().putBoolean("use_transmiter_pl_bluetooth", false).apply(); // Disable TX PL support
+                    prefs.edit().putBoolean("use_rfduino_bluetooth", false).apply(); // Disable RF Duino
+                    prefs.edit().putBoolean("pref_dex_collection_polling", false).apply(); // Disable xBridge polling mode
+                    prefs.edit().putBoolean("run_service_in_foreground", true).apply(); // Enable running collector in foreground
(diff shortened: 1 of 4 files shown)


jamorham:
I'm not sure what the best way to handle this or if we can enforce a one-size-fits-all to all of the settings you've identified and disallowing changes to them.  Would this be better included as part of the wizard setup so that users can go back through the wizard to reset settings to defaults?

Navid200:
@jamorham The only way we can go for a one size fits all approach is if we can be sure that there will be absolutely no device that needs to modify any of these settings in order to function properly.

Is that your feeling about these settings?

Navid200:
@jamorham xDrip has got an image, at least, in the facebook group, that it is difficult to set up.  
Many advise people who ask for help to abandon xDrip and switch to the Dexcom app.

Some of the frustration can be avoided by removing the settings that almost no one touches.
This PR grays out those settings.  When a new user installs xDrip, those settings will be grayed out.
Two exceptions are enabling OB1 and fallback to xDrip.  Those will not be grayed out.  But, they will be reset after a restart.  The rest will be grayed out.

The settings can be categorized in different groups.  
There are settings that should be hard coded and removed altogether.  As an example, I don't know why anyone would need to disable the OB1 collector.  
So, I am looking at this as a first step.
If you can identify any of these settings that you think no one needs to touch, we can hard code them and remove the option completely.  
This would make the settings look less intimidating to a new user also.

tolot27:
Why not hide these settings if not in engineering mode? Graying them out does not reduce the number of visible settings.

Navid200:
We decided not to use the engineering mode and create a new mode, Experiment, instead.  The thinking is that ideally people will leave both disabled.  But, if they were going to leave one enabled, we prefer if they leave experiment enabled rather than engineering.  So, this PR leave engineering mode alone.

Even though the settings remain visible, graying them out helps psychologically.  If I am in panic and am doubting which settings I changed that may have caused my current problem, I may change any setting I can change because I wonder if changing any of them accidentally could have caused the problem I am experiencing.
But, if some of the settings are grayed out, I know that I could not have changed those.  So, I will leave those alone.

I searched, Google, for removing (hiding) settings if Experiment was off.  All I found was the need to recreate the settings after Experiment is reenabled.  It sounded really complicated to me.  
If you can give me reading on how to do it in a simple manner, I would love to hide the settings when experiment is disabled.

Navid200:
This is a work-around:  https://github.com/NightscoutFoundation/xDrip/pull/2196

PR #1568, opened 2020-12-22
add foregroundServiceType to services using Bluetooth

tolot27:
With Android 10+ some privacy features have been [implemented](https://developer.android.com/about/versions/10/privacy/changes), especially, background services accessing Bluetooth devices also need [location permissions](https://developer.android.com/about/versions/10/privacy/changes#app-access-device-location).

If this permission is not granted, exceptions are thrown as reported with #1566 and as I discovered with InPen authorization. Because we are using compileSdkVersion 29 (even if our targetSdkVersion is 23) we have to the foregroundServiceType.

This PR adds this to all services using Bluetooth and this should fix #1566.

PR #1832, opened 2021-09-23
declares Foreground service types for FollowServices

### LINKED ISSUE #1831: Android 10 NightScoutFollowService exception
Hi,

I am running xDrip+ (stable version) and Android APS on Jelly 2 with Android 10.

My set up if following.

I have NightScout receiving BG from Medtronic 640 uploader.

xDrip reads BG from NightScout and should send it locally to Android APS where xDrip is set as BG source.

Problem is that sending BG from xDrip to APS is not working or works once and then not.

In xDrip event logs I see following exception

"Got exception trying to use Android 10+ service starting for NightscoutFollowService java.lang.IllegalArgumentException: foregroundServiceType 0x0000....08 is not subset of foregroundServiceType attribute 0x00000000 in service element of manifest file"

Please help :)

tolot27:
Android 10 and newer require the declaration of the foreground service type. This is added for NightscoutFollowService and ShareFollowService as FOREGROUND_SERVICE_TYPE_DATA_SYNC.

Fixes #1831

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -710,11 +710,15 @@
         <service android:name=".watch.miband.MiBandService"
             android:foregroundServiceType="connectedDevice|location"
             />
-        <service android:name=".cgm.nsfollow.NightscoutFollowService" />
+        <service android:name=".cgm.nsfollow.NightscoutFollowService"
+            android:foregroundServiceType="dataSync"
+            />
         <service android:name=".insulin.inpen.InPenService"
             android:foregroundServiceType="connectedDevice|location"
             />
-        <service android:name=".cgm.sharefollow.ShareFollowService" />
+        <service android:name=".cgm.sharefollow.ShareFollowService"
+            android:foregroundServiceType="dataSync"
+            />
 
         <activity
             android:name=".EventLogActivity"
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/ForegroundServiceStarter.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/ForegroundServiceStarter.java
@@ -9,7 +9,7 @@ import com.eveningoutpost.dexdrip.Models.JoH;
 import com.eveningoutpost.dexdrip.Models.UserError;
 import com.eveningoutpost.dexdrip.Models.UserError.Log;
 
-import static android.content.pm.ServiceInfo.FOREGROUND_SERVICE_TYPE_LOCATION;
+import static android.content.pm.ServiceInfo.FOREGROUND_SERVICE_TYPE_MANIFEST;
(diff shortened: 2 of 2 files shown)


tolot27:
Here is an unsigned debug build based on [Nightly build 2nd Sep 2021](https://github.com/NightscoutFoundation/xDrip/releases/tag/2021.09.02) which requires reinstalling xDrip from untrusted source. Backup of DB and settings are highly recommend beforehand, otherwise you will loose your data.

https://www.dropbox.com/s/5vjmjj0nlq4ry7q/xDrip_20210923_727aeaffc_DataForegroundServiceTypes_debug.apk?dl=0

jrsokolow:
I have tested and bug is still there

tolot27:
Hmm, cf5500f introduced a fixed foregroundServiceType (`FOREGROUND_SERVICE_TYPE_LOCATION`) for any ForegroundService, regardless of the settings in the manifest:

https://github.com/NightscoutFoundation/xDrip/blob/38fee03be56da6ce9cbe90e08dbe8e4b651c6edc/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/ForegroundServiceStarter.java#L68

Hence, the exception shown in #1831 changed from "Got exception trying to use Android 10+ service starting for NightscoutFollowService java.lang.IllegalArgumentException: foregroundServiceType 0x00000008 is not subset of foregroundServiceType attribute 0x00000000 in service element of manifest file" to 
"Got exception trying to use Android 10+ service starting for NightscoutFollowService java.lang.IllegalArgumentException: foregroundServiceType 0x00000008 is not subset of foregroundServiceType attribute 0x00000001 in service element of manifest file"

0x00000001 = FOREGROUND_SERVICE_TYPE_DATA_SYNC
0x00000008 = FOREGROUND_SERVICE_TYPE_LOCATION

According to the [docs](https://developer.android.com/guide/components/foreground-services) `Service.startForeground(notification, foregroundServiceType)` is used to limit the access of the foreground service only to a subset of the types declared in the manifest, at runtime. IMHO, we do not need this additional limitation as long as the types are declared in the manifest. But this was not the case for `NightscoutFollowService` and `ShareFollowService` and also not for many other services after cf5500f was introduced. For the other services it was fixed with #1568.
IMHO, https://github.com/NightscoutFoundation/xDrip/commit/cf5500fcc53b55680d09ee22f83c853422c4a6c7#diff-14bc6a55e69088bd392ea8dfb3dd7bab1da28826850f60f395427ffc7cac66d3 is completely unnecessary and just a workaround for providing the foregroundServiceType programmatically at runtime instead of declaring it in the manifest.

tolot27:
The new apk is available at https://www.dropbox.com/s/gsmxydjq8jqoocm/xDrip_20210923_7282febf4_DataForegroundServiceTypes_debug.apk?dl=0.

jamorham (approved):
Looks like it makes sense

Issue #1658, opened 2021-02-25
Don't delay ascending volume alerts by 3 minutes


Navid200:
There have already been complaints about not being able to control the alarm volume:
https://github.com/NightscoutFoundation/xDrip/issues/813

In response to "please add a low profile as well", the suggested solution has been to use the ascending volume profile.
But, that profile starts the volume at "silent".  So, if you want to get an alarm when your glucose crosses a threshold, you will have to add the time it takes for the volume to rise from silent to the next level to know when you will really hear the alarm.  But, by then, your glucose may be at a different level.  

Would you please change the profile so that it would start from very low volume instead of starting from silent?

samhedgerows:
Personally I like that the ascending volume starts silently - with the "vibrate on alerts" option enabled it starts with vibration only before starting to make a noise. I find the vibration loud enough most of the time, but if I miss it it then starts to make a noise. So I'd definitely not want this functionality to completely disappear.

However with vibrations off the alarm doesn't make any noise for a few minutes - which isn't great as you say. A temporary work-around would be to enable vibrations, if possible.

Navid200:
One consequence of the alarm starting silent is that is you tap on test alarm, it makes no sound!  Many think there is an error and stop giving the ascending volume a chance.

Are you saying if you have the vibration as you like to set it and also have the minimum volume, instead of silent, it will be too loud for you?  The vibration is actually much louder than minimum volume, isn't it?

I'm not asking for the ability to enable vibration to be removed.  Only for the sound volume for the selected sound file to start from an extremely low volume instead of completely silent.  Would that be a problem?

Navid200:
@tzachi-dar @jamorham  @AdrianLxM 

Looking at the AlertPlayer.java file (from the March 8, 2021 release), I see this:
![1 (2)](https://user-images.githubusercontent.com/51497406/111714975-3b9b4400-8829-11eb-9043-8d9adcbfbb34.png)

The _if_ statement on line 530 keeps the alert (for ascending profile only) silent for 3 minutes.  Only the vibrate option will function in that period.  But, if the user does not choose to enable the vibrate option, the 3-minute delay remains.   Then, the user receives no alert in any form for 3 minutes.
However, if they choose a medium or high volume profile, they will not experience that delay.
Is there a logic behind this discrepancy?

I am happy to create a pull request to make the medium, high and ascending volume profiles neither to have a delay.
But, would you please tell me if there is a reason for this and therefore, such a pull request will never be accepted?

PR #1834, opened 2021-09-23
No delay for the ascending volume profile

Navid200:
**Why do we need this?**
Currently, the alert sound for the ascending volume profile is delayed by 3 minutes.
This PR adds a setting (Delay Ascending Volume) allowing the user to change the behavior (by disabling it) and remove the delay.  
If you are in a meeting, everything is perfect as is and you can leave the new setting enabled.  Then, with the phone in your pocket, only you will feel the vibration without disrupting the meeting.
But, if you are going to bed and don't want to wake up anyone else, you can disable the new setting.  

I see ascending volume profile useful for two completely different scenarios and the new setting allows the user to choose which scenario is more important for them. 

The following image shows the new setting at the bottom of the "Glucose Alerts Settings" page.
Screenshot_20211129-090204

The setting is enabled by default.  By default, the behavior is identical to what we have now.  To change the behavior, the user must disable the setting as shown in the above image.  

The following table shows how the volume is adjusted now or after this PR with the new setting enabled (default), and after this PR with the new setting disabled, versus time.

| Time [minutes] | Volume now, or after <br/> this PR with the <br/> new setting enabled | Volume after this PR <br/> with the new setting <br/> disabled | Vibrate on alert enabled <br/> with or without this PR | 
|------------------|---------------|------------------------|----------------------------|
| 0 (Alert trigger) | 0 | 30% | Vibrate |
| 1 | 0 | 30% unless snoozed | Vibrate unless snoozed |
| 2 | 0 | 30% unless snoozed | Vibrate unless snoozed |
| 3 | 33% unless snoozed | 33% unless snoozed | Vibrate unless snoozed |
| 4 | 67% unless snoozed | 67% unless snoozed | Vibrate unless snoozed |
| 5 | 100% unless snoozed | 100% unless snoozed | Vibrate unless snoozed |

The time points of 0, 1, 2, 3 and 4 minutes represent one read cycle.  So, 5 is a part of the next read cycle.
I am assuming that an alert is triggered on the reading performed at time 0.  At time 5 minutes, we will have the next reading.  It is always possible that the next reading, at time 5 minutes, may trigger another alert.
The ascending volume does not reach its maximum in one read cycle.  And, I am not changing that.
The purpose of the table is to compare what we have now with what we will have after.

Fixes https://github.com/NightscoutFoundation/xDrip/issues/1658

**Is there a workaround?**
If you enable vibration for the alert, the vibration triggers at time 0 and will not be delayed.  But, if you disable the vibration, there is no workaround.  
Having the vibration enabled defeats the purpose of the ascending volume in some circumstances.  The point is not to wake up someone next to you.  A phone vibrating on a nightstand, rattling everything else on the table, can be very loud.  

 

No

**Is xDrip behavior changing under default settings?**
No

**Tests**
Tested with a real Android 11 collector from G6 as well as with the fake data source on an Android 8 phone.  
Tested Ascending volume profile under both settings, Medium volume profile and High volume profile.  

**Anything you don't like about this PR**
I would have liked to make the new setting available only when the ascending volume profile is selected.  But, that is not selected with a checkbox.  It is selected using an array.  So, it would have required changing more than a few files to implement.  I chose to keep this PR as simple as possible.  This PR adds a new setting that is available regardless of what volume profile is selected.  But, it only affects the behavior when ascending volume profile is selected.  

**Changed lines**  
There are some lines that are flagged as changed only because the initial tab size has changed due to formatting.  The following shows the lines that have really changed or are new in AlertPlayer.java.  
261
537 The conditional on this line is gone.  It's as if the condition is always true.
539 The formula on line 539 is now on line 537 with no change
Since the original conditional on line 537 is gone, the formula on this line could result in a negative value now.
The new limit on line 539 sets the result to 0 if it is negative.

540 is what it was before just with an added comment.

Current lines 541, 542 and 543 are new.

Everything that is on and follows the old line 541 is unchanged.
But, because the conditional on line 537 is removed, everything following it (up to line 555) has moved up on the conditional branches and is flagged as modified.


tolot27:
Looks good so far. Even if I have "Vibrate on alert" disabled, it vibrates, if I test the alert.

Navid200:
I don't think it vibrates if you disable vibrate on alert.
But, that's not something I have changed.  That aspect is identical to before.
This is my understanding of what happens unless you have set up your phone (not xDrip) to vibrate for any notification:
If you enable vibrate on alert, it will vibrate every minute until you snooze starting from time 0.  This happens now and after.
If you disable vibrate on alert, it will never vibrate and only play the MP3 file with the volume shown in the table above.

tzachi-dar:
The main idea in asceding was to have it vibrate for two times before it starts making noise.
This is needed if you are in a meeting (for example)

Navid200:
@tzachi-dar Thanks,
That's what I wanted to know.  Awesome!

I will add a setting.  The setting will allow the original intended behavior as you describe as a default.  
Using the setting to activate the new mode, will allow what I prefer as at night the phone will be on a table and rattling caused by the vibration will effectively be louder than the low volume alert.

So, the user will have the option to choose one way or the other.

Navid200:
**Why do we need this?**
Currently, the alerts using this profile are delayed by 3 minutes.
This PR will add a setting (Delay Ascending Volume) allowing the user to change the behavior and remove the delay.

The following image shows the new setting at the bottom of the "Glucose Alerts Settings" page.
Screenshot_20211001-232514  

The setting is enabled by default.  By default, the behavior is identical to what we have now.  To change the behavior, the user must disable the setting.
The following table shows how the volume is adjusted now or with the setting enabled and after this PR with the new setting disabled versus time.

| Time [minutes] | Volume now | Volume after this PR <br/> with the new setting <br/> disabled | Vibrate on alert enabled <br/> with or without this PR | 
|------------------|---------------|------------------------|----------------------------|
| 0 (Alert trigger) | 0 | 17% | Vibrate |
| 1 | 0 | 33% unless snoozed | Vibrate unless snoozed |
| 2 | 0 | 50% unless snoozed | Vibrate unless snoozed |
| 3 | 33% unless snoozed | 67% unless snoozed | Vibrate unless snoozed |
| 4 | 67% unless snoozed | 83% unless snoozed | Vibrate unless snoozed |
| 5 | 100% unless snoozed | 100% unless snoozed | Vibrate unless snoozed |

The time points of 0, 1, 2, 3 and 4 minutes represent one read cycle.  So, 5 is a part of the next read cycle.
So, I am assuming that an alert is triggered on the reading performed at time 0.  At time 5 minutes, we will have the next reading.  So, it is always possible that the next reading, at time 5 minutes, may trigger another alert.
So, the ascending volume does not reach its maximum in one read cycle.  And, I am not changing that.
The purpose of the table is to compare what we have now with what we will have after.

Fixes https://github.com/NightscoutFoundation/xDrip/issues/1658

**Is there a workaround?**
If you enable vibration for the alert, the vibration triggers at time 0 and will not be delayed.  But, if you disable the vibration, there is no workaround.  
Having the vibration enabled defeats the purpose of the ascending volume.  The point is not to wake up someone next to you.  A phone vibrating on a nightstand, rattling everything else on the table, can be very loud.

No

**Tests**
Tested with a real Android 11 collector from G6 as well as with the fake data source on an Android 8 phone.

**Anything you don't like about this PR**
I would have liked to make the new setting available only when the ascending volume profile is selected.  But, that is not selected with a checkbox.  It is selected using an array.  So, it would have required changing more than a few files to implement.  I chose to keep this PR as simple as possible.  This PR adds a new setting that is available regardless of what volume profile is selected.  But, it only affects the behavior when ascending volume profile is selected.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
```diff
@@ -534,9 +535,10 @@ private void VibrateNotifyMakeNoise(Context context, AlertType alert, String bgV
                 .setDeleteIntent(snoozeIntent(context, minsFromStartPlaying));
 
         if (profile != ALERT_PROFILE_VIBRATE_ONLY && profile != ALERT_PROFILE_SILENT) {
-            if (minsFromStartPlaying >= MAX_VIBRATING_MINUTES) {
-                // Before this, we only vibrate...
-                float volumeFrac = (float) (minsFromStartPlaying - MAX_VIBRATING_MINUTES) / (MAX_ASCENDING_MINUTES - MAX_VIBRATING_MINUTES);
+            if (minsFromStartPlaying >= MAX_VIBRATING_MINUTES || (minsFromStartPlaying > Vibrating_Calc && !Pref.getBoolean("delay_ascending_3min", true))) {
```
tolot27, line 538: This could be rewritten to
`if (minsFromStartPlaying >= MAX_VIBRATING_MINUTES || (Vibrating_Calc > 0 && minsFromStartPlaying > Vibrating_Calc)) {`

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
```diff
@@ -115,6 +115,7 @@ boolean isUiThread() {
     // when ascending how many minutes since alert started do we wait before escalating
     final static int MAX_VIBRATING_MINUTES = 2;
     final static int MAX_ASCENDING_MINUTES = 5;
+    public int Vibrating_Calc = Pref.getBoolean("delay_ascending_3min", true)? MAX_VIBRATING_MINUTES : -1; // Set to 2 if enabled and -1 if disabled
```
tolot27, line 118: Please apply code formatting at least to your inserts/changes.

tolot27 (changes requested):
Just some minor requests.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
Navid200: But, it will make the code harder to understand.  
Someone who is looking at this for the first time, will have to go look for Vibrating_Calc to see what its value is in order to find out that its value depends on a setting.  Whereas what we have now right away shows that we are checking a setting.
I see that if I make the change you have suggested, the line will become a little bit shorter.  But, do you think it is worth the price of making the code harder to understand?

tolot27: I don't think it make it harder to understand. IMHO, the opposite is the case and code duplication is avoided. One has to lookup `Vibrating_Calc` always to understand the code. From this time, it knows that it depends on a preference.

Avoiding code duplication does not only save space and improves readability. It also increases maintenance and reduces the likelihood of errors. 

@jwoglom What do you think about code readability in this context?

Navid200: Some day, someone may need to rewrite the formula on line 539.  And they may need to change the values of Vibrating_Calc.  Then, they will have one more restriction they will have to worry about if we make this suggested edit, which will be the logic in the suggested change.

But, the intended logic is extremely simple; either the user has enabled the setting or not.  It is true that right now, Vibrating_Calc only has two possible values and one of them is greater and the other is smaller than 0.  But, that function of Vaibrating_Calc does not constitute that condition will always be true.

The suggested change adds a restriction to the code, on the values of Vibrating_Calc.  
I chose to write the formula (line 539) the way I did.  There is no reason it cannot be written in a simpler way.  But, imposing more restrictions on it seems to be unnecessary.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
```diff
@@ -534,9 +535,10 @@ private void VibrateNotifyMakeNoise(Context context, AlertType alert, String bgV
                 .setDeleteIntent(snoozeIntent(context, minsFromStartPlaying));
 
         if (profile != ALERT_PROFILE_VIBRATE_ONLY && profile != ALERT_PROFILE_SILENT) {
-            if (minsFromStartPlaying >= MAX_VIBRATING_MINUTES) {
-                // Before this, we only vibrate...
-                float volumeFrac = (float) (minsFromStartPlaying - MAX_VIBRATING_MINUTES) / (MAX_ASCENDING_MINUTES - MAX_VIBRATING_MINUTES);
+            if (minsFromStartPlaying >= MAX_VIBRATING_MINUTES || (minsFromStartPlaying > Vibrating_Calc && !Pref.getBoolean("delay_ascending_3min", true))) {
+                float volumeFrac = (float) (minsFromStartPlaying - Vibrating_Calc) / (MAX_ASCENDING_MINUTES - Vibrating_Calc);
```
tolot27, line 539: This line is an elegant mathematical solution but much more dangerous because it is not easy to understand and it becomes invalid (as well as the comment) as soon as one changes one of the two underlying constants.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
```diff
@@ -115,6 +115,7 @@ boolean isUiThread() {
     // when ascending how many minutes since alert started do we wait before escalating
     final static int MAX_VIBRATING_MINUTES = 2;
     final static int MAX_ASCENDING_MINUTES = 5;
+    public int Vibrating_Calc = Pref.getBoolean("delay_ascending_3min", true) ? MAX_VIBRATING_MINUTES : -1; // Set to 2 if enabled and -1 if disabled
```
tolot27, line 118: This comment is also wrong because the variable is set to a predefined constant and not to 2.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
tolot27: I think there is an easy solution without introducing a new variable, just with an if-else clause like

```java
float volumeFrac = 0;
if (!Pref.getBoolean("delay_ascending_3min", true)) {
    volumeFrac = (float) (minsFromStartPlaying + 1)/6;
} else {
    if (minsFromStartPlaying >= MAX_VIBRATING_MINUTES) {
                // Before this, we only vibrate...
                volumeFrac = (float)(minsFromStartPlaying - MAX_VIBRATING_MINUTES) / (MAX_ASCENDING_MINUTES - MAX_VIBRATING_MINUTES);
   }
}
if (volumeFrac > 0) {
    ...
}
```

tolot27 (changes requested):

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
Navid200: I don't follow why you have made this comment.
The code as it stands, before this PR, exactly fits your description.

Do you want me to rewrite this code from scratch throwing away what tzachi wrote a decade ago?

I'm not going to do that.
If you want to change the code, you can.  My intent is to remove the delay.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
Navid200: Corrected it.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
Navid200: I like that.  I need time to test it.
Thanks a lot.

Navid200:
**Why do we need this?**
Currently, the alert sound for the ascending volume profile is delayed by 3 minutes.
This PR adds a setting (Delay Ascending Volume) allowing the user to change the behavior (by disabling it) and remove the delay.

The following image shows the new setting at the bottom of the "Glucose Alerts Settings" page.
Screenshot_20211001-232514  

The setting is enabled by default.  By default, the behavior is identical to what we have now.  To change the behavior, the user must disable the setting as shown in the above image.  

The following table shows how the volume is adjusted now or after this PR with the new setting enabled (default), and after this PR with the new setting disabled, versus time.

| Time [minutes] | Volume now, or after <br/> this PR with the <br/> new setting enabled | Volume after this PR <br/> with the new setting <br/> disabled | Vibrate on alert enabled <br/> with or without this PR | 
|------------------|---------------|------------------------|----------------------------|
| 0 (Alert trigger) | 0 | 30% | Vibrate |
| 1 | 0 | 30% unless snoozed | Vibrate unless snoozed |
| 2 | 0 | 30% unless snoozed | Vibrate unless snoozed |
| 3 | 33% unless snoozed | 33% unless snoozed | Vibrate unless snoozed |
| 4 | 67% unless snoozed | 67% unless snoozed | Vibrate unless snoozed |
| 5 | 100% unless snoozed | 100% unless snoozed | Vibrate unless snoozed |

The time points of 0, 1, 2, 3 and 4 minutes represent one read cycle.  So, 5 is a part of the next read cycle.
I am assuming that an alert is triggered on the reading performed at time 0.  At time 5 minutes, we will have the next reading.  It is always possible that the next reading, at time 5 minutes, may trigger another alert.
The ascending volume does not reach its maximum in one read cycle.  And, I am not changing that.
The purpose of the table is to compare what we have now with what we will have after.

Fixes https://github.com/NightscoutFoundation/xDrip/issues/1658

**Is there a workaround?**
If you enable vibration for the alert, the vibration triggers at time 0 and will not be delayed.  But, if you disable the vibration, there is no workaround.  
Having the vibration enabled defeats the purpose of the ascending volume in some circumstances.  The point is not to wake up someone next to you.  A phone vibrating on a nightstand, rattling everything else on the table, can be very loud.

No

**Tests**
Tested with a real Android 11 collector from G6 as well as with the fake data source on an Android 8 phone.

**Anything you don't like about this PR**
I would have liked to make the new setting available only when the ascending volume profile is selected.  But, that is not selected with a checkbox.  It is selected using an array.  So, it would have required changing more than a few files to implement.  I chose to keep this PR as simple as possible.  This PR adds a new setting that is available regardless of what volume profile is selected.  But, it only affects the behavior when ascending volume profile is selected.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
Navid200: I changed a lot.  There is no new formula any longer.
My phone would make no noise when the volume was lower than 30%.
So, I changed the behavior and updated the description.

tolot27:
Now, we have three times the PR description. Please merge it with the first one. It is always possible to compare a comment with any of its previous version, in case it is necessary. IMHO, it is much easier if the PR description is maintained at the top of the conversation.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
```diff
@@ -534,23 +534,25 @@ private void VibrateNotifyMakeNoise(Context context, AlertType alert, String bgV
...
-                if (notSilencedDueToCall()) {
-                    if (overrideSilent || isLoudPhone(context)) {
-                        playFile(context, alert.mp3_file, volumeFrac, forceSpeaker, overrideSilent);
-                    }
-                } else {
-                    Log.i(TAG, "Silenced Alert Noise due to ongoing call");
+            float volumeFrac = (float) (minsFromStartPlaying - MAX_VIBRATING_MINUTES) / (MAX_ASCENDING_MINUTES - MAX_VIBRATING_MINUTES);
+            // While minsFromStartPlaying <= MAX_VIBRATING_MINUTES, we only vibrate ...
+            volumeFrac = Math.max(volumeFrac, 0); // Limit volumeFrac to values greater than and equal to 0
+            volumeFrac = Math.min(volumeFrac, 1); // Limit volumeFrac to values less than and equal to 1
```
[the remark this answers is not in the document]
tolot27: I believe we should reevaluate the complete formula because currently, this line always returns 0 for minsFromStartPlaying ∈ [0..2] and always returns 1 for minsFromStartPlaying ∈ [0..2].

| `minsFromStartPlaying` | `volumeFrac` | `Math.min(Math.max(volumeFrac, 0), 1)` |
|:----------------------:|-------------:|---------------------------------------:|
|            0           |         -2.0 |                                    0.0 |
|            1           |         -1.0 |                                    0.0 |
|            2           |          0.0 |                                    0.0 |
|            3           |          1.0 |                                    1.0 |
|            4           |          2.0 |                                    1.0 |
|            5           |          3.0 |                                    1.0 |

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
```diff
@@ -534,23 +534,25 @@ private void VibrateNotifyMakeNoise(Context context, AlertType alert, String bgV
...
+            if (!Pref.getBoolean("delay_ascending_3min", true) && volumeFrac < 0.3) {
+                volumeFrac = (float) 0.3; // If delay_ascending_3min is disabled, we never only vibrate.
+            }
+            if (profile == ALERT_PROFILE_MEDIUM) {
```
tolot27, line 544: This condition check should be moved before line 537 and the remaining `volumeFrac` calculation in the else part, because line 545 is a constant assignment.

tolot27 (changes requested):
Many thanks for considering my previous review comments, @Navid200! You found an elegant solution which avoids the declaration of an additional variable it is very readable.

It can be further improved as commented and should be reevaluated because I believe there is no real ascending volume.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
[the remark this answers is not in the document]
Navid200: I have tested the PR and it works.  I have been using it since I made the last commit change.

I suggest you test my PR as is and if you find a problem, please let me know.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
```diff
@@ -534,23 +534,25 @@ private void VibrateNotifyMakeNoise(Context context, AlertType alert, String bgV
...
-                Log.d(TAG, "VibrateNotifyMakeNoise volumeFrac = " + volumeFrac);
-                boolean overrideSilent = alert.override_silent_mode;
-                boolean forceSpeaker = alert.force_speaker;
-                if (notSilencedDueToCall()) {
-                    if (overrideSilent || isLoudPhone(context)) {
-                        playFile(context, alert.mp3_file, volumeFrac, forceSpeaker, overrideSilent);
-                    }
-                } else {
-                    Log.i(TAG, "Silenced Alert Noise due to ongoing call");
+            float volumeFrac = (float) (minsFromStartPlaying - MAX_VIBRATING_MINUTES) / (MAX_ASCENDING_MINUTES - MAX_VIBRATING_MINUTES);
```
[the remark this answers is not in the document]
Navid200: Any change necessary for the original code (before a PR), should be submitted as a separate PR.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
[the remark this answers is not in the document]
Navid200: The volume behaves exactly as it did before except, with the new setting disabled, it starts from 30% instead 0f 0.  That's the only change.
So, it is incorrect to say that there is no ascending volume.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
Navid200: Please explain what you mean by this.

tolot27:
We should really have a unit test for this PR. Please can you create one? Otherwise, I will do it.

Navid200:
> We should really have a unit test for this PR. Please can you create one? Otherwise, I will do it.

I don't know how.  
I also wonder what is special about this PR, compared to may others, that makes you say that.  In other words, why only some PRs should have a unit test?

tolot27:
> I also wonder what is special about this PR, compared to may others, that makes you say that.

It has nothing to do with your PR. I've also requested unit tests for other PRs. The main problem is that many developers don't take unit testing into account or find it hard to implement. If I find code not easy to understand or with potentially problematic/dangerous code lines, I recommend unit tests.

> In other words, why only some PRs should have a unit test?

I wish we had unit tests for all PRs and a much higher code coverage because that avoids further errors in xDrip and detects already existing errors, semi-automatically. That's one of the biggest concerns external developers had about our code basis.

Navid200:
There is a conditional in the original code, line 537, that only let's access to the formula and sounding the alarm when the numerator of the formula is non-negative.  
I removed that original conditional.  So, now, the executable runs all the time.  What I gained by removing that conditional is that I can create sound, if the new setting is disabled, using the same exisiting executable.
The disadvantage of removing that conditional is now we may end up with negative values. So I added volumeFrac = Math.max(volumeFrac, 0).  Now, when the numerator becomes negative, this formula changes it to 0.

What's left is to create sound even when the formula results in 0 if the new setting is disabled.   A single new conditional does that.  If the new setting is not enabled, it raises the volume to 30% if it is lower.  If it is enabled, it leaves the volume alone and everything works as it did before.

tolot27:
If I understand the ascending volume profile correctly, it increases the volume every minute by "x" until the next reading is processed. Is my understanding correct?

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
tolot27: There is no need to set the `volumeFrac` before the condition `profile == ALERT_PROFILE_MEDIUM` evaluates to `true`. That's why I suggest to change it to:

```java
            float volumeFrac;
            if (profile == ALERT_PROFILE_MEDIUM) {
                volumeFrac = (float) 0.7;
            } else
                volumeFrac = (float) (minsFromStartPlaying - MAX_VIBRATING_MINUTES) / (MAX_ASCENDING_MINUTES - MAX_VIBRATING_MINUTES);
                // While minsFromStartPlaying <= MAX_VIBRATING_MINUTES, we only vibrate ...
                volumeFrac = Math.max(volumeFrac, 0); // Limit volumeFrac to values greater than and equal to 0
                volumeFrac = Math.min(volumeFrac, 1); // Limit volumeFrac to values less than and equal to 1
                if (!Pref.getBoolean("delay_ascending_3min", true) && volumeFrac < 0.3) {
                    volumeFrac = (float) 0.3; // If delay_ascending_3min is disabled, we never only vibrate.
                }
            }
```

Navid200:
> If I understand the ascending volume profile correctly, it increases the volume every minute by "x" until the next reading is processed. Is my understanding correct?

That is correct if we look at it literally.  However, effectively, the current setting will set the volume to 0 and keep it like that for 3 consecutive minutes.  Then, it raises it to 33%.  Next minute, it raises it to 0.67%.  The following minute, it will raise it to 100%.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
Navid200: We have to ask the original developer who wrote this.
If I was to speculate, I would say the following:
If the ascending volume profile is selected, the volume is updated every minute.

But, if the medium volume profile is selected, any modifications made by the ascending volume conditional need to be overwritten by 70%.  Otherwise, the ascending volume scheme will overwrite the medium volume.

By placing 

if (profile == ALERT_PROFILE_MEDIUM) {
                    volumeFrac = (float) 0.7;

after the entire ascending volume code, if medium volume is selected, it will overwrite.

What I don't understand is how High volume overwrites it if selected.

But, I have not changed that.  So, it should still work.

I am now testing Medium setting on my device.
After, I will test High.

Navid200:
I just completed testing Medium and High volume profiles after this PR.
They work as intended setting the volume to 70% and 100% respectively.
I conclude from this that the following is redundant.

if (profile == ALERT_PROFILE_MEDIUM) {
volumeFrac = (float) 0.7;

But, it has been a part of the xDrip code since June 8, 2015.
Someone can open a PR and submit a request to remove it from the code.  I have no idea if it will be merged or not.
But, please let's not make that change a part of this PR.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
tolot27: > But, if the medium volume profile is selected, any modifications made by the ascending volume conditional need to be overwritten by 70%.

Thats true. `volumeFrac = (float) 0.7` is an assignment of a constant and is completely independent of the ascending volume profile. That's why it is not necessary to do the ascending calculations if the medium profile is selected.
Hence, I don't think it is necessary to ask the original developer because my suggestion does not change the logic. BTW: I've simplified it further which is more readable.

tolot27:
@Navid200 Please can you enable modifications of this PR. Afterwards, I can do extract method refactoring and adding a unit test.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
Navid200: Thanks for the suggestions.
I would like to leave it.  
This PR changes something that has been in the code for longer than 6 years.
I'd like to minimize the number of lines being changed to increase the chance of it being merged.  I do that by only changing what is necessary.
Changing the Medium volume assignment, which seems to be redundant, is not necessary to accomplish the objective of the PR, to avoid delaying the ascending volume alerts.

There could be another PR that would delete the medium volume assignment here, and perform tests to confirm functionality remains unchanged.  I believe if medium or high volume profiles are selected, they are assigned somewhere else in the code.
There are many lines in the code that can be removed.
But, proposing to change something that makes no difference to functionality only jeopardizes its approval.

A PR that is not merged, no matter how correct and complete it is, is useless.

Navid200:
I prefer to make any required changes myself.

tolot27:
> I prefer to make any required changes myself.

Then please implement the unit test your self.

app/src/main/res/values/strings.xml
```diff
@@ -1432,6 +1432,8 @@
     <string name="title_ongoing_notification_channel">Use ongoing notification channel</string>
     <string name="summary_notification_channels_grouping">Whether notifications should be grouped together. Incompatible with \'Number Item in Notification Area\' option</string>
     <string name="title_notification_channels_grouping">Group notifications together</string>
+    <string name="summary_delay_ascending_3min">For ascending volume profile, delay the start of the alert sound by 3 minutes</string>
```
tzachi-dar, line 1435: I would write here something like:
"For ascending volume profile, only vibrates for the first 3 minutes"

tzachi-dar:
Apart from my comment, LGTM.

app/src/main/res/values/strings.xml
Navid200: Thanks for the review.

I don't understand your comment.
I just tested xDrip.  It doesn't vibrate just for the first 3 minutes.  It continues to vibrate until the alert is snoozed.

This PR does not make any changes to the vibration behavior.
The only change this PR makes is with the sound behavior, and only if the new setting is disabled.

Would you please explain what you mean by your suggestion?

Thanks again

Navid200: May be I'm starting to understand your point.  Do you mean this summary says nothing about the vibration and the user may think that the vibration is also delayed?

If yes, what if I change the summary to the following?

For ascending volume profile, delay the start of the alert sound by 3 minutes.  The vibration (if the vibration is enabled) starts as soon as the alert triggers regardless of this setting.

tzachi-dar: yes, your last suggestion seems what I meant.

jamorham:
Thanks for taking the time to do this and describe and test the potential behavior changes as well as give the user control over the change. Thanks to the other reviewers also. It looks good to me.

PR #1835, opened 2021-09-24
Don't ask to enable oop if libreoop2 is found.

tzachi-dar:
oop2 actually does not work when oop is enabled, so we should not offer this option to the users.

Actually, there was another problem here:
The code does a check for which external application is installed, and if one of the oop is found, it is set in the preference (EXTERNAL_ALG_PACKAGES).
Later, EXTERNAL_ALG_PACKAGES is used as a target when sending intents. But this check was only done when external_blukon_algorithm is not turned on.
This is wrong, the check has to be done always. Asking to enable oop should only be done if oop is off.


tolot27 (commented):
LGTM

tolot27 (approved):

Issue #1631, opened 2021-01-31
New feature: Import settings from file system


master3395:
It might be easy to keep it in 1 folder and 1 file to export/import.
But sadly not everyone has "SD" to save to or load from.
And I can't seem to find a good way to restore the settings.
If I click export settings, it goes to the "download" folder.
But if I try and import, it does not work, the image below will show that having the new config file in downloads does not let it get imported.
So maybe let us find the file and click import instead of an auto check for 1 location?
(The app has all permissions the phone can give, OnePlus 8 Pro EU, android 11)

![image](https://user-images.githubusercontent.com/4199669/106401182-0aa2b380-6423-11eb-9fd8-5f1a5e3b48f0.png)

SimonPhilpott:
This is similar to the following and would address both things. https://github.com/NightscoutFoundation/xDrip/issues/1622

Navid200:
Why do you think it doesn't work?

I just tested this on Jan.29, 2021.  But, I don't think there is anything new about it.  It worked with much older versions as well.

I emptied the Downloads folder.
Under Import/Export features, tapped on Load/Save Settings on SD card.

Tapped on LOAD all settings from SDCARD.  Got the message could not load preferences.

Then, tapped on Save all settings to SD card.  It saved xDrip-export in the Downloads folder.
Changed the Hardware data source from disabled (test phone) to Nightscout follower.
Went back to Load/Save Settings on SD card and tapped on Load all settings from SDcard. 
It worked.  The data source was back to disabled.

master3395:
> Why do you think it doesn't work?
> 
> I just tested this on Jan.29, 2021. But, I don't think there is anything new about it. It worked with much older versions as well.
> 
> I emptied the Downloads folder.
> Under Import/Export features, tapped on Load/Save Settings on SD card.
> 
> Tapped on LOAD all settings from SDCARD. Got the message that could not load preferences.
> 
> Then, tapped on Save all settings to SD card. It saved xDrip-export in the Downloads folder.
> Changed the Hardware data source from disabled (test phone) to Nightscout follower.
> Went back to Load/Save Settings on SD card and tapped on Load all settings from SDcard.
> It worked. The data source was back to disabled.

I re-installed the app, I hit export on the default settings.
Yes, it does save to the download folder, if you don't have an SD card.
But then I changed several settings to see if it changed back to default when importing.

And the error message came, and the settings were not changed.

Navid200:
[...]

PR #1837, opened 2021-09-24
Remove emphasis from SD card

Navid200:
**Why we need this**
Users get the impression that if their phone has no SD card, they cannot save their settings (https://github.com/NightscoutFoundation/xDrip/issues/1631#issue-797838599).  This is because the menu items for saving or loading settings refer to the SD card.   
The following images show what we have now.
Screenshot_20210924-183219

Screenshot_20211002-234720

Screenshot_20210924-181743

But, xDrip can also save to and load from the internal storage.  This PR replaces the wordings of the menu items to reflect that.
The following images show what we will have after this change.
Screenshot_20210924-183632

Screenshot_20211002-233845

Screenshot_20211004-101147

There is no change to any functionality in this PR; only wordings of the existing buttons and menu items.

**Is there a workaround?**
No

No

**Tests**
Tested on an Android 8 phone

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -369,10 +369,10 @@
     <string name="version_details">Version details</string>
     <string name="download_now">Download Now</string>
     <string name="automatically_check_for_updates">Automatically check for updates</string>
-    <string name="here_you_can_save_the_settings_to_the_external_storage_sdcard">Here, you can save the settings to the internal storage (or SD card).\n\nOnce saved to a file, it is potentially possible for any app to read the settings which might contain sensitive information.\n\nYou are advised to delete the settings once you have imported them again if you are concerned about this.</string>
+    <string name="here_you_can_save_the_settings_to_the_external_storage_sdcard">Here, you can save the settings to the internal storage (Download/xDrip-export/com.eveningoutpost.dexdrip_preferences.xml).\n\nOnce saved to a file, it is potentially possible for any app to read the settings which might contain sensitive information.\n\nYou are advised to delete the settings once you have imported them again if you are concerned about this.</string>
     <string name="save_all_settings_to_sdcard">Save all Settings to the settings file</string>
     <string name="load_all_settings_from_sdcard">Load all settings from the settings file</string>
-    <string name="delete_any_settings_on_sdcard">Delete settings file</string>
+    <string name="delete_any_settings_on_sdcard">Delete settings folder (Download/xDrip-export)</string>
     <string name="use_qr_codes_to_transfer_settings_using_settings_auto_configure_feature"><![CDATA[Use QR codes to transfer settings using Settings->Auto Configure feature]]></string>
     <string name="xdrip_plus_security_key_settings_only">xDrip Plus Security Key Settings only</string>
     <string name="show_general_and_collection_settings">Show General and Collection Settings</string>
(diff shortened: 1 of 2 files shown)


nightscout-ricsi:
Another great polish to make xDrip userfriendly!. I have modified this string in the translation, than one can save it to a folder or a card.

emp-00:
Great modification - one additional idea: 
1) Can you permanently show the path+filename where the data will be saved as well? I had a hard time finding the export in my file explorer since xDrip only shows the path for 1 or 2 seconds after saving and then it disappears again.
2) Currently we have this set of options in the menu:
a) "Load / Save Settings on SD card" ==> showing the additional screen (see above)
b) "Export database" and "Restore database" ==> no additional info screen, direct export/restore
For consistency, can we change the two b) options into one item "Export / Restore database" which then also shows a summary screen (including path/filename, see 1) ) and the user can choose between Export/Restore on the additional screen? Then both options would be 100% consistent and improved from user experience from my perspective.

Navid200:
@emp-00 I am fine with making the changes you have suggested to xDrip.
However, the more changes are included in a single PR, the less likely it is that it will be merged.
So, I am going to just wait and see where this PR goes as it stands.  

1- If it is merged, I am happy to talk about making another PR incorporating what you have asked.
2- If jamorham asks that I add the changes you have suggested now before he merges this, I will make the changes now.

I hope that makes sense.

emp-00:
@Navid200 : Thanks for the efforts and structured way of working for the community. Much appreciated.

jamorham:
This looks good, the whole meaning of SD card in android has completely changed over the years so these changes make sense.

app/src/main/res/values/strings.xml
```diff
@@ -369,10 +369,10 @@
...
     <string name="automatically_check_for_updates">Automatically check for updates</string>
-    <string name="here_you_can_save_the_settings_to_the_external_storage_sdcard">Here you can save the settings to the external storage (SD card).\n\nOnce saved to the external storage it is potentially possible for any app to read the settings which might contain sensitive information.\n\nYou are advised to delete the settings once you have imported them again if you are concerned about this.</string>
-    <string name="save_all_settings_to_sdcard">Save all Settings to SD card</string>
-    <string name="load_all_settings_from_sdcard">Load all Settings from SD card</string>
-    <string name="delete_any_settings_on_sdcard">Delete any settings on SD card</string>
+    <string name="here_you_can_save_the_settings_to_the_external_storage_sdcard">Here, you can save the settings to the internal storage (or SD card).\n\nOnce saved to a file, it is potentially possible for any app to read the settings which might contain sensitive information.\n\nYou are advised to delete the settings once you have imported them again if you are concerned about this.</string>
```
tolot27, line 372: I would remove the brackets around "or SD card" because "or" is a grammatical conjunction and I don't see any meaning of the brackets, here.

tolot27 (changes requested):
LGTM so far.

app/src/main/res/values/strings.xml
tolot27: Maybe, we can shortened it to "Here, you can save the settings.". Alternatively, we can think about improving consistency and replace all Load/Save staff with Import/Export because it is below "Import/Export features".

Navid200: Thanks for looking at this.

After I submitted this, I wasn't happy with leaving or SD card.  When I use this to save settings, it is always saved to the Download folder.  I don't know how you can save to the SD card even if you want to.  So, unless there is any objection, I would like to remove or SD card and the brackets.  But, I'd rather not just remove the brackets.

I wouldn't change to "here, you can save the settings".  The fact that the settings are saved to the internal storage, which is outside the boundary of xDrip, is very important for the user.  That's what this paragraph is getting at by the warning that the user should be aware of the possibility of revealing important information if someone copies the saved file without their knowledge.

tolot27 (approved):

PR #1840, opened 2021-09-29
Enable override silent mode and force speaker for the 2 existing alarms

Navid200:
**Why we need this**
When xDrip is installed, there are two existing alarms.  They have override silent mode and force speaker disabled.
If you create a new alert, override silent mode and force speaker are enabled by default.
This is a discrepancy in defaults that this PR resolves.
If this is merged, after installing xDrip, override silent mode and force speaker will be enabled on the two existing alerts.

I kept looking for where the settings for the existing alerts came from.  But, couldn't find it, and eventually, gave up and reached out to Adrian.  He pointed out where the settings were.
Thanks, @AdrianLxM. 

**Is there  a workaround?**
No.

No

**Tests**
Tested on an Android 8 phone.


tolot27 (approved):
This PR solves a minor inconsistency of the alarm settings.

jamorham (approved):

Issue #303, opened 2018-02-12
Complication displays permanently null null ()


zackallan:
After reinstalling the xDrip+ app on my Ticwatch E the complication on the watch face fails to display the measurement values.
OS watch: Android Wear 2.8.0 Android 7.1.1
watch: xDrip+ build 8d09da5-2018.02.11 version 1.0.2
OS phone: Android 8.1.0
phone: xDrip+ build 8d09da5-2018.02.11

zackallan:
@gregorybel Thanks for the tips. I did reboot on both the phone and the watch. Android wear is activated as described in the manual. I did reconfigure the complication. Unfortunately these steps did not solve the problem. Clicking the complication does not show any reaction.

gregorybel:
then need to pass! or maybe you should look at the inter-app settings.

one thing is sure, it works for me.

ins5736:
I'm running the 2/13 Xdrip nightly/1.58 NSClient as a follower. Data works on the dedicated Xdrip watches, but not on the complication.

The complication works for my wife who is running Xdrip as a collector and uploader. We both have Ticwatch e watches.

jamorham:
Okay I think this bug report is that complication doesn't work on the follower device. It is possible because of how the sync works.

ins5736:
Yes, that's correct. The complication works perfectly for my wife using the uploader phone.

Ian Smith BS, RN
Programmer Analyst eRecord
________________________________

zackallan:
It would be interesting how the synchronisation works here. In my case there is no Nightscout involved in the dataflow. xDrip+ on my phone sends the data directly to the watch.

ins5736:
Sorry for the confusion. I'll start a different issue.

gitanO0:
Just just looking at getting a Ticwatch e.  Is this watch compatible with xdrip+ G5 setup?

ins5736:
Displays data for my wife, but she hasn't used it for a collector.

jamorham:
Please try the May 11th Nightly. This should now be resolved.

zackallan:
Updated to May 11th Nightly and complication still displays "null null ()". But as mentioned above I am not using my phone as a follower device.

jamorham:
related: https://github.com/NightscoutFoundation/xDrip/issues/325

jamorham:
[...]

PR #1846, opened 2021-10-08
Bugfix: Wear complication null with Share and Nightscout follow sources

JasonJShuler:
Fix for issue #303 - confirmed fixed using Follow Share source.

1. Updated Listener service on WearOS to apply the calibration-ignoring logic previously exclusive to Dexcom Native, to now include Dexcom Share Follow. This is appropriate as Follow BG values are identical to Native values and do not require calibration. Credit to @gregorybel for finding this one!
2. Populating the BgReading uuid in the ShareFollow and NightscoutFollow source entry processors

The null uuid in the BgReading values created by the follow sources were causing a nulPointerException in BgReading.getByUuid(). This was causing BgReading.last() to always return null.
The most visible consequence was the complication provider always getting null values, however BgReading.last() is used extensively throughout the Wear OS side code (one example is in checking to see if it needs to perform an update - without the uuid, the answer was always yes)

These are the only instances in the app-side code of "new BgReading" (outside of the BgReading class itself) that do not populate the uuid. The same method is used here as in the Native Dex source.

Confirmed working on Wear OS emulator and Galaxy Watch 4.

NOTE: If it was intended that the uuid be null, a condition to check for a null parameter can be added to getByUuid(). (This is what I did first, but it felt like a hack given my limited understanding of the codebase)

Demo_Follow_Complication_Working

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NSClientReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NSClientReceiver.java
@@ -199,7 +199,6 @@ public class NSClientReceiver extends BroadcastReceiver {
         JSONObject jsonObject = new JSONObject();
         try {
             jsonObject.put("uuid", UUID.randomUUID().toString());
-            //TODO: Determine impact of making this not-null - scan all code
             jsonObject.put("source_info", "NSClient Follow");
 
             jsonObject.put("timestamp", sgv_map.get("mills"));
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NSClientReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NSClientReceiver.java
@@ -199,6 +199,8 @@ public class NSClientReceiver extends BroadcastReceiver {
         JSONObject jsonObject = new JSONObject();
         try {
             jsonObject.put("uuid", UUID.randomUUID().toString());
+            //TODO: Determine impact of making this not-null - scan all code
+            jsonObject.put("source_info", "NSClient Follow");
 
             jsonObject.put("timestamp", sgv_map.get("mills"));
             jsonObject.put("calculated_value", sgv_map.get("mgdl"));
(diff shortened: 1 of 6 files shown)


FreDiabetics:
Does this work for dexcom readings recieved by the local broadcast? datasource 640G/Eversense? as I'm using the build-your-own-dexcom-app to recieve my bg and push it to xDrip+ via AndroidAPS.

Navid200:
@JasonJShuler No, that's not what I meant at all.
I thought you wanted to make some changes to it.  I am not talking about the question whether to edit the existing PR or closing it and opening a new PR.  I am not talking about that question.
But, that question gave me the impression that you needed to make a change.

If you are happy with your PR as is and if it is merged, you will be happy with it and you will not need to open another PR to correct any errors in it, then leave it be.

On the other hand, if there is anything in it that you need to change, please change this to a draft.  Otherwise, it may be merged as is.  

When you say this is ready for a review, I hope you mean you are happy with this if it is merged as is.

app/src/main/java/com/eveningoutpost/dexdrip/NSClientReceiver.java
```diff
@@ -199,6 +199,8 @@ private String toBgReadingJSON(HashMap<String, Object> sgv_map) {
         JSONObject jsonObject = new JSONObject();
         try {
             jsonObject.put("uuid", UUID.randomUUID().toString());
+            //TODO: Determine impact of making this not-null - scan all code
```
tolot27, line 202: I do not fully understand this TODO comment. Please can you explain it and maybe solve it, too?

tolot27 (changes requested):

jamorham:
Looks good to me

andreaontheweb:
I'm on the last nightly build 20230503 but I have still this problem. Null null on watch 5 complications, I'm a follower of a xdrip g6. Native watchface xdrip works well

andreaontheweb:
If I swap hardware source to Dexcom share it works, xdrip sync follow gives null null..

PR #1852, opened 2021-10-15
Disable Predictive simulation by default

Navid200:
**Why we need this**  
New users often post on facebook asking what everything shown on screen means.  When, they should focus on learning the basics of xDrip and not be distracted by the simulation parameters or curves.  

**Is there a workaround?**  
No  

No  

**Tests**  
Tested on an Android 8


jamorham:
I wouldn't plan to integrate this as is because of the importance of the predictive simulations, but if you can give an example of the type of new user problem which might occur then we could maybe do something with the `Experience` feature to gate it until later on.

PR #1858, opened 2021-10-20
Speak twice bug in German

Navid200:
**Why we need this**
The bug is described here:
https://github.com/NightscoutFoundation/xDrip/commit/52b1ef5937f0816bbd99d99f47911952879fe7c1#commitcomment-57648510
It results in a wrong spoken word in German.  There is no issue with English.
This is on my phone:
https://drive.google.com/file/d/1rEff_6gDdLQGc9hzr5QYZzkrWwfaWi4P/view

This is after the change:
https://drive.google.com/file/d/160eJWnpk7YJAkQFszAR6dfg9wjm-TqpI/view?usp=sharing

**Is there a workaround?**
No

No

**Tests**
Tested with an Android 8 follower


jamorham:
Looks good to me

Issue #1077, opened 2019-11-02
numberwall_background not loaded on app start/phone reboot


zjedr:
numberwall_background is not loaded on reboot and is not displayed in configuration page after reboot.
I believe people using this feature are wanting to display medic alert information on their lock screen, therefore this is an important feature to fix. 
Found on Galaxy S9 running 9.0

Navid200:
@zjedr Would you please explain how this problem can be recreated?
Assume you are talking to someone who has never used xDrip.  Don't expect them to know what any acronym means.  Explain every single step required to recreate this problem.
Your post contains 3 lines.  You may need to create a post that contains 30.  Use as many lines as you need to help us recreate the problem.

Alternatively, if the issue has been resolved already, would you please close this issue?
Thanks

zjedr:
Number wall on lock screen does not display image after reboot.

Number wall on lock screen allows a background image to be display under the BG reading. 
The image is not loaded on the lock screen or in xdrip setting after your phone is rebooted and xdrip has started.

- number wall and image are set in Settings/xDrip+ Display Settings/Number Wall on Lock Screen
- set 'Use on Lock screen' on
- select 'Configure Number Wall', then select the folder icon and choose an image to display 

Lock screen now displays image and BG reading correctly until phone is rebooted.

After reboot (and xdrip starting) the background image is not displayed on the lock screen and is not shown in the 'Number Wall on Lock Screen' settings area.

This occurs on multiple phones including Samsung S9, Unihertz Atom and both Android 9 and 10, stable and nightly release.

Example image:-
![t1logo](https://user-images.githubusercontent.com/13398921/107864441-3d7f8980-6eb0-11eb-8cd1-8a73519bcb6f.png)

Navid200:
After I restart the phone, I can still see the image until xDrip starts.  As soon as xDrip starts, after a reboot, the image disappears.  
It is as if the image is stored in a cache that is cleared by xDrip upon a restart.

jamorham:
The numberwall is done using lockscreen wallpaper so it is quite likely the system will store this and display the old/previous one on reboot. I believe the issue being reported is that a customized background image is no longer applied after reboot.

[...]

PR #1867, opened 2021-10-30
Fix NumberWall Background image

marciogranzotto:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/1077

After rebooting the content image was not pointing to the correct image file.
The solution is to copy the image content to a new image file on internal storage.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
@@ -72,8 +72,7 @@ public class BitmapUtil {
                     FileDescriptor fileDescriptor = xdrip.getAppContext().getContentResolver().openFileDescriptor(background_uri, "r").getFileDescriptor();
                     image_bitmap = BitmapFactory.decodeFileDescriptor(fileDescriptor);
                 } else {
-                    File file = new File(background_file);
-                    inputStream = new FileInputStream(file);
+                    inputStream = new FileInputStream(background_file);
                     image_bitmap = BitmapFactory.decodeStream(inputStream);
                     inputStream.close();
                 }
@@ -128,7 +127,7 @@ public class BitmapUtil {
         String destinationFileName = "background" + new Date().getTime() + ".png"; // unique name to avoid caching
         File destination = new File(destinationFolder, destinationFileName);
 
-        Bitmap sourceImage = BitmapUtil.getBackgroundBitmap(getScreenWidth(), getScreenHeight(), sourceUri.toString());
+        Bitmap sourceImage = getBackgroundBitmap(getScreenWidth(), getScreenHeight(), sourceUri.toString());
         String output = destination.getPath();
         if (sourceImage != null) {
             FileOutputStream out = null;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
@@ -15,6 +15,7 @@ import android.net.Uri;
 
 import com.eveningoutpost.dexdrip.Models.JoH;
 import com.eveningoutpost.dexdrip.UtilityModels.Pref;
+import com.eveningoutpost.dexdrip.utils.FileUtils;
 import com.eveningoutpost.dexdrip.xdrip;
 
 import java.io.File;
@@ -58,6 +59,7 @@ public class BitmapUtil {
 
     public static Bitmap getBackgroundBitmap(final int xSize, final int ySize, final String background_file) {
         Bitmap tiledBitmap = null;
+        FileInputStream inputStream = null;
         try {
             tiledBitmap = BitmapLoader.bitmapFromBundleCache(background_file);
             if (tiledBitmap == null) {
@@ -71,9 +73,9 @@ public class BitmapUtil {
                     image_bitmap = BitmapFactory.decodeFileDescriptor(fileDescriptor);
                 } else {
                     File file = new File(background_file);
-                    FileInputStream in = new FileInputStream(file);
-                    image_bitmap = BitmapFactory.decodeStream(in);
-                    in.close();
+                    inputStream = new FileInputStream(file);
+                    image_bitmap = BitmapFactory.decodeStream(inputStream);
+                    inputStream.close();
                 }
 
                 Bitmap rotated_bitmap = image_bitmap;
@@ -84,10 +86,8 @@ public class BitmapUtil {
                         FileDescriptor fileDescriptor = xdrip.getAppContext().getContentResolver().openFileDescriptor(background_uri, "r").getFileDescriptor(); // reset
                         exif = new ExifInterface(fileDescriptor);
                     } else {
-                        File file = new File(background_file);
-                        FileInputStream in = new FileInputStream(file);
-                        exif = new ExifInterface(in);
-                        in.close();
+                        inputStream = new FileInputStream(background_file);
+                        exif = new ExifInterface(inputStream);
                     }
                     int rotation = exifOrientationToDegrees(exif.getAttributeInt(ExifInterface.TAG_ORIENTATION, ExifInterface.ORIENTATION_NORMAL));
                     android.util.Log.d("NumberWall", "Rotation: " + rotation);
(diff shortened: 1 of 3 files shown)


Navid200:
I can confirm that this PR addresses the issue https://github.com/NightscoutFoundation/xDrip/issues/1077.
Restarted the phone multiple times, and the chosen lockscreen image remains active, after using this PR.

tolot27 (changes requested):
Please remove all unnecessary changes like variable renaming. This distracts the review process.

You can add such refactoring using a separate commit.

BTW: Code formatting using the standard format is okay.

Navid200:
@marciogranzotto I spent at least a week full time looking at this and eventually, I tabled it because it was too complicated for me to figure out.  I thought may be some day, I will figure out how to fix it.
It seemed that you looked at it for only a few hours and you had a solution already.  Awesome!
I was pleasantly surprised, and very happy and hopeful that it seems another very competent developer is joining us.  I am delighted.

If I may state a few words, I would say that we need to minimize the risk of errors.  So, we only change what absolutely needs to change.  Now, I personally, may correct a spelling error in a comment if I notice it while I am making my necessary changes.  But, that's as far as it goes for me.  I never change anything else.
This is because I want to make it as easy as possible for those reviewing to feel confident that there are no mistakes.  The more extensive the changes are, the harder it becomes for them to verify.

Thanks a lot for fixing this.
Please follow the requests made by tolot27.

marciogranzotto:
@Navid200 thanks for the kind words!~

@tolot27 I've rollback the variable name changes. I just kept the android studio's autoformat changes

app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
```diff
@@ -86,6 +56,84 @@ public static Bitmap getTiled(final Bitmap source, final int xSize, final int yS
...
+        String destinationFileName = "background" + new Date().getTime() + ".png"; // unique name to avoid caching
+        File destination = new File(destinationFolder, destinationFileName);
+
+        Bitmap sourceImage = BitmapUtil.getBackgroundBitmap(getScreenWidth(), getScreenHeight(), sourceUri.toString());
+        if (sourceImage != null) {
+            try {
+                FileOutputStream out = new FileOutputStream(destination);
+                sourceImage.compress(Bitmap.CompressFormat.PNG, 100, out);
+                out.flush();
+                out.close();
```
tolot27, line 127: You should close the output stream in a finally block after catching a possible IOException.

app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
```diff
@@ -86,6 +56,84 @@ public static Bitmap getTiled(final Bitmap source, final int xSize, final int yS
...
+            // cannot load bitmap
+            android.util.Log.e("NumberWall", "Cannot load bitmap: " + e);
+        }
+        return tiledBitmap;
+    }
+
+    public static String copyBackgroundImage(Uri sourceUri) {
+        File destinationFolder = new File(xdrip.getAppContext().getFilesDir().getAbsolutePath() + File.separator + "numberWall");
+        destinationFolder.mkdirs();
+        String destinationFileName = "background" + new Date().getTime() + ".png"; // unique name to avoid caching
```
tolot27, line 118: What happens with the PNGs created during a previous session? Or more precisely, where and when are the unique background images cleaned?

app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
```diff
@@ -86,6 +56,84 @@ public static Bitmap getTiled(final Bitmap source, final int xSize, final int yS
...
+                    if (background_file.startsWith("content://")) {
+                        FileDescriptor fileDescriptor = xdrip.getAppContext().getContentResolver().openFileDescriptor(background_uri, "r").getFileDescriptor(); // reset
+                        exif = new ExifInterface(fileDescriptor);
+                    } else {
+                        File file = new File(background_file);
+                        FileInputStream in = new FileInputStream(file);
+                        exif = new ExifInterface(in);
+                        in.close();
+                    }
+                    int rotation = exifOrientationToDegrees(exif.getAttributeInt(ExifInterface.TAG_ORIENTATION, ExifInterface.ORIENTATION_NORMAL));
```
tolot27, line 92: `exif` could still be `null` here.

app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
```diff
@@ -86,6 +56,84 @@ public static Bitmap getTiled(final Bitmap source, final int xSize, final int yS
...
+                Bitmap rotated_bitmap = image_bitmap;
+                final Matrix image_matrix = new Matrix();
+                if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N) {
+                    ExifInterface exif = null;
+                    if (background_file.startsWith("content://")) {
+                        FileDescriptor fileDescriptor = xdrip.getAppContext().getContentResolver().openFileDescriptor(background_uri, "r").getFileDescriptor(); // reset
+                        exif = new ExifInterface(fileDescriptor);
+                    } else {
+                        File file = new File(background_file);
+                        FileInputStream in = new FileInputStream(file);
```
tolot27, line 88: Please combine it with the previous line into `FileInputStream in = new FileInputStream(new File(background_file));`

app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
```diff
@@ -86,6 +56,84 @@ public static Bitmap getTiled(final Bitmap source, final int xSize, final int yS
...
+                        exif = new ExifInterface(in);
+                        in.close();
```
tolot27, line 90: If you close the input stream here, you cannot access the exif data anymore. Please change it and use proper exception handling (try, catch finally). An example can be found at https://blog.mindorks.com/exifinterface

tolot27 (changes requested):
Thanks for keeping the commit simple. I've added some comments.

After addressing them and review from our side, we are open for a last commit changing the variables as you proposed.

app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
marciogranzotto: yeah, you are right! My first implementation was just overriding the same `background.png` file, but that was not working because of the bitmap caching. I'll add a call to clean that folder before saving the new file.

app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
marciogranzotto: actually, if I add an `if (exif == null)` there, Android Studio shows a warning that it's not necessary. I think that because on the `else` case it will throw an exception if it fails to open the file, otherwise it will not be null

app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
```diff
@@ -86,6 +56,84 @@ public static Bitmap getTiled(final Bitmap source, final int xSize, final int yS
...
+        return tiledBitmap;
+    }
+
+    public static String copyBackgroundImage(Uri sourceUri) {
+        File destinationFolder = new File(xdrip.getAppContext().getFilesDir().getAbsolutePath() + File.separator + "numberWall");
+        destinationFolder.mkdirs();
+        String destinationFileName = "background" + new Date().getTime() + ".png"; // unique name to avoid caching
+        File destination = new File(destinationFolder, destinationFileName);
+
+        Bitmap sourceImage = BitmapUtil.getBackgroundBitmap(getScreenWidth(), getScreenHeight(), sourceUri.toString());
```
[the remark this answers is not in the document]
marciogranzotto: I think so... this way we already save the proper bitmap instead of copying the file.

app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
marciogranzotto: apparently I can just `new FileInputStream(background_file)` with the path for the file
image

marciogranzotto:
@tolot27 I've made the changes that you suggested, plus another one in a different commit, also deleting the image files when the user taps on the "folder" image button

app/src/main/java/com/eveningoutpost/dexdrip/ui/helpers/BitmapUtil.java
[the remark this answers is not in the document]
tolot27: What I meant is: is it necessary to call the function `getBackgroundBitmap` with the class name. Even if the function is static, it should be callable inside the the same class without the class name.

tolot27 (approved):
LGTM. Thanks for the contribution and review request handling.

@Navid200 Please can you test whether this modified PR still fixes the issue?

tolot27 (approved):
All review comments are addressed. This PR is ready for merging. Thanks again, for the contribution!

jamorham:
There is something weird with this merge which causes problems on one of my machines. Basically the lombok processing stops working.  I can't see any reason for it so I'm going to trace through each of the changes to try to discover what could be causing it and try resetting gradle/caches etc.  Until then I have temporarily reverted this merge.

tolot27:
> There is something weird with this merge which causes problems on one of my machines. Basically the lombok processing stops working.

Did you try updating Lombok from 1.18.10 to 1.18.22 as I did with cd333fe? I'll have a look at it as well.

jamorham:
This change was necessary to get this to work on all my machines: https://github.com/NightscoutFoundation/xDrip/commit/84eb5aa098173ac06f9eed999e358d9139affe5a

I think this is because static imports are processed before annotation processors. I don't fully understand it though as it doesn't affect every gradle installation. Regardless, everywhere else in the code I believe static imports are declared last in the import list so we should probably continue with that convention to avoid compatibility problems.

PR #1871, opened 2021-11-02
Ask users to use discussions

Navid200:
I have only edited issues.  Not feature requests yet because I have questions.  After I understand, because there are similarities between the two, I will make the necessary changes once and then create the feature requests form from it.

I'm not clear about the role of the link to the discussion.  I have no problem with it.   But, what should we do if we see a comment in the Gitter room that raises a need for a new issue? 

I have changed the page that showed facebook and Gitter rooms as well as the options to open issues and feature request.
I added the option to post in discussions and put that up as the first option.
Is there a way to move the three options up before the options to open an issue and  a feature request?  I don't know how to do that.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,11 +1,12 @@
 ---
 name: "\U0001F41BBug report\U0001F41B"
-about: Create a report to help us improve things
+about: Create a report only if the developers asked you to do so
 label: bug
 ---
 
-### If you need support for xDrip, PLEASE DO NOT FILE A TICKET HERE
-- For support, please post a question to the xDripG5 group at Facebook (https://www.facebook.com/groups/xDripG5) or visit the Gitter room at https://gitter.im/jamorham/xDrip-plus.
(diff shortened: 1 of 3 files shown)


tolot27:
> I'm not clear about the role of the link to the discussion. I have no problem with it. But, what should we do if we see a comment in the Gitter room that raises a need for a new issue?

In that case, the Gitter comment can be linked. But in the future, it would be great if we can avoid Gitter and use the Discussions instead.

> Is there a way to move the three options up before the options to open an issue and a feature request?

I don't think so and I did not find any additional configuration for moving `contact_links` before the template links.

But I found [Issue Forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms). That's a cool feature to add required fields like Android and xDrip version to an issue description.

Navid200:
Thanks
And thanks for catching the typos.
I will now edit the feature request file.

tolot27 (approved):
LGTM

jamorham:
Yes we can try this and see how it goes. Thanks

PR #1873, opened 2021-11-06
Tzachi libretrendutilfix

tzachi-dar:
The main issue fixed here is that fact that on a first call of LibreTrendUtil.getData the value of m_libreTrendLatest.timestamp might be zero. In this case the data that will be read from the DB will be the entire data.
Depending on how big your DB is, it might take a long time, and creates an ANR.
After that, xdrip might also remove LibreTrendUtil from memory.

The fix is simple: only read the last 60 minutes if nothing else was specified.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
@@ -780,8 +780,11 @@ public class NFCReaderX {
         return true;
     }
 
+    // Sensor structure is described at  https://github.com/UPetersen/LibreMonitor/wiki
     public static ReadingData parseData(int attempt, String tagId, byte[] data, Long CaptureDateTime) {
-
+        final int FRAM_RECORD_SIZE = 6;
+        final int TREND_START = 28;
+        final int HISTORY_START = 124;
         int indexTrend = data[26] & 0xFF;
 
         int indexHistory = data[27] & 0xFF; // double check this bitmask? should be lower?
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -797,8 +797,8 @@ public static ReadingData parseData(int attempt, String tagId, byte[] data, Long
             //       getGlucose(new byte[]{data[(i * 6 + 125)], data[(i * 6 + 124)]});
```
tolot27, line 797: Please can you remove the comment if it it not necessary anymore?

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -797,8 +797,8 @@ public static ReadingData parseData(int attempt, String tagId, byte[] data, Long
...
 
             // If the data is decoded for some reason, we might have a wrong index.
-
-            if(i * 6 + 125 >= data.length) {
+            // The 6 is because we read up to 6 bytes.
```
tolot27, line 800: Are up to or exactly six bytes more read? It's not obvious to me where these extra bytes are read. Maybe, a constant should be used here and at the other place.

tolot27: A few lines later `data[(i * 6 + 125)]` and `data[(i * 6 + 124)]` are passed to a function. That's hard to understand. Maybe both `6` and `124` could be defined as a constant for more readability.

tolot27 (changes requested):
Just some minor changes for better readability.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
tzachi-dar: Done.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
tzachi-dar: I have replaced the code reading the data with constants. Also added a link to documentation.
I hope things are more readable now.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -849,12 +856,8 @@ public static ReadingData parseData(int attempt, String tagId, byte[] data, Long
     }
 
 
-    private static int getGlucoseRaw(byte[] bytes, boolean thirteen) {
-        if (thirteen) {
+    private static int getGlucoseRaw(byte[] bytes) {
             return ((256 * (bytes[0] & 0xFF) + (bytes[1] & 0xFF)) & 0x1FFF);
```
tolot27, line 860: 256 should also be a constant in the future.

jamorham:
Good work with this so far, but is it possible to avoid instantiating the elements of such a big list in `LibreTrendUtil`, for example with some code which only instantiates an empty object when it is needed? (I haven't tested the code below)

```
    class LazyLibreList extends ArrayList<LibreTrendPoint> {

        public LazyLibreList(int max_points) {
            super(max_points);
        }

        @Override
        public LibreTrendPoint get(int index) {
            LibreTrendPoint v = super.get(index);
            if (v == null) {
                v = new LibreTrendPoint();
                super.set(index, v);
            }
            return v;
        }
    }

    void ResetPoints() {
        if (m_points == null) {
            m_points = new LazyLibreList(MAX_POINTS);
        } else {
            m_points.clear();
        }
    }
```

I don't know enough about the data flow overall, but are all these elements really indexed each time? I'm assuming not which is why I'd suggest the optimization above. Also depending on how this list is used it might be possible to use a different data structure which doesn't even create the empty index for each record and reduce memory footprint that way. `ArrayList` however is very fast for sequential element retrieval.

tolot27 (approved):
Thanks for addressing my review comments. This will improve the code quality.

I've fixed the code formatting for `LibreOOPAlgorithm.java`. As a last commit, please do it for `NFCReaderX.java` as well and try to format your future changes with the default code format. Even if it looks like perfectionism, it ensures more readable code. Thanks!

tolot27:
Is there an open issue which is addressed by this PR?

tolot27 (approved):

app/src/main/java/com/eveningoutpost/dexdrip/utils/LibreTrendUtil.java
```diff
@@ -70,20 +71,40 @@ public String toString() {
     }
 }
 
+class LazyLibreList extends ArrayList<LibreTrendPoint> {
```
tolot27, line 74: Why do we need this extra class? IMHO, we can simply create the array list and replace the checks like at line https://github.com/NightscoutFoundation/xDrip/blob/e65299d2f5655ef797175a5c2a54e397fa76dfb8/app/src/main/java/com/eveningoutpost/dexdrip/utils/LibreTrendUtil.java#L203 to

```java
            if (m_points.get((int)id) == null) {
                m_points.get((int)id) = new LibreTrendPoint(id, glucoseData.glucoseLevelRaw, glucoseData.flags, glucoseData.source);
            }
```

And line https://github.com/NightscoutFoundation/xDrip/blob/e65299d2f5655ef797175a5c2a54e397fa76dfb8/app/src/main/java/com/eveningoutpost/dexdrip/utils/LibreTrendUtil.java#L162 should be replaced with

```java
                if (m_points.get(i) != null) {
```

Is `m_points` used somewhere else or just dumped, if `debug_per_minute` is `true`?

And if the `m_points` is sparse, a `HashSet` can be used instead and even a `TreeSet` if ordered traversal is necessary.

tolot27 (changes requested):

jamorham (approved):

PR #1875, opened 2021-11-12
Improve GRADLE build Performance

shisheng-1:
[Compiler daemon](https://docs.gradle.org/current/userguide/performance.html#compiler_daemon). The Gradle Java plugin allows you to run the compiler as a separate process by setting `options.fork = true`. This feature can lead to much less garbage collection and make Gradle’s infrastructure faster. This project has more than 1000 source files. We can consider enabling this feature.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.


tzachi-dar:
Can you please share improvement numbers for a clean build and an incremental build?

build.gradle
```diff
@@ -23,8 +23,12 @@ buildscript {
...
+    tasks.withType(JavaCompile).configureEach {
+        options.fork = true
+    }
+
     repositories {
         google()
         jcenter()
     }
-}
+}
```
tolot27, line 34: Please do not remove the newline at EOL.

shisheng-1:
hi @tzachi-dar  !
|                | clean build  |  incremental build |
| -------- | ------------ | ------------------ |
|with fork | [54s](https://github.com/i-Taozi/xDrip/actions/runs/1488364476)  |  [33s](https://github.com/i-Taozi/xDrip/actions/runs/1488368626)|
|without fork| [1m 16s](https://github.com/i-Taozi/xDrip/actions/runs/1488346340) | [36s](https://github.com/i-Taozi/xDrip/actions/runs/1488352398)|

build.gradle
shisheng-1: We have modified this file

tolot27 (approved):
This PR improves build performance, especially for clean builds. The impact on incremental builds (which is the default since Gradle 4.10) is far less, especially on faster machines.

jamorham:
What is the likely increase in ram usage? My reason for asking is I'm not sure what this will do to CI builds as there are memory constraints like 4GB ram etc.

tolot27:
> What is the likely increase in ram usage? My reason for asking is I'm not sure what this will do to CI builds as there are memory constraints like 4GB ram etc.

The JVM memory used by Gradle is [limited to 4 GB](https://github.com/NightscoutFoundation/xDrip/blob/master/gradle.properties#L14). If the memory consumption would increase much by the fork parameter and reaches the 4 GB boundary, the garbage collector would interact. But that would slow down the build process. Since the overall build time decreases, no memory issues occur.

Issue #1796, opened 2021-08-02
Issues with nightly build 2021.08.01


nathan990315:
The glycemic value is not shown on the main screen; and in the system notification. Even if the sensor is active. However you can see the sensor readings from the graph.
The nightly build is that of August 1st.
<img width="400" src="https://user-images.githubusercontent.com/17655407/127819483-45b797a0-4eb9-4442-b445-79f864316dfc.png" /><img width="400" src="https://user-images.githubusercontent.com/17655407/127819498-783f3d50-a7ab-4c11-a493-7061bf0375e9.png" />

tolot27:
Please try to revert to [2021.07.17](https://github.com/NightscoutFoundation/xDrip/releases/tag/2021.07.17) and report if the problem went away.

nathan990315:
I tried the 2021.07.31 and in this version I do not have this problem

jamorham:
@nathan990315 Can you let me know which handset model and android version this is with? thanks

nathan990315:
Nokia X20 and Android 11

parapenT1sta:
The same issue for me, Dexcom G5, Google Pixel 4a 5G with Android 12.

The numeric value wasn't updated but the graph was and for a while AndroidAPS was receiving the readings from xDrip matching with the graph dots values.

jamorham:
@parapenT1sta The main glucose value wasn't updated or it wasn't showing at all?

I'm preparing a new version I would like you both to try to see if you get the same issue.

parapenT1sta:
@jamorham The graph dots were being updated and working fine, only the value on the top right was stuck to the value I had before the update.

parapenT1sta:
@jamorham I will test the new version as soon you release it.

jamorham:
@parapenT1sta @nathan990315 Please try the latest nightly, it is also available here: https://github.com/NightscoutFoundation/xDrip/releases/download/2021.08.02/xDrip-plus-20210802-0af3ba0.apk
If it doesn't work just roll back to the 31st July version by simply installing that over the top.

nathan990315:
In the new version this problem has been solved for me

parapenT1sta:
Working fine here. Thank you!

Thomas-Wolff:
Hat leider nicht funktioniert, habe wieder auf den Bau vom 31.07 zurück gesetzt.

tolot27:
[...]

PR #1876, opened 2021-11-13
add Java 11 compatibility and cleanup build scripts

tolot27:
To make xDrip forward compatible and use features made available with Android 12 (i. e. new Bluetooth Scan without enabling GPS) it have to be compiled against Android SDK 31 (`compileSdkVersion 31`). This SDK requires at least Java 11. Unfortunately, the [Java Architecture for XML Binding (JAXB)](https://www.oracle.com/technical-resources/articles/javase/jaxb.html) required for data binding is not part of Java 11 anymore [^1]<sup>,</sup>[^2]. This results in the following error during build:

```
org.gradle.execution.MultipleBuildFailures: Build completed with 1 failures. Caused by:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':app:compileDebugJavaWithJavac'.
Caused by: java.lang.NoClassDefFoundError: javax/xml/bind/JAXBException
```

For Java 9 and 10 it can be enabled with a JVM argument `--add-modules` but this is not possible for Java 11 anymore.

This PR adds the JAXB dependencies if required (Java version >= 9) and also cleans up the Gradle build files.

[^1]: [JEP-320](https://openjdk.java.net/jeps/320)
[^2]: [JAXB on Java 9, 10, 11 and beyond](https://jesperdj.com/2018/09/30/jaxb-on-java-9-10-11-and-beyond)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,7 +12,7 @@ buildscript {
 apply plugin: 'com.android.application'
 apply from: 'token-replace.gradle'
 apply from: '../dependencies.gradle'
-
+apply from: '../common.gradle'
 
 if (project.file('local.gradle').exists()) {
     apply from: 'local.gradle'
@@ -36,83 +36,6 @@ repositories {
     flatDir {
         dirs 'libs'
     }
-}
-
-def generateVersionNumber = { ->
-    StringBuilder stringBuilder = new StringBuilder()
-    stringBuilder.append((new Date()).format('yyMMddHHmm'))
-    return Integer.parseInt(stringBuilder.toString())
-}
-
-def generateVersionNumberString = { ->
-    StringBuilder stringBuilder = new StringBuilder()
-    stringBuilder.append((new Date()).format('21yyMMddHH'))
-    return stringBuilder.toString()
-}
-
-def generateTimestamp = { ->
-    StringBuilder stringBuilder = new StringBuilder()
-    stringBuilder.append((new Date()).getTime())
-    stringBuilder.append("L")
-    return stringBuilder.toString()
-}
-
-def generateRandomUUID = { ->
-    StringBuilder stringBuilder = new StringBuilder()
-    stringBuilder.append('"' + UUID.randomUUID().toString() + '"')
-    return stringBuilder.toString()
-}
-
-def generateVersionName = { ->
-
-    StringBuilder stringBuilder = new StringBuilder()
-    try {
-        def stdout = new ByteArrayOutputStream()
-        exec {
-            commandLine 'git', 'describe', '--always'
-            standardOutput = stdout
-        }
-        String commitObject = stdout.toString().trim()
-        try {
-            stdout = new ByteArrayOutputStream()
-            exec {
-                commandLine 'git', 'describe', '--tags'
-                standardOutput = stdout
-            }
-            //stringBuilder.append(stdout.toString().trim())
-            //stringBuilder.append("-")
-        } catch (ignored) {
-            // no tags
-        }
-
-        stringBuilder.append(commitObject)
-        stdout = new ByteArrayOutputStream()
-        exec {
-            commandLine 'git', 'rev-parse', '--show-toplevel'
-            standardOutput = stdout
-        }
-        if (stdout.toString().trim().contains("xDrip-Experimental")) {
-            stringBuilder.append("-experimental")
-        }
-
-        stdout = new ByteArrayOutputStream()
-        exec {
-            commandLine 'git', 'rev-parse', '--abbrev-ref', 'HEAD'
-            standardOutput = stdout
-        }
-        String branch = stdout.toString().trim()
-        if (!branch.equals("master")) {
-            stringBuilder.append('-')
-            stringBuilder.append(branch)
-        }
-
-    } catch (ignored) {
-        return "NoGitSystemAvailable"
-    }
-    stringBuilder.append('-')
-    stringBuilder.append((new Date()).format('yyyy.MM.dd'))
-    return stringBuilder.toString()
-
 }
 
 /**
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -11,7 +11,6 @@ buildscript {
 }
 apply plugin: 'com.android.application'
 apply from: 'token-replace.gradle'
-apply from: '../dependencies.gradle'
 apply from: '../common.gradle'
 
 if (project.file('local.gradle').exists()) {
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -219,6 +219,8 @@ android {
 
     packagingOptions {
         exclude 'META-INF/INDEX.LIST'
+        exclude 'META-INF/NOTICE.md'
+        exclude 'META-INF/LICENSE.md'
     }
 
     compileOptions {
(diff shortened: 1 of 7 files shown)


app/build.gradle
```diff
@@ -302,17 +211,12 @@ dependencies {
     }
 
     implementation project(':localeapi')
-    //implementation 'com.nispok:snackbar:2.10.8'
 
-
-    // use specific version of data binding until google resolves the bugs with mixed case package names
```
jamorham, line 308: When these databinding versions were changed before, it apparently introduced significant compatibility problems with some handsets and those changes were reverted to resolve that issue. Can this java 11 support be added without altering other dependency versions?

tolot27: > When these databinding versions were changed before, it apparently introduced significant compatibility problems with some handsets and those changes were reverted to resolve that issue. 

The version of the databinding library is defined by the Android Gradle Plugin in the root `build.gradle` and is `classpath 'com.android.tools.build:gradle:3.4.3'`. You can check this if you look at the file `META-INF\com.android.databinding_library.version` in the apk file. The implementations I've removed here are never used. Maybe, they were required for former databinding build tools. Currently, you only need to enable databinding and use the correct AGP.

Hence, we are on 3.4.3. for nearly a year (55c54b367) and on 3.4.1 for nearly two years (54e30f255).

>Can this java 11 support be added without altering other dependency versions?

In general, yes. Just cherry pick 4782eef. But I think it is worth to cleanup the files we have and remove old and redundant code. It is also important to fix at least known bugs in the dependencies by updating to patched/minor versions. Any compatibility issues should be avoided. That's why I check any dependency.

Most of the libraries I've updated are just build time libraries (lombok, gradle). In general, I've just upgraded to a patched version (lombok, google-services). Only New Relic is updated to a newer minor version. I've always linked the change log in the corresponding commit (if any) and checked them carefully.

tolot27:
I've merged #1954 into this PR because half of its changes are already included in this PR.

In addition to the upgrade of [setup-java](https://github.com/actions/setup-java) it also upgrades [gradle-build](https://github.com/gradle/gradle-build-action) actions to version 2 and aggregates the Gradle executions into one call and builds only the production release APKs. These changes results in a workflow runtime decrease by about 50% as it can be seen here: https://github.com/NightscoutFoundation/xDrip/actions/runs/1758735602.

BTW: GitHub Actions can be tested locally using [act](https://github.com/nektos/act).

jamorham:
I'm happy with the addition of the java 11 support and functionally identical cleanup etc of the gradle scripts. 

I'm extremely wary of changing dependency versions because of previous experience where even small sub-revision changes can break behaviors.  Just as bugs can be cured in a sub-revision they also can be added in.  I would prefer to do these one at a time and really unless there is a specific issue being addressed I'm not sure how useful it is.

I can cherry pick the java 11 change as a first step and look at the other elements but bundling them in to a single PR makes it a lot harder to test each version so I would prefer not to do that.

tolot27:
Only the New Relic and Google Service library updates can have a potentially negative impact on xDrip's runtime behavior. Is it sufficient for you if I just revert these two updates?
Lombok and Gradle are just build time tools used in hundert thousand of other projects even in much newer versions.

tolot27:
@jamorham 
> Only the New Relic and Google Service library updates can have a potentially negative impact on xDrip's runtime behavior. Is it sufficient for you if I just revert these two updates? Lombok and Gradle are just build time tools used in hundert thousand of other projects even in much newer versions.

Please let me know which libraries I should remove from this PR and whether I should split this PR into different once and if yes which commits should be separated.

jamorham:
I don't want to create additional work for you but if you can provide a version which doesn't change any of the dependency versions then I would expect to approve that. Dependency changes need to be handled very carefully and so I would like to deal with those in a separate PR.

app/build.gradle
jamorham: It is unfortunate that databinding is apparently dependent on the gradle version. I hadn't realised that it wasn't respecting the specific implementation versioning and was silently promoting the version. The commit which fixed the databinding related issues that affected certain handsets by reverting the use of V2 and changing the gradle version is acfa86f240006ee595ec82a4a262e5db8f2ff521 from Aug 2021

tolot27: > It is unfortunate that databinding is apparently dependent on the gradle version.

That is true but it is bundled with the Android Gradle Plugin as clearly stated at https://developer.android.google.cn/jetpack/androidx/releases/databinding:

>The databinding library is bundled with [the Android Gradle plugin](https://developer.android.google.cn/studio/releases/gradle-plugin). You do not need to declare a dependency on the library, but you must enable it.
> To enable data binding, set the dataBinding build option to true in your module's build.gradle file, ...

I've confirmed this by checking `META-INF\com.android.databinding_library.version` in the apk file. If you don't trust, verify it by yourself.

I did not touch `android.databinding.enableV2=false` and kept the AGP at 3.4.3.

tolot27:
> I don't want to create additional work for you but if you can provide a version which doesn't change any of the dependency versions then I would expect to approve that. 

I've removed all dependency updates and [explained](https://github.com/NightscoutFoundation/xDrip/pull/1876#discussion_r809881610) why https://github.com/NightscoutFoundation/xDrip/pull/1876/commits/9017c7485e48c280d0207b823d7ed571481094aa#diff-51a0b488f963eb0be6c6599bf5df497313877cf5bdff3950807373912ac1cdc9L315-L323 is just a clean-up.

build.gradle
```diff
@@ -5,12 +5,12 @@ buildscript {
         google()
         mavenCentral()
         maven { url "https://jitpack.io" }
-
     }
     dependencies {
+        // Upgrading AGP to a newer version also updates the data binding library.
```
jamorham, line 10: I'm not sure this is correct. It builds okay since the pre-processor plugin but the resulting binary doesn't work correctly on some handsets.

jamorham: I believe it is this issue: https://github.com/NightscoutFoundation/xDrip/issues/1796

As far as I know the problem is with the later databinding versions and not anything to do with mixed package names. I believe it only affects specific handsets or perhaps configurations. 

In regards to the tests inside android studio, its only the test filtering which doesn't work correctly if the test package name has mixed case? eg, running all tests works fine. I think this could be quite easy to resolve simply by adjusting the package names (on the first line of the test class) to lower case in the tests only.

tolot27: > In regards to the tests inside android studio, its only the test filtering which doesn't work correctly if the test package name has mixed case? eg, running all tests works fine. I think this could be quite easy to resolve simply by adjusting the package names (on the first line of the test class) to lower case in the tests only.

No, executing all tests fails as well. Renaming just the test package is not possible because test classes are tightly bound to there target classes and their package names must be identical. Otherwise, testing would not be possible the tests have to be completely rewritten and target classes have to be imported individually. Just try to rename the test package. The refactoring preview will show you hundreds of affected files.

tolot27: I'll do the tests and report in #2005. Let us focus on the changes in this PR.

keencave:
Hm, dont want to really jump in here. This PR solved all my problems generating the latest nigthly build. Using a fresh install of Android Studio Bumblebee I got errors which I was unable to solve. Not being a really app expert maybe I am not the only one.

jamorham (approved):
Looks good, thanks

Issue #1062, opened 2019-10-18
Unparseable date when Nightscout has DE_NORMALIZE_DATES enabled


inventor96:
After enabling setting the `DE_NORMALIZE_DATES` config var in Nightscout to `true` (required for proper data upload from OpenAPS 0.6), xDrip+ is unable to download treatments from Nightscout. The error listed in the "Uploaders" section of the System Status in xDrip+ is
```
Unable to do REST API Download
java.text.ParseException:
Unparseable date: "2019-10-18T20:21:22.781+00:00"
Unparseable date: "2019-10-18T20:21:22.781+00:00"
url: https://....herokuapp.com/api/v1/
```

inventor96:
I'm wondering if this is related to #309?

Navid200:
@inventor96 Is this still the case?

inventor96:
Yes, this is still a current issue.

![Screenshot_20210226-020034](https://user-images.githubusercontent.com/7132744/109279243-bea31e80-77d6-11eb-8c06-d988a3e33c54.png)

tolot27:
The reason and possible solutions are documented here: https://stackoverflow.com/a/36798830

tolot27:
@inventor96 What is your time zone and time offset from UTC, -7?

inventor96:
I live in the US Mountain timezone. The fact that the treatment had a timezone other than UTC in the second screenshot could be due to a Nightscout update and/or because I was running OpenAPS 0.7 at that time.

I wouldn't think the timezone difference would be part of the issue, though; the timestamp is still in the same format.

tolot27:
I've fixed this issue and will create a PR soon. The timezone matters during format conversation and if you travel into a different timezone.

tolot27:
@inventor96 Does the latest build fixes the issue for you as well?

inventor96:
It seems to be getting all of the treatments from nightscout, and I haven't seen any errors on the respective status page regarding parsing dates. So it looks like we're good!

tolot27:
Great! Thanks for the feedback.

PR #1879, opened 2021-11-25
fix for Unparseable date format

### LINKED ISSUE #309: Unparseable dateformat usinf Xdrip+ in follower mode with nightscout.
img_20180217_140511
Using xdrip not on uploader phone but in follow mode, I can't sync from Nightscout due to parse errors on date format. 

Is there something I can change in the uploaded phone to become 'dateformat-compliant'?

tolot27:
Change ISO 8601 date string parsing to be able to read Nightscout data if DE_NORMALIZE_DATES is enabled.

A unit test is implemented as well. Test strings and their correction can be seen at https://regex101.com/r/utF21U.

A detailed description is added to the individual commits, especially to 5a30fc2. The unit test is designed to cover also the Android SDK condition check.

Fixes #309, #1062


jamorham:
I think you need to merge the google truth update to the gradle build to resolve the conflicts on this

app/src/main/java/com/eveningoutpost/dexdrip/Models/DateUtil.java
```diff
@@ -1,81 +1,44 @@
...
-        }
-    }
-
-    public static Date tolerantFromISODateString(String isoDateString)
-            throws Exception {
-        try {
-            return fromISODateString(isoDateString.replaceFirst("\\.[0-9][0-9][0-9]Z$", "Z"));
-        } catch (java.text.ParseException e) {
-            return fromISODateString2(isoDateString);
+    public static Date tolerantFromISODateString(String isoDateString) {
```
jamorham, line 35: I haven't examined this in any detail but I'm kind of nervous about modifying this code. Are you confident the new unit tests cover whatever might have been seen and be accepted by this method?

tolot27: I've added an additional test case to cover date/time strings without seconds but offset value (FORMAT_DATE_ISO3). Now all former string formats are covered and the regex fixes much more "wrong" date/time strings. See https://regex101.com/r/utF21U/1 for a clean impression.

jamorham (approved):

PR #1883, opened 2021-11-27
WIP: Set variable max height for insulin graph

vnniyB:
The purpose of this is for Basal pen users.
The insulin graph by default has a maximum value of 11.4
My Lantus + Humalog insulin on board combination spends most of the day above this maximum value. The modification will enable users to set their own max value for their insulin line graph


Navid200:
If this is about "insulin on board", I suggest you edit the above description to clarify.  Right now, your description only states "insulin", which can be misleading.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
```diff
@@ -1653,7 +1654,7 @@ private synchronized void addBgReadingValues(final boolean simple) {
                     final double cobscale = 0.2 * bgScale;
                     final double initial_predicted_bg = predictedbg;
                     final double relaxed_predicted_bg_limit = initial_predicted_bg * 1.20;
-                    final double cob_insulin_max_draw_value = highMark * 1.20;
+                    final double cob_insulin_max_draw_value = Pref.getStringToInt("max_insulin_draw_value", 204);
```
tolot27, line 1657: Why not using `highMark * 1.20` instead of `204`?

app/src/main/res/xml/xdrip_plus_prefs.xml
```diff
@@ -650,6 +650,13 @@
                     android:summary="Plot Basal activity on graph and incorporate in predictive model"
                     android:title="Use Basal Activity"/>
 
+                <EditTextPreference
+                    android:defaultValue="204"
+                    android:inputType="number"
+                    android:key="max_insulin_draw_value"
+                    android:summary="When using Basal injections, increase this value. Default is 204mg/dl or 11.3mmol/L"
```
tolot27, line 657: Please put a space between number and unit and lower case "Basal".

app/src/main/res/xml/xdrip_plus_prefs.xml
```diff
@@ -650,6 +650,13 @@
                     android:summary="Plot Basal activity on graph and incorporate in predictive model"
                     android:title="Use Basal Activity"/>
 
+                <EditTextPreference
+                    android:defaultValue="204"
```
tolot27, line 654: This value is static and confusing for "mmol/L" users.

tolot27 (changes requested):
In general, the graph height depends on the high threshold (default 170) which is multiplied by 1.2). Hence, this PR removes the dependency to the high threshold (expect for the default setting), which should be noted in the PR description.

IMHO, a better approach would be increasing the graph height based on the max BG value within the view area (mostly 24 hours). That would avoid the introduction of an additional preference setting.

tolot27:
Please add some screenshots to your PR description.

PR #1885, opened 2021-12-02
DexShareFollow service to support trend as a number and as a string

### LINKED ISSUE #1884: Dex Share follower does not receive data (with older EU account)
**Subject of the issue**
Dex Share follower (using xDrip+) doesn't receive any data (worked fine for many months until November 30th). 

**Your environment**
I don't believe this is related to my hardware/settings (Android 11). It worked fine until November 30th. No changes made to hardware and Android OS.
Software: xDrip+ using Dex Share follower
xDrip+ version is: 9ff5104-2021.11.23  (has been updated today because of the described problem; previous version showed the same problem)
xDrip+ is used by me (= follower of my daughter) using "Dex Share Follower" as data source
The data is uploaded by the Dexcom G6 App (used by my daughter).

**Expected behavior**
Dex Share follower (using xDrip+) should receive data from Dex share servers.

**Actual behavior**
No data is received by xDrip+. Error message in the log file:
"Get Share Data Failed: java.lang.NumberFormatException: For input string: "Flat"
The Dexcom follow app functions normally.

**Steps to reproduce the behavior:**
Set xDrip+ to Dex Share follower with matching credentials and server.

tolot27:
On 1 December 2021, Dexcom changed their EU servers to serve the trend arrow in text representation (i. e. "Flat") instead of the former numerical representation (i. e. 4 [= flat]). This PR supports both representations.

Fixes #1884

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/ImportedLibraries/dexcom/Dex_Constants.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/ImportedLibraries/dexcom/Dex_Constants.java
@@ -93,7 +93,7 @@ public class Dex_Constants {
         SINGLE_DOWN(6,"\u2193", "SingleDown", -2d),
         DOUBLE_DOWN(7,"\u21CA", "DoubleDown", -3.5d),
         NOT_COMPUTABLE(8, "", "NotComputable"),
-        OUT_OF_RANGE(9, "", "OutOfRange");
+        OUT_OF_RANGE(9, "", "RateOutOfRange");
 
         private String arrowSymbol;
         private String trendName;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/ImportedLibraries/dexcom/Dex_Constants.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/ImportedLibraries/dexcom/Dex_Constants.java
@@ -92,8 +92,8 @@ public class Dex_Constants {
         DOWN_45(5,"\u2198", "FortyFiveDown", -1d),
         SINGLE_DOWN(6,"\u2193", "SingleDown", -2d),
         DOUBLE_DOWN(7,"\u21CA", "DoubleDown", -3.5d),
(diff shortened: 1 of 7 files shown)


tolot27:
The various (Dex_)Constant.java files are a nightmare. We should fix this. To not forget this, I've created a [task](https://github.com/NightscoutFoundation/xDrip/projects/2#card-74029052) in the project plan.

app/src/main/java/com/eveningoutpost/dexdrip/ImportedLibraries/dexcom/Dex_Constants.java
```diff
@@ -92,8 +92,8 @@
         DOWN_45(5,"\u2198", "FortyFiveDown", -1d),
         SINGLE_DOWN(6,"\u2193", "SingleDown", -2d),
         DOUBLE_DOWN(7,"\u21CA", "DoubleDown", -3.5d),
-        NOT_COMPUTABLE(8, "", "NOT_COMPUTABLE"),
-        OUT_OF_RANGE(9, "", "OUT_OF_RANGE");
+        NOT_COMPUTABLE(8, "", "NotComputable"),
```
jamorham, line 95: Will these changes break things like `isSlopeNameInvalid()` and other string comparisons vs the friendly name?

tolot27: No, because `isSlopeNameInvalid()` [is part of](https://github.com/NightscoutFoundation/xDrip/blob/a439093a021b2183676290fbe3e726841c109372/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java#L706) the `BgReading` class and still does not use the `TREND_ARROW_VALUES` enum.

jamorham (approved):
LGTM

anglosid:
It seems that this is the reason why my XDrip+ is down for 5 days. When it is fixed, will it be an automatic update or do I need to do something? Thanks!

PR #1887, opened 2021-12-04
Tzachi libre no calibrations

tzachi-dar:
Allow libre 2 to work without calibrations. 
By default, users still have to calibrate. After installation, users can activate the new feature.
No issues are expected with OOP1 or bluecon users.

Still needs testing.
To try the new version please use:
https://drive.google.com/uc?id=1vqFnHulnXXBYggydqTv8KYY8NVZ9T3Ur&export=download

A new version of oop2 is also needed:
https://drive.google.com/uc?id=1aeWvBiFCidOhreCMoW1S0ARa5HvNNf2m&export=download

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -938,12 +938,12 @@ public class BgReading extends Model implements ShareUploadableBg {
                 .execute();
     }
 
-    public static BgReading readingNearTimeStamp(double startTime) {
-        double margin = (4 * 60 * 1000);
+    public static BgReading readingNearTimeStamp(long startTime) {
+        long margin = (4 * 60 * 1000);
         return readingNearTimeStamp(startTime, margin);
     }
 
-    public static BgReading readingNearTimeStamp(double startTime, final double margin) {
+    public static BgReading readingNearTimeStamp(long startTime, final long margin) {
         final DecimalFormat df = new DecimalFormat("#");
         df.setMaximumFractionDigits(1);
         return new Select()
@@ -1293,7 +1293,7 @@ public class BgReading extends Model implements ShareUploadableBg {
     }
 
     // TODO this method shares some code with above.. merge
-    public static void bgReadingInsertFromInt(int value, long timestamp, double margin, boolean do_notification) {
+    public static void bgReadingInsertFromInt(int value, long timestamp, long margin, boolean do_notification) {
         // TODO sanity check data!
 
         if ((value <= 0) || (timestamp <= 0)) {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -939,7 +939,11 @@ public class BgReading extends Model implements ShareUploadableBg {
     }
 
     public static BgReading readingNearTimeStamp(double startTime) {
-        final double margin = (4 * 60 * 1000);
+        double margin = (4 * 60 * 1000);
+        return readingNearTimeStamp(startTime, margin);
+    }
+
+    public static BgReading readingNearTimeStamp(double startTime, final double margin) {
         final DecimalFormat df = new DecimalFormat("#");
         df.setMaximumFractionDigits(1);
         return new Select()
@@ -1289,7 +1293,7 @@ public class BgReading extends Model implements ShareUploadableBg {
     }
 
     // TODO this method shares some code with above.. merge
-    public static void bgReadingInsertFromInt(int value, long timestamp, boolean do_notification) {
+    public static void bgReadingInsertFromInt(int value, long timestamp, double margin, boolean do_notification) {
         // TODO sanity check data!
 
         if ((value <= 0) || (timestamp <= 0)) {
@@ -1319,11 +1323,11 @@ public class BgReading extends Model implements ShareUploadableBg {
             }
 
             try {
-                if (readingNearTimeStamp(bgr.timestamp) == null) {
+                if (readingNearTimeStamp(bgr.timestamp, margin) == null) {
                     bgr.save();
                     bgr.find_slope();
                     if (do_notification) {
-                       // xdrip.getAppContext().startService(new Intent(xdrip.getAppContext(), Notifications.class)); // alerts et al
+                        // xdrip.getAppContext().startService(new Intent(xdrip.getAppContext(), Notifications.class)); // alerts et al
                         Notifications.start(); // this may not be needed as it is duplicated in handleNewBgReading
                     }
                     BgSendQueue.handleNewBgReading(bgr, "create", xdrip.getAppContext(), false, !do_notification); // pebble and widget
@@ -1334,7 +1338,7 @@ public class BgReading extends Model implements ShareUploadableBg {
                 Log.e(TAG, "Could not save BGR: ", e);
             }
         } else {
-            Log.e(TAG,"Got null bgr from create");
+            Log.e(TAG, "Got null bgr from create");
         }
     }
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreOOPAlgorithm.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreOOPAlgorithm.java
@@ -14,9 +14,7 @@ import com.eveningoutpost.dexdrip.UtilityModels.Constants;
 import com.eveningoutpost.dexdrip.UtilityModels.Intents;
 import com.eveningoutpost.dexdrip.UtilityModels.LibreUtils;
 import com.eveningoutpost.dexdrip.UtilityModels.PersistentStore;
-import com.eveningoutpost.dexdrip.UtilityModels.Pref;
 import com.eveningoutpost.dexdrip.xdrip;
-import com.eveningoutpost.dexdrip.NFCReaderX;
 
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.TimeUnit;
@@ -343,7 +341,7 @@ public class LibreOOPAlgorithm {
         String SensorSN = LibreUtils.decodeSerialNumberKey(patchUid);
 
         Log.e(TAG, "handleDecodedBleResult Created the following object " + readingData.toString());
-        NFCReaderX.SendLibrereadingToFollowers(SensorSN, readingData.raw_data, timestamp, patchUid, null);
+        NFCReaderX.sendLibrereadingToFollowers(SensorSN, readingData.raw_data, timestamp, patchUid, null);
         boolean bg_val_exists = trend_bg_vals != null && history_bg_vals != null;
         LibreAlarmReceiver.processReadingDataTransferObject(readingData, timestamp, SensorSN, true /*=allowupload*/, patchUid, null/*=patchInfo*/, bg_val_exists);
     }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreOOPAlgorithm.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreOOPAlgorithm.java
@@ -249,24 +249,16 @@ public class LibreOOPAlgorithm {
             Log.e(TAG, "HandleData cought exception ", e);
             return;
         }
-        boolean use_raw = Pref.getBooleanDefaultFalse("calibrate_external_libre_algorithm");
         ReadingData readingData = new ReadingData();
 
         readingData.trend = new ArrayList<GlucoseData>();
-
-        double factor = 1;
-        if (use_raw) {
-            // When handeling raw, data is expected to be bigger in a factor of 1000 and 
-            // is then devided by Constants.LIBRE_MULTIPLIER
-            factor = 1000 / Constants.LIBRE_MULTIPLIER;
-        }
 
         // Add the first object, that is the current time
         GlucoseData glucoseData = new GlucoseData();
         glucoseData.sensorTime = oOPResults.currentTime;
         glucoseData.realDate = oOPResults.timestamp;
-        glucoseData.glucoseLevel = (int) (oOPResults.currentBg * factor);
-        glucoseData.glucoseLevelRaw = (int) (oOPResults.currentBg * factor);
+        glucoseData.glucoseLevel = (int) (oOPResults.currentBg);
+        glucoseData.glucoseLevelRaw = (int) (oOPResults.currentBg);
 
         verifyTime(glucoseData.sensorTime, "LibreOOPAlgorithm", null);
         readingData.trend.add(glucoseData);
@@ -280,8 +272,8 @@ public class LibreOOPAlgorithm {
             if (historicBg.quality == 0) {
                 glucoseData = new GlucoseData();
                 glucoseData.realDate = oOPResults.timestamp + (historicBg.time - oOPResults.currentTime) * 60000;
-                glucoseData.glucoseLevel = (int) (historicBg.bg * factor);
-                glucoseData.glucoseLevelRaw = (int) (historicBg.bg * factor);
+                glucoseData.glucoseLevel = (int) (historicBg.bg );
+                glucoseData.glucoseLevelRaw = (int) (historicBg.bg);
                 readingData.history.add(glucoseData);
             }
         }
@@ -290,12 +282,12 @@ public class LibreOOPAlgorithm {
         // TODO: Base this on real BG values.
         glucoseData = new GlucoseData();
         glucoseData.realDate = oOPResults.timestamp;
-        glucoseData.glucoseLevel = (int) (oOPResults.currentBg * factor);
-        glucoseData.glucoseLevelRaw = (int) (oOPResults.currentBg * factor);
+        glucoseData.glucoseLevel = (int) (oOPResults.currentBg);
+        glucoseData.glucoseLevelRaw = (int) (oOPResults.currentBg);
         readingData.history.add(glucoseData);
 
         Log.d(TAG, "handleData Created the following object " + readingData.toString());
-        LibreAlarmReceiver.CalculateFromDataTransferObject(readingData, false, use_raw);
+        LibreAlarmReceiver.CalculateFromDataTransferObject(readingData, false, false);
     }
 
     public static SensorType getSensorType(byte[] SensorInfo) {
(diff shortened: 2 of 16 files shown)


FuLiDu:
Hi, thanks for sharing.
Installed it and working correctly.
I do not understand though what did you mean by "no calibrations".
I had to to two initial calibrations as usual. 
Thanks for your time.

tzachi-dar:
There is a new key called "Apply xDrip calibration methods on OOP2 data."
set it to false, and you will not be asked for calibrations.

Lore0033:
Hi tzachi-dar, I've subscribed only now to this platform but the help you've been giving to me with your explanation has been immense. But now I can't solve a problem. I recently switched to Libre 2, and I can't find a way how to send values to xdrip+ and calibrate them into xdrip. I have a MiaoMiao 2, and I've been using it together with OOP1, it works pretty well but values are not too precise. So, I decided to try OOP2, which, if I understood well, allows calibration, but I can't find a way how to make it work... I started the libre sensor with librelink app, then I uninstall it and tried to nfc scan with xdrip+, but nothing happened ("not valid nfc values" message appeared). Since I'm not able to patch the librelink app, is there a method which could enable me to calibrate my libre values in xdrip, with or without using MM2? Thank you so much 
Lore

tzachi-dar:
@Lore0033 Did you follow the instruction from https://www.minimallooper.com/post/how-to-setup-freestyle-libre-2-and-oop2-to-use-a-native-bluetooth-connection-in-xdrip?

Can you please write what the exact message that you are getting? I can't seem to find the message "not valid nfc datas" in the code.

Lore0033:
@tzachi-dar yes, I've followed that instructions, I only didn't fully reset xdrip because I was afraid to lose my datas but I activated or disactivated every settingtthe guideline mentioned. As for the message, it actually appears in Italian as it is the language of my phone: in Italian, after "Scansione in corso" (scanning), appears "Dati NFC non validi - riprovare" (Invalid NFC data - try again), and then "Funzione Any-tag disattivata" (Any - tag function disabled).

tzachi-dar:
@Lore0033 the issue that you are seeing is only related to wrong nfc settings.
Please make sure that your settings look like:
Screenshot_20211220-231713_xDrip+

tolot27:
Please continue support discussions somewhere else. This is a pull request and only for developers.

app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreOOPAlgorithm.java
```diff
@@ -183,277 +184,285 @@ static public  UnlockBuffers sendGetBluetoothEnablePayload(boolean increaseConne
...
-        glucoseData.glucoseLevel = (int)(oOPResults.currentBg * factor);
-        glucoseData.glucoseLevelRaw = (int)(oOPResults.currentBg * factor);
+        glucoseData.glucoseLevel = (int) (oOPResults.currentBg);
+        glucoseData.glucoseLevelRaw = (int) (oOPResults.currentBg);
         readingData.history.add(glucoseData);
-        
+
         Log.d(TAG, "handleData Created the following object " + readingData.toString());
-        LibreAlarmReceiver.CalculateFromDataTransferObject(readingData, false, use_raw);
+        LibreAlarmReceiver.CalculateFromDataTransferObject(readingData, false, false);
```
tolot27, line 290: Is the parameter `use_raw` still in use, elsewhere, if you set it to a constant `false` here? If not, please update the function definition.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -263,38 +268,39 @@ public static void SendLibrereading(final String tagId, byte[] data1, final long
...
+        if (LibreOOPAlgorithm.isDecodeableData(patchInfo) && decripted_data == false
                 && !Pref.getBooleanDefaultFalse("external_blukon_algorithm")) {
             // Send to OOP2 for decryption.
             LibreOOPAlgorithm.logIfOOP2NotAlive();
             LibreOOPAlgorithm.sendData(data1, CaptureDateTime, patchUid, patchInfo, tagId);
             return true;
         }
 
-        SendLibrereading(tagId, data1, CaptureDateTime, patchUid, patchInfo);
+        SendLibrereadingToFollowers(tagId, data1, CaptureDateTime, patchUid, patchInfo);
```
tolot27, line 303: If you touch this function name, please change it to full camel-case except first character like `sendLibreReadingsToFollowers'. See [Java Naming Conventions](https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html) for details.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -831,24 +839,39 @@ public static ReadingData parseData(int attempt, String tagId, byte[] data, Long
...
 
             glucoseData.realDate = sensorStartTime + time * MINUTE;
             glucoseData.sensorTime = time;
-            if(verifyTime(time, "parseData trendList", data)) {
+            if (verifyTime(time, "parseData trendList", data)) {
                 trendList.add(glucoseData);
             }
         }
 
+        Collections.sort(trendList);
```
tolot27, line 860: How often is this called? Maybe a treeset is a suitable data structure, if sorting occurs frequently.

tolot27 (changes requested):
I really like the improvements in code documentation and formatting. Thanks!

Please remove all the translated files. Translation changes must be done at Crowdin. They will by synced only unidirectional. I can replace OOP by OOP2 there, if you like.

Please document any potential backward compatibility breaks like with L1 or Bluekon.

Lore0033:
@tolot27 oh sorry, I just subscribed to this platform and I didn't know that this wasn't the right place to ask for support... @tzachi-dar, thank you so much for your help, now everything seems to work well, but OOP2 currently works for me only if I switch on "Use OOP algorithm" in xdrip, instead the guideline said to switch it off. Anyway, now I can calibrate! Thank you so much !

app/src/main/java/com/eveningoutpost/dexdrip/Models/ReadingData.java
```diff
@@ -212,11 +214,18 @@ private boolean calculateSmoothDataPerPoint(GlucoseData glucoseData, List<LibreT
                 continue;
             }
             sum += libreTrendPoint.rawSensorValue;
-            Log.d(TAG, "Using  point for some" + libreTrendPoint);
             points_used++;
+            Log.d(TAG, "Using  point for some" + libreTrendPoint);
```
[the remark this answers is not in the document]
tzachi-dar: Done.

app/src/main/java/com/eveningoutpost/dexdrip/NSEmulatorReceiver.java
```diff
@@ -247,83 +248,116 @@ private JSONObject extractParams(final Bundle bundle) {
...
+                for (int i = 0; i < computed_bg.length(); i++) {
+                    trend_bg_vals[i] = computed_bg.getInt(i);
+                }
+                computed_bg = json_object.getJSONArray(Intents.HISTORIC_BG);
+                history_bg_vals = new int[computed_bg.length()];
+                for (int i = 0; i < computed_bg.length(); i++) {
+                    history_bg_vals[i] = computed_bg.getInt(i);
+                }
+            }
+
```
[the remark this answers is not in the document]
tzachi-dar: It can be factored to one functoin, but it will have a boolean saying weather this is FRAM or BLE and then it will have 3-4 if branches testing what option is involved. I think that seperating the two pathes makes the code easier to read as each is handeling its own intent.

app/src/main/res/xml/pref_advanced_settings.xml
```diff
@@ -1471,12 +1471,10 @@
                     android:title="@string/use_external_blukon_algorithm" />
 
                 <CheckBoxPreference
-                    android:defaultValue="false"
-                    android:dependency="external_blukon_algorithm"
-                    android:enabled="false"
-                    android:key="calibrate_external_libre_algorithm"
-                    android:summary="@string/calibrate_external_libre_algorithm_summary"
-                    android:title="@string/calibrate_external_libre_algorithm_title" />
+                    android:defaultValue="true"
```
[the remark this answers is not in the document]
tzachi-dar: Before my change, calibrating was mandatory, so nothing is realy changing until the user will activate the new feature.
I have also updated the PR description.

wear/src/main/java/com/eveningoutpost/dexdrip/Models/LibreOOPAlgorithm.java
```diff
@@ -92,24 +92,16 @@ static public void HandleData(String oopData) {
             Log.e(TAG, "HandleData cought exception ", e);
             return;
         }
-        boolean use_raw = Pref.getBooleanDefaultFalse("calibrate_external_libre_algorithm");
```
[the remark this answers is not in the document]
tzachi-dar: This was the oop1 calibration option. (it does not have 2 in it). Indeed for OOP1 it was not possible to calibrate.
As far as I can see it does not exist any more.

wear/src/main/java/com/eveningoutpost/dexdrip/Models/LibreOOPAlgorithm.java
```diff
@@ -92,24 +92,16 @@ static public void HandleData(String oopData) {
...
         ReadingData.TransferObject libreAlarmObject = new ReadingData.TransferObject();
         libreAlarmObject.data = new ReadingData();
         libreAlarmObject.data.trend = new ArrayList<GlucoseData>();
         
-        double factor = 1;
```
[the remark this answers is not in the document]
tzachi-dar: On OOP1 there was no way to activate calibrations, so use_raw was always false, the factor was always 1, and this PR changes nothing.

app/src/main/java/com/eveningoutpost/dexdrip/Models/LibreOOPAlgorithm.java
tzachi-dar: On line 206 of LibreAlarmReceiver.java in this pr (Sorry, I did not find a way to create a link there) it might be called with true.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
tzachi-dar: Done.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
tzachi-dar: This is only happening per reading, so probably only once in 5 minutes (or once in a minute, depend on sensor type). In any case this lists are short (around 15 items in list) and complexity of the sort is o(n log n). Inseting n objects to a tree is also n log n complexity, so I don't think that we can gain anything from it.

tzachi-dar:
> Please remove all the translated files. Translation changes must be done at Crowdin. They will by synced only unidirectional. I can replace OOP by OOP2 there, if you like.

We had an old key called calibrate_external_libre_algorithm. This was disabled by defaut since it wasn't good enough.
As it is clear now that it will never be fixed, I'm removing it.
I'm now adding a new key called calibrate_external_libre_2_algorithm. Since they are very close, I thought that I would change that also in all translations. If you fill it will break something, please let me know, so I'll revert this and translaters will do this in Crowdin.

tolot27:
> > Please remove all the translated files. Translation changes must be done at Crowdin. They will by synced only unidirectional. I can replace OOP by OOP2 there, if you like.
> 
> We had an old key called calibrate_external_libre_algorithm. This was disabled by defaut since it wasn't good enough. As it is clear now that it will never be fixed, I'm removing it. I'm now adding a new key called calibrate_external_libre_2_algorithm. Since they are very close, I thought that I would change that also in all translations. If you fill it will break something, please let me know, so I'll revert this and translaters will do this in Crowdin.

At Crowdin it is not enabled to fetch translated files from the repository. They will be pushed only. Hence, to not loose the translations it is best to not rename `calibrate_external_libre_algorithm_summary` and `calibrate_external_libre_algorithm_title`. Please revert this renaming and remove the translated xml files. After merging this PR I'll replace "OOP" with "OOP2" in these two strings at all translations on Crowdin.

app/src/main/java/com/eveningoutpost/dexdrip/NSEmulatorReceiver.java
```diff
@@ -247,83 +248,116 @@ private JSONObject extractParams(final Bundle bundle) {
...
             return;
         }
         JSONObject json_object = extractParams(bundle);
-        if(json_object == null) {
+        if (json_object == null) {
             return;
         }
         String decoded_buffer;
         String patchUidString;
+        JSONArray computed_bg;
```
tolot27, line 316: No need to declare it here. It can be declared at line 327. In fact, it is identical to the corresponding part in `handleOop2DecodeFramResult`.

app/src/main/java/com/eveningoutpost/dexdrip/NSEmulatorReceiver.java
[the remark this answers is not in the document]
tolot27: Sorry, I cannot follow you. I've harmonized the two functions and a diff looks like:
image

As you can see, the inner part is completely identical. Extracting this into a private sub function increases the readability and maintainability quite a lot, IMHO.
[handleOop2DecodeFramResult.txt](https://github.com/NightscoutFoundation/xDrip/files/8036896/handleOop2DecodeFramResult.txt)
[handleOop2BluetoothEnableResult.txt](https://github.com/NightscoutFoundation/xDrip/files/8036897/handleOop2BluetoothEnableResult.txt)

tolot27 (changes requested):
Still some part to improve by removing code duplication.

app/src/main/java/com/eveningoutpost/dexdrip/NSEmulatorReceiver.java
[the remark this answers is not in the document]
tolot27: After thinking about a little more, these two functions can be merged into a single one and testing for `json_object.has(Intents.TAG_ID)` can distinguish the different handling.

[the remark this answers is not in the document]
tzachi-dar: I do not agree that using one function for both is a good idea.
Right now there are two code flows: one for using ble data and one for using fram data.
Both are sent to OOP2 using different functions.
The code that handles them is in a function called handleOop2DecodeFramResult (for example). it starts with nsemulatorreciver where the intent is being handled, and continues to oop2 where data is handled.
There is a similar code for BLE.
Everyone reading the code can see that there is a path for BLE and a path for FRAM and it is easy to follow.
Now, if I would create one function from both of them 3 things will happen:

1. The code will not be clear. For example the name of the function will not tell if it is BT or FRAM. 
2. Drawing a code graph will show that the BT data might be called from a FRAM intent.  and vice versa.
3. Since one function has to handle an extra parameter, there will be ifs in the code to say if the parameter should exist or should not exist. The code becomes harder to be understood.

One more point to understand is that in no case we can use json_object.has(Intents.TAG_ID) to say if this should be a BLE data or FRAM data. Doing so, will cause the code to be very fragile. We will be using some unrelated parameter to decide if this is BT or FRAM. For example, In the case of a change in OOP2 we might have the code treat BLE data as FRAM data or vice versa. We can pass a boolean to make the decision more clear, but this will look ugly.

Bottom line, I don't file that removing 30 lines of code is worth the effort.

jbr7rr:
Hi @tzachi-dar 

I have tested your branch/PR "tzachi-libre-no-calibrations" (pulled it in a branch up to date with current master), and noticed that the BG data gets updated every 4 minutes when no calibrations are used instead of 5 minutes when calibrations are used. This results in some weird graphic things in xDrip, but also results in more readings then expected ( > 288 per day).
I've traced this error down in the code.

readingNearTimeStamp is used in function BgReading.bgReadingInsertFromInt() which is used with !use_raw data. and has a resolution/timeslice fixed at 4 minutes
for raw_data createBGfromGD is used, which in turn uses getForPreciseTimestamp and sets the resolution/timeslice to 4.5 minutes

So far i see 2 solutions to this problem:
- set the margin for readings to 4.5 minutes instead of 4 (like "segmentation_timeslice" which is also 4.5)) see my commit: https://github.com/jbr7rr/xDrip/commit/6003d323d485fb4b289d0eac5ad447eff4ca6489
- Use getForPreciseTimestamp instead of readingNearTimeStamp. see my commit: https://github.com/jbr7rr/xDrip/commit/5cf866e4cf05c7d3af675ca65be87d4f3d18f12a

I have tested both solutions independently, and both seem to fix the problem

tzachi-dar:
Thanks @jbr7rr.
I'll take a deeper look at what is the best way to fix this.

tzachi-dar:
Thanks again @jbr7rr I ended doing something similar to what you have suggested.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -939,7 +939,11 @@ public static boolean isDataStale() {
     }
 
     public static BgReading readingNearTimeStamp(double startTime) {
-        final double margin = (4 * 60 * 1000);
+        double margin = (4 * 60 * 1000);
+        return readingNearTimeStamp(startTime, margin);
+    }
+
+    public static BgReading readingNearTimeStamp(double startTime, final double margin) {
```
jamorham, line 946: FYI there is: `getForPreciseTimestamp` but it has slightly different search criteria so I'm not certain if it is what you need here. `readingNearTimeStamp` should probably at some point also be refactored to use `long` instead of `double`

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -1289,7 +1293,7 @@ public static BgReading bgReadingInsertFromJson(String json, boolean do_notifica
     }
 
     // TODO this method shares some code with above.. merge
-    public static void bgReadingInsertFromInt(int value, long timestamp, boolean do_notification) {
+    public static void bgReadingInsertFromInt(int value, long timestamp, double margin, boolean do_notification) {
```
jamorham, line 1296: I'm slightly concerned about adding more use of `double` to represent `milliseconds`

jamorham:
I made a couple of very small comments. I assume that sufficient testing has been done that this is known to be working. Otherwise it looks good to me.

tzachi-dar:
I have moved readingNearTimeStamp to use long. I hope this is what is you meant. If not, please let me know.

jamorham:
Yes, please just check that it all still works okay. I'm pretty sure it should but its been like that for a while..

PR #1899, opened 2022-01-03
execute unit tests just once

tolot27:
Currently, unit tests are executed for every combination of build type and product flavor. This results in unnecessary recurring tests and elongates the CI workflows significantly.
With this PR only the unit tests of the production release builds are executed but after the compile tests. This reduces the total workflow time by 65 % from about 15m 45s to about 6m 40s.

Take the wasted CPU time and the CO2 impact into account! :smile:

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -19,9 +19,9 @@ jobs:
         java-version: 1.8
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
-    - name: Run Unit Tests
-      run: ./gradlew :app:testProdReleaseUnitTest :wear:testProdReleaseUnitTest
     - name: Run Compile Tests
       run: ./gradlew assembleRelease
+    - name: Run Unit Tests
+      run: ./gradlew :app:testProdReleaseUnitTest :wear:testProdReleaseUnitTest
     - name: Check Output
       run: bash ./etc/CheckBuild/check-release.sh test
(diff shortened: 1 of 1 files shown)


jamorham:
I agree that there are unnecessary unit tests being run because of the different flavors. But do we need to also do unit tests for the production release flavors because of differences with how proguard rules are applied in that we must make sure unit tests have been done on the actual version which will be installed by users?

tolot27:
> But do we need to also do unit tests for the production release flavors because of differences with how proguard rules are applied in that we must make sure unit tests have been done on the actual version which will be installed by users?

We need the debug unit tests because typically it gives more detailed output and ProGuard likely removes/shrinks code in the release builds which is only necessary for testing (see [Configuring ProGuard for Android Instrumentation Tests](https://medium.com/wix-engineering/configuring-proguard-for-android-instrumentation-tests-120d58cba155) for instance). 

The unit tests are designed to test specific part/logic of the code; they do not test the whole application (yet).

But I'll review the ProGuard rules in the context of the unit/instrumentation tests again.

jamorham:
I think we definitely need the release versions also tested because proguard could remove something that breaks something and there can be subtle issues created which we want to expose via unit tests. There are also robolectric integration style tests which can utilize the database and particularly this can be a gotcha where proguard can strip unused fields from object mapping systems so its essential to include the most stripped flavor in the unit testing in my opinion.

tolot27:
I've changed it to use release builds.

jamorham (approved):

PR #1933, opened 2022-01-13
Update stop sensor warning

Navid200:
This is what the user sees after tapping on stop sensor:

This is confusing for someone who wants to restart a G6.  There was a post on facebook asking if they could stop sensor to restart considering there is a warning not to.
The fact is that you can stop sensor any time in order to do a restart.

This is what the user will see after this PR if they tap on stop sensor:

@tzachi-dar 
Would this change make the warning confusing for a Libre user?

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -901,7 +901,7 @@
     <string name="bluetooth_scan">Bluetooth Scan</string>
     <string name="level_alerts">Level Alerts</string>
     <string name="override_calibration">Override Calibration</string>
-    <string name="only_stop_your_sensor_when_you_actually_plan">Only stop your sensor when you actually plan to remove it, otherwise leave it running!</string>
+    <string name="only_stop_your_sensor_when_you_actually_plan">Only stop sensor if you actually intend to end the session. Otherwise, leave it running!</string>
     <string name="restart_collector">Restart Collector</string>
     <string name="forget_device">Forget Device</string>
     <string name="other_notes">Other Notes:</string>


tolot27:
I believe "session" is confusing for many users.

tolot27:
> Can you think of something that would be less confusing?

Not yet because I don't understand what the confusion is? If it is about how to restart a sensor, your proposal is not less confusing either.

The majority of users don't want to restart a sensor, the actually intend to replace it by a new one.

Maybe a discussion about your proposal should be created?

Navid200:
Read the current warning.  It says the only time you should stop sensor is if you want to remove it.
But in fact, you can stop a sensor even if you don't intend to remove it.  For example, if you want to restart.
That's why the warning is incorrect and needs to be changed.

Navid200:
If you look at the G6 user's guide, you will see the word "session" used 65 times.
The following shows it on the cover.
Screenshot 2022-01-14 124111

I think we need to change the warning.  
I don't insist that my PR offers the best alternative.  I am open to suggestions.  But, I hope that the point is clear that the existing warning is wrong and should be replaced.

Navid200:
> The majority of users don't want to restart a sensor, the actually intend to replace it by a new one.
> 

Would you mind stating what that statement is based on?
Have we got a poll from xDrip users?

Even if that was true, why would that matter?
If someone suggests to add translation for a language that is not currently included in Crowdin, should we tell them that the majority of people don't speak that language?

tolot27:
You are the master and you decide. Be happy.

Navid200:
> You are the master and you decide. Be happy.

Can I remove the type-discussion label from this PR then?

tzachi-dar:
I think that putting "End the session" will be confusing for libre users.
It should probably something like 
"only stop the sensor in the following cases:
1) you intend to remove it.
2) You intend to reuse the sensor for 10 more days. (G6 only)".

Navid200:
@tzachi-dar There are times we also restart a G5.
I don't know about a G4.
So, a Libre is never restarted?

How about this?
Only stop sensor if you intend to remove or restart it.  Otherwise, leave it running.

Navid200:
@tzachi-dar If someone wants to try both patched and Libre 2, shouldn't they stop before switching to the other mode?

tzachi-dar:
I guess that they better change that, but I'm not sure that this is something that happens too much times.
Adding this sentence will confuse anyone else.

Navid200:
The existing message discourages users from attempting to restart.
That is the right thing to do.

If someone chooses to restart, they need to know what they are doing, specially the impact of a restart on accuracy and factory calibration due to the effects of immune system on the sensor filament.   The existing warning should not be an issue for someone who chooses to restart knowing the risks and facts.

Issue #1890, opened 2021-12-10
G6 On new sensor start, data with timestamp before new sensor was started is sent to NS and AAPS


andyrozman:
### Subject of the issue
When new G6 sensor is started, old data (data accumulated before new sensor was started) is being added to NS and send over broadcast. This seems like similar issue to https://github.com/NightscoutFoundation/xDrip/issues/1778 and solution might be the same. Once new sensor is started, all data from before should be ignored and dropped. Is this something that could be easily implemented? 

### Expected behavior
Data before sensor start is dropped.

### Actual behavior
When sensor starts all data is sent to NS and broadcast.

### Steps to reproduce the behavior:
Start new G6 sensor and look at data

### Additional notes
So new version of AAPS (3.0 - current betas), stops Loop as soon is it receives such data (new BgQualityPlugin). This is why all AAPS users are being warned not to use Xdrip anymore. If this could be fixed sooner, rather the later, warning about Xdrip probably won't go in release...

Navid200:
When you attempted to open this issue, you saw a request that you consult with us before opening the issue.
Did you not see that note?

Would you please not use these issues as if this was a forum?

Navid200:
As your questions like "is the solution the same as an existing issue", or "is this something that can be easily implemented" in the discussions:
https://github.com/Navid200/xDrip/discussions

I assure you having an open issue does not expedite resolving the issue.  It only clutters this platform. 
Please work with us so that we can efficiently maintain xDrip.  We need your help.
Please pay attention to the requests we make (notes you saw when you opened this issue).

andyrozman:
I did ask on Gitter, but there was no response... It is true I didn't have all the details yet at that point, but this is valid issue, that will soon affect a lot of users of Xdrip (everybody that is using AndroidAPS that is).

Navid200:
@andyrozman That may be the case.
But, if the cause is already included in an existing open issue, there is no reason to open another issue.
You can open a discussion and bring it to our attention.
We can label an existing issue and increase its priority if necessary.
But, having redundant issues doesn't help.

And if have a conversation, in discussions, and it turns out that the matter needs a dedicated issue, we can then open an issue.

We are just asking "please don't pull the trigger and ask questions later".
There are too many closed issues and people refuse to search them.  All because of so many unnecessary issues (some redundant, some just user error) that have been opened.

If you open a discussions, we can talk.

savek-cc:
[...]

PR #1955, opened 2022-01-26
G6 backfill prior to start AAPS fix copied to wear

Navid200:
I have just copied the mods (f98d57b and 6f7c366) to the corresponding wear files.

I have no way of testing this as I have no Android wear device.

Please feel free to edit, or tell me to edit.

Fixes https://github.com/NightscoutFoundation/xDrip/issues/1890

### DIFF
--- a/wear/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
+++ b/wear/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
@@ -762,6 +762,12 @@ public class Ob1G5StateMachine {
     }
 
     private static void backFillIfNeeded(Ob1G5CollectionService parent, RxBleConnection connection) {
+        final Sensor sensor = Sensor.currentSensor();
+        if (sensor == null) {
+            UserError.Log.e(TAG, "Cannot process backfill evaluation as no active sensor");
+            return;
+        }
+
         final int check_readings = nextBackFillCheckSize;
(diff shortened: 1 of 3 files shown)


tolot27:
@Navid200 Thanks for taking this over. I've added the corresponding commit hashes to simplify the review.

As long as we have no common library module for app and wear, we have to increase the code duplication, unfortunately. :unamused:

Navid200:
@tolot27 Thanks

tolot27 (approved):
Files are identical and build succeeds.

LCTM

jamorham:
Just to add, I don't think this will have any practical effect regarding resolving any particular issue as AAPS isn't running on wear.

Navid200:
@jamorham In that case, please let's not merge it.

Navid200:
@jamorham There is a comment in the open issue.
I am going to respond to the comment and state there should be no need to edit the wear files.
And, I am also going to ask the individuals to confirm the issue has been resolved and hopefully close the issue in a week or so.

jamorham:
There might be some combination of starting/stopping the sensor from elsewhere with wear running as a collector that could result in duplicated data but I don't think xDrip running on the phone would allow that data to be emitted but it is still worth ensuring that the wear collector doesn't even request the data in the first place. This assumes that the Sensor record is kept up to date on wear as well but either way I would normally just intend to keep the collector code in sync between app and wear. Its designed to work like that with identical files.

Navid200:
Understood.  Thanks

Issue #1882, opened 2021-11-27
Dex Share follower doesn't receive data with new EU accounts having an email for username


psonnera:
### Subject of the issue
New Dexcom users aren't proposed a username during account creation (only for linked accounts) and the username is the email address used to create the account. Dex Share follower doesn't receive any data with the email address in account credentials (checked correct server, username and password). Using a kid linked account with a username is functional.

### Your environment
I don't believe this is related to xDrip+ but Dex server recent changes.

### Expected behavior
Dex Share follower should receive data from Dex share servers.

### Actual behavior
No data is received by xDrip+ even if they are present in the vendor app followers.

### Your environment
I don't believe this is related to xDrip+ but Dex server recent changes.

### Expected behavior
Dex Share follower should receive data from Dex share servers.

### Actual behavior
No data.

### Steps to reproduce the behavior:
Create a new Dex account, log in the vendor master app, add a follower (either vendor app or xDrip+) and enable share. Accept the invitation in the vendor follower app to check functionality, set xDrip+ to Dex Share follower with matching credentials and server. And wait...

tolot27:
Likely, Dexcom tries to enforce their API documented at https://developer.dexcom.com.

I still don't know the source of the xDrip implementation, because nothing is documented.

psonnera:
@tolot27 I'm pretty sure it was documented there but even time travel machine doesn't have the original site... https://web.archive.org/web/20170919194707/https://developer.dexcom.com/ maybe @StephenBlackWasAlreadyTaken remembers or has some legacy documentation?

psonnera:
@tolot27 @Navid200 which debug tags could be enabled to allow me to give more information on what's going on?

fzelins:
I can confirm this bug. I have created new dexcom account, now my sensor expired so I deleted dexcom app, installed again and created new account. All new dexcom account have email as login. I have also completely reinstalled xDrip app and I am still getting error:

ShareFollowCallback
01:55 Get Share Data Failed: java.io.EOFException: End of input at line 1 column 1 path $

ShareFollowDL
01:55 Interceptor Body size: 0
01:55 Session id appears valid so going direct to get data
01:55 Setting message: 01: 55 Connectivity problem reaching Share servers

ShareFollowCallback
01:55 Get Share Data Failed: java.io.EOFException: End of input at line 1 column 1 path $

ShareFollowDL
01:55 Interceptor Body size: 0
01:55 Session id appears valid so going direct to get data

[...]

PR #1974, opened 2022-01-31
Dex share email ID

Navid200:
New Dex share IDs are email addresses.
This adds email format as one of the accepted options.

I have only compiled this.  But, cannot test it because I don't have a Dex share account.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/DexCollectionHelper.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/DexCollectionHelper.java
@@ -86,7 +86,7 @@ public class DexCollectionHelper {
                 textSettingDialog(activity,
                         "shfollow_user", "Dex Share Username",
                         "Enter Share Follower Username",
-                        InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS | InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD,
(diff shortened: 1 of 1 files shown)


Navid200:
If you are experiencing the problem described in the attached open issue, you can test this PR.

BUT, please note there is no guarantee this will fix the problem.

You cannot install this over your existing xDrip.  You will need to uninstall your current xDrip first.  But, if you uninstall xDrip, you will lose all your settings and data.  Therefore, you should save the database before uninstalling.
https://navid200.github.io/xDrip/docs/Backup-Database.html

Don't uninstall before saving the database.  If you have saved the database, uninstall xDrip.
Install: https://drive.google.com/drive/folders/12AZfKOCcGs_6o0_8f-eaJeY2BjO_-SlO?usp=sharing

After installing the test version, restore the database:  https://navid200.github.io/xDrip/docs/Restore-Database.html

To undo, uninstall the test version.  Install an official release.  Then, restore the database again.

tolot27:
This does not fix #1882.

Navid200:
Great.
Thanks

Navid200:
I thought you meant it did not fix the problem.
Which one is it then?  Does it fix the problem but does not have tests?
Or, does it not fix the problem?

Navid200:
@tolot27 There are parents who cannot follow their children.
You can always add tests.  Can we get this fixed first please?
Please don't label the PR invalid without explaining clearly why it is invalid.

Navid200:
> If you cannot test your code, please don't open a PR and claim that it will fix an open issue.

That's fair

Navid200:
If I can help with the progress, please let me know.

PR #1975, opened 2022-01-31
Keep G6 battery voltage logs for 7 days

Navid200:
This will keep the G6 battery voltage logs for 7 days instead of just one day.

In simplest ways, it's enough to know what the battery voltage is to judge the health of the battery.
But, there are many factors affecting the recorded voltage.  Those include temperature, voltage monitor accuracy, battery production variation, battery production batch, noise at the voltage sample time ...

Keeping the logs for 7 days let's us more confidently judge if the battery is the cause of malfunction or not.

This is what the log looks like now before this PR:

This is what the log looks like after this PR:

So, the log will become teal.

This will be helpful for those who help people use xDrip.  Low battery voltage is one of many factors causing malfunction.  It (having a 7-day period instead of only 1 day) will be very helpful to confirm that with certainty instead of just a possibility.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
@@ -1537,7 +1537,7 @@ public class Ob1G5StateMachine {
         if (transmitterId.length() != 6) return false;
         if (data.length < 10) return false;
         final BatteryInfoRxMessage batteryInfoRxMessage = new BatteryInfoRxMessage(data);
-        UserError.Log.e(TAG, "Saving battery data: " + batteryInfoRxMessage.toString());
+        UserError.Log.ueh(TAG, "Saving battery data: " + batteryInfoRxMessage.toString());
         PersistentStore.setBytes(G5_BATTERY_MARKER + transmitterId, data);
         PersistentStore.setLong(G5_BATTERY_FROM_MARKER + transmitterId, tsl());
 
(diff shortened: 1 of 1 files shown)


tolot27:
Would a `UserError.Log.uel()` be sufficient as well? Its severity is still `>3` and the log entry will be retained for seven days.

Navid200:
As long as it is kept for 7 days, I am OK with it.
But, I checked and alert snooze log is also ueh.  I don't know why a snooze alert is more severe than battery voltage.  Is it?  Does it really matter?

Navid200:
I changed it to uel.  I am testing it.  It will take me a few days to confirm that it really keeps them for 7 days.

tolot27:
It will definitely kept for seven days because the severity of uel is [5](https://github.com/NightscoutFoundation/xDrip/blob/f43d0377467dcbfa968c43a0b776bee838798548/app/src/main/java/com/eveningoutpost/dexdrip/Models/UserError.java#L95) and everything above 3 will be [kept for seven days](https://github.com/NightscoutFoundation/xDrip/blob/f43d0377467dcbfa968c43a0b776bee838798548/app/src/main/java/com/eveningoutpost/dexdrip/Models/UserError.java#L140).

Navid200:
This is what it will look like after a search in the logs:
Screenshot_20220209-161715
  
It will be very useful for users as well as for those helping users.

tolot27:
> This is what it will look like after a search in the logs:

Why do the values not change even after seven days? That's a bit surprising to me.
Nevertheless, it is a good adition.

tolot27 (approved):
An improvement to the battery monitoring.

Navid200:
Under normal circumstances, the voltages don't change much between days 10 to 90.
Thanks

PR #1986, opened 2022-02-08
Fix a bug that prevented users from selecting audio files.

tzachi-dar:
See https://github.com/NightscoutFoundation/xDrip/discussions/1983 for more details.

The change in xDrip code is simple: We ask for all files and not only audio files.
Once a file is received, we call an existing library to get the name and location of the file.
The File "AndroidFileNamePicker.java" is based on https://stackoverflow.com/questions/13209494/how-to-get-the-full-file-path-from-uri/60642994#answer-60642994

This has been testing on an android emulator and on a samsung with a few a few programs that handle the selce intent.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/AndroidFileNamePicker.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/AndroidFileNamePicker.java
@@ -1,0 +1,385 @@
+package com.eveningoutpost.dexdrip;
+// Based on https://stackoverflow.com/questions/13209494/how-to-get-the-full-file-path-from-uri/60642994#answer-60642994
+
+import android.annotation.SuppressLint;
+import android.content.ContentUris;
+import android.content.Context;
+import android.content.Intent;
+import android.database.Cursor;
+import android.net.Uri;
+import android.os.Build;
+import android.os.Environment;
+import android.provider.DocumentsContract;
+import android.provider.MediaStore;
+import android.provider.OpenableColumns;
+import android.text.TextUtils;
+import android.util.Log;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.InputStream;
+
+public class AndroidFileNamePicker {
+    private static Uri contentUri = null;
+
+    Context context;
+
+    public AndroidFileNamePicker(Context context) {
+        this.context = context;
+    }
+
+    @SuppressLint("NewApi")
+    public String getPath(final Uri uri) {
+        // check here to KITKAT or new version
+        final boolean isKitKat = Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT;
+        String selection = null;
+        String[] selectionArgs = null;
+        // DocumentProvider
+        if (isKitKat) {
+            // ExternalStorageProvider
+
+            if (isExternalStorageDocument(uri)) {
+                final String docId = DocumentsContract.getDocumentId(uri);
+                final String[] split = docId.split(":");
+                final String type = split[0];
+
+                String fullPath = getPathFromExtSD(split);
+                if (fullPath != "") {
+                    return fullPath;
+                } else {
+                    return null;
+                }
+            }
+
+
+            // DownloadsProvider
+
+            if (isDownloadsDocument(uri)) {
+
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                    final String id;
+                    Cursor cursor = null;
+                    try {
+                        cursor = context.getContentResolver().query(uri, new String[]{MediaStore.MediaColumns.DISPLAY_NAME}, null, null, null);
+                        if (cursor != null && cursor.moveToFirst()) {
+                            String fileName = cursor.getString(0);
+                            String path = Environment.getExternalStorageDirectory().toString() + "/Download/" + fileName;
+                            if (!TextUtils.isEmpty(path)) {
+                                return path;
+                            }
+                        }
+                    } finally {
+                        if (cursor != null)
+                            cursor.close();
+                    }
+                    id = DocumentsContract.getDocumentId(uri);
+                    if (!TextUtils.isEmpty(id)) {
+                        if (id.startsWith("raw:")) {
+                            return id.replaceFirst("raw:", "");
+                        }
+                        String[] contentUriPrefixesToTry = new String[]{
+                                "content://downloads/public_downloads",
+                                "content://downloads/my_downloads"
+                        };
+                        for (String contentUriPrefix : contentUriPrefixesToTry) {
+                            try {
+                                final Uri contentUri = ContentUris.withAppendedId(Uri.parse(contentUriPrefix), Long.valueOf(id));
+
+
+                                return getDataColumn(context, contentUri, null, null);
+                            } catch (NumberFormatException e) {
+                                //In Android 8 and Android P the id is not a number
+                                return uri.getPath().replaceFirst("^/document/raw:", "").replaceFirst("^raw:", "");
+                            }
+                        }
+
+
+                    }
+                } else {
+                    final String id = DocumentsContract.getDocumentId(uri);
+
+                    if (id.startsWith("raw:")) {
+                        return id.replaceFirst("raw:", "");
+                    }
+                    try {
+                        contentUri = ContentUris.withAppendedId(
+                                Uri.parse("content://downloads/public_downloads"), Long.valueOf(id));
+                    } catch (NumberFormatException e) {
+                        e.printStackTrace();
+                    }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
```diff
@@ -55,10 +57,13 @@
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
+import java.util.Objects;
 
 import static com.eveningoutpost.dexdrip.Home.startWatchUpdaterService;
 import static com.eveningoutpost.dexdrip.xdrip.gs;
 
+import androidx.annotation.RequiresApi;
```
tolot27, line 65: We should avoid Android X unless we decide to migrate to it.

app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
```diff
@@ -682,10 +685,9 @@ public void onTimeSet(TimePicker timePicker, int selectedHour, int selectedMinut
 
     }
 
-    private void chooseFile()
-    {
+    private void chooseFile() {
         final Intent fileIntent = new Intent();
-        fileIntent.setType("audio/mpeg3");
+        fileIntent.setType("*/*");
```
tolot27, line 690: Why do we allow all files here but filter it later? I thought the new file picker will handle it better.

app/src/main/java/com/eveningoutpost/dexdrip/AndroidFileNamePicker.java
```diff
@@ -0,0 +1,385 @@
+package com.eveningoutpost.dexdrip;
+// Based on https://stackoverflow.com/questions/13209494/how-to-get-the-full-file-path-from-uri/60642994#answer-60642994
```
tolot27, line 2: Did you check the license or is this file completely written by you?

app/src/main/java/com/eveningoutpost/dexdrip/AndroidFileNamePicker.java
```diff
@@ -0,0 +1,385 @@
...
+        return file.getPath();
+    }
+
+    /***
+     * Used for Android Q+
+     * @param uri
+     * @param newDirName if you want to create a directory, you can set this variable
+     * @return
+     */
+    private String copyFileToInternalStorage(Uri uri, String newDirName) {
```
tolot27, line 285: It looks like this class copies files in some circumstances. We should test this and also mention this.

tolot27 (changes requested):
Thanks for this contribution. Applying code formatting to the whole should be a separate commit and only pushed after the review is complete. Otherwise, it is quite hard to review.

tolot27:
@Navid200 Is there any issue which is addressed by this PR? If yes, please can you link it?

Navid200:
There is no existing (open or closed) issue.

app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
tzachi-dar: It is the same picker. Using it like this allows people to choose all files and later we have to filter them.
Not ideal, but does make life for users better. I did not find an easy way around this.

app/src/main/java/com/eveningoutpost/dexdrip/AndroidFileNamePicker.java
tzachi-dar: The file is copied almost as is from stackoverflow.
As far as I can tell, we are doing a fair use of this code. I'm based on answers like [this](https://www.ictrecht.nl/en/blog/what-is-the-license-status-of-stackoverflow-code-snippets).
Obviously, there are other opinions on the internet. I'm not a lower, but I fill that in a GPL project there shouldn't be a problem to use it.

app/src/main/java/com/eveningoutpost/dexdrip/AndroidFileNamePicker.java
tzachi-dar: Where should we mention this?

Navid200:
The AndroidFileNamePicker file seems to be a very large file.
By placing a custom sound file in the correct folder (Notifications) on the phone, the sound file can easily be used as is with no change to the xDrip code.  Isn't that a better, cleaner, simpler, solution?

tzachi-dar:
@Navid200 The complexity in this PR is for the developer. Hopefully from the user perspective this makes things simpler. 
The file added seems to be mentioned in a lot of places, so it should be fine (well tested).
If this is the case, than this PR will solve the issue for users without making too much troubles for developers.

tzachi-dar:
@Navid200 Did you check this on your systems? Did it solve the issue.

Navid200:
@tzachi-dar Sorry, I haven't.  
I don't understand what we want to address by this.  
Any custom file can be used easily by placing it in the "Notifications" folder.  Why is that not an adequate solution?

There may be multiple ways to accomplish the same objective.  
Putting the file in the Notifications folder seems to be the simplest solution.  It does not require adding more to the already complicated code.  It removes some unnecessary option from the complicated user interface.

What are the advantages of this PR?

Navid200:
I was going to compile this to test.
But, I keep thinking of things that don't make sense.

For example, when I go to select the sound file for the Forecast low alert or the Persistent high alert, I am not given the choice to select any categories.  I am taken right to the list of alerts.

Adding the sound file to the "Notifications" folder allows me to use it regardless.  I can use it for Forecast low, or persistent high or my glucose level alerts.  It is a universal solution.

How is this PR going to address the Forecast low or persistent high alerts?

jamorham:
Thanks to @tzachi-dar for this and @Navid200 for the well documented description of the issue and @tolot27 for the initial review.

I don't think this PR is the way to resolve this problem. There is a lot of hard-coded stuff and workarounds here which makes me very uneasy, but also I don't think it is that complicated to resolve and Android should provide a much cleaner way to do this.

Firstly to resolve the picker problem I think we just set a wildcard audio type, eg: 

`fileIntent.setType("audio/*");`

The problem with playback on android 9+ will be because the files themselves are not directly accessible any more due to permissions and need to be played via the content provider.

So when selecting, we shouldn't store the real path but instead the original URI and then in `AlertPlayer` instead of setting the URI we use the content resolver to open a file descriptor and pass that to media player.

I'm pretty sure this will fix it but it will need testing and because this touches `AlertPlayer` it must be handled carefully, especially as there could easily be a migration issue if we move from using paths to URIs.

I can put together a PR for this and interested parties can verify it works cleanly.

Navid200:
@jamorham Will your suggestion work for Forecast Low and Persistent High?

This does:  https://github.com/NightscoutFoundation/xDrip/pull/2285

jamorham:
The Forecast Low and Persistent High sound selection uses a different method which doesn't suffer from this issue.

Navid200:
But, how would you select a custom sound for the Forecast Low alert?

I'm saying the list that comes up will include your custom sound files if you place them in the notifications folder.

How is a user supposed to use a custom sound file for the Forecast low alert in the solution you have in mind?

jamorham:
Well on my Android 11 phone, on Forecast Low it brings up android sound picker and I click the + in the top right and add any sound file as required. But that is outside the scope of the issue this PR is to address which is that android 9 probably broke the custom sound file selection within `EditAlertActivity`

Navid200:
That's my point:

My suggestion solves the problem of adding a custom file for all alerts and makes the approach consistent for all alerts.

Navid200:
On my Android 11, I get this when want to choose a sound file for Forecast Low:
Screenshot_20220823-160736

After tapping on add ringtone at the bottom, I am able to choose a custom file.
However, the file is not effectively selected.  The next time I do the same, I still see the previously selected file.

The only way I can have the selected file really selected is if I place it in the Notifications folder.

jamorham:
Please see how I have addressed this for reminders. https://github.com/NightscoutFoundation/xDrip/commit/9f3ebd1b2c7981a5dde3e56ca77bff4a0c4feced

I would plan to produce a PR that would do the same sort of thing for custom sounds with alerts.

There are a few changes required to make it work, the wildcard mime time for selection and then a request to be able to access uri's after device reboot and then opening using file descriptor. I've attempted to implement this with fallback to different mechanisms and if absolute paths are already stored so I'm hopeful it should work on previous installations and android 4+ I have only tested this on Android 11 and 12 so far though.

Navid200:
@tzachi-dar 
I think this has been fixed now.
If you agree, do you think we should close this PR?

tzachi-dar:
Closing this, as it was fixed in another pr.

PR #1987, opened 2022-02-09
closes the media player correctly and releases resources

tolot27:
This is an extended PR of the code @gruoner provided with #1715.

Basically, it defines an error handler and an complete listener to release resources of the media player after errors and completion of media playing. This is the recommend usage according to the [Android MediaPlayer docs](https://developer.android.com/reference/android/media/MediaPlayer#release()).

Gruoner reported that "after a while this results in silent alarm without any sound" if not properly handled.

The last two commits are preparations for unit testing.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -128,6 +128,7 @@ android {
         animationsDisabled = true
         unitTests {
             includeAndroidResources = true
+            returnDefaultValues = true
         }
     }
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
@@ -1061,25 +1061,44 @@ public class JoH {
         return ContentResolver.SCHEME_ANDROID_RESOURCE + "://" + xdrip.getAppContext().getPackageName() + "/" + id;
     }
 
-    public static synchronized MediaPlayer playSoundUri(String soundUri) {
+    private static MediaPlayer player;
+    public static synchronized void playSoundUri(String soundUri) {
         try {
             JoH.getWakeLock("joh-playsound", 10000);
-            final MediaPlayer player = MediaPlayer.create(xdrip.getAppContext(), Uri.parse(soundUri));
-            player.setLooping(false);
+            if (player != null) {
+                UserError.Log.i(TAG, "playSoundUri: media player still exists. Releasing it.");
+                player.release();
+                player = null;
+            }
+            player = MediaPlayer.create(xdrip.getAppContext(), Uri.parse(soundUri));
             player.setOnCompletionListener(mp -> {
                 UserError.Log.i(TAG, "playSoundUri: onCompletion called (finished playing) ");
+                player.release();
+                player = null;
+            });
+            player.setOnErrorListener((mp, what, extra) -> {
+                UserError.Log.e(TAG, "playSoundUri: onError called (what: " + what + ", extra: " + extra);
+                // possibly media player error; release is handled in onCompletionListener
+                return false;
+            });
+            player.setLooping(false);
+            player.start();
+        } catch (Exception e) {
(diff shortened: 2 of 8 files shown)


tolot27:
@Navid200 Do you now any issues which could be related to this PR?

Navid200:
I have experienced scenarios when I expected an alert but never got one.
I have also seen complaints about alerts not making a sound.  
But, I have never had a way to identify the reason.  The fact is that there are several factors that could cause an alert not to make a sound.  

So, this could be a reason.  
Can we test?
Can we undo a PR after merge if we conclude that it did not fix what we expected?

tolot27:
> Can we test?

I've tried to implement instrumentation tests but this is a huge effort because a lot of classes (i. e. `UserError`) need to be mocked. Hence, two test-related commits. Indeed, we can test it during our daily usage. The PR does not change the general function of the media player. It only releases it as recommend in the [docs](https://developer.android.com/reference/android/media/MediaPlayer#release()).

>Can we undo a PR after merge if we conclude that it did not fix what we expected?

It is not possible to undo/revert a PR as a whole, but reverting single commits is a standard practice in git. Nevertheless, I recommend not to revert things which follow SOPs and improve code quality in general, just because it does not (fully) fix what we expect.
But likely you mean that we should revert commits which make things worse. :smile:

app/src/main/java/com/eveningoutpost/dexdrip/Models/JoH.java
```diff
@@ -1062,16 +1062,44 @@ public static String getResourceURI(int id) {
...
-            return player;
         } catch (Exception e) {
             Log.wtf(TAG, "Failed to play audio: " + soundUri + " exception:" + e);
-            return null;
+        }
+    }
+
+    public static synchronized void stopSoundUri() {
+        if (player != null) {
+            if (player.isPlaying()) {
```
tolot27, line 1094: I could not find information whether it is necessary to stop the player before releasing it, or not.

tzachi-dar: Well in any case stopping the player should be fine in any case.

tolot27: According to the [state diagram](https://developer.android.com/reference/android/media/MediaPlayer#state-diagram) calling stop is also possible from prepared and paused state. Should we test for this as well?

app/build.gradle
```diff
@@ -128,6 +128,7 @@ android {
         animationsDisabled = true
         unitTests {
             includeAndroidResources = true
+            returnDefaultValues = true
```
jamorham, line 131: Is there a reason not to just use `RobolectricTestWithConfig` ?

jamorham:
I'm happy to add the additional locking I mentioned to this PR.

jamorham:
I'm happy to add the additional locking I mentioned to this PR.

jamorham (approved):

jamorham:
I've extended this in today's nightly to safely handle audio stopping and thread safety. For example to handle the following problems:

- Instances can get set to null by completion events out of sequence - I've added a semaphore lock to prevent this
- I don't think references need to be set to null once the media player instance is released (as that is what frees the resources)
- MediaPlayer is really awkward, for example if you check if audio is playing then tell it to stop and it stopped by itself between those two calls then an exception is created. The only way is to wrap exception handlers around wherever this logic is applied - I've done that.
- The event for audio playback completion actually fires before audio has finished playing (yes really) and so I've delayed the release by 200ms to avoid clipping the end of audio sequences.
- I've tested the various places where audio playback occurs, alerts, reminders, nfc sync and it all seems to work reliably now. I tried 1000 concurrent audio playbacks and these also appeared to leave the media player in a good state.

See: b5a2e8867cbfc59bcd66bc102a682499c9a42d54 and da6bb4f7fa9dc55e339000fb00209a5b0fc7dfd7

PR #2005, opened 2022-02-18
remove token pre-processor

tolot27:
It does nothing useful until AGP gets updated to 3.5.x or higher, but prevents Android Studio to execute unit tests inside the IDE.


tolot27:
> As described in the comments for the other PR it doesn't prevent tests being executed. Please let me know if you can't reproduce the steps I described. Even without modifications, running all tests appears to work fine to me.

I've tried to reproduce what you commented, also with a fresh git clone:

> If I right click on the green java folder under test and run Tests in xDrip.app it works fine.

After building the app and start of the tests (as you described) I immediately get the error "No tests were found" and the output log shows what it is executed:
```
"C:\Program Files\Java\jdk1.8.0_311\bin\java.exe" -ea -Didea.test.cyclic.buffer.size=1048576 "-javaagent:C:\Program Files\Android\Android Studio\lib\idea_rt.jar=62730:C:\Program Files\Android\Android Studio\bin" -Dfile.encoding=UTF-8 -classpath C:\Users\Walter\AppData\Local\Temp\classpath1912145624.jar com.intellij.rt.junit.JUnitStarter -ideVersion5 @w@C:\Users\Walter\AppData\Local\Temp\idea_working_dirs_junit1.tmp @C:\Users\Walter\AppData\Local\Temp\idea_junit1.tmp -socket62729
Exception in thread "main" java.lang.NoClassDefFoundError: com/eveningoutpost/dexdrip/Models/JoHTest (wrong name: com/eveningoutpost/dexdrip/models/JoHTest)
	at java.lang.ClassLoader.defineClass1(Native Method)
	...
Process finished with exit code 1
```
As you can see, the ReplaceTokenPreprocessor lowercases the package name which is not known to the Android Studio execution engine. If I comment out `apply from: 'token-replace.gradle'` in `app\build.gradle` unit tests run fine in AS.

I've also tried other build variants but could not find any combination which works as long as the token preprocessor is in use.

> but my point is that it is the android studio filtering of test selection which is the problem here rather than the execution of the tests.

I do not agree. Android Studio loads the complete directory structure and knows the (uppercase) package names from it and also from the first line of the java class files. During test build, the package names are converted to lowercase but AS still executes it with its uppercase naming.

It would be interesting which AS and Java version you are using and why you cannot reproduce this error. I'm using AS 4.2.2 with jdk1.8.0_311.

Nevertheless, my point is that the token preprocessor is currently not needed and it is save to remove it.

BTW: If I execute the test with the default settings, I get "Error running 'dexdrip in xDrip.app': Command line is too long. Shorten command line for dexdrip in xDrip.app." I have to change the "Shorten command line" option to either "JAR manifest" or "classpath file" to get it working:

image

jamorham:
I'm testing using `Android Studio Bumblebee | 2021.1.1 Patch 1
Build #AI-211.7628.21.2111.8139111, built on February 1, 2022` on the Linux platform. Could this be a windows specific issue especially as you are mentioning things like shortening command lines which I don't think is an issue on linux?

tolot27:
After upgrading from 4.2.2 to 2021.1.1 Patch 2 (Bumblebee) no more unit test execution errors occur. The reason is obvious: [Unit tests now uses Gradle test runner](https://developer.android.com/studio/releases#test-configurations). That explains why you did not get those errors.

Again: the token pre-processor currently does nothing useful. It just increases the build time. Hence, why not removing them – at least for now?

PR #2022, opened 2022-02-28
change icon of auto configure setting to be a QR code scan

tolot27:
Extracted from #1151. The red color is identical to that of the xDrip logo.

Here is a side-by-side comparison with the new QR code icon:
Screenshot_1628020238 128077861-a52b6fa6-0f0b-4f21-aa62-2832a89bfe33


tolot27:
@jamorham mentioned:
>The barcode scan icon is present in screenshot documentation and as mentioned in another comment I don't think the auto-configuration feature is limited to QR codes.

My biggest question in this context is how to generate a configuration **barcode** automatically. I've never seen this and the implemented feature generates a QR code. Hence, if there is never a barcode generated, I suggest to use the QR code. This is visually more consistent.

I've someone points me to the barcode scan icon in the documentation, I'll update it as well.

jamorham:
The auto-configuration feature can read various types of barcode. A QR code is a type of 2D barcode and the icon provided here is probably more appropriate as the vast majority of scans will involve QR codes.

jamorham (approved):

tolot27:
> The auto-configuration feature can read various types of barcode.

But what is the source of the generated barcode? Is it documented somewhere? Can I scan the Dexcom sensor barcode for instance?

PR #2057, opened 2022-03-23
Feature/settings reorganization

Lupum1001:
Ok, so these changes are based on the discussion here: (The subset that I referred to as "Part 1")
https://github.com/NightscoutFoundation/xDrip/discussions/1991#discussioncomment-2213674

The goal is essentially to various find ways to tidy up the preferences menus somewhat.

### Changes:
1. I've created a consolidated "Backup and Update" page and **_moved_** the following pages into it:
   -- "Auto configure" (the barcode scanner)
   -- "xDrip+ Predictive Simulation Settings" --> "End User License Agreement"
   -- "xDrip+ Update Settings"
   -- "Show Settings QR codes"
   -- "Load / Save settings"
2. I've removed the "Check for updated version" from the home screen dropdown menu in favor of one in the new menu.
 I also enhanced it by adding the app version string under it as the summary (e.g. `92b6d8d-2022.02.09`).
3. Changed name/summary of "Auto Configure" functionality.

image
image

### Notes/Rationale:
I'll try to fill this section in later; especially as questions/critiques start coming in.


app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/UpdateActivity.java
```diff
@@ -191,7 +191,7 @@ public static void checkForAnUpdate(final Context context) {
 
     public static void forceUpdateCheckNow() {
         JoH.static_toast_long(xdrip.gs(R.string.checking_for_update));
-        UpdateActivity.last_check_time = -2;
+        UpdateActivity.last_check_time = -1;
```
tolot27, line 194: What's the rationale behind this change? Does `-1` have a special meaning?

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -915,15 +914,15 @@ public void onCreate(Bundle savedInstanceState) {
...
-            setupBarcodeConfigScanner();
-            setupBarcodeShareScanner();
             bindPreferenceSummaryToValue(findPreference("cloud_storage_mongodb_uri"));
             bindPreferenceSummaryToValue(findPreference("cloud_storage_mongodb_collection"));
             bindPreferenceSummaryToValue(findPreference("cloud_storage_mongodb_device_status_collection"));
             bindPreferenceSummaryToValue(findPreference("cloud_storage_api_base"));
 
             addPreferencesFromResource(R.xml.pref_advanced_settings);
             addPreferencesFromResource(R.xml.xdrip_plus_prefs);
+            setupBarcodeConfigScanner();
```
tolot27, line 924: Is this code movement necessary and why?

app/src/main/res/values/strings.xml
```diff
@@ -37,8 +37,8 @@
     <string name="title_activity_double_calibration">Initial Calibration</string>
     <string name="title_activity_calibration_override">Calibration Override</string>
     <string name="pref_header_cloud_storage">Cloud Storage</string>
-    <string name="auto_configure_title">Auto configure</string>
-    <string name="prefs_auto_config_summary">Auto configure using a barcode.</string>
+    <string name="barcode_scanner_title">Barcode Scanner</string>
+    <string name="prefs_barcode_scanner_summary">Import settings from QR or barcode</string>
```
tolot27, line 41: As far as I know, the settings are not stored as QR or barcode. The QR code just contains a shared link. Hence, "Import settings using a QR or barcode" or the previous text is more appropriate.

Also, changing the string name removes all translations from Crowdin. A change in the string text just flags the translations as unapproved.

tolot27 (changes requested):
@Lupum1001 Thanks for your effort and contribution. 😄 

I've some comments and suggestions. Please create screenshots from before and after your PR to compare it side-by-side. That helps a lot during review and understanding what the PR changes.

> * -- "xDrip+ Predictive Simulation Settings" --> 

The predictive simulation settings is still shown at the same place - at least at your screenshot. The "End User License Agreement" is moved from the top of the settings list.

>   -- "xDrip+ Update Settings"
>   -- "Show Settings QR codes"
>   -- "Load / Save settings"

Just to let you know, my settings also contain "xDrip+ Motion Tracking" between "xDrip+ Sync Settings" and "xDrip+ Update Settings".

In general, I believe this PR is a reasonable improvement in the xDrip settings "hell".

Lupum1001:
> The predictive simulation settings is still shown at the same place - at least at your screenshot. The "End User License Agreement" is moved from the top of the settings list.

Sorry, I just edited the wording in my post to make it a bit less confusing; I didn't move the predictive settings, but there was a second EULA entry _within_ that menu that I moved.

> my settings also contain "xDrip+ Motion Tracking" between "xDrip+ Sync Settings" and "xDrip+ Update Settings".

Yeah, I also have that page on the stable release of the app, but it seems to have been removed in the current master branch since then? The nightly build for Xdrip (without my changes) doesn't seem to have it either.

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
Lupum1001: I believe I had to move it down below the line that loads the `R.xml.xdrip_plus_prefs`, because I'm essentially moving the "Auto configure" scanner further down the settings list into the "xDrip+ Extra Settings" section.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/UpdateActivity.java
Lupum1001: The code is a little convoluted, but it basically interacts with this existing check here:
image
If I had to translate it into pseudocode, I'd probably describe it as:
Do we skip the update check?
- If the last_check_time is set to -1, _never_ skip.
- If it's anything else, **SKIP** the check if the "automatic updates" switch is turned **OFF**


The function forceUpdateCheckNow() _is_ used in one other place (a function called `testAndSetNightly(..)`), which now won't care if "automatic updates" are switched off. I've been considering tidying up this wider logic before I formally submit this PR for review.

app/src/main/res/values/strings.xml
Lupum1001: My impression was that it was encoded json, based on the [original tool](https://nightscout.github.io/pages/configure/) for sync url setup, but I can take another look at the scanner code.

Lupum1001: I can confirm that everything is self-contained within the QR codes.

- the xDrip QRs are taking the selected preferences as json, zip-compressing them, and then encoding the data as a base64 string. The "xdpref:" prefix is used to differentiate it from the other QR codes below.
- The Bluejay watch QRs appear to encode their data in lower-level hexadecimals.
- The Nightscout QR codes (for Mongo, MQTT, etc) don't appear to be encoded at all (the QR codes contain straight-up json).

tolot27 (changes requested):
Thank you for your great contribution!

Unfortunately, you were changing to much files in a single PR. It is very unlikely that this PR will ever be merged. Please separate your work into smaller PRs.

tolot27:
> Unfortunately, you were changing to much files in a single PR.

Ah, that's likely caused because you changed the translated XML files. That's not necessary. As soon as a string is removed from the English strings.xml, it will be removed by Crowdin from all other translations. Hence, just change the original strings.xml.

Lupum1001:
> that's likely caused because you changed the translated XML files. That's not necessary. As soon as a string is removed from the English strings.xml, it will be removed by Crowdin from all other translations. Hence, just change the original strings.xml.

Ah, ok. That certainly simplifies things a lot. I'll just revert those file changes so they don't clutter the PR. 👍

app/src/main/res/xml/xdrip_plus_prefs.xml
```diff
@@ -684,19 +685,6 @@
...
-            <Preference
-                android:key="I_understandb"
-                android:summary="@string/not_for_medical_use"
-                android:title="@string/end_user_license_agreement">
-                <intent
-                    android:action="android.intent.action.MAIN"
-                    android:targetClass="com.eveningoutpost.dexdrip.LicenseAgreementActivity"
-                    android:targetPackage="@string/local_target_package" />
-            </Preference>
-        </PreferenceCategory>
```
Lupum1001, line 699: This extra EULA button has been moved. The Preference hasn't changed, but the PreferenceCategory surrounding it was removed.
It now lives in the "xdrip_version_category".

app/src/main/res/values/strings.xml
tolot27: > Also, changing the string name removes all translations from Crowdin. A change in the string text just flags the translations as unapproved.

This is still an issue.

Lupum1001: Is the idea to let this specific feature remain "Auto configure" in other languages for the time being?
I changed other strings as well, and figured if they had to be re-translated anyway it made sense to rework the key-strings while the chance was there.
Side question: am I correct that they just render in English if there's no translation?

For example, it's probably fine to go back and keep using the "copying_settings" translations, but "use_qr_codes_to_transfer_settings_using_settings_auto_configure_feature" would have been a bad one to keep by contrast.

Lupum1001: I rolled back three string names in the following commit:
0e0c5818caf837a6c3e2caf195dfda7ad37d7301
If we can update the translations when we get the chance, that would be preferable, but I guess it's ok if it takes a bit longer.

tolot27: I've asked for clarification at https://community.crowdin.com/t/rename-of-source-string-identifiers.

tolot27 (commented):
Looks good so far. I'm not really happy with the renaming of "Auto Configure". The new summary is fine but "Barcode scanner" sounds like a barcode scanner app.

@Navid200 @jamorham What do you think about this renaming?

Lupum1001:
> The new summary is fine but "Barcode scanner" sounds like a barcode scanner app.

Hmm... I can change it from a noun to a verb ("Barcode Scanner" -> "Scan Barcode"). Is that any better?
I'm essentially concerned about keeping the name Auto Configure for a couple reasons:

- I feel like "scan" or "scanner" needs to be somewhere in the name/description to make it instantly click with users. The same was true of "Import" and "QR", which I already included.
- Within the context of backup/restore tools, "auto" starts to feel awkward to me. It was ok in its old location, but here it just sort of feels redundant.

The new name _would_ be too vague as an isolated setting, but I think the new "Backup and Restore" sub-category provides enough context.
image

Navid200:
So many changes in a single PR.  
If jamorham goes through all the proposed changes here, only one change he doesn't like is enough to hold up the entire PR.

I personally don't understand why the end user license agreement should be on the backup page.

If I were you, I would identify a single setting that is constantly being missed by users because it is in an unexpected place.  And I would open a PR just to move that single setting to where it should be.
If that is a success, you can then take another small step to correct something else that is broken just like that.

But, I don't know if we should rearrange everything.  
I hate it if we start getting complaints about where is this and where is that?

Lupum1001:
What this PR boils down to is a new menu "Backup and Update" to bring things together and reduce clutter in _other_ places.
If I only moved in _some_ of the corresponding settings, I don't feel like I'd be improving things. My goal is a broader cleanup, as you know I've mentioned [here](https://github.com/NightscoutFoundation/xDrip/discussions/1991), but I wanted to start with this category for the moment.

---
> I personally don't understand why the end user license agreement should be on the backup page.

It's technically under "App Version", but I understand your reflex. I consider it to be an improvement over its last position within "xDrip+ Predictive Simulation settings", but since it's technically a duplicate we could also just remove it outright.

Navid200:
> My goal is a broader cleanup, as you know I've mentioned [here](https://github.com/NightscoutFoundation/xDrip/discussions/1991), 

You have mentioned it there.  But, I see no post from jamorham there.
There is no one here who can promise you a merge other than jamorham.

Lupum1001:
>There is no one here who can promise you a merge other than jamorham.

Understood. And the more feedback I get from you guys, the smoother I expect it to go. 😉
I re-did the screenshots in the summary post to try and make it more concise, and I'm waiting for @tolot27's response on the "Scan Barcode" re-naming question. I'll probably wait for feedback from @jamorham before I look at splitting up the PR, since I don't know his stance on the various parts yet.

tolot27 (approved):
LGTM.
I'm fine with barcode scanner. Regarding the string identifier renaming I've asked Crowdin support for a git based solution. We can rename string identifiers within Crowdin but we have to sync this with our source code. This can be done with two PRs. The first one duplicates the string and the second PR removes the old one after a certain amount of time (about one week).

Navid200:
xDrip is supposed to not offer any medical advice.  If you consider that and anything related to simulation included in xDrip, you may see that there is a contradiction.  Telling someone how much insulin they should take is effectively medical advice.  You can spin it all you want.  A CGM tells you what your blood glucose is not how much insulin to take.

Anyway, the reason the license agreement is on the predictive simulation page as well may be to remind users that they should look at it (simulations) only for research.

Please note I am only speculating.  I didn't put it there.  
Why everything is where it is may have to do with so many different factors.

Lupum1001:
The 2nd EULA entry appears to have been added in this commit: 6205fe81703f8 (changes in the last file)
It appears that even then, it was probably - as you guessed - just a kind of reminder. I say that because the `android:dependency` field was still targeting the main EULA, not the duplicate one.

If you feel strongly about keeping it on the simulations page, I can instead just move it further up to the top so it doesn't look like "Low level prediction values" is part of the EULA PreferenceCategory.
image

tolot27:
> If you feel strongly about keeping it on the simulations page

If we **really** want to remind the user that xDrip is not a medical device/application we should show a popup regularly (i.e. monthly) or right after enabling a feature requiring this reminder (would be a separate PR). Putting additional links to the EULA in certain sub menues does not fulfill this purpose because it is likely overseen, especially because "Not for medical use" is much smaller text than "End User License Agreement". It is like with manuals – they will be rarely read.

Remember: the EULA is not removed and right after installing xDrip you have to agree.

Maybe @jamorham can comment/agree on the suggested change.

Lupum1001:
I ended up moving the EULA back to the top of the Simulations page instead. This at least corrects the issue with the PreferenceCategory visual.
Ultimately, I'd still like to get it out of there in favor of something more explicit, but it's arguably better than no warning at all in the interim.

Lupum1001:
@tolot27 @Navid200 Are there any more requests/concerns on this PR before it goes to @jamorham? I've been working on my next branch, but didn't want to let this one get too stale in the meantime.

Lupum1001:
Back after a hiatus, and trying to revive a few of my branches. I'm pretty sure this branch is up to date again, but I'm going to try and split off a few changes, in case that gets them reviewed.

Navid200:
> @Navid200 Are there any more requests/concerns.

I was never in support of this PR.

There are a million different ways to arrange a menu.  And there isn't a right or wrong.  
If there is something that is extremely confusing, you can open a PR and suggest to change that.   But, that would be one or two lines, not like this.

Navid200:
@Lupum1001 
Fortunately, xDrip users can change the colors:  https://navid200.github.io/xDrip/docs/Legend.html.

But, imagine if there was no way to change the colors.
Then, imagine if someone opened a PR to change the colors.  And imagine if the PR was merged.
Then, next month, someone else would open another PR asking to change the colors again.
And next month, there would be another.
This wouldn't go on and on because we couldn't make up our mind.  Rather, because there isn't a right or a wrong color.
That's why we let users change the colors as they wish.

But, I don't think we can allow users to change the menu.
Considering that we also allow users to choose their language, if we also allow them to move the items on the menu, it will become a nightmare for me as someone who helps users troubleshoot xDrip.  How can I figure out what they are talking about if I don't know which menu item they are talking about.

You are obviously a competent java developer.
I wish you hadn't chosen to modify the menu as your way of contributing to xDrip.
xDrip can definitely benefit from your expertise:  https://navid200.github.io/xDrip/docs/Issues.html

Lupum1001:
@Navid200 you are absolutely correct that there dozens of ways to organize a menu, and part of what I'm doing is considering less aggressive alternatives than the one I'm proposing here. The translation strings seem to very locked in, and resistant to change. "Backup and Update" fits neatly on one page for now, but if the list of backup/restore options keeps growing and growing, I might find myself wishing I'd kept them separate.

That being said, the xDrip settings _are_ quite cluttered atm. Some of that has been correctly characterized as "too many settings", and some of it comes from inconsistent groupings after years and years and development.

It's worth keeping in mind though, that grouping "Backup and Update" settings together wasn't necessarily because THOSE settings were hard to find. It was because it reduces the clutter on the main settings list, which makes everything _else_ a bit easier to find as well.

moberer:
> There are a million different ways to arrange a menu. And there isn't a right or wrong.

Well, not "right" or "wrong", but there absolutely is a real gradient between "better" and "worse" user interfaces.

> If there is something that is extremely confusing

I do not think this is the standard user interfaces should be held by.
We should strive for excellence, not just "barely enough".

In my opinion the reorganization of the settings is very-much needed.

And - just for the sake of the users - it makes sense to do it in a single change.
Repeatedly finding out a setting has moved is not a good experience. - In my opinion, reorganizing the settings once and then be done with it for a while would have much less friction for end users.

Is there anything that can be done for this or another improvement to the settings to be merged?

PR #2063, opened 2022-04-02
Log the call stack of exception for log.e calls

tzachi-dar:
Log the call stack of exceptions for all `log.e(TAG, message, exception)` calls.

Here is an example:


tolot27 (approved):
LGTM.

Just a side note: adding a single character to a StringBuilder object is lightly faster if it is added as a character using `append('\n')` instead of as a string (`append("\n")`) because it is directly added to the underlying character array.

jamorham (approved):

PR #2066, opened 2022-04-04
fix translationTest path replacement

tolot27:
Currently, the `translationTest` fails if run in Windows environments with the assertion No locales added from resources - this seems unlikely.

This PR fixes the path replacement regular expression.


jamorham (approved):
Thanks

PR #2073, opened 2022-04-10
Libre OOP2 temperature and calibration noise detection.

jbr7rr:
# Libre OOP2 temperature and calibration noise detection.

I've been using the OOP2 algorithm with the non-calibratable value's for quite a while, and i noticed that the readings can go down or go up highly during temperature changes. This is most notable when in the shower, of when putting and icepack on the sensor. I've noticed that in all cases the raw value's stay level or go in the other direction, so instead of up they go down.

This is unwanted when looping, as we will get data that is not representable of the actual BG's. There are multiple ways to address this problem. Ideally in the OOP2 algorithm itself.

An example of me showering, where you can clearly see the fall and rise of the BG levels compared to the raw values:
(The bigger dots are the BG values from OOP2 (unsmoothed), the trend dots are the raw values, readings are shown every minute for testing purposes)
https://imgur.com/a/gmU21fJ

In the mean time we can detect that this is happening and report it as noise. That is where this PR comes in. 

# Detection 

The idea is that we compare the last 5 rawBg and last 5 bg values from OOP2. and see if they diverge too much.

We do this by getting the delta's of these values for the past 5 readings, then we create a linear trendLine for these readings and we get a slope of the delta's. When the trend of the raw readings and the bg readings are the same this slope should be zero. 

The value of this slope is converted into a noise parameter which is reported.

# Questions:

- Ideally we also want noise detection for the minute data of the Libre raw value's and bg values (whatever is in use). I could add this to this PR. But to make this more review friendly i would like to add general noise detection in a separate PR. Please let me know what you think
- Noise is now cut off at 200 for local broadcast. Do we want to do anything further with the noise levels?

# Todo for this PR:
- Get feedback
- Create unit tests


tzachi-dar:
Thanks for creating this PR, I'll need a couple of days to go over it.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
```diff
@@ -1326,11 +1330,7 @@ public static void bgReadingInsertFromInt(int value, long timestamp, long margin
                 if (readingNearTimeStamp(bgr.timestamp, margin) == null) {
                     bgr.save();
                     bgr.find_slope();
-                    if (do_notification) {
```
tzachi-dar, line 1329: What happened to this part? Why is it removed?

tzachi-dar:
A few notes. To be honest, I don't know how noise is realy been used in other applications. If I understand correctly, xDrip is not realy using it.

So:
1) The feature can not be turned off. Should be enabled by some switch.
2) The feature only works with raw enabled. The calculations can be done also without it.
3) The definition of noise is quite weird here: IIUC It is based on changes on temperature. I think that noise should be based mainly on changes in the raw values. Here is an example of what I think as noise (to be honest, this is quite a small noise... It can get much bigger). 
image

Are their examples of what would happen to openaps in cases of noise?

And one harder question...
Is there a way to find the real influence of temperature (Without extreme cases of ice / shower) Maybe something like sleeping under the blankets? 
Can we compare two sensors with one person at the same time (one hand wrapped with a blanket and the other one not). (This one is not realy related to this PR directly).

This is initial feedback, need to think about it more.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
jbr7rr: This is moved to bgr.postProcess(!do_notification, true); see other insert functions as well

jbr7rr:
@tzachi-dar Thank you for your initial comments.

1. Honestly I've thought about this, but there is already a setting overload in xDrip, but we can add a switch, which should ideally switch off all noise detection. I will see if i can find a way
2. I dont quite follow, This detection is only relevant if no calibrations are used, and temperature has an influnce on the display glucose
3. This really depends on how you define noise, yes in a sense most of the time it are jumps in the data, but one can view it as any unwanted distortions of the original signal/source. And in this case sometimes it is really unwanted and does not follow the original signal (a persons blood glucose value)

Also I would like to add detection for the noise you described as well. Tell me if you are ok with this or if the PR will get too large.

jbr7rr:
@tzachi-dar 

About APS, i cannot comment on openAPS because i have not looked in their sources yet. I can comment on AndroidAPS (which uses the same algorithm afaik).
There is some noise handling, but it is not "activated".
These commits activate it:
https://github.com/jbr7rr/AndroidAPS/commit/b381a9d483a1628169a6a566cdbb2990b71a3aa8
https://github.com/jbr7rr/AndroidAPS/commit/aaf0c718f45ba2774332b83ed46aa3494c42e21a

When it is activated I have observed the following behavior:
- If NS noise level = 2 the target BG is raiseed 30%
- If NS noise level >= 3 The loop is disabled, and if the temp basal is below the profile basal it is shortend to 30 min, if it is above the profile basal the temp basal is canceled.

I do intent on making a PR to AndroidAPS if the PR into xDrip is more mature and/or merged.

app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
jbr7rr: This is moved to bgr.postProcess(!do_notification, true); see other insert functions as well

tzachi-dar:
A few notes:
1) I meant to say Androidaps (OAPS "market share" is probably too small this days.)
2) As for a switch. I think it is a must. This PR will make significant changes to android APS behavior. I don't think that there is a way this can happen just by a version update without the user doing something to activate it.
3) The fact if a sensor is noisy or not noisy, is something that is happening in the sensor. This is why I think that it doesn't realy matter if one is using the temperature for this. [This opens two questions actually: 1) Can a sensor be noisy in the temperature field but not in the raw field (and vice versa), or actualy why the temperature meters for the noise].

Do you have an estimation on the influence of this pr (in other words if activated) how long will the noise level be 2 or 3.

Thanks
Tzachi

jbr7rr:
@tzachi-dar 
Thank you for your feedback so far!

I will implement the switch and also some additional sensitivity settings.
As for point 3: From my experience the noisyness of the different values overlap but are not entirely the same.
However I agree that the BG values play a huger role in the noise so we should include a noise estimator for the bg values as well.

I will update this PR with noise estimation for the BG values, and also have to tweak and reconsider the part of the delta noise estimation. And will report back.

As for an estimation on the noise levels, I'm still tweaking the value's, But for what I'm testing now with BG + delta noise estimation: When the sensor is starting up its really noisy and a lot of noise is reported, after it settled (1-2 days) there is a lot less noise, and noise is reported 1 to 5 times a day, ranging from 5 to 20 minutes per incident. I think a good balance is really important for this, I will also let other users test it and report on the sensitivity of the noise estimation. I have never used a dexcom so i cannot compare it to their noise estimation myself.

tzachi-dar:
Finally got time to review the new code.
Generally this seems like it would work, but there are some cases that will not be catched by this (for example, a nice curve of bg going up and than starting to go down.)
In any case, doing something perfect is obviously impossible. I will therefore try and add a new table with raw values, bg values, temperature, and both noise values, and try to see how this looks on real life examples.

jbr7rr:
@tzachi-dar Thank you for your time and response.

Yes it is not perfect, but the goal is to detect as much erroneous readings as possible and I think this enables that.
The temperature parameter does not affect the code directly. But  it will indirectly.

Do you want a unit test for this? Also if you have the table we could use that as input for the unit test?

Since my last comment I have acquired a Dexcom one. And that sensor is also somewhat sensitive to temperature but in a lesser degree then the Libre2 with OOP2 native readings. it is more comparable to the OOP2 raw readings. Do you have any tips on how to easily extract the raw data from the xDrip? Currently I use the database export but that is a pain to get the actual data I want.

jbr7rr:
Closed, out of date, superseded by other improvements

Navid200:
Forgive me for posting here now that it is closed.
I have seen it for the first time considering it is related to Libre and I often leave Libre to others.
But, now that it was closed, I had a look at it and noticed for the first time that it is related to noise.  I would like to say one thing considering for the record.

It can be very misleading if you only look at a graph after the fact.  It will be obvious then what was noise and what was real.
But, the whole idea I believe is to be able to decide a new reading is noise and not showing its whole deviation at the time.  The problem is that at the time you get a new reading that has moved significantly out of the sequence, there is no way to know if it is noise that has caused it to move away or if it is a real deviation caused by insulin or just having eaten something.
Both of those require the human factor.  There is no way for the software to decide if a deviation is noise at the time that it has just received it before it receives the subsequent readings.

As much as there are frequent requests for adding such intelligence to xDrip, I have always stated we cannot.
If we do, we could cause a real deviation not to be reported or to be reported later than it could have been reported.

Another way of thinking about the problem is a choice we have to make between two imperfect outcomes:
1- Occasionally, we have to put up with a false alert.
2- Occasionally, we have to put up with a delayed alert.

If you compare 1 and 2, I hope you see we have no choice but to go for 1.
A delayed low alert could result an unforgiving outcome.

Thanks for your contribution.
Everything I state is only my opinion.  I am not the owner of this repository.

jbr7rr:
> Forgive me for posting here now that it is closed. I have seen it for the first time considering it is related to Libre and I often leave Libre to others. But, now that it was closed, I had a look at it and noticed for the first time that it is related to noise. I would like to say one thing considering for the record.
> 
> It can be very misleading if you only look at a graph after the fact. It will be obvious then what was noise and what was real. But, the whole idea I believe is to be able to decide a new reading is noise and not showing its whole deviation at the time. The problem is that at the time you get a new reading that has moved significantly out of the sequence, there is no way to know if it is noise that has caused it to move away or if it is a real deviation caused by insulin or just having eaten something. Both of those require the human factor. There is no way for the software to decide if a deviation is noise at the time that it has just received it before it receives the subsequent readings.
> 
> As much as there are frequent requests for adding such intelligence to xDrip, I have always stated we cannot. If we do, we could cause a real deviation not to be reported or to be reported later than it could have been reported.
> 
> Another way of thinking about the problem is a choice we have to make between two imperfect outcomes: 1- Occasionally, we have to put up with a false alert. 2- Occasionally, we have to put up with a delayed alert.
> 
> If you compare 1 and 2, I hope you see we have no choice but to go for 1. A delayed low alert could result an unforgiving outcome, which should be avoided at all cost.
> 
> Thanks for your contribution. Everything I state is only my opinion. I am not the owner of this repository.

Well the intent of this PR was to only report the noise level to AAPS (there was/is an ancient mechanism in oref1 that could handle that and adjust the predictions accordingly). So it could be handled there.

Indeed I agree noise detection has some issues, however the algorithm could account for normal spikes as it did an polynomial fit. It could detect deviations from a normal looking pattern so to say. Main issue would still be indeed false positives caused by the sensitivity of the detection. This could be reduced by tweaking of the parameters but would require a larger sample size then I had available.

But I have moved on from Libre since then, and the development continued, AAPS has accepted Libre 2/3 readings now for SMB Always, so I think things should have been improved on that front and thus this PR does not add much anymore.

Issue #1329, opened 2020-05-05
Support Libre Pro sensors


ipsita1917:
Hi.. am using the xdrip+ latest nightly build and trying to connect to a libre pro sensor with a BluCon. But every time after I do the initial calibration, xdrip gives a flat line of bg values i.e. the same value I calibrated with  What could I be doing wrong?

ipsita1917:
![Screenshot_20200506-000016](https://user-images.githubusercontent.com/64782570/81102069-af43a380-8f2c-11ea-963d-128d61586abb.jpg)
This is the update log screenshot

tzachi-dar:
What is libre pro ?
What country? How many days do they work?

בתאריך יום ג׳, 5 במאי 2020, 21:32, מאת ipsita1917 ‏<notifications@github.com

ipsita1917:
I'm in India. The sensor works for 14 days, it's meant for healthcare professionals to assess patient values, and is the only libre available here. Have been getting readings on the BluCon app but not on the xdrip.

tzachi-dar:
I'm not sure xDrip supports the indian sensors. Does it work for other people in India?
Does it work when scanning it with the phone?

ipsita1917:
I'm not sure it works for other people as I don't know anyone else using it right now. It does work with scanning with Glimp and it works on the LinkBluCon app. I just wanted to use it with xdrip....

Navid200:
@ipsita1917 You can ask in the support channels.
Many more frequent those and you have a better chance of finding a solution if one exists.
https://www.facebook.com/groups/xDripG5
https://gitter.im/jamorham/xDrip-plus

Navid200:
Closing due to inactivity

anirudhreddy28:
@Navid200 copying the feature request 1677 (duplicate) I had submitted. 

**Is your feature request related to a problem? Please describe.**
We notice that Xdrip currently does not support Libre pro sensors. We in India primarily use Libre Pro as our CGM and we are struck with using the Native apps like Tomato (for Miao Miao) and Link Blucon (Ambrosia Blucon). Most of them are happy with the Native apps but with the recent server issues Miao Miao had, we are interested to see a possibility of using these DIY apps. There are hundreds (possibly thousands) who could benefit with this addition.

**Describe the solution you'd like**
We know that Xdrip already supports Libre 1, Libre US 14 day, Libre US 10 day etc.. We would like the similar support for Libre Pro as well

[...]

PR #2075, opened 2022-04-12
Initial support for libre pro.

tzachi-dar:
Based on information and code that I have received from @keencave 

This is not tested as I don't have libre pro sensors.
Would be nice if someone with such a sensor can check this.
Should work with direct nfc scan or devices like miaomiao.

Can be downloaded from: https://drive.google.com/file/d/1VGVvnx-ktEsvHfuxSHm9RJ8KsJtWuIsF/view?usp=sharing

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
@@ -86,6 +86,16 @@ public class NFCReaderX {
     private static boolean nfc_enabled = false;
     private static final String ENABLE_BLUETOOTH_TIMESTAMP = "enable_bluetooth_timestamp";
 
+    // Constants for libre1/2 FRAM
+    final static int FRAM_RECORD_SIZE = 6;
+    final static int TREND_START = 28;
+    final static int HISTORY_START = 124;
+
+    // Constants for libre pro
+    static final int LPRO_SENSORMINUTES = 74;
+    static final int LPRO_TRENDPOINTER = 76;
+    static final int LPRO_TRENDOFFSET = 80;
+
     // For libre2 emulation only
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
```diff
@@ -862,9 +862,13 @@ public static ReadingData parseData(int attempt, String tagId, byte[] data, Long
                 historyList.add(glucoseData);
             }
         }
+        return historyList;
+    }
 
-
+    // Get the trend data for libre1/2
+    private static ArrayList<GlucoseData> parseTrendData(byte[] data, int sensorTime, long sensorStartTime, Long CaptureDateTime) {
```
tolot27, line 869: `parseHistoryData` and `parseTrendData` are almost identical with the exception of the index (26 vs 27) and data size (16 vs 32). To reduce code duplication these two functions could be merged.

tolot27 (approved):
LGTM

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
gui-dos: @tolot27: No: in the Libre Pro parsing the history data is a totally different matter: the historic data for the whole 14-day sensor life are stored in a contiguous memory space following the body and the two-byte pointer is stored at `78/79`. To read the FRAM blocks with an index higher than 255 you have to use the NFC custom commands `B0` (single block) or `B3` (maximum three blocks).

keencave:
It should fix #1329. As referenced there the LimiTTer transmitter supports Libre Pro. That code base was used for this PR.

app/src/main/java/com/eveningoutpost/dexdrip/NFCReaderX.java
tolot27: @gui-dos I've just compared `parseHistoryData` and `parseTrendData` and the code is identical with the exception of the two mentioned cases. These can be parameterized. If the Libre Pro history ist totally different than something else should be implemented.

tzachi-dar:
I think that parsing the history is not realy feasible. We are using devices like miaomiao and their fw needs to change in order to support that. From what I read in their website they do not support history as well.

gui-dos:
The MiaoMiao is still sending 21 further blocks for a usual total of `5 + 4 +13 + 21 = 43` blocks. It is not clear when from the past those partial 28 historic measurements come from (`21 * 8 / 6 = 28`) but the @bubbledevteam clearly is converting the Libre Pro data into the Libre 1/2 layout: https://github.com/bubbledevteam/bubble-client-swift/blob/master/BubbleClient/LibrePro.swift

A really intriguing thing I discovered is that the NFC custom command `B0` and `B3` are also supported by the Libre 2 and you can read a maximum of 89 encrypted blocks. After sending the encoded NFC custom command `A1 1A` the sensor replies with decrypted data (you can dump up to 1252 blocks). You can notice that there are two copies of the 43 FRAM blocks and that the first copy of the extended header and footer correspond to the sizes `5 + 4` of the  Libre Pro ones (null bytes are appended): 43 + 2 + 1 + 43 = 89! 

Notice that the additional 2 blocks of the Pro header contain the serial number of the reader (14 ASCII chars) and further two bytes that you have to append to the command `A0` to activate a Libre Pro sensor.

keencave:
I agree on history readings. The need to have modified firmware in the given transmitter models to even access the history data structures  is not feasible.

prakashdevadas69:
Hi,
I tried but was not able connect to Libre Pro and Miao Miao 3
It was stuck with collecting initial data
3B487B81-F445-4952-8237-3818927EF3A5

tzachi-dar:
Can you please see what you have in your logs? 
Please go to the 3 dots on the right, press view events log and let us see what you are getting.

Thanks
Tzachi

prakashdevadas69:
This is the log

tzachi-dar:
There was a problem with the checksum.
Please try again, I think that I have found the error.
https://drive.google.com/file/d/1VGVvnx-ktEsvHfuxSHm9RJ8KsJtWuIsF/view?usp=sharing

prakashdevadas69:
Still stuck at receiving data
3613211B-EEC7-4F84-B3FB-A9ED4E6C5FA4

prakashdevadas69:
After stopping sensor and restart sensor got this
7A267908-D7B0-45E6-8C85-D9A6B50ED672
95FE52D0-0D09-4FCB-957F-6EAA9B7684C6

tzachi-dar:
Good thing is that the logs show that you have received two different points, so we are making progresses.

Next step is to enable more logging, collect data of 30 minutes, and send it to me:
Please go to settings->less common settings ->logging->extra tags for logging and add there:
"BgReading:d,jamorham librereceiver:v,LibreOOPAlgorithm:v,jamorham nsemulator:v,DexCollectionService:v"

Than please press on the 3 points on the right side, import export features->export database.
When back is complete, please press share and send it to tzachi.dar@gmail.com

tzachi-dar:
A short update.

The latest version has been working well on libre pro sensors. I have added more checksum checks and more tests to see that nothing is broken. 

Here is a link to the latest version:
https://drive.google.com/file/d/1ORxDCa-A9Aig3cv3zUjKgXuNOr9qNbVX/view?usp=sharing

saurabh1480:
Thank to all developers for libre pro version for india,thabk you very much from all

Navid200:
@saurabh1480 Does it work?

prakashdevadas69:
> @saurabh1480 Does it work?

Yes it does work like a GEM..!!..thank you for making this change..

Navid200:
Thanks go to @tzachi-dar and @keencave

jamorham:
Please confirm this is tested and ready for merge? I have other changes relating to Libre so I will merge this first if its looking good. @tzachi-dar

prakashdevadas69:
> Please confirm this is tested and ready for merge? I have other changes relating to Libre so I will merge this first if its looking good. @tzachi-dar

Yes..XDrip is working fine in all aspects with libre pro..
Thank you .. and you may kindly merge with main release

jamorham (approved):

terra-alex:
Just a quick comment on this - can we still not get history values from the libre pro? It seems there are a few major caveats with the current code for it

1. The trend value only returns a single reading
2. The history values are not being processed (the manual reader seems to be able to read up to all 14 days of data)

This means that we'd have to keep scanning the sensor every 15 mins? Whereas the whole point of the sensor is not to need to scan it all the time

PR #2076, opened 2022-04-12
Move everything from the title to the summary

Navid200:
@jamorham Is this what you meant?

This is an alternative solution that fixes: https://github.com/NightscoutFoundation/xDrip/issues/553  

This was my original attempt:  https://github.com/NightscoutFoundation/xDrip/pull/1702

This is on the virtual device in Android Studio with a resolution of 240x400.

|Before:|After:|
|:---|:---|
|Screenshot_20220412_234640|Screenshot_20220412_192249|

 And this is what it looks like on a large screen phone.
Screenshot_20220412-191105

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
@@ -822,7 +822,37 @@ public class Preferences extends BasePreferenceActivity implements SearchPrefere
         }
     }
 
+    private static void bindPreferenceSummaryAppendToIntegerValueFromLogSlider(Preference preference, NamedSliderProcessor ref, String name, boolean unitize) {
 
+        final Preference.OnPreferenceChangeListener listener = new Preference.OnPreferenceChangeListener() {
+            @Override
+            public boolean onPreferenceChange(Preference preference, Object value) {
+
+                boolean do_update = false;
+                // detect not first run
+                if (preference.getSummary().toString().contains("(")) {
+                    do_update = true;
+                }
+                final int result = ref.interpolate(name, (int)value);
+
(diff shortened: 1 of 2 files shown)


tolot27 (changes requested):
This is a much better solution than the previous one. Great!

`bindPreferenceTitleAppendToIntegerValueFromLogSlider` is also used for one more preference: "tidepool_window_latency" (Data Age Mins). Please can you change it there as well. Afterwards, `bindPreferenceTitleAppendToIntegerValueFromLogSlider` can be removed completely.

Navid200:
@tolot27 
There are many more seekbars in xDrip.  I can list them all for you if you like.  The seekbar title does not wrap for any of them.

We have an instruction from @jamorham not to create large PRs.  That has always been my approach because I agree with that direction.  
If a PR improves things, please don't delay it because it does not fix everything.

I respectfully request @jamorham to approve your request before I execute it.  
I personally have no objection in editing all the xDrip seekbars.  But, if that is going to be the reason this PR will never be merged, because it is too big, I'd rather not do it in one step.

tolot27:
> @tolot27 There are many more seekbars in xDrip. I can list them all for you if you like. The seekbar title does not wrap for any of them.

But after your PR only "tidepool_window_latency" remains which uses `bindPreferenceTitleAppendToIntegerValueFromLogSlider`. I did not request to fix all of them. I just request to modify all affecting seekbars which is just one in this case.

> We have an instruction from @jamorham not to create large PRs.

This is not a large PR and my suggestion would further reduce it. Currently, your PR duplicates code which increases maintenance efforts.

> If a PR improves things, please don't delay it because it does not fix everything.

I don't delay anything here.

jamorham:
I think this looks nicer and should solve the original issue. I haven't tested it myself. Smaller PRs are easier to merge.

jamorham (approved):

PR #2079, opened 2022-04-14
Replace silent mode with DND

Navid200:
If you search for difference between dnd and silent mode, you will get some hits.
There are individuals who may think that if they lower the volume on their phone to 0 (silent mode) and enable "Override silent mode", the alert will still play.

To avoid such misunderstandings, it is best to rename the setting to Override DND.
The following images show what it looks like after this update.
Screenshot_20220414-104836
Screenshot_20220414-104856
Screenshot_20220414-104923

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -535,7 +535,7 @@
     <string name="select_time_for_alert_colon">Select time for alert:</string>
     <string name="all_day">all day</string>
     <string name="tap_to_change">(tap to change)</string>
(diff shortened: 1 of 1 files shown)


jamorham:
I'm not sure that all Samsung and MIUI phones actually use DND and instead have a concept of silent alert volume similar to Android 4 & 5 which we also still support.

Most significantly though is that xDrip doesn't override DND mode. To do that it would need to use the interruption filter features from sdk 23+ which it doesn't afaik currently do. All it does is adjust the audio volume, eg bypass silent mode.

This wording change seems inappropriate to me.

Navid200:
@jamorham Understood.

PR #2081, opened 2022-04-16
Time span feature

Navid200:
This is a fresh attempt at an older PR I am closing now:  https://github.com/NightscoutFoundation/xDrip/pull/1729

If you scroll back on screen, you can go back to 24 hours ago. After this PR is merged, you can go back more or less based on a setting.

This is the new setting:
Screenshot_20220419-230109

Screenshot_20220419-225737
  
The default 24 hours, which is the current value before this feature is added, is marked on the page.  So, the user can clearly see what the default is.

Fixes https://github.com/NightscoutFoundation/xDrip/issues/650

Why do we need this?
We don't! It is a feature request. It can be handy.

| Time span setting | Main screen |  
|:--------------------:|:--------------:|  
| 12 hrs | Screenshot_20220418-085929 |  
| 24 hrs | Screenshot_20220418-085850 |  
| 26 hrs | Screenshot_20220418-085803 |  
| 30 hrs | Screenshot_20220418-085718 |  
| 36 hrs | Screenshot_20220418-085607 |  

When your blood glucose goes too high, the vertical axis max is automatically increased so that the high readings are visible on screen.  The max will remain modified as long as the high readings are available on screen.  So, by default, you will have to wait 24 hours for the vertical axis max to return to normal.  Changing the time span will increase/decrease how long you will have to wait in that case.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -2019,7 +2019,7 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
             Viewport moveViewPort = new Viewport(chart.getMaximumViewport());
             float tempwidth = (float) moveViewPort.width() / 4;
             holdViewport.left = moveViewPort.right - tempwidth;
-            holdViewport.right = moveViewPort.right + (moveViewPort.width() / 24);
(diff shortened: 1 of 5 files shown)


Navid200:
Let's table this for now.  I need to fix the conflicts.  But, let's see if this is really a priority before spending the time.

PR #2082, opened 2022-04-19
Ensure current device is Dexcom

Navid200:
This is a bug that could cause hardware devices other than Dexcom, like Libre, malfunction if the user switches from Dexcom.

To test this in action, one needs to have a functioning Libre as well as a FireFly G6 transmitter that still has battery.
I don't have a Libre device.  So, I cannot reproduce the problem myself.

If you have those and are willing to test, please install an xDrip release after August 4, 2021 like the current stable release; set xDrip to connect to your G6 transmitter and make sure it establishes connectivity and shows the battery voltages.  The Firmware should confirm that it is FireFly (Greater than 1.6.5.27).

Now, switch to your Libre.  Attempt a calibration.  It should malfunction as the logic (my mistake) still detects a G6 even though you have switched to Libre.

Now, update to this PR.  Your Libre should calibrate properly now.


jamorham:
This seems okay in my tests

Navid200:
@jamorham Thanks for identifying the bug, suggesting a solution, and testing it as well.

jamorham (approved):

PR #2083, opened 2022-04-20
Remember long pressed time buttons

Navid200:
If you bring the time buttons on screen and long press any of them, you will be defining the default time span.
After that you can pinch the screen to change the span.  As soon as you minimize xDrip and reopen, the time span will go to your chosen default.

However, if you remove the time buttons from the screen, xDrip ignores your chosen default.

This PR changes that last behavior.  Even if you remove the time buttons from the main screen, the one that you have long pressed will still define your default time span.

Fixes: https://github.com/NightscoutFoundation/xDrip/discussions/1910

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -2025,15 +2025,10 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
             chart.setCurrentViewport(holdViewport);
             previewChart.setCurrentViewport(holdViewport);
         } else {
-            if (homeShelf.get("time_buttons")) {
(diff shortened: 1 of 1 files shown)


Navid200:
@jamorham I would like to set a preferred value.  But, I prefer not to have those button take so much space on the main screen.
Right now, the second preference wins for me.  So, I cannot choose a preferred span.

Am I the only one?
Would you entertain making this a setting?  I can do it.

What do you suggest?

jamorham:
I think the best way to handle this is with an additional home shelf preference to use the time period always. I can add this.

jamorham:
See here: https://github.com/NightscoutFoundation/xDrip/commit/4e868c49fc7e82526be2d53a1e106b6a6d01cd58 please let me know if this meets your needs.

Issue #2074, opened 2022-04-11
Testers needed for NovoPen NFC syncing


jamorham:
I have a working open implementation for syncing dose information from the NovoPen 6 and NovoPen Echo Plus which I've been testing internally.

I am planning to make this available shortly in a nightly. This issue ticket is to find people willing to test the feature, receive feedback and diagnose any problems or incompatibilities.

So far, the way it works is that the preference feature must be enabled from within the `Insulin Pens` preference screen and then the pen display is touched against the NFC receiver of the phone while the xDrip+ home screen is open. 

At this point it will attempt to retrieve the last 7 days worth of dose information from the pen and then present these as xDrip treatment items within the display/history log. Duplicate data will be discarded.

Each treatment item contains a note entry giving the pen serial number which it was retrieved from. I would say that this is best suited for logging rapid-acting doses. For long-acting you can use the multiple insulins feature of xDrip and designate each pen to the type of insulin it contains.

In a similar way that the glucose meter feature works, optional noises are generated to indicate data being successfully synced.

To delete all synced pen treatments there is a voice command for testing: `delete all pen data`

Coexisting with Libre should now be working.

If you prime your pen with a small dose before the real dose then you can specify the maximum size of a priming dose and the time window maximum between the priming and real dose and then priming doses can automatically be removed from the data.

For each dose you can tap on the item on the graph and select options such as adding a note, zeroing the value (if it was not actually a real dose) or with multiple insulins changing the future designation for that pen. (If you change what is in which pen etc)

If you are interested in testing this, please add a comment below with your phone model and android version and pen type. Also please make sure you have switched to the Nightly update channel so that you can receive the update when it is available.

@madsviggers @retroneo @PieterDeBruijn @utzaki @yoshimo @jtxgit @PaarnTh

aleks372:
Interested, Galaxy S21 Ultra (SM-G998B) and Novopen Echo Plus

utzaki:
WOW! Good to hear! Sure I will do everything to help to test.

Echo Plus, Galaxy S7. Unfortunately, the S7 is my old mobile, which is not in regular use anymore. But for a test, I could install xDrip+ there. My actual mobile is not compatible with Novopen 6.

PaarnTh:
I'l help. 
Samsung S21FE (SM-G990B/DS)
Latest android (Android 12)
Novopen ehco plus. 2 of them.

madsviggers:
[...]

PR #2086, opened 2022-04-24
Allow Nightscout followers to view treatments added by other xDrip Nightscout followers.

adamecker:
Append unique app instance id to XDRIP_TAG so that multiple xdrip followers can add treatments that are propagated to everyone.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/Treatments.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/Treatments.java
@@ -31,6 +31,7 @@ import com.eveningoutpost.dexdrip.insulin.MultipleInsulins;
 import com.eveningoutpost.dexdrip.utils.jobs.BackgroundQueue;
 import com.eveningoutpost.dexdrip.watch.thinjam.BlueJayEntry;
 import com.eveningoutpost.dexdrip.xdrip;
+import com.google.firebase.iid.FirebaseInstanceId;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.annotations.Expose;
@@ -70,7 +71,7 @@ public class Treatments extends Model {
     public static final String SENSOR_STOP_EVENT_TYPE = "Sensor Stop";
     private static final String DEFAULT_EVENT_TYPE = "<none>";
 
-    public final static String XDRIP_TAG = "xdrip";
+    public final static String XDRIP_TAG = "xdrip_" + FirebaseInstanceId.getInstance().getId();
 
     //public static double activityMultipler = 8.4; // somewhere between 8.2 and 8.8
     private static Treatments lastCarbs;


tolot27 (commented):
In general, this PR seams a straightforward solution to the problem but  I'm not sure whether `lastNotFromXdrip` and `lastEventTypeFromXdrip` will work correctly afterwards.
From the local perspective they will work correctly, indeed, but not to distinguish xDrip from non-xDrip treatments.
We have to check where this functions are used. Indeed, they can be reimplemented by using the original tag.

Navid200:
Just two lines?!  
Can you fix all the other xDrip problems too?   :o)

Have you tested this?  
I don't understand why the checks have failed.

jamorham:
I think the unit test failure illustrates that it isn't good practice to add a dependency like firebase to the static initialization of the treatments class. If we need a unique id then this needs to be handled carefully. I could add a unique id class which could use persistent store for the backing. This would only be unique per install/uninstall of xDrip but I believe that is all that is needed for this use case? I'm also concerned about the potential for problems in the de-duplication to be an unwanted side effect of a change like this. What happens if xDrip sync is also enabled? Does that impact on anything? What about uninstall/reinstall of an instance, does that correctly work?

tolot27:
> I think the unit test failure illustrates that it isn't good practice to add a dependency like firebase to the static initialization of the treatments class. If we need a unique id then this needs to be handled carefully. I could add a unique id class which could use persistent store for the backing. This would only be unique per install/uninstall of xDrip but I believe that is all that is needed for this use case? I'm also concerned about the potential for problems in the de-duplication to be an unwanted side effect of a change like this. What happens if xDrip sync is also enabled? Does that impact on anything? What about uninstall/reinstall of an instance, does that correctly work?

I don't think we need a global ID appended to all treatments. We need it just for the manually inserted treatments. The insulin treatments read from a NovoPen (introduced with #2074) are already synced to the NS followers. 

The article [Best practices for unique identifiers](https://developer.android.com/training/articles/user-data-ids) describes the different identifier use cases in Android and [Resettability and persistence](https://developer.android.com/training/articles/user-data-ids#resettability-persistence) describes the durability of the IDs. I think *Install-reset* would be sufficient for our use case but *FDR-reset* would be slightly better because every nightly update changes the ID. The Firebase installation ID (FID) should be *FDR-reset* persistent.

@jamorham Please provide the suggested unique id class. This should solve all concerns also mentioned in https://github.com/NightscoutFoundation/xDrip/pull/1403#issue-679451352.

jamorham:
Please see the new `UniqueId` class: https://github.com/NightscoutFoundation/xDrip/commit/8ee13f894b1341534528b4662e624d55d969b605

To use to obtain a per-installation 64 bit (16 character) string, simply use `UniqueId.get()`

Its also possible to specify the number of characters (up to 32) like: `UniqueId.get(8)` depending on how much uniqueness you need and the chance of a collision vs how nice and short the string is. So for things like nightscout syncing between multiple handsets I'm thinking a 32 bit (8 character) string would be more than sufficient.

Navid200:
If you set up a follower to upload treatments to Nightscout, other followers can see the treatments as long as you disable the switch:
"Skip items from xDrip".  

I don't believe there is a problem anymore in that respect.

Would you please tell me if you disagree?
Otherwise, would you please close this PR?

PR #2094, opened 2022-04-30
treat raw as float for recent g6

tynbendad:
for https://github.com/NightscoutFoundation/xDrip/discussions/2080
compiles, builds, but i still haven't seen raw data from xdrip+

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/FirmwareCapability.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/FirmwareCapability.java
@@ -15,7 +15,7 @@ public class FirmwareCapability {
     private static final ImmutableSet<String> KNOWN_G5_FIRMWARES = ImmutableSet.of("1.0.0.13", "1.0.0.17", "1.0.4.10", "1.0.4.12", "1.0.4.14", "1.0.4.15");
     private static final ImmutableSet<String> KNOWN_G6_FIRMWARES = ImmutableSet.of("1.6.5.23", "1.6.5.25", "1.6.5.27");
     private static final ImmutableSet<String> KNOWN_G6_REV2_FIRMWARES = ImmutableSet.of("2.18.2.67", "2.18.2.88", "2.18.2.98", "2.24.2.88", "2.27.2.98", "2.27.2.103");
-    private static final ImmutableSet<String> KNOWN_G6_REV2_RAW_FIRMWARES = ImmutableSet.of("2.18.2.67");
+    private static final ImmutableSet<String> KNOWN_G6_REV2_RAW_FIRMWARES = ImmutableSet.of("2.18.2.67", "2.27.2.103");
     private static final ImmutableSet<String> KNOWN_G6_PLUS_FIRMWARES = ImmutableSet.of("2.4.2.88");
     private static final ImmutableSet<String> KNOWN_TIME_TRAVEL_TESTED = ImmutableSet.of("1.6.5.25");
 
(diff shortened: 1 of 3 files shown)


tynbendad:
I was able to test w/anubis and the raw (in the BG data table) is still showing as "-.13", so there's something else going on besides this. Its beyond me what unless someone else has suggestions for me to look at. We will need this change as well if the other issue is ever fixed.

tynbendad:
ahah, i see raw isn't being requested because of a firmware version check (isFirmwareRawCapable) that the anubis isn't passing... any tips on how to change this would be appreciated, for now i'll try just adding the anubis' version number to the current check.

tolot27:
@joshbohmer Any ideas?

> ahah, i see raw isn't being requested because of a firmware version check (isFirmwareRawCapable) that the anubis isn't passing... any tips on how to change this would be appreciated, for now i'll try just adding the anubis' version number to the current check.

joshbohmer:
> isFirmwareRawCapable

rather than detecting by firmware version, the correct method would be to issue 0x2E and check for a valid 0x2F response. This would make the raw check version agnostic.

The other option would be to detect the 180 day expiry, but that feels very workaround(ish) to me.

tynbendad:
raw is working for me now with the latest changes in this PR, and i've updated the test so it should also pass.

changing the version checks is beyond me - there are way too many versions and checks, i don't feel comfortable removing them, but as Josh says, the sensor tx/rx message exchange should be enough to know whether raw is working or not.

Navid200:
You cannot do this.
You have modified the firmware detection fine.  This affects all users, not just the users of modified transmitters.

2.27.2.103 is a real firmware.
Either change the modified transmitters so that they identify themselves, with a unique firmware code, or please leave that file alone.

jamorham:
@tynbendad can you provide some extracts from logs with the raw values being received from the transmitter prior to being interpreted by the scaling. These should be visible with OB1 debugging enabled.

tynbendad:
> You cannot do this. You have modified the firmware detection file. This affects all users, not just the users of modified transmitters.
> 
> 2.27.2.103 is a real firmware. Either change the modified transmitters so that they identify themselves, with a unique firmware code, or please leave that file alone.

What does xdrip do when the sensor tx message gets an error instead of a reply?

tynbendad:
> @tynbendad can you provide some extracts from logs with the raw values being received from the transmitter prior to being interpreted by the scaling. These should be visible with OB1 debugging enabled.

There is a bunch of data for these in the discussion thread - let me know if that is sufficient, https://github.com/NightscoutFoundation/xDrip/discussions/2080.

Navid200:
You can ignore my message.  

I suggest that you follow what jamorham has asked you.  Convince him that you are not breaking xDrip for anyone else, which is what all of us have to do when we open a PR.

tolot27:
> You cannot do this. You have modified the firmware detection file. This affects all users, not just the users of modified transmitters.
> 
> 2.27.2.103 is a real firmware. Either change the modified transmitters so that they identify themselves, with a unique firmware code, or please leave that file alone.

He did nothing wrong and did not remove it from the general firmware detection

Navid200:
@tolot27 Did you see my post?  Or, do you just like to argue?

jamorham:
I plan to include additional support in `FirmwareCapability` which will detect whether the transmitter can provide raw without relying solely on the version number.

jamorham:
I've added passive support for detecting transmitters which will likely work for raw. It doesn't need any changes to the version number list or additional bluetooth sequences. https://github.com/NightscoutFoundation/xDrip/commit/de0a352294b4d040b2a9d83d061e83061eeb8a6a It should now just work and those related changes in this PR can be backed out.

I'd still like to see a bunch of the numbers actually being returned via bluetooth before they are processed so that we can verify correctness of the processing/scaling. Thanks.

tynbendad:
@jamorham see my reply above for sensor tx data packets

jamorham:
@tynbendad thanks, I hadn't spotted that link. I will review.

tynbendad:
@jamorham have you had a chance to look at the data?

tynbendad:
@tolot27 i see you added enhancement tag but i think the current impl of raw for recent g6 is broken so this should be considered a fix.

Navid200:
> i think the current impl of raw for recent g6 is broken so this should be considered a fix.

There is no implementation of raw for recent G6 as the recent G6 does not transmit raw.
Raw works perfectly fine for what it was meant for, G5 and the old G6.

This, indeed, would be an enhancement and not a fix.  And it will need to be done very carefully avoiding any negative impacts on G5 and old G6.
Have you tested your PR with G5?  How about with an old G6?

tynbendad:
There was a version of g6 firmware that is already allowing raw with the broken calculation. And this change only affects that version, or the Anubis with the new check @jamorham recently added. I will not be testing any other devices so if you're waiting for that then we can close this

Navid200:
>  if you're waiting for that then we can close this

It doesn't matter what I am waiting for as long as jamorham is happy with the submission.  That's all you need for it to be merged.
I post here to communicate my thoughts for anyone who cares to listen.

G5 is still being sold.  There are many who still use it.  In my opinion, one way or another, we should be certain this will not have a negative impact on G5 before merging it.

tynbendad:
fwiw G5 should be going thru the "default" case in RawScaling.java, which is not touched by my change, since DType has 3 enums (G5, G6v1, G6v2). whether or not "should" actually happens I don't know.

Issue #1580, opened 2020-12-28
treatment bg values are silently used for calibration despite answering "no" or setting to "never"


gfornax:
Like the headline says, my xdrip uses treatment bg values for calibration no matter what I do.
I noticed this several months ago but stopped caring about it since the G6 calibration still seems broken and doesn't work for me since the firefly transmitters. Maybe this affects other data sources as well.
Nonetheless, it is a major safety issue, as usually users have a good reason for not wanting a certain treatment value to be used for calibration - e.g., when they now the slope is very high. I personally ended up with some sensors that were dangerously off due to this bug.

### environment
same on stable releases, 20200713 nightly , 20201001 nightly, 20210217 nightly.
Dexcom G6 native, 8H transmitter.
Setting "Use Treatment BG values" set to "Ask me every time" or "Never use them"
Automatic Calibration off
no Calibration Plugin
Using mg/dl

### Expected behavior
If I answer the dialogue with "No" or set the setting to "Never", I expect a treatment BG value to show up on the graph and nothing else.

### Actual behavior
After entering the value, instantly there is a calibration inserted to the queue items. And this is (as long as it was still working with the older transmitters) actually sent to the transmitter.

### Steps to reproduce the behavior:
Set settings like described above
Main screen, tap treatment, select fingertip button, enter a value, confirm
If xdrip is configured to "ask", select "No" in the dialogue box.
Check "System Status", calibration is queued.

Navid200:
Can you try this with a more recent release?  For example, January 29, 2021.

Navid200:
Are you sure calibrations are submitted to the transmitter.  Do you see the log saying calibration was accepted by the transmitter?

gfornax:
I just tried with the latest nightly build. It still happens, I can answer both questions with a definite yes.
Despite answering "no" and despite xdrip showing the _Sensor data fails sanity test - Cannot Calibrate! raw:-0.1..._ message.
The calibration is shown right after the next sync in the dexcom transmitter.

Navid200:
Would you mind posting a screenshot that shows the receiver showing the result of the calibration?

gfornax:
Sure, here you are. Oh and it is already up now, no worries.

![IMG_20210223_183010](https://user-images.githubusercontent.com/5310025/108883291-8ef0ec80-7605-11eb-9b49-fe699ca55414.jpg)
[...]

PR #2098, opened 2022-05-03
No, don't use for calibration

Navid200:
To reproduce the problem, go to Settings &#8722;> Less common settings &#8722;> Advanced Calibration &#8722;> Use Treatment BG values
Select "Ask me every time".

Now, go to the main screen and use the treatment menu to enter a blood glucose measurement.
When it asks you if you want to use the value for calibration, answer "No".  It will still use it for calibration.
You can clear the queue to undo calibration.

This PR fixes: https://github.com/NightscoutFoundation/xDrip/issues/1580

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -770,6 +770,9 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
             Log.d(TAG, "ProcessCalibrationNoUI number: " + glucosenumber + " offset: " + timeoffset);
 
             final String calibration_type = Pref.getString("treatment_fingerstick_calibration_usage", "ask");
+            Log.d(TAG, "Creating blood test record from input data");
+            BloodTest.createFromCal(glucosenumber, timeoffset, "Manual Entry");
+            GcmActivity.syncBloodTests();
             if (calibration_type.equals("ask")) {
                 AlertDialog.Builder builder = new AlertDialog.Builder(this);
                 builder.setTitle(gs(R.string.use_bg_for_calibration));
@@ -783,9 +786,6 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
                 });
 
                 builder.setNegativeButton(gs(R.string.no), (dialog, which) -> {
-                    Log.d(TAG, "Creating bloodtest record");
-                    BloodTest.createFromCal(glucosenumber, timeoffset, "Manual Entry");
-                    GcmActivity.syncBloodTests();
                     dialog.dismiss();
                 });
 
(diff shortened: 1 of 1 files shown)


jamorham:
Thanks for this, appears at first glance as a well overdue todo. Can you confirm that after you say no you can then click on the blood test and calibrate later from it?

Navid200:
@jamorham I just tested:
After answering no, it becomes a square marker on the main screen.  Tapping on it, shows time and value at the bottom as well as "BLOOD TEST" in blue at the bottom right of the screen.  Tapping on blood test, brings up a form that lets you do nothing, delete or calibrate.

I have to add one thing here:
I don't know why after tapping on no, nothing shows on screen.  One has to go to logs and come back to the main screen to see the square marker.  Or, one has to minimize xDrip and open it up again to see the square marker.
I don't remember this behavior.  

I just tested with never and the marker appears right away.  I need to figure out why.  I will find out and correct.

jamorham:
add a `staticRefreshBGChartsOnIdle()` call to your dialog lambda

Navid200:
Thanks
But, why does the same problem not occur for never?!
Is this an obvious demonstration of me not being a programmer?  :o)

Navid200:
@jamorham auto also has a dialog.  Why does it not happen with auto?

Navid200:
The behavior of the marker is now consistent across all possible options.
I have tested them all.

jamorham:
Can you describe how the logic flow is changed by this and why this fix works? thanks

Navid200:
@jamorham 
Regardless of which setting the user chooses on the form shown next, we need a marker to be placed.


But, we have repeat code for that and I am actually responsible for that:
https://github.com/NightscoutFoundation/xDrip/pull/1814/commits/e25d6e62ae47643426a70b28f67f80125631c66c
In fact @tolot27 caught that then and pointed it out.  But, I didn't get it.

When I first opened this PR yesterday, I had added a third instance of this code, which happened to be inside a dialog.  The consequence of that was that it caused an inconsistent behavior for the marker placement.

So, I have taken the part of the code that creates the marker and have moved it ahead of all the conditionals:
Screenshot 2022-05-04 081856
Now, all marker placements inside the conditionals are redundant.  So, I have removed those.

Now, the dialog has no impact on the marker.
I hope this makes sense.

Navid200:
If you want to test this, please note:

1- After you enter a blood glucose value using the treatment menu, if you delete it, you cannot apply another one right away.  You need to wait for the next read cycle (5 minutes).  This limitation is not caused by this PR.  If you don't like that behavior, we will need another PR to change it.

2- If you are testing the "Auto" option, please pay attention to this memory:
Screenshot 2022-05-04 100523
That is one month!  

That means after you test it and get a dialog asking you if you want to enable automatic calibration or not, and you respond, it's going to remember your response for one month and will never ask you again.
So, if you try again and don't see the question, again, please don't think this is caused by this PR.
The only way to test it again is to uninstall xDrip, which I don't recommend.  Do it on a test phone, not your mission mode phone.

PR #2103, opened 2022-05-06
remove last usage of bindPreferenceTitleAppendToIntegerValueFromLogSlider

tolot27:
This is a code cleanup of #2076 to remove code duplication. The function `bindPreferenceTitleAppendToIntegerValueFromLogSlider` is not used anywhere else.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
@@ -1849,7 +1849,7 @@ public class Preferences extends BasePreferenceActivity implements SearchPrefere
 
 
             final NamedSliderProcessor tidepoolProcessor = new UploadChunk();
-            bindPreferenceTitleAppendToIntegerValueFromLogSlider(findPreference("tidepool_window_latency"), tidepoolProcessor, "latency", false);
+            bindPreferenceSummaryAppendToIntegerValueFromLogSlider(findPreference("tidepool_window_latency"), tidepoolProcessor, "latency", false);
 
 
             wifiRecievers.setOnPreferenceChangeListener(new Preference.OnPreferenceChangeListener() {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
@@ -1,5 +1,7 @@
 package com.eveningoutpost.dexdrip.utils;
 
+import static com.eveningoutpost.dexdrip.xdrip.gs;
+
 import android.Manifest;
 import android.annotation.SuppressLint;
 import android.app.Activity;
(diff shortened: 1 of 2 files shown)


jamorham:
This doesn't make any sense?

This PR actually changes the UI behavior to bind the `Data Age Mins` value to the Summary instead of the Title which I don't think is necessary as its already a short line. 

Even if we did that and removed the single usage of this method I don't want to remove `bindPreferenceTitleAppendToIntegerValueFromLogSlider` from the source because its a UI helper which might be used at any time in the future. Proguard will automatically remove unused instances to reduce binary size so we don't have to worry too much about that.

Additionally `NamedSliderProcessor` is an interface and not a class and is definitely in use.

tolot27:
> This doesn't make any sense?

You are partially right. I've forgotten to change the xml file as well and did not test it. May fault.

> This PR actually changes the UI behavior to bind the `Data Age Mins` value to the Summary instead of the Title which I don't think is necessary as its already a short line.

In English, it is short but not in other languages like Spanish ("Minutos de antiguedad de datos"). 

> Even if we did that and removed the single usage of this method I don't want to remove `bindPreferenceTitleAppendToIntegerValueFromLogSlider` from the source because its a UI helper which might be used at any time in the future. 

I agree partially. It is very likely that we do not need exact this function again because in that case we would always run into the problem we are trying to solve. Also, we have git to track code changes. I don't see the value of dead code. And we have removed `SystemStatus.java` with 0d0cf2b18 exactly for this reason. I've tried to generalize the function to accept a property parameter but don't came to a solution.
Anyway, I've moved the deletion into a separate commit and can remove it if you really like.

> Additionally `NamedSliderProcessor` is an interface and not a class and is definitely in use.

I did not remove `NamedSliderProcessor`, just the function `bindPreferenceTitleAppendToIntegerValueFromLogSlider`.

tolot27:
> This doesn't make any sense?

You are partially right. I've forgotten to change the xml file as well and did not test it. May fault.

> This PR actually changes the UI behavior to bind the `Data Age Mins` value to the Summary instead of the Title which I don't think is necessary as its already a short line.

In English, it is short but not in other languages like Spanish ("Minutos de antiguedad de datos"). 

> Even if we did that and removed the single usage of this method I don't want to remove `bindPreferenceTitleAppendToIntegerValueFromLogSlider` from the source because its a UI helper which might be used at any time in the future. 

I agree partially. It is very likely that we do not need exact this function again because in that case we would always run into the problem we are trying to solve. Also, we have git to track code changes. I don't see the value of dead code. And we have removed `SystemStatus.java` with 0d0cf2b18 exactly for this reason. I've tried to generalize the function to accept a property parameter but don't came to a solution.
Anyway, I've moved the deletion into a separate commit and can remove it if you really like.

> Additionally `NamedSliderProcessor` is an interface and not a class and is definitely in use.

I did not remove `NamedSliderProcessor`, just the function `bindPreferenceTitleAppendToIntegerValueFromLogSlider`.

jamorham:
The PR description references `NamedSliderProcessor` as being removed.

jamorham:
The PR description references `NamedSliderProcessor` as being removed.

PR #2107, opened 2022-05-09
System Status cleanup

Lupum1001:
Noticed a couple unused files from the old SystemStatus activity and removed them. For example, `SystemStatus.java` was replaced with `SystemStatusFragment.java` back in 2017.
I also tweaked some formatting on the System Status page for the sake of readability.

Old
image
New
image

The visual changes are quite subtle, but I didn't want to bundle much else with a file deletion.


tolot27:
Good catch. Looks like you are working on #1522, even if you did not find this issue/task list. 😄

Issue #1781, opened 2021-07-14
RuntimeException: Cannot find preference screen within search result


tolot27:
Searching the settings for "back-fill" and clicking on the top-most hit "Back-fill data Tap to send historical data to Nightscout" (if English is the UI language) results in a `java.lang.RuntimeException: Cannot find preference screen within search result`.

The log shows:
```
jamorham PREFS
Got error trying to highlight search results:
java.lang.RuntimeException: Cannot find preference screen
within search result
```

Any version of xDrip is affected.

tolot27:
I've traced the bug down to the library `search-preference.aar`. This library is self-compiled from repository [ByteHamster/SearchPreference](https://github.com/ByteHamster/SearchPreference) and there are many never versions available. Unfortunately, non of these versions work with xDrip. It looks like `search-preference.aar` was cloned and modified around version [v1.2.6](https://github.com/ByteHamster/SearchPreference/releases/tag/v1.2.6).

@jamorham Do you know who modified the original library?

tolot27:
I've found [Adapted for using in PreferenceActivity Context](https://github.com/CrazyXacker/SearchPreference/commit/d8f0558b68a23b14f81d97971a09b5fd3a7347f9) but this commit introduces some other changes and it will not introduce `SearchConfiguration.getKeysList()`.

jamorham:
This is easy to resolve see: https://github.com/NightscoutFoundation/xDrip/commit/87ac6fc1f57ac40d89f79fa77a6daa13e8379074

Lupum1001:
Opened another pull request with more fixes.


Is there a way to screen for this issue automatically, either by linting the xml or writing unit tests?

Lupum1001:
https://github.com/NightscoutFoundation/xDrip/pull/2114#issuecomment-1122833173
https://github.com/NightscoutFoundation/xDrip/pull/2114#issuecomment-1122837054
Sorry about the relocation @tolot27; I just have a preference for discussing further fixes in the context of the issue than the individual pull request if that's ok.

> This will also find root nodes of `PreferenceScreen`. I don't know whether they need an `android:key` attribute or not.

To my knowledge anything of the type Preference, or its subclasses will need the `android:key` defined. There's at least one exception which is **PreferenceCategory**, but PreferenceGroup may be another.
https://developer.android.com/reference/android/preference/Preference

tolot27:
[...]

PR #2114, opened 2022-05-10
Further fixes for #1781 (preference search exceptions)

Lupum1001:
Added some more missing Preference keys to fix the behavior in preference searches. Fixed some whitespace and attributes along the way.

### DIFF
--- a/app/src/main/res/xml/xdrip_plus_prefs.xml
+++ b/app/src/main/res/xml/xdrip_plus_prefs.xml
@@ -98,8 +98,6 @@
                         app:colorpicker_noneSelectedSummaryText="@string/default_color_selected"
                         app:colorpicker_selectNoneButtonText="@string/revert_to_default"
                         app:colorpicker_showHex="false" />
-
-
                 </PreferenceCategory>
 
                 <PreferenceCategory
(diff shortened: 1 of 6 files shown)


tolot27:
> Is there a way to screen for this issue automatically, either by linting the xml or writing unit tests?

Not really automatically, but you can use XPath to query nodes which does not have a specific attribute. You can test it with http://xpather.com and load `pref_data_sync.xml`. The xpath query would be `//Preference[not(@android:key)]` or `//PreferenceScreen[not(@android:key)]`.

Initially, I thought the lint check `MissingDefaultResource` will do the job but it does not.

tolot27:
> `//PreferenceScreen[not(@android:key)]`

This will also find root nodes of `PreferenceScreen`. I don't know whether they need an `android:key` attribute or not.

app/src/main/res/xml/xdrip_plus_prefs.xml
```diff
@@ -1198,7 +1181,6 @@
                 android:maxLength="4"
                 android:title="@string/title_pendiq_pin" />
 
-
         </PreferenceScreen>
         </PreferenceScreen>
```
tolot27, line 1185: Here is still an indentation missing. Did you use code formatting?

Lupum1001: I didn't, no. I just manually corrected some of the more obvious ones. There's enough inconsistency in the formatting that the whitespace changes would have gotten out of hand really fast. As long as it was readable and more or less consistent with its surroundings, I tended not to mess with it.

PR #2121, opened 2022-05-12
Fix for Android 12 unwanted unbonding

jbr7rr:
This should fix connection problems for some android 12 phones.
As discussed here: https://github.com/NightscoutFoundation/xDrip/discussions/2064

This has been tested by myself on the following devices:
- Pixel 6 Pro (Android 12)
- Fossil 5e (Wear OS 2.34)

The problem is that when a unknown bond state gets reported and even if unbonding is not allowed the collector will initiate the bond process which gets aborted (due to the setting), and the next reading will be attempted in about 5 minutes. this will lead to a loss of reading. if unbonding is allowed, the current reading will be missed due to unbonding and with the next reading the user gets a pair request

This fix checks for the flag returned which is sometimes returned by these device and if the setting does not allow unbonding the code will continue to the GET_DATA state. And the data can be received successfully.

Old situation (left) and after this fix (right):
 

If you want to try it out:
https://drive.google.com/file/d/1jUGnn-fVNcxMaQJgl3whQ2-83Zq45w5t/view?usp=sharing

Beware that this is not compatible with the main xDrip+ builds as it uses a different keystore so you need remove xDrip+ first before installing this version. Also make sure you backup and restore your database

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/AuthStatusRxMessage.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/AuthStatusRxMessage.java
@@ -34,7 +34,7 @@ public class AuthStatusRxMessage extends BaseMessage {
         return bonded == 1;
     }
 
-    public boolean isBondUnknown() { // Android 12 sometimes returns this, in this case data can still be received
+    public boolean isBondUnknown() { // Sometimes this is returned, but data can still be get if we dont want to unbond
         return bonded == 2;
     }
 }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/G5Model/Ob1G5StateMachine.java
@@ -303,7 +303,7 @@ public class Ob1G5StateMachine {
                             parent.authResult(true);
                             parent.changeState(Ob1G5CollectionService.STATE.GET_DATA);
                             throw new OperationSuccess("Authenticated");
-                        // Android 12 sometimes returns this, in this case data can still be received
+                        // Sometimes this is returned, but data can still be get if unbonding is not allowed
                         } else if (status.isBondUnknown() && ! Pref.getBoolean("ob1_g5_allow_resetbond", true)) {
                             parent.msg("Authenticated, but unknown bond state, setting prevents us to reset bond, trying to get data");
                             parent.authResult(true);


Navid200:
> This should fix connection problems for some android 12 phones.  

How about the other Android 12 phones?
How about Android 11 phones?
What happens to the phones that are working fine now?

In other words, how many devices have you tested this on?

ChanceHarrison:
It would be useful to be able to have this as an engineering mode option in an actual xDrip nightly so more people can test without having to reinstall completely. Is that at all feasible / worth the development effort? It doesn't seem like a high-risk change.

jbr7rr:
> 

@Navid200 I updated the PR and included which devices I tested myself. I also added the wear code.
It seems that not only Android 12 are affected but other devices as well. (In my case my wear watch) I asked around for ppl to test. but almost everyone has the setting "ob1_g5_allow_resetbond" enabled and in that case this has no effect. So to me that setting is already a workaround which in this case was not working. 

One would expect that based on the description of the setting. That in this case unbonding is also not allowed an the data is received anyway.

I will try on an old Xioami in the next days and will update this PR once I have the results.

@ChanceHarrison There are already a lot of options in xDrip so I do not want to have another setting for this in engineering mode. I think the setting "ob1_g5_allow_resetbond" is satisfactory in this case.

Navid200:
>  but almost everyone has the setting "ob1_g5_allow_resetbond" enabled 

That's because it is a recommended setting.
There are only a few devices that need to disable it.  For example Samsung S21.

We cannot change xDrip imposing a change on everyone unless we are certain it will help the most.  I didn't say unless it helps the majority because that's not enough.
If a change breaks a working system, it is an unacceptable change even if it is a minority.   

It is great that you are interested and are trying to help.
My suggestion is to communicate with @jamorham to understand why things are the way they are.
For all I know he is aware of all of this and has made a choice to have the settings the way they are because the alternative would be much worse.

If there is only one person who is successfully using the same phone you are using with the existing xDrip as is, there is no reason to make this change.
But, if everyone who is using your phone is having this problem and we want to support this phone, we have to do it making sure that no other phone is going to be negatively affected.

jbr7rr:
@Navid200 I agree we have to be certain to a some degree that it will not mess things up for other perfectly working devices. 

As seen in this https://github.com/NightscoutFoundation/xDrip/discussions/2064 a lot of ppl are having issue's with the pixel range of phones. some don't as you pointed out there but i think that may have to do with software version/region etc. Maybe even the transmitter may play a role in this. As I use a Dexcom One transmitter. 

It would be great to have some feedback from @jamorham and I hope my PR will at least contribute to a solution on some form or another. Also there was/is an open TODO in the part of the code I changed. And see my PR as a proposal to fix that.

Also do you know if these bonded flags/messages are specific to the Dexcom transmitters or are they general for BLE? I could not find this in the BLE specifications.

Navid200:
@jbr7rr Just to let you know it turns out jamorham has a Pixel with Android 12 and is testing.
He will add a mod to the collector to address connectivity issues.  
But, it is not simple.  So, I don't know when.

jbr7rr:
Fixed in https://github.com/NightscoutFoundation/xDrip/commit/0d2269cd741701b09438149a05f92177688a6b33

Thanks!

Closing this PR!

Navid200:
Thanks

Issue #2119, opened 2022-05-11
OOP1new and Libre 2 with Sync followers


psonnera:
### Subject of the issue
On a non NFC master phone (KingKong Mini 2), when using a Libre 2 EU with OOP1new (fa218ef 2019-11-22) , versions above https://github.com/NightscoutFoundation/xDrip/releases/tag/2022.03.26 do not allow followers to receive data.
Versions above https://github.com/NightscoutFoundation/xDrip/releases/tag/2021.10.13 start showing NFC block errors in master and follower.

### Environment
- Master above 2022.03.26
- Master KingKong Mini 2, NSClient
- Bubble bridge
- All followers (various Android versions from 6 to 12) in Sync follower
- Master using OOP1 enabled

### Expected behavior
Followers should display BG.

### Actual behavior
Libre blocks are received but ignored.

### Screenshots
Screenshots won't add value, what's the best way to share some logs?

tzachi-dar:
As far as I can say, when using OOP1 there was no way to interrupt the block data. So, I'm not surprised that there are errors regarding them.
That said, the BG values that are being sent from the master should be displayed on the follower as well.

If you can connect the follower phone to a pc and run 
adb logcat > file
for about 15 minutes and send this file, that would be great.

psonnera:
Here you go, using May 12th 2022 on both. Logcats at different moments. Follower on a Samsung Galaxy 5 but had same results on other followers with Android 9 and 12.
[master+follower.zip](https://github.com/NightscoutFoundation/xDrip/files/8690735/master%2Bfollower.zip)
.

tzachi-dar:
Strange, the issue seems to happen on the master code.

Please try this version:
https://drive.google.com/file/d/1gf2RAlpbl-YuyfgrQ8-n_zpj8sz-Ogii/view?usp=sharing

Here is the log:

05-13 20:54:27.013  3147  5534 I jamorham graph: Noise: Not enough data to get sensible noise value
05-13 20:54:27.014  3147  3147 D BestGlucose: No noise option slope by minute: -.04732
05-13 20:54:27.014  3147  5534 D BestGlucose: No noise option slope by minute: -.04732
05-13 20:54:27.014  3147  5533 E BgReading: Could not save BGR: 
05-13 20:54:27.014  3147  5533 E BgReading: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.Class java.lang.Object.getClass()' on a null object reference
[...]

PR #2127, opened 2022-05-16
Make sure that xDrip will not crash in the case that parsing a trade or history sensor will fail.

tzachi-dar:
This scenario happens when using OOP1 with EU encrypted sensors.

This fixes https://github.com/NightscoutFoundation/xDrip/issues/2119


tolot27 (approved):
Quite simple and straightforward.

PR #2138, opened 2022-05-20
Only if automatic calibration is disabled

Navid200:
I only added a new line at 776 and 822.  And, I did an auto format to make sure I have not messed up formatting.
But, it looks like I have modified more than 40 lines.      
I have only modified 2 lines.

If automatic calibration is enabled, you want xDrip to judge for you and decide if the blood glucose measurement you have entered should be used for calibration or not.
In that case, asking the user if they want it to be used for calibration or not makes no sense.
So, those options will be ignored, after this PR, if automatic calibration is enabled.


jamorham (approved):
I will trust that you have sufficiently tested this

Navid200:
@jamorham Yes, I have tested this.  
It will prioritize automatic calibration choice (if enabled) over manual blood glucose entry through the treatment menu options selection.

PR #2140, opened 2022-05-22
don't age adjust raw for any G6

tynbendad:
see https://github.com/NightscoutFoundation/xDrip/discussions/2137
I will post more justification in the comments here tomorrow.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -1661,7 +1661,7 @@ public class BgReading extends Model implements ShareUploadableBg {
 
     public void calculateAgeAdjustedRawValue(){
         final double adjust_for = AGE_ADJUSTMENT_TIME - time_since_sensor_started;
-        if ((adjust_for > 0) && (!DexCollectionType.hasLibre())) {
+        if ((adjust_for > 0) && (!DexCollectionType.hasLibre()) && (!Ob1G5CollectionService.usingG6())) {
             age_adjusted_raw_value = ((AGE_ADJUSTMENT_FACTOR * (adjust_for / AGE_ADJUSTMENT_TIME)) * raw_data) + raw_data;
             Log.i(TAG, "calculateAgeAdjustedRawValue: RAW VALUE ADJUSTMENT FROM:" + raw_data + " TO: " + age_adjusted_raw_value);
         } else {


Lupum1001:
can the G6 be used in non-native mode? I haven't tried it myself, though the user-facing settings certainly appear to allow for it.

If it can be used like a G5 in that way, I'd double check that you don't want `&& !Ob1G5CollectionService.usingNativeMode()` instead of `&& !Ob1G5CollectionService.usingG6()`.
I haven't personally delved into the data processing section of the Dexcom code yet, so my ability to make suggestions is extremely limited atm.

tynbendad:
G6 can be used in non-native mode, if you calibrate it. However I don't want this particular "aged raw" on any G6, native or non-native as the raw does not need to be adjusted in this way - I will post data showing this shortly.

tynbendad:
this chart should show why aged raw doesn't make sense - NativeG6 is the "gold standard", and Aged raw doesn't come close for G6 (I have not seen any codes or firmware versions do anything different than this, at least from my limited sampling.)

image

Navid200:
This PR affects all G6 devices.
There are G6 devices in circulation from years ago that can be used in non-native mode.  So, this PR, as it stands, should not be merged.

A capability was added to the xDrip code so that you can identify modified transmitters.
Please use it to limit your modifications to those devices.

If you indeed intend to modify the behavior of xDrip on oldG6 devices, you need to reach out to users and show us why they are complaining about xDrip as is.

Your curves show that the aged raw replicates the impact of the immune system on the sensor sensitivity.  It was done for a reason.

tynbendad:
> This PR affects all G6 devices. There are G6 devices in circulation from years ago that can be used in non-native mode. So, this PR, as it stands, should not be merged.
> 
> A capability was added to the xDrip code so that you can identify modified transmitters. Please use it to limit your modifications to those devices.
> 
> If you indeed intend to modify the behavior of xDrip on old G6 devices, you need to reach out to users and show us why they are complaining about xDrip as is.

correct, the chart is using one of the older G6 devices, whether aged raw is applicable or not applies to all G6, not just modified transmitters.

> 
> Your curves show that the aged raw replicates the impact of the immune system on the sensor sensitivity. It was done for a reason.

I will update the chart with another line showing what aged raw would look like if it were calibrated at the end of warmup to 147. that would be a better comparison. I'll also wait for more data to see what happens with nativeG6 over more time.

I don't know if this PR can be moved back to a draft PR for the moment, in any case, just hang on...

Navid200:
> I don't know if this PR can be moved back to a draft PR for the moment 

At the top, in the right pane,  under "Reviewers", you can see "Still in progress? Convert to draft".
"Convert to draft" is a link that let's you convert it.

tynbendad:
image

image

https://docs.google.com/spreadsheets/d/1yYr3y__srzGMfbPouyLTfWtddOrWqkWk8-mLRKEQmJ4/edit?usp=sharing

Navid200:
There are three parameters that affect the accuracy of a calibrated result.  Those are slope, intercept (offset), and time.
Some individuals overlook the importance of time.

The reason time is a factor is that the sensor is affected by the immune system.
A user of G5 (or G6 in non-native mode) needs to frequently calibrate.  The reason it is necessary to calibrate frequently is to account for the impact of the immune system on the sensor.

Looking at your curves, I have to assume the reason aged raw was created was to account for the impact of the immune system.
But, that cannot be perfect.  So, frequent calibrations are still required.  But, I assume with this, the error will be much smaller if someone delays their calibrations.

I am not sure what you are demonstrating by the curves and what you like to conclude from them.

Navid200:
Reading my own post, I can see that it may be taken the wrong way.
Even with aged raw, calibrations are still required.  Initial calibrations are absolutely necessary.  But, after you have done your initial calibration and after you have done a few more, you get the right slope and intercept.  Now, with aged raw, your results are going to be really good for much longer than if you just used raw.
I assume if we disable aged raw, just after 12 hours, your results will be way wrong and you will absolutely need to calibrate or your results will keep drifting and the error will increase dramatically.

With aged raw, your results will be much more accurate if you only calibrate once a day instead of once every 12 hours.  
But, I don't think you can only calibrate once a day with just raw.

tynbendad:
the MARD table shows what happens with a simple single-point calibration (in spreadsheet, not xdrip's) every 12 hours, 24 hours, or just once. It is based on the same 48 hours of data collection that the chart is, I just didn't want to keep putting lots of lines in the chart. All data and calculations are available at the spreadsheet link.

Navid200:
Do you also have access to a G5 for your tests?

It is possible that this was done mainly for G5.  The use of non-native mode for G6 I don't believe has ever been a priority.  

If we make this change with no consultation with those who are using old G6, we may be modifying their systems and they may never know.  Not a very good idea.

tynbendad:
I am using an "old" G6, 819xxx code (and fwiw as far as I have seen all G6, from 80 to 8H at least, have generally the same raw & native behaviors.) I don't plan to collect data for G5 although I probably have one that I could use. Collecting the data takes a long time and I'm still running the current collection since I'm interested in looking at more than just aged raw. But if there is a concensus that something specific is needed to go further then I might be able to collect it.

Navid200:
It would be a good idea to calibrate once a day and show what the slope and intercept are for each case after the first days.
Those parameters have some limits.  If your change possibly could make one of those parameters go out of range, that would be a show stopper.

tynbendad:
an alternative to discarding aged raw for g6 is to mellow it - using the same spreadsheet you can just tweak the values of AGE_ADJUSTMENT_TIME and AGE_ADJUSTMENT_FACTOR to see how MARD would be affected... these values seem a lot better.
image
image

I will continue to collect data on other sensor codes but it will take a while.

tynbendad:
@Navid200 the table shows it calibrated.

jamorham:
I agree that it may need some adjustments if the dynamics are different. The current adjustment value is based on experiments seen with G4 and G5 afaik. 

There is one other assumption that you've made though which is that the native value is the correct number! We have observed that this isn't always the case (although the 1.x firmware was very good!)

Beware also that there are some additional complexities, see `estimated_raw_bg()`

tynbendad:
>There is one other assumption that you've made though which is that the native value is the correct number! 

i'm just trying to make sense of the current xdrip+ g6 raw values, and native g6 makes a pretty good gold standard. i won't be doing medical studies comparing to meter bg in humans or animals.

>Beware also that there are some additional complexities, see estimated_raw_bg()

that looks like some kind of curve fitting of the raw values to the current timestamp. and if the result is quite different (>20) from aged raw then aged raw is used, and from there the result is used with current bg to update calibration. if i'm far off please let me know. given this, getting aged raw right still seems fairly important.

PR #2141, opened 2022-05-22
Refactor/assorted cleanup

Lupum1001:
Cleaning up a few strings

- "pref_ringtone_silent" was the only one in use in its file, so I moved it to the main strings.xml file for future translation.
- "need_storage_permission" was a duplicate of "need_storage_permission_to_access_all_ringtones". It was a toss-up, but I kept the second one.
image

I also added a few comments in StartNewSensor and bypassed `startSensorOrAskForG6Code()` for the Libre code, since it's redundant. 
The parameters I removed in `activity_start_new_sensor.xml` were largely marked as invalid by android studio.


jwoglom (approved):
I believe @jamorham may be the only one able to approve PR merges still but this looks like a simple fix.

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -2255,8 +2254,7 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
 
                     } else if (preference instanceof RingtonePreference) {
                         if (TextUtils.isEmpty(stringValue)) {
-                            preference.setSummary(R.string.pref_ringtone_silent);
-
+                            preference.setSummary(gs(R.string.pref_ringtone_silent));
```
jamorham, line 2257: Why changing this to use `gs` instead of passing the reference?

Navid200:
Can you please address jamorham's question?

Lupum1001:
> Can you please address jamorham's question?

I haven't looked at the code recently, but I would have been using nearby code as a guide/reference. If I remember correctly, any UI elements that used the un-translated strings file just used R.string.whatever, while the ones that pulled from the translated set seemed to wrap it in gs() first.
I don't remember if I _actually_ confirmed that gs() was abstracting away a translation step, but I certainly considered it a possibility, and would've wanted to be on the safe side.

An alternative convention that I saw getting used for translated strings was xdrip.getAppContext().getString(R.string.whatever), and I probably _did_ assume that gs() was functionally identical shorthand. If one was pass-by-reference and the other was pass-by-value, I didn't notice. Both were being used, so I went with the abbreviated/more readable one.

A lot may have changed since I submitted this PR though. If the old strings file is still kicking around and you want me to re-check this PR against the latest code, let me know and I'll try to get to it.

PR #2142, opened 2022-05-23
Adding the ability to set vibration patterns per individual alerts

erobinson:


usernamebeep:
Could we use 0,100,50,100,50,100,50,100,50,100,50,100 for low or as additional pattern?
It is more aggerssive and distinctive for low

erobinson:
> Could we use 0,100,50,100,50,100,50,100,50,100,50,100 for low or as additional pattern? It is more aggerssive and distinctive for low

Added as an extra option.

erobinson:
Screenshot_20220523-222527

Navid200:
Looks very good.

What exactly happens if someone installs xDrip for the very first time after this?  What will the vibration type be set to by default?  Will it be identical to the current vibration pattern before this PR?

Have you tested this in conjunction with the Ascending volume profile?

Disabling "Vibrate on alert" disables them all, right?

I don't know if we have a standard method for marking the default.  May be @tolot27 can tell us.
But, it's a good idea to help the user know what the default setting is.  It will be a shame if the only way for a user to go back to the default settings would be to reinstall xDrip.

I have used > as a marker for the default in a PR that has not been merged.  So, I cannot tell you to use it unless we can an approval.

usernamebeep:
I have no knowledge of java but couldn't one use arrays or lists to generate all the radio buttons and vibration patterns as there seems to be so many code repetitions? Not sure what is stored after clicking save but if you stored the pattern itself, you could create an input so others could type in custom pattern.

usernamebeep:
How is it going? Still looking very forward to this feature.

hartlijn:
What is the problem with the PR? I'd love to use this feature

usernamebeep:
Yea, same, still waiting this to be implemented.

Navid200:
@erobinson How would you feel about creating a sub page that would contain everything related to the vibration patterns?
The advantage of that is that the alert edit page that everyone is used to will not be modified much.  it will only have a single new button added to it.

I am happy to collaborate with you on this as well.

PR #2151, opened 2022-05-28
Rename engineering mode to test mode

Navid200:
Users enable engineering mode and leave it enabled.
This is not what engineering mode is meant for.
Engineering mode should only be used when absolutely necessary, which should be for testing.
If a setting that is only available in engineering mode needs to be enabled continuously, the users should inform the developers with an explanation.  In that case, the developers can move the setting out of engineering mode when possible.

This PR changes the name of the setting, only from the point of view of the user, to "Test mode".   The parameter is still called the same.  So, anyone updating to this release with engineering mode enabled, will have Test mode enabled instead automatically.
Screenshot_20220528-000626

The name "engineering" has a positive connotation.  There are individuals who feel having it enabled means they have earned a badge.  We shouldn't be encouraging people to enable it.
Calling it test mode is more representative of what it is meant for.

There are posts in the facebook group asking for help.  After spending time going through the settings, there are times we find out the engineering mode is enabled and that's the only reason the system is broken.  But, most users are not aware that they should not use the engineering mode under normal conditions for day-to-day use.

This PR also places a text on the main screen reading "Testing" as shown below when the Test mode is enabled.
This let's those who provide support see it on the main screen.  But, more importantly, it is hoped that it would encourage the users to disable it.
The text is placed right under the predicted low text in white.
Screenshot_20220528-074457

  
Screenshot 2022-05-28 000301

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -2286,9 +2286,11 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
         initializeGraphicTrendArrow();
         final TextView notificationText = (TextView) findViewById(R.id.notices);
         final TextView lowPredictText = (TextView) findViewById(R.id.lowpredict);
+        final TextView testText = (TextView) findViewById(R.id.testtext);
         if (BgGraphBuilder.isXLargeTablet(getApplicationContext())) {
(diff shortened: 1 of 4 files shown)


jamorham:
The terminology isn't going to change nor should the home screen have any text on it. It is something I and other developers have permantently enabled and is hidden behind a voice command already. We can provide an addendum on the status screen if we need to make it visible that it is enabled.

PR #2159, opened 2022-05-30
Event log access from MegaStatus menu

Lupum1001:
This PR is based on the following discussion/request: https://github.com/NightscoutFoundation/xDrip/discussions/2128

it enables the dropdown menu for MegaStatus, and adds a link to the event log, just like in the Home activity.

---

I also did a bit of refactoring in Home.java->onOptionsItemSelected() as well, based on some linting advice.
image
I got rid of the initial switch-case and converted everything into one big if-else-if statement.


app/src/main/res/menu/menu_mega_status.xml
```diff
@@ -1,9 +1,12 @@
 <menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     tools:context="com.eveningoutpost.dexdrip.MegaStatus">
-    <item
-        android:id="@+id/action_settings"
```
tolot27, line 5: Does the burger menu to access the settings still work if you remove these lines here?

tolot27 (approved):
Currently, we cannot upgrade the AGP because of the wrong casing of package names (#1522) and it is unlikely that this will change in the near future because we had to modify a lot of classes (#1539).
Hence, I would separate this into another PR which could contain all fixes for this lint waring.

Lupum1001:
> Hence, I would separate this into another PR which could contain all fixes for this lint waring.

I figured I might as well change it while I was already there, but there are some efficiency improvements in there as well.

- `item.getItemId()` only really needs to be called once
- if an 'if' statement evaluates as true, the use of else-ifs prevents the subsequent checks from having to run anyway.

 
In theory, I actually prefer a switch-case for the whole function (replacing the 'if' statements), because I feel it's cleaner syntax when you're just checking the id against a bunch of different options. I figured it was better to go with the future-proof format though.

jamorham:
If they're going to be fussy about resource id's (and I don't know why the integer can't be used in a switch statement) then wouldn't it be better just to use the `onClick` xml parmeter instead as this is closer to the databinding ui method which is used in later parts of the code anyway and is less susceptible to merge conflicts than an if/else chain?

Lupum1001:
Ok, I've moved the refactoring out of this PR; it just includes the new menu item now.

PR with the refactoring coming momentarily.

app/src/main/java/com/eveningoutpost/dexdrip/MegaStatus.java
```diff
@@ -439,21 +439,20 @@ protected void onResume() {
...
         // as you specify a parent activity in AndroidManifest.xml.
-        int id = item.getItemId();
-
-        //noinspection SimplifiableIfStatement
-        if (id == R.id.action_settings) {
-            return true;
-        }
+        //int id = item.getItemId();
 
         return super.onOptionsItemSelected(item);
```
tolot27, line 457: If you just call the super method, you can remove this overriding function completely.

tolot27 (changes requested):

app/src/main/java/com/eveningoutpost/dexdrip/MegaStatus.java
Lupum1001: I think it's called behind the scenes, instead of explicitly, so the way to do that would presumably be to delete the function entirely, or at least comment it out. Do you have a preference for one or the other?

tolot27: If a function does nothing else than its super function it is better to remove it. This removes code readability. In our case, the click is simply handled and we don't do anything special.

Lupum1001: I've removed the function. 👍

Navid200:
There is now a screenshot button on the status page.
How practical is it to place the screenshot button inside the 3-dot menu that this PR proposes?

Please note that I cannot merge a PR.  So, even if you do that, this may still not be merged.

I am sorry about the delay.

PR #2161, opened 2022-06-02
Warn user about transmitter age on sensor start

Navid200:
This PR addresses # 2 from the list identified here:
https://github.com/NightscoutFoundation/xDrip/pull/2143#issuecomment-1140510928

In addition, it also provides a warning ahead of time so that the user can plan the timing of the following few sensors carefully and plan to place an order for a new transmitter or hard reset their transmitter depending on what kind of transmitter they have.

@jamorham I have not tested this yet.  I will have a sensor start coming up and will test then.  That's why this is still a draft.
But, would you please tell me if you see anything in the code that you don't like like code duplication or ...?


Navid200:
Never mind.  A toast message is not good as it disappears and the user may never see it.  
I am going to change the existing text that comes up and add to it depending to the number of day.

Navid200:
Don't know how to resolve conflicts in github.
It seems the only way is to use git.
It's easier for me to just open a  new PR.

PR #2164, opened 2022-06-05
Add Range Settings to the first statistics page

Navid200:
The time in-range parameter, included on the first statistics page is a very important parameter.  However, it is meaningless without specifying what the range is.
Currently, that page does not show the range.  If you change your High and Low Values, you can see that the time in-range as well as PGS change accordingly.

This PR adds the High and Low values to the first statistics page as shown below.
Screenshot_20220705-115855

Edit:
I have changed the font size to be identical to everything else on the screen.  I personally believe the font should be smaller because the high and low values are not values that have been calculated.  Rather those are parameters the user has entered.  So, it's best to have a way of separating them from the rest of the data.

I have opened another PR to make all the blood glucose units in xDrip including the ones on the statistics page consistent:
https://github.com/NightscoutFoundation/xDrip/pull/2166

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/stats/FirstPageFragment.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/stats/FirstPageFragment.java
@@ -73,11 +73,11 @@ public class FirstPageFragment extends Fragment {
             TextView lowView = (TextView) localView.findViewById(R.id.textView_stats_low);
             //update stats_high/low
             if (!mgdl) {
-                updateText(localView, highView, (Math.round(stats_high * 10) / 10d) + " mmol/L");
-                updateText(localView, lowView, (Math.round(stats_low * 10) / 10d) + " mmol/L");
+                updateText(localView, highView, (Math.round(stats_high * 10) / 10d) + " mmol/l");
+                updateText(localView, lowView, (Math.round(stats_low * 10) / 10d) + " mmol/l");
             } else {
-                updateText(localView, highView, Math.round(stats_high) + " mg/dL");
-                updateText(localView, lowView, Math.round(stats_low) + " mg/dL");
+                updateText(localView, highView, Math.round(stats_high) + " mg/dl");
+                updateText(localView, lowView, Math.round(stats_low) + " mg/dl");
             }
 
             if (context == null) {
@@ -110,10 +110,10 @@ public class FirstPageFragment extends Fragment {
                 TextView medianView = (TextView) localView.findViewById(R.id.textView_median);
 
                 if (mgdl) {
-                    updateText(localView, medianView, Math.round(median * 10) / 10d + " mg/dL");
+                    updateText(localView, medianView, Math.round(median * 10) / 10d + " mg/dl");
 
                 } else {
-                    updateText(localView, medianView, Math.round(median * Dex_Constants.MG_DL_TO_MMOL_L * 100) / 100d + " mmol/L");
+                    updateText(localView, medianView, Math.round(median * Dex_Constants.MG_DL_TO_MMOL_L * 100) / 100d + " mmol/l");
 
                 }
 
@@ -127,9 +127,9 @@ public class FirstPageFragment extends Fragment {
                 TextView meanView = (TextView) localView.findViewById(R.id.textView_mean);
                 //update mean
                 if (mgdl) {
-                    updateText(localView, meanView, (Math.round(mean * 10) / 10d) + " mg/dL");
+                    updateText(localView, meanView, (Math.round(mean * 10) / 10d) + " mg/dl");
                 } else {
-                    updateText(localView, meanView, (Math.round(mean * Dex_Constants.MG_DL_TO_MMOL_L * 100) / 100d) + " mmol/L");
+                    updateText(localView, meanView, (Math.round(mean * Dex_Constants.MG_DL_TO_MMOL_L * 100) / 100d) + " mmol/l");
 
                 }
                 //update A1c
@@ -145,9 +145,9 @@ public class FirstPageFragment extends Fragment {
                 stdev = Math.sqrt(stdev);
                 TextView stdevView = (TextView) localView.findViewById(R.id.textView_stdev);
                 if (mgdl) {
-                    updateText(localView, stdevView, (Math.round(stdev * 10) / 10d) + " mg/dL");
+                    updateText(localView, stdevView, (Math.round(stdev * 10) / 10d) + " mg/dl");
                 } else {
-                    updateText(localView, stdevView, (Math.round(stdev * Dex_Constants.MG_DL_TO_MMOL_L * 100) / 100d) + " mmol/L");
+                    updateText(localView, stdevView, (Math.round(stdev * Dex_Constants.MG_DL_TO_MMOL_L * 100) / 100d) + " mmol/l");
                 }
 
                 TextView coefficientOfVariation = (TextView) localView.findViewById(R.id.textView_coefficient_of_variation);
--- a/app/src/main/java/com/eveningoutpost/dexdrip/stats/FirstPageFragment.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/stats/FirstPageFragment.java
@@ -7,6 +7,8 @@ import android.preference.PreferenceManager;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v4.app.Fragment;
+
+import com.eveningoutpost.dexdrip.Models.UserError;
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/stats_general.xml
```diff
@@ -14,6 +14,62 @@
...
+
+                <TextView
+                    android:id="@+id/textView20"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:layout_weight="1"
+                    android:paddingStart="5dp"
+                    android:paddingEnd="0dp"
+                    android:text="@string/stats_high_value"
+                    android:textAppearance="?android:attr/textAppearanceMedium" />
```
tolot27, line 32: Is there a reason for the smaller font size?

app/src/main/res/values/strings.xml
```diff
@@ -204,6 +204,8 @@
     <string name="general_settings">General Settings</string>
     <string name="high_value">High Value</string>
     <string name="low_value">Low Value</string>
+    <string name="stats_high_value">High Value: </string>
```
tolot27, line 207: This duplicates the resource strings and increases the translation effort. We should think about different solutions, i. e. the [Android Stem](https://github.com/LikeTheSalad/android-stem) Gradle plugin. And example usage is shown at https://stackoverflow.com/a/59767841.

app/src/main/java/com/eveningoutpost/dexdrip/stats/FirstPageFragment.java
```diff
@@ -96,10 +110,10 @@ public void run() {
                 TextView medianView = (TextView) localView.findViewById(R.id.textView_median);
 
                 if (mgdl) {
-                    updateText(localView, medianView, Math.round(median * 10) / 10d + " mg/dl");
+                    updateText(localView, medianView, Math.round(median * 10) / 10d + " mg/dL");
```
tolot27, line 113: If the unit of measurements are correct here, shouldn't they fixed it anywhere? Otherwise, it is out of scope of this PR (which you stress often). I know that the wrong unit of measurement would be repeated in lines 76--80. This is unsatisfactory to me as well. I recommend to separate the change of the units at least into a separate commit.

tolot27 (changes requested):

app/src/main/res/values/strings.xml
Navid200: I wanted to append a colon to the existing strings "High Value" and "Low Value".  But, I didn't know how to do it easily.
Assuming this is merged, I hope some time in the future to propose a second step for this where instead of using the existing High and Low values to change these to be inputs so that the user can specify what they want to use for their statistical analysis not necessarily the same as the values used for their alarms.
But, I didn't want to make this a very large PR.  So, that's a second step.

So, in the future, we will need separate strings with respect to the existing high and low values.

app/src/main/java/com/eveningoutpost/dexdrip/stats/FirstPageFragment.java
Navid200: The units are correct.  There is just a typo.
There would be a discrepancy on the page if I used mmol/L for the new High and Low values but left mmol/l for mean and standard deviation on the same page.

app/src/main/res/layout/stats_general.xml
Navid200: >other statistics

The two new values are not statistics.  

I appreciate your point about it getting attention, which is not what I want.  In fact, I want them to take less attention.
Can you think of any other way to accomplish this?

app/src/main/java/com/eveningoutpost/dexdrip/stats/FirstPageFragment.java
tolot27: I meant "corrected". Fixing a typo is some kind of correction and a typo in a unit of measurement lead to an incorrect unit with respect to the value.
If "L" is used just on this page a discrepancy is introduced compared to the other places in xDrip were "mg/dl" or "mmol/l" is used. We should keep this in mind for a future PR, at least.

Navid200: I agree.  I would like all instances of units to be mg/dL and mom/L.
For some reason, we have mmol/l and mg/L in some places.  I would like to fix them all.
As you say, I would never have done this because it is a different fix.  I always make my PRs as small as possible.

The only reason I did it here is that otherwise, there would have been a discrepancy on the page.

In fact, I will be happy to search xDrip and open a PR correcting them all.

app/src/main/res/layout/stats_general.xml
tolot27: IMHO, it is not necessary to distinguish the setting values from the statistic values. If you really want to, maybe one way is to write "Range setting: 3.9 ‒ 10 mmol/L".

app/src/main/java/com/eveningoutpost/dexdrip/stats/FirstPageFragment.java
tolot27: Just for the future I recommend to copy & paste the erroneously written text, create the commit with the intentional change and fix the errors in a separate commit with it own descriptive commit message. Later on, we can identify it much easier based on the commit message.

Navid200: I agree.
I am going to change the units including the ones I am adding to all be incorrect but consistent.
I will open another PR correcting all instances in xDrip.

tolot27 (approved):

jamorham:
Conceptually I approve of this change but I feel that having `High Value` and `Low Value` at the top of the page could be misunderstood as the maximum and minimum values for that time range. 

I think it would be better to have, below the time in range %'s, to say something like `Range Settings: 3.9-10.0 mmol`

Navid200:
> I feel that having `High Value` and `Low Value` at the top of the page could be misunderstood as the maximum and minimum values for that time range.

I never thought of that.  Makes perfect sense and now, I am concerned about the possibility of that misunderstanding as well.  Thanks.
    
<br/>  

> I think it would be better to have, below the time in range %'s, to say something like `Range Settings: 3.9-10.0 mmol`  

Perfect suggestion.  Thanks

Navid200:
Please don't hesitate to ask for more changes.

jamorham (approved):

Der-Schubi:
@Navid200 This PR broke the build:
`> Task :app:compileProdReleaseJavaWithJavac
Gradle may disable incremental compilation as the following annotation processors are not incremental: databinding-compiler-3.4.3.jar (androidx.databinding:databinding-compiler:3.4.3).
Consider setting the experimental feature flag android.enableSeparateAnnotationProcessing=true in the gradle.properties file to run annotation processing in a separate task and make compilation incremental.
/home/cs/Android/Projekte/xDrip/app/build/preprocessor/replace/main/java/com/eveningoutpost/dexdrip/deposit/DepositCallback.java:46: warning: as of release 10, 'var' is a restricted local variable type and cannot be used for type declarations or as the element type of an array
            var msg = "Error";
                ^
warning: vm.showButton.get() is a boxed field but needs to be un-boxed to execute !vm.showButton.get(). This may cause NPE so Data Binding will safely unbox it. You can change the expression and explicitly wrap vm.showButton.get() with safeUnbox() to prevent the warning
warning: vm.showButton.get() is a boxed field but needs to be un-boxed to execute vm.showButton.get() ? prefs.getbool("engineering_mode") : false. This may cause NPE so Data Binding will safely unbox it. You can change the expression and explicitly wrap vm.showButton.get() with safeUnbox() to prevent the warning
vm.showButton.get() is a boxed field but needs to be un-boxed to execute !vm.showButton.get(). This may cause NPE so Data Binding will safely unbox it. You can change the expression and explicitly wrap vm.showButton.get() with safeUnbox() to prevent the warning

vm.showButton.get() is a boxed field but needs to be un-boxed to execute vm.showButton.get() ? prefs.getbool("engineering_mode") : false. This may cause NPE so Data Binding will safely unbox it. You can change the expression and explicitly wrap vm.showButton.get() with safeUnbox() to prevent the warning

/home/cs/Android/Projekte/xDrip/app/build/preprocessor/replace/main/java/com/eveningoutpost/dexdrip/stats/FirstPageFragment.java:94: error: cannot find symbol
            TextView rangeView = (TextView) localView.findViewById(R.id.textView_stats_range_set);
                                                                       ^
  symbol:   variable textView_stats_range_set
  location: class id
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error
3 warnings`

PR #2165, opened 2022-06-05
adjust age_adjusted_raw_value for g6

tynbendad:
all of the latest data is available in chart and table form in spreadsheets available here:
current aged calc: https://docs.google.com/spreadsheets/d/1yYr3y__srzGMfbPouyLTfWtddOrWqkWk8-mLRKEQmJ4/edit?usp=sharing
proposed aged calc: https://docs.google.com/spreadsheets/d/1xEK8DSyUAVhuD-DIbk7oPLHs0pmVI7n1s4OLYGIdbBw/edit?usp=sharing

each sheet has a chart as well as a MARD table for 3 coded starts (codes suggested by JoshuaB that should represent a wide range of G6 codes) of at least 48 hours data and a no-code start (of approx. 12 hours data.)

i've built the code but have not tested it (and i don't plan to as i'm otherwise occupied for a bit.)

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/BgReading.java
@@ -68,6 +68,8 @@ public class BgReading extends Model implements ShareUploadableBg {
     private final static String PERSISTENT_HIGH_SINCE = "persistent_high_since";
     public static final double AGE_ADJUSTMENT_TIME = 86400000 * 1.9;
     public static final double AGE_ADJUSTMENT_FACTOR = .45;
+    public static final double AGE_ADJUSTMENT_TIME_G6 = 86400000 * 1.9 / 1.8;
+    public static final double AGE_ADJUSTMENT_FACTOR_G6 = .45 / 3;
     //TODO: Have these as adjustable settings!!
     public final static double BESTOFFSET = (60000 * 0); // Assume readings are about x minutes off from actual!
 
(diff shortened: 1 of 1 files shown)


Navid200:
For transparency and clarity, the questions that were asked in the previous PR, that you have closed, should be asked again:

What would be the reason for a G6 user to use it in non-native mode for regular use?
A G6 cannot be used in non-native mode unless in engineering mode, which is meant for testing.

tynbendad:
> What would be the reason for a G6 user to use it in non-native mode for regular use?

the reasons are the same as they are for any CGM with raw values - to get data when the CGM would not otherwise give you data.

Navid200:
There are existing G6 devices (old G6) that will be affected by this change.
Those using such devices may be surprised by this change.

Is there a reason not to limit this to the devices that currently are not using non-native mode (modified transmitters)?
Isn't that your intent?  Aren't you using a modified transmitter?
Why do you want to impose this on others when you yourself are not going to be affected?
Can you explain your reasoning for this approach and insistence on it?

tynbendad:
> Can you explain your reasoning for this approach and insistence on it?

Look at the math in my spreadsheets, that is the entirety of the reasoning, i.e., that the current aged calculation is just plain wrong. The data there comes from non-modified G6 (80xxxx and 81xxxx) but I believe they apply to all non-modified & modified.

Navid200:
Imagine a left-handed person learns to play the guitar using a regular (right-handed) guitar and keeps playing.
He keeps doing that for decades and enjoys the music he plays.
And one day, you take away the guitar from him and give him a left-handed guitar.   Even though you may feel that he should be thankful, you may have actually caused the music to die for him.
In this case, instead of focusing on what matters, the music, we have focused on conventions being right.   We have actually caused a disservice.  
It may have been the right thing to do if we did it just when he started to learn to play the  guitar.  But, not decades after.

Our focus should be people's control of their diabetes rather then the conventions.
Please limit this change to the modified transmitters for the sake of people who are happily controlling their diabetes today using xDrip as is now.  

There are so many things I would like to change in xDrip.  But, I don't make a proposal because I don't want to possibly screw up people's lives.

tolot27 (approved):
Thank you for your contribution. This PR looks like a major correction of the underlying calculation which results in much more correct values than the previous adjustments [introduced for G4/G5](https://github.com/NightscoutFoundation/xDrip/pull/2140#issuecomment-1138753061).
IMHO, this is an improvement for all users using G6 with raw values. tynbendad made a lot of tests with different calibration codes and provide all data - much more data then previously obtained.

jamorham (approved):

PR #2166, opened 2022-06-08
Consistent glucose unit throughout

Navid200:
If I look up the units for blood glucose, this is what I find:
https://en.wikipedia.org/wiki/Blood_sugar_level

Screenshot 2022-06-07 211032

What worries me specially is this:
https://github.com/NightscoutFoundation/xDrip/blob/4173437074e4c5c1c307dd57d9658a6da7bf065c/wear/src/main/java/com/eveningoutpost/dexdrip/BIGChart.java#L471

Of course, I don't know anything about wear.  So, may be there is a reason to use this instead of using the original parameter.
I mean instead of asking what delta ends with, why not ask what the blood glucose unit is?

Anyway, I am not going to leave this as an open PR because I am really worried about this breaking things.  So, I will leave it as a draft.
But, please let me know what you think.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/AddCalibration.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/AddCalibration.java
@@ -131,7 +131,7 @@ public class AddCalibration extends AppCompatActivity implements NavigationDrawe
                                                     if (convertedBg != null) {
                                                         final long calibration_timestamp = JoH.tsl() - (bgAgeNumber * 1000);
                                                         NativeCalibrationPipe.addCalibration(convertedBg.intValue(), calibration_timestamp);
-                                                        UserError.Log.uel(TAG, "Sending native calibration pipe value: " + convertedBg.intValue() + " mg/dl taken at timestamp: " + JoH.dateTimeText(calibration_timestamp) + " source: " + cal_source);
+                                                        UserError.Log.uel(TAG, "Sending native calibration pipe value: " + convertedBg.intValue() + " mg/dL taken at timestamp: " + JoH.dateTimeText(calibration_timestamp) + " source: " + cal_source);
                                                     }
 
                                                     final Calibration calibration = Calibration.create(calValue, bgAgeNumber, getApplicationContext(), (note_only.equals("true")), localEstimatedInterstitialLagSeconds);
--- a/app/src/main/java/com/eveningoutpost/dexdrip/BestGlucose.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/BestGlucose.java
@@ -126,7 +126,7 @@ public class BestGlucose {
         }
 
         public String humanSummary() {
-            return unitized + " " + (doMgDl ? "mg/dl" : "mmol/l") + (isStale() ? ", " + minutesAgo(true).toLowerCase() : "");
+            return unitized + " " + (doMgDl ? "mg/dL" : "mmol/L") + (isStale() ? ", " + minutesAgo(true).toLowerCase() : "");
         }
 
     }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
@@ -133,7 +133,7 @@ public class CalibrationGraph extends ActivityWithMenu {
         Axis axisX = new Axis();
         Axis axisY = new Axis().setHasLines(true);
         axisX.setName("Raw Value");
-        axisY.setName("Glucose " + (doMgdl ? "mg/dl" : "mmol/l"));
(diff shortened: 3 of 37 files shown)


Navid200:
This actually already exists:
https://github.com/NightscoutFoundation/xDrip/blob/4173437074e4c5c1c307dd57d9658a6da7bf065c/app/src/main/res/values/arrays.xml#L141-L144  

A better solution is to replace all of the text entries in the 37 files with pointers to these existing array values.
Then, if we ever in the future need to change it, we only need to edit the existing array values instead of all the many instances of text all over the place.

jamorham:
I personally think that the capitalized L at the end is ugly, like having a capital at the end of a sentence. That's why its as it is how it is right now, its purely for aesthetics. I don't believe any confusion is caused by the existing lower casing. I'm not really sure what need this addresses, but please tell me if you think differently.

Navid200:
@jamorham I agree with you that this is not a critical matter.

The  only thing is this:
We have an inconsistency.  
If you go to Settings &#8722;> Glucose Units &#8722;> Glucose Units, you will see this.
Screenshot_20220608-184322
I have not done that.  Someone has used the correct case (capital).
If you look at the existing translations for some of the text that I have found in this PR, you can see that some translators have used mmol/L and mg/dL.  And we cannot blame them.  it is just a standard and it is inevitable that we will face this over and over again.

This whole thing started because I opened this PR:
https://github.com/NightscoutFoundation/xDrip/pull/2164
and I wanted to use capital L.

If you say that we should use lower case l, my first comment will be that I wish you would change your mind.
But, my second comment will be that we should either use L everywhere, or we should use l everywhere.

tolot27:
https://en.wikipedia.org/wiki/Litre#Symbol explains why the lower case "l" is the correct SI unit and why "L" was adapted as an alternative (can be better distinguished from "1").
As a scientist I favor native SI units and consistency. In our case we should be consistent as well. Likely, we have to change fewer files. After we finally decide wich variant we use, I can replace the strings at Crowdin in a single batch.

Navid200:
@tolot27 Thanks for the link.  I didn't know that.

We should find out what is acceptable for the blood glucose unit not just the unit of volume, liter.
I now understand what the SI unit promotes.
But, what I wonder is what the majority of doctors expect to see.

Another way of stating this is even if most people use l to represent liter in the UK, what really matters is what the doctors in the UK or other countries expect to see.  
If the unit mg/dL is what the doctor expects to see even though it is scientifically incorrect, we should consider the outcome rather than being scientifically correct at any cost.

I am not suggesting that the majority of the doctor expect to see mg/dL and mmol/L.  What I am saying is that in order to make our decision, I hope that we consider that as an important factor.

Navid200:
This started here:  https://github.com/NightscoutFoundation/xDrip/pull/2164

There are people who use xDrip to create their reports as I have suggested here:  https://navid200.github.io/xDrip/docs/Report.html

My intent is to also add the high and low values on the third statistics page (AGP).
In that case, it will be what the doctors will see.

Navid200:
It seems that different nationalities may have different preferences.  And that must be the reason for the current inconsistency in xDrip.
I can fix this.  But, it will need yet one more setting.
I am not in favor of adding more settings unless absolutely necessary.  This is not a necessity.
Therefore, I close this PR.

tolot27:
I really favor consistency. Since doctors have studied and read many publications and books they should know the correct case or otherwise don't care about it.

Navid200:
I closed this PR because it was about deciding one way or the other for everyone.
I don't think that's a good idea.   I agree that consistency is better than inconsistency.  
But, that's only the first step.  Even if we make everything consistent, we will be dictating to some what they may not prefer.

At least, that was the message I got.  I personally prefer L because it is the official unit where I live.  But, I believe there are places where l is used by almost everyone.
If that is true, wouldn't it be better for us to give the choice to the user?

Navid200:
Giving the option to the user requires something that seems to me to be too complicated for me to implement.  I am concerned about how to use a variable inside config preferences.  Considering it is difficult for me to implement, that's why I opened an issue hoping to get support for implementing it.  

This PR was just about replacing some text with another.  I can do that.
But, using a variable everywhere even inside forms is not something I am sure how to do.

Navid200:
I am trying, not very cleanly, to explain why I opened an issue instead of continuing here.
Both this closed PR and the new issue I opened are about bringing consistency.
But, the issue brings something more that this PR does not.  That is to add a setting allowing user to choose which form to use consistently.

Even if we make everything consistent now, for all we know, a developer will open a PR with the other case.
We will need to spend a lot of time maintaining the consistency.

But, if we add the switch, and if the developers know the  switch exists, they only need to add a pointer to the switch instead of adding the unit itself.  Regardless of what case the developer prefers, using the switch will allow them to use their preferred case.

PR #2170, opened 2022-06-08
Add hyperlink to xDrip Documentation & Support

petervanrijt:
[This site](https://navid200.github.io/xDrip/) of Navid Fo could help a lot of users with (basic) settings, and can prevent questions on Facebook groups like [xDrip](https://www.facebook.com/groups/1021594524580709), [AndroidAPS](https://www.facebook.com/groups/1900195340201874) or [Looped](https://www.facebook.com/groups/1782449781971680). Please review idea / code since this is my first PR for xDrip :) All the best and many thanks for all effort, Peter


app/src/main/res/values/strings.xml
```diff
@@ -408,6 +408,7 @@
     <string name="temporary_work_around_disable_all_sync_detail">Temporary work-around option, completely stops all synchronization. May need reboot after being re-enabled. Use with care!</string>
     <string name="automatic_updates_crash_reports_and_feedback">Automatic Updates, crash reports and Feedback to Developer</string>
     <string name="crowd_sourced_translation">Crowd-sourced translation</string>
+    <string name="documentation_support">Documentation & support</string>
```
tolot27, line 411: You have to use `&amp;` instead of `&`.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -1114,6 +1114,11 @@ public void crowdTranslate(MenuItem x) {
        // startActivity(new Intent(this, LanguageEditor.class).setFlags(Intent.FLAG_ACTIVITY_NEW_TASK));
         startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("https://crowdin.com/project/xdrip")).addFlags(Intent.FLAG_ACTIVITY_NEW_TASK));
     }
+    
+    public void documentationSupport(MenuItem x) {
+        // startActivity(new Intent(this, LanguageEditor.class).setFlags(Intent.FLAG_ACTIVITY_NEW_TASK));
+        startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("https://navid200.github.io/xDrip/")).addFlags(Intent.FLAG_ACTIVITY_NEW_TASK));
```
tolot27, line 1120: As parapenT1sta mentioned, https://xdrip.readthedocs.io fits better because Navids pages are more like for faq/support.

tolot27 (changes requested):
Thanks for your contribution! IMHO, it is a good idea to provide a link to the documentation within the app.

As far as I know, it is not necessary to update `app/src/main/res/values/translation-editor.xml`. I'll check this and likely remove this file and its dependencies.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
petervanrijt: Great! Love the hyperlink to the official documentation. I will do another PR to xDrip documentation to include a link in the troubleshoot menu section to site of Navid. Currently both sites are a valuable resource, which - unfortunately - are separate websites.

Navid200: @petervanrijt Thanks for the contribution.  
  
> unfortunately - are separate websites

Those two being two different websites is a strength in my opinion, not a weakness.
Patrick and I work closely and provide feedback and suggestions to each other.
We constantly tag each other when there is a question that the other person can better handle.

I can ensure accuracy by limiting the material on my site to what I have expertise in.  Same goes I believe for Patrick.

We don't even have links to those sites on the wiki.  In my opinion, we should.

Navid200:
Look at Settings &#8722;> Cloud Upload &#8722;> Nightscout Sync (Rest-API) &#8722;> Use mobile data  

This allows user to remove the possibility of using data service.  
If we want to offer a similar and consistent courtesy to the user community, we shouldn't have any links in the app.  Otherwise, the user could accidentally access a web page not realizing and end up using data without realizing.

If we want to offer the same level of control, we need more switches and the links will become complicated (conditional on the switch).

petervanrijt:
@Navid200 suggestion to change menu-items to:

- Hamburger menu add option Help:
      - Suboption Read the Docs > https://xdrip.readthedocs.io/
      - Suboption Documentation & Support > https://navid200.github.io/xDrip/

What are your thoughts on this?

jamorham:
Please instead create a new `Help.md` file in the `Documentation` folder detailing external websites with help/documentation and for any in-app link point to this file on the xDrip github repository. For the `Readme.md` create also a link to this file instead of linking directly to external sites.

PR #2175, opened 2022-06-11
AAPS profile import Log instead of toast message

Navid200:
Those who use AAPS have an option to disable the sound.  But, there is no way to disable the toast message.
It seems the toast message can be annoying.  https://github.com/NightscoutFoundation/xDrip/discussions/2092 
This changes it to a log message.  

I cannot 100% test this because I don't use AAPS!
I have compiled and am using it on my mission mode device right now.
But, I cannot tell you that I have verified the log.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/profileeditor/ImportAapsProfile.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/profileeditor/ImportAapsProfile.java
@@ -71,7 +71,7 @@ public class ImportAapsProfile {
             if (Pref.getBooleanDefaultFalse("profile_import_sound")) {
                 BackgroundQueue.post(() -> JoH.playSoundUri(JoH.getResourceURI(R.raw.labbed_musical_chime)));
             }
-            JoH.static_toast_long("xDrip imported AAPS profile");
+            UserError.Log.e(TAG, "xDrip imported AAPS profile");
         }
     }
 


app/src/main/java/com/eveningoutpost/dexdrip/profileeditor/ImportAapsProfile.java
```diff
@@ -71,7 +71,7 @@ public static void importAndSaveFromMap(final HashMap<String, Object> tmap) {
             if (Pref.getBooleanDefaultFalse("profile_import_sound")) {
                 BackgroundQueue.post(() -> JoH.playSoundUri(JoH.getResourceURI(R.raw.labbed_musical_chime)));
             }
-            JoH.static_toast_long("xDrip imported AAPS profile");
+            UserError.Log.e(TAG, "xDrip imported AAPS profile");
```
tolot27, line 74: This logs it as an error but I believe it is just an info message. Maybe `uel` or `ueh` is more appropriate.

tolot27 (changes requested):

jamorham:
When this feature was put in I didn't realize that people would have automated systems switching profiles and that it would be more than a manually initiated thing and that was the reason for the sound. I don't know how frequently these profile switches happen for some people but I think `.e` log profile is fine as that is a low level and people either are going to be having a lot of profile changes or very few so those who want to be notified can use the sound. If the synchronization with xDrip is working then it is likely to work every single time so I don't think we need to highlight more than it has happened in the event log.

jamorham (approved):

PR #2176, opened 2022-06-11
Don't allow xDrip to work with OOP2 versions before 1.2 (unless calibrate raw is used)

tzachi-dar:
There have been two versions before the 1.2.
The first version only supported raw calibration mode. Users of this version will not see any difference, and if they try to move to a new mode will get a message saying that they need to upgrade OOP2, or return to "calibrate based on raw" mode.

The second version had a bug and was removed. Still some users might be using it in "no calibration" or "calibrate based on glucose" mode. This users will see xDrip stops working, but this is the preferred way since the banned OOP2 had a bug that might make it show BG values that are very inaccurate (300 instead of 100).
OOP2 does not have an auto-update feature, so this is the only way to revoke it.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NSEmulatorReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NSEmulatorReceiver.java
@@ -29,6 +29,7 @@ import java.util.ArrayList;
 import java.util.UUID;
 
 import static com.eveningoutpost.dexdrip.Models.BgReading.bgReadingInsertFromJson;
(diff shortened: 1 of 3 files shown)


jamorham (approved):
I think this is an acceptable change in behavior as you are eliminating a bug by enforcing a third party component update.

sschamp:
Great!
But where can we grab an updated version of OOP2?
All links seem to lead to versions of december or januari (including the ones from minimallooper and previous no calibration PR)

Navid200:
@sschamp 
https://github.com/NightscoutFoundation/xDrip/discussions/2177

PR #2186, opened 2022-06-25
allow up to two decimal digits for insulin

### LINKED ISSUE #1878: 0.05 unit insulin resolution
I only use 19 units of Humalog per day and often do microboluses of 0.05-0.1 units since my Tandem t:Slim allows for that small of a bolus.  When I go into xDrip+ to add a treatment it allows me to enter 0.05 but displays it as 0.1.  

Any chance you guys can simply go in and reduce the resolution size for that entry?

Thank you

tolot27:
Currently, treatment values with a size of up to six digits are accepted but only one decimal digit is shown on the input summary dialog and more importantly, only one decimal digit is passed downstream. 

That means for example that 1.2 IE FIASP and 0.04 IE Lantus can be entered but they are reported wrongly. Just 0.0 Lantus in the top notification line but 1.2 Lantus in the treatment list (total insulin line is missing). Also, the treatment confirm/cancel buttons are missing (see screenshots below, before and after PR).

 

This PR half-rounds the decimal digits to two (i. e. 0.007 to 0.01) but does not print trailing zeros.

It fixes #1878.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -1,5 +1,15 @@
 package com.eveningoutpost.dexdrip;
 
+import static android.Manifest.permission.WRITE_EXTERNAL_STORAGE;
+import static com.eveningoutpost.dexdrip.Models.JoH.msSince;
+import static com.eveningoutpost.dexdrip.Models.JoH.tsl;
+import static com.eveningoutpost.dexdrip.UtilityModels.ColorCache.X;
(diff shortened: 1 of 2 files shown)


tolot27:
> This looks like it should work.

It indeed works on at least two smartphones.

> Am I right in thinking that the inconsistencies regarding the confirmation button etc only affect's multiple insulins feature?

No, it also affects if multiple insulins are deactivated. I've fixed it, too.

jamorham:
How do you reproduce that bug without multiple insulins enabled?

jamorham (approved):

PR #2188, opened 2022-06-25
Transmitter ID

Navid200:
When I ask people to post a screenshot of their system status page, I add "You can blackout the transmitter ID if you like".
But, not everyone knows what transmitter ID means looking at the system status page as it is currently listed as "Sensor Device".
This PR changes it to Transmitter ID as shown below.
Screenshot_20220625-174151


tolot27 (approved):
Good idea!

Maybe you can also implement switching the blackout the transmitter ID if someone taps on this line. That would avoid additional image editing.

PR #2196, opened 2022-06-28
Notify inadvisable settings

Navid200:
Notify that engineering mode is on.
This notification is issued once a month.


jamorham (approved):
Looks okay to me. Please confirm you have tested it.

Navid200:
I have tested a different version of this that had a 10-minute service instead of a daily service and with the renotify time set to 5 minutes instead of 24 hours to make the test practical.

However, testing this very version will take 30 days.  I will have results in 30 days.  But, I'm afraid, I need the phone for other tests!
Please advise.

Navid200:
I just had an idea, which is probably what you were thinking.  I disabled time sync with internet on the phone and changed the time to move time forward to confirm it will trigger.
I am testing now.

Navid200:
It just showed the 24-hour notification.

Navid200:
Thanks

Issue #682, opened 2018-11-12
Red On Black


Diabeticliberty:
Does anybody have a way of changing the red text colour that appears during warnings of 'low blood sugar' events? At times of near hypoglycemia the text is extremely difficult to read. I have tried at length changing colour settings but the red text seems preset.

Navid200:
@Diabeticliberty Would you please explain exactly what text you mean?
Do you mean the main blood glucose value shown almost at the top of the screen?

Diabeticliberty:
It appears I didn't explain myself adequately. I meant every piece of information that shows up on screen in red at times of low blood sugar.

Navid200:
Do you mean this?
![Screenshot_20210401-112405](https://user-images.githubusercontent.com/51497406/113317414-5962b180-92dd-11eb-950d-75df103b27a6.png)

Diabeticliberty:
Hello, yes that's it. During times of hypoglycemia it is extremely difficult to read. In fairness I struggle with red text against a black background at most other times too. When low on glucose it is really challenging

Navid200:
There is no significant contrast of dark red on black, no matter how good someone's vision is.  
So, I'm not surprised.  I agree with you and would like to change the text color to white regardless of value.  
But, that would be a long-term project and may not be approved considering how everyone has got used to everything as is.
I will work on that.

But, it is possible to change the background color.  If you change the background color from black to gray, you may get better contrast.
You can change colors:
Settings -> xDrip+ Display Settings -> xDrip+ Color Settings

The background color is "Main Chart Background".

If you mess things up, you can just tap on "Reset to Default" for each color individually.

Diabeticliberty:
Hello again, thanks for getting back to me. I tried some time ago changing everything I could possibly find to try and make the colour contrasts work better for me. Nothing really seemed to work all that well which was in essence why I posted the issue on here. In all fairness Xdrip+ has brought me nothing but good since the very first day I started using it. It would of course be extremely good to be able to find a workaround the red and black issue but ultimately it is the highly enhanced way Xdrip+ allows me to monitor and control my type 1 condition that really matters most to me.

All the very best to you and all involved in Xdrip+

Navid200:
@Diabeticliberty Would you please post a screenshot of the following page?

Settings -> Alarms and Alerts -> Extra Alerts (xDrip+)

Diabeticliberty:
[...]

PR #2212, opened 2022-07-07
Add option to disable colored BG Reading text on widget / AOD

Der-Schubi:
The red BG Reading of the widget is very hard to read on a dim screen, e.g. on the Always On Display Widget.
This was discussed here: https://github.com/NightscoutFoundation/xDrip/discussions/2208
Maybe the strings of the option could be a bit better. Any suggestions?

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
@@ -2422,6 +2422,7 @@ public class Preferences extends BasePreferenceActivity implements SearchPrefere
             findPreference("widget_range_lines").setOnPreferenceChangeListener(new WidgetListener());
             findPreference("extra_status_line").setOnPreferenceChangeListener(new WidgetListener());
             findPreference("widget_status_line").setOnPreferenceChangeListener(new WidgetListener());
+            findPreference("widget_disable_colored_text").setOnPreferenceChangeListener(new WidgetListener());
             findPreference("status_line_calibration_long").setOnPreferenceChangeListener(new WidgetListener());
             findPreference("status_line_calibration_short").setOnPreferenceChangeListener(new WidgetListener());
             findPreference("status_line_avg").setOnPreferenceChangeListener(new WidgetListener());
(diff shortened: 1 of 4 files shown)


Richspiral:
Thanks for putting this here. I realise now that the blood sugar number font changes to red everywhere in the app - main display, AOD & widget. It's programmed to go red when a low is reached. While being almost impossible to read on a dimmed screen it's also unnecessary. We don't need the number to be red to tell us we are in trouble. I thought I'd found a temporary work around by stipulating my low number as 0.1 but then i found i lost the calculation 'x minutes to low ' which i find useful. So would still appreciate an option to turn off the display number in red when low program. Thx!

Navid200:
Red means danger.  
I assume that's why it was chosen at the beginning of xDrip years ago.

The xDrip background color is black.

Unfortunately, red on black provides a very low contrast.  As a result, those who need reading glasses, those who have just woken up by a low alert, or those who are visually impaired suffer from this low contrast.

You can see an example of something relevant here in this old issue:  https://github.com/NightscoutFoundation/xDrip/issues/682

Regardless of what we want to do, we have to be very careful that a change should not confuse anyone.
One option would be to add a shade to red so that it becomes red on something less dark than black without changing the text color so that it would still be red representing danger.

Der-Schubi:
We could also use pure red (#FF0000) instead of the toned down one (#C30909):
brightRed
Its a bit better, but not that much.

tolot27 (commented):
I support this PR but suggest an improvement which does not require an additional setting.
It is possible to [detect programmatically](https://stackoverflow.com/questions/44170028/android-how-to-detect-if-night-mode-is-on-when-using-appcompatdelegate-mode-ni/44170179#44170179), if the night mode is active and disable the coloring just if it is on or choose a different color. Another way is to simply have a resource variant below in the folder `res/values-night` which have a different background or the suggested font border.

Der-Schubi:
> I support this PR but suggest an improvement which does not require an additional setting. It is possible to [detect programmatically](https://stackoverflow.com/questions/44170028/android-how-to-detect-if-night-mode-is-on-when-using-appcompatdelegate-mode-ni/44170179#44170179), if the night mode is active and disable the coloring just if it is on or choose a different color.
Do you mean the ExtraDim option from Android 12? I already tested this once and it worked fine.
My problem with this was that the text is also way to dark if the display just dims down during the day. Happens for example in my car.
I'm trying to make the colors configurable now. We have a dozen colorpickers for the graph anyways, no harm having three more.

Der-Schubi:
I made a new PR #2214 with selectable text colors.
Should I close this one here?

tolot27:
Closed in favor of #2214.

PR #2214, opened 2022-07-08
Configurable text color for high-, in-range- and low-clucose values

Der-Schubi:
There are three new color pickers for changing text and trend arrow colors of high-, in-range and low-glucose values on main-screen, widget and also AOD widget. The default colors are set to the previous colors.

Screenshot of the colorpickers with default colors:

Screenshot of main screen with changed color setting for in-range values:

Screenshot of widget with changed color setting for in-range values:

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/BestGlucose.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/BestGlucose.java
@@ -113,11 +113,11 @@ public class BestGlucose {
             if (isStale()) wholeSpan(ret, new StrikethroughSpan());
             if (color) {
                 if (isLow()) {
-                    wholeSpan(ret, new ForegroundColorSpan(getCol(ColorCache.X.textcolor_low_values)));  //TESTING
+                    wholeSpan(ret, new ForegroundColorSpan(getCol(ColorCache.X.color_low_bg_values)));
                 } else if (isHigh()) {
-                    wholeSpan(ret, new ForegroundColorSpan(getCol(ColorCache.X.textcolor_high_values))); //TESTING
+                    wholeSpan(ret, new ForegroundColorSpan(getCol(ColorCache.X.color_high_bg_values)));
                 } else {
-                    wholeSpan(ret, new ForegroundColorSpan(getCol(ColorCache.X.textcolor_inrange_values)));
+                    wholeSpan(ret, new ForegroundColorSpan(getCol(ColorCache.X.color_inrange_bg_values)));
                 }
             }
             return ret;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/BestGlucose.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/BestGlucose.java
@@ -13,12 +13,14 @@ import com.eveningoutpost.dexdrip.Models.JoH;
 import com.eveningoutpost.dexdrip.Models.SensorSanity;
 import com.eveningoutpost.dexdrip.Models.UserError;
 import com.eveningoutpost.dexdrip.UtilityModels.BgGraphBuilder;
+import com.eveningoutpost.dexdrip.UtilityModels.ColorCache;
 import com.eveningoutpost.dexdrip.UtilityModels.Pref;
 import com.eveningoutpost.dexdrip.calibrations.CalibrationAbstract;
 import com.eveningoutpost.dexdrip.utils.DexCollectionType;
 
 import java.util.List;
 
+import static com.eveningoutpost.dexdrip.UtilityModels.ColorCache.getCol;
 import static com.eveningoutpost.dexdrip.calibrations.PluggableCalibration.getCalibrationPluginFromPreferences;
 
 /**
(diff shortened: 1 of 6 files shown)


Der-Schubi:
@Richspiral, @tolot27, @Navid200
I think this is ab better solution as the last PR (#2212).

Richspiral:
Sounds like a great suggestion to me and would fix the current issue. Thx!

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/ColorCache.java
```diff
@@ -70,7 +70,10 @@ public enum X {
         color_smb_line("color_smb_line"),
         color_number_wall("color_number_wall"),
         color_number_wall_shadow("color_number_wall_shadow"),
-        color_basal_tbr("color_basal_tbr");
+        color_basal_tbr("color_basal_tbr"),
+        textcolor_high_values("textcolor_high_values"),
+        textcolor_inrange_values("textcolor_inrange_values"),
+        textcolor_low_values("textcolor_low_values"),;
```
tolot27, line 76: Please can you put the semicolon on a new line? In that case a future diff would be smaller. 

Also, is the color limited to text? The arrow is also changed (indeed, it is a Unicode character). If the prefix 'text' is removed, it would be in line with the other colors and the usage can be generalized.

app/src/main/java/com/eveningoutpost/dexdrip/BestGlucose.java
```diff
@@ -111,11 +113,12 @@ public SpannableString spannableString(String str, boolean color) {
             if (isStale()) wholeSpan(ret, new StrikethroughSpan());
             if (color) {
                 if (isLow()) {
-                    // TODO should colors be configurable?
-                    wholeSpan(ret, new ForegroundColorSpan(Color.parseColor("#C30909")));
+                    wholeSpan(ret, new ForegroundColorSpan(getCol(ColorCache.X.textcolor_low_values)));  //TESTING
                 } else if (isHigh()) {
-                    wholeSpan(ret, new ForegroundColorSpan(Color.parseColor("#FFBB33")));
-                } // else default to whatever default is?
+                    wholeSpan(ret, new ForegroundColorSpan(getCol(ColorCache.X.textcolor_high_values))); //TESTING
```
tolot27, line 118: Is the comment `TESTING` still required?

app/src/main/res/values/strings.xml
```diff
@@ -325,6 +325,7 @@
     <string name="high_glucose_values">High Glucose Values</string>
     <string name="in_range_glucose_values">In-range Glucose Values</string>
     <string name="default_color_selected">Default Color Selected</string>
+    <string name="glucose_values_textcolor">Text Colors for Glucose Values</string>
```
tolot27, line 328: Same as above regarding 'text'.

tolot27 (changes requested):
This PR will indeed fix #682 but is still a static solution. I would approve it as a first step. Later on we could implement a dynamic solution based on day/night mode and display brightness.

app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/ColorCache.java
Der-Schubi: color_[high/inrange/low]_values is already taken for graph colors. Can't really think of better names. Suggestions?

Richspiral:
Apologies, I didn't realise that the proposed solution hard coded the colours until time to low. I mistakenly thought it was a colour picker . We need at least one option where the colour is white (or can be selected to be white) no matter what the number. Anything which doesn't include this as an option risks the number being misidentified either due to poor contrast or colour blindness or some other reason.

Der-Schubi:
> Apologies, I didn't realise that the proposed solution hard coded the colours until time to low. I mistakenly thought it was a colour picker .

This PR makes the three possible colors of the BG readings configurable with a color picker. I just stated that this does not include the text in the upper left of the home screen that says "low predicted in XX minutes". The colors of this text are still hard-coded.

Richspiral:
Ah I see, thx for clarifying for me.

Der-Schubi:
As I said, this "low predicted" text can have three colors (dependent on the number of minutes to low predicted), green for more than 60 minutes, yellow for less than 60 and red for a setting named "low_predicted_alarm_minutes".
My PR includes three color pickers, low clucose color is normally red, in-range is white and high is yellow. So there is none left for the green color of this low predicted message.
My suggestion is to use the white / in-range color instead of green. Then we can make this text configurable too.

jamorham:
Can the PR be updated to explain the rationale for the changes and what will be affected. Before and after screenshots are also good for a UI feature. In an ideal world I would prefer that nothing is changed to existing behavior, but new color options being introduced, so long as they don't clutter, are good to allow users to customize the look and feel.

One thing this might address for my own usage is when the decimal point in the glucose value in mmol can be confused for one of the graph dots as they are the same color when high.

Der-Schubi:
@tolot27 I did the changes you wanted.
@jamorham Existing behavior is not changed, the default colors are set to the previous colors.
Affected are the BG value itself and the trend arrow on main screen, widget and AOD widget.

Screenshot of the colorpickers with default colors:
Screenshot_20220713-142031

Screenshot of main screen:
Screenshot_20220713-135631

Screenshot of widget:
Widget

tolot27 (approved):
LGTM. As a last step please move the description and screenshots of https://github.com/NightscoutFoundation/xDrip/pull/2214#issuecomment-1183165795 to the first comment. It is best to have the (improved) description at the top of the PR.

jamorham (approved):

tolot27:
@Der-Schubi Thank you so much for your contribution!

PR #2245, opened 2022-07-26
Ability to choose the Dexcom Communication Slot on WearOS

funkstille:
Ability to choose the communication slot between wearOS watch (tested Sony SWR50) and Dexcom G6. 
In Watch collector mode. 

I use this feature for reading CGM Values from Dexcom G6 (on Slot 1) while it is also connected to a mobile phone (with CamAPS) on Slot 2. Or if Pump and phone are while swimming in a safe place and still have the ability to read CGM Values with a waterproof watch (SWR50).


wear/src/main/res/values/strings.xml
```diff
@@ -269,4 +278,6 @@
     <string name="ob1_session_restarted_title">Sensor session restarted</string>
     <string name="ob1_session_restarted_msg">The sensor session was preemptively restarted and set to 3 days, please check if next readings are off and calibrate if needed</string>
     <string name="ob1_session_restarted_note">Transmitter session was restarted</string>
+    <string name="pref_dex_slot">Select Dexcom Bluetooth Slot</string>
+    <string name="pref_summary_dex_slot">Dexcom G6 has 2 Slots. 1 - Medical Device, 2 - Phone. Each SLot can only be used by one End device. Default: 2</string>
```
tolot27, line 282: SLot -> Slot
End -> communication

Maybe this description is too long.

wear/src/main/res/values/strings.xml
```diff
@@ -27,6 +27,15 @@
         <item>1</item>
         <item>2</item>
     </string-array>
+    
+    <string-array name="dex_slot">
+        <item>1</item>
+        <item>2</item>
+    </string-array>
+    <string-array name="dex_slot_values">
```
tolot27, line 35: Why is the second array necessary. It is identical to the first one.

Please can you also add `3` for customized transmitters?

tolot27 (changes requested):

wear/src/main/res/values/strings.xml
funkstille: What’s a customized transmitter? Haven’t heard of it? Alternative Firmware? More slots ?

tolot27: > What’s a customized transmitter? Haven’t heard of it? Alternative Firmware? More slots ?

PM me

jamorham:
This should already be supported by selecting the `Only use Wear` option. I'm wondering if there could be a bug that this doesn't interoperate properly with the OB1 collector.

wear/src/main/res/values/strings.xml
funkstille: Will add slot3 asap

funkstille:
> 

At least it did not work for me. So I started digging and @tolot27 pointee me to the commit where this feature added to xDrip. I just copy & pasted it to wearOS

PR #2264, opened 2022-08-07
Some refactoring and test coverage

vin:
I just put together a couple of small changes to get the hang of developing in this codebase, running tests, etc.

I extracted some named helper functions and added tests for them, then refactored the code and ensured that the tests stayed green.


vin:
Ah, oops I just realized these commits have my employer's email address on them -- let me change that before we do anything with this.

vin:
Email address fixed, ready for review. Thanks!

vin:
Is there another process for requesting a code review?

Navid200:
> Is there another process for requesting a code review?

I could be wrong and may be corrected.

It may help if you explain what's wrong and why it is necessary to fix it.
There are many bugs already.
There are many requests for new features etc.
They cannot all be addressed right away.  That's why it may help to highlight why this is an important issue that needs to be fixed.

Presenting steps for reproducing the problem can be very helpful.
 
Then, you can follow that by explaining how your changes are fixing it.

Then, you can explain the steps you have taken to verify and test your changes.

Reading your first post in this PR, I see none of any of that.  So, I have no idea what this is for.
But, I cannot emphasize enough that I am not the person who can decide one way or another what will happen to your PR even if you follow my points.

jwoglom:
@Navid200 this seems like a pretty straightforward refactoring which adds a helper method for padding a string and then adds a unit test for it -- this doesn't add any new features or fix any bugs. I don't see any benefit to not merging it. Assuming we want to expand the number of xDrip developers, and incrementally make the app better and more testable, it's to everyone's benefit to merge trivial PR hotfixes like this.

tolot27 (approved):
No complexity and straightforward to merge.

PR #2269, opened 2022-08-12
Feature more import export

gruoner:
this new feature exports the database-export and the pref-file (and any other relevant file to come) either to a local directory (e.g. the users Download) or to a webdav server.
To configure this new feature, xdrip got a new dialog to enable and configure these two export-targets.
When enabled, the export of the database (existing menu-item) is redirected to the new export-function - if not, the old-school export takes place
When importing/restoring a backup, one can select which style to use (old-school, local directory or webdav-server) and the restore takes place as selected.


Navid200:
A problem I see we have with both database and settings is that I don't know if we have the option to import either.

I mean imagine if I have a database save today.
Tomorrow I travel somewhere and my phone stops working.
I get a replacement phone and set it up and use it with xDrip.

In 2 months, I get my hands on the database save I made, which is now 2 months old.
The only option I have is to overwrite the current database with it.  IN that case, I will lose all the data I have collected over the past two months.
But, if I could just import the backup without overwriting anything, I could effectively get everything into xDrip.
<br/>  

We have the same problem, as far as I know, with settings.
Imagine if I want to save only 10 particular settings that are important to me.  When I install xDrip on a new phone and set everything up, I want to be able to add those 10 settings without changing any of the existing settings.
As far as I know, the only way to create a backup of the settings is to export the settings, which exports all of them.
And when you import the file, it overwrites all the settings, not just some.
<br/>  

Does this PR address either of those two shortcomings?

gruoner:
@Navid200 yes of course -- thats the main reason.
We need the have a simultaneous backup of database and config(file) in a single representation (i choosed a zip file but there might be other options) -- and i didn't want it on google (drive) because my companion device is totally google-free

Navid200:
So, this PR adds a new menu where the user can choose which specific setting to save and which not to?
Or, does it save all the settings, but when importing them, the user gets a menu where they choose which settings to import and which ones to leave alone?

Navid200:
@gruoner Do you have any idea for adding the capability to xDrip so that it can import settings?

Let's say you have customized your xDrip display settings;  someone sees it and asks you to give them your xDrip display settings.
If you save your settings and provide the saved settings file to the individual and they import it into their xDrip, they will overwrite their settings with all of yours, not just your xDrip display settings.

As far as I know, xDrip can only overwrite all settings with the settings that you load.  But, it cannot just import some settings and leave the rest alone.

That was just an example.  There may be other cases where  a user may need to just import some settings while leaving the rest alone.
Do you have any idea for something like that?

gruoner:
@Navid200 there are a lot of options but all of them are tickering in preference class like QR codes do. For me it looks like a merge of two preference files and someone needs to decide which pref has precedence.
If we can control the creation of the settings file to get imported on the other xdrip, precedence can be presumed and we can do a dirty merge just via replacement within the preference file....

PR #2284, opened 2022-08-20
Minor repositioning. No functional changes.

Lupum1001:
I'm essentially putting this here to try and split it off from a separate PR.
All this does is reorder two preferences, so that "low level prediction values" doesn't look like it's part of the EULA category.

### BEFORE
<image width="300" src="https://user-images.githubusercontent.com/52087143/185745304-be11c440-332e-47a4-ac61-bf05eaa03309.png">

### AFTER

<image width="300" src="https://user-images.githubusercontent.com/52087143/185745194-1ae1c782-2bde-45f1-8f10-f093e30f3169.png">

### DIFF
--- a/app/src/main/res/xml/xdrip_plus_prefs.xml
+++ b/app/src/main/res/xml/xdrip_plus_prefs.xml
@@ -717,19 +717,6 @@
                 </intent>
             </Preference>
         </PreferenceCategory>
-        <PreferenceCategory
-            android:key="title_I_understandb"
-            android:title="@string/eula">
-            <Preference
-                android:key="I_understandb"
-                android:summary="@string/not_for_medical_use"
-                android:title="@string/end_user_license_agreement">
-                <intent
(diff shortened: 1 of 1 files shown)


philou78:
EULA is also present at the top of the main settings page.

Why not just remove EULA from the predictive simulation settings page?

Lupum1001:
That's what I did in my last PR, but concerns were raised about removing it, because it's sort of acting as a crude reminder to not trust those features in particular. Happy to replace it with a pop-up warning or something later.
https://github.com/NightscoutFoundation/xDrip/pull/2057#issuecomment-1121814837

Just trying to keep this as uncontroversial as possible, so it goes through quickly. It's a minor tweak, and I don't want it to eat up @jamorham's time.

philou78:
Thank's for your reply.
I agree EULA issue is another subject

jamorham (approved):

PR #2285, opened 2022-08-21
Custom Sound File

Navid200:
This is a second attempt at an older PR (https://github.com/NightscoutFoundation/xDrip/pull/1827).  

**Why we need this**
If you tap on "Choose File" on the edit alert page in order to change the sound file for one of your main glucose level alerts, you will see this:
Screenshot_20210915-140641

If you then tap on custom sound and navigate to an mp3 file, you will see that it is grayed out.  So, you cannot choose a custom sound file.
The solution is to add the custom sound file to your Notifications folder (as explained here: https://navid200.github.io/xDrip/docs/Custom-sound-grayed-out.html) and use the "System Sound/Alarm" option to choose it.

The custom sound option on the form is not functional on many phones and confusing to the users.
You can see an example here:  https://github.com/NightscoutFoundation/xDrip/discussions/2282

In case you don't experience this problem on your phone, you can recreate it using virtual device as explained here:  https://github.com/NightscoutFoundation/xDrip/discussions/1983 

There is an identical problem if you go to xDrip reminders and attempt to change the sound file for one of the reminders.

If you experience this problem on your phone, you will have the same problem if you choose to use a custom sound file for any other alert like the Forecast Low alert or the Persistent High Alert.

This PR removes the "Custom sound/alarm" option and makes tapping on the choose file behave as shown below.  
Screenshot_20211117-225530

And for reminders:
Screenshot_20211117-230119

This PR also removes the all caps category title of Alerts and notifications from the Alarms and Alerts page, which is redundant with respect to the existing title of "Alarms and Alerts".  
And adds a note on the same page explaining where custom files should be placed, as shown below:
Screenshot_20220821-181119

**Is there a workaround?**
The only workaround is to avoid using the custom sound option.  

No

**Tests**
Tested with Android 8 and 11 and virtual Android 11

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
@@ -589,12 +589,12 @@ public class EditAlertActivity extends ActivityWithMenu {
         buttonalertMp3.setOnClickListener(new View.OnClickListener() {
             public void onClick(View v) {
                 AlertDialog.Builder builder = new AlertDialog.Builder(mContext);
-                builder.setTitle("What type of Alert?")
+                builder.setTitle(getString(R.string.select_new_sound))
                         .setItems(R.array.alertType, new DialogInterface.OnClickListener() {
                             public void onClick(DialogInterface dialog, int which) {
                                 if (which == 0) {
                                     Intent intent = new Intent(RingtoneManager.ACTION_RINGTONE_PICKER);
-                                    intent.putExtra(RingtoneManager.EXTRA_RINGTONE_TITLE, "Select tone for Alerts:");
+                                    intent.putExtra(RingtoneManager.EXTRA_RINGTONE_TITLE, getString(R.string.choose_alert_sound_file));
                                     intent.putExtra(RingtoneManager.EXTRA_RINGTONE_SHOW_SILENT, true);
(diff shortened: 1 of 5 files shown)


Navid200:
The following image shows a screenshot of the list of sound files I see when I attempt to change the sound for one of my alerts.  I have highlighted two custom sound files I have placed in the Notifications folder, which appear in the list with all the other existing sounds that came with the phone.
Screenshot_20220823-074613

tzachi-dar:
I have tried testing this with a galaxy S9+ with android 8, and I could not add an mp3 file to be played.
I have put the file under the notifications folder, but xdrip could not see it.

Navid200:
@tzachi-dar You need to play the file with any player to see what the title of the clip is.  It may be different than the filename.
What will appear in the list is the title of the sound clip.

And the folder's name  has to be Notifications with capital N at the beginning.

tzachi-dar:
I'm afraid that the file was still not found (it was in the Notifications folder, and even with the name of the clip still not found).
So,  https://github.com/NightscoutFoundation/xDrip/pull/1986 seems like a better option right now.

Navid200:
Thanks for testing.  I wish I had that phone.
Yes, let's see what happens with that PR.  Hopefully, it will stop the complaints.  As long as that happens, I don't care one way or the other.

Thanks

PR #2294, opened 2022-08-25
Add predicted iob data to broadcastservice

bigdigital:
Some people would like to see predicted(simulated) iob data on the watch. So this PR extends the broadcast data with a  predicted IOB and BWP info.
Also it enables **not** simple mode in _defaultLines_  function in order to receive simulated data including predicted lgraph lines.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/broadcastservice/BroadcastService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/broadcastservice/BroadcastService.java
@@ -527,6 +527,9 @@ public class BroadcastService extends Service {
                 bundle.putLong("treatment.timeStamp", treatment.timestamp);
             }
 
+            predictedIOB = "";
+            predictedBWP = "";
+
             if (settings.isDisplayGraph()) {
                 long graphStartOffset = settings.getGraphStart();
                 long graphEndOffset = settings.getGraphEnd();
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Services/broadcastservice/BroadcastService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Services/broadcastservice/BroadcastService.java
@@ -12,6 +12,8 @@ import android.os.IBinder;
 import android.os.Parcelable;
 import android.os.PowerManager;
 import android.preference.PreferenceManager;
+import android.support.v4.content.LocalBroadcastManager;
+import android.util.Log;
 
 import com.eveningoutpost.dexdrip.BestGlucose;
 import com.eveningoutpost.dexdrip.Models.Accuracy;
@@ -29,6 +31,7 @@ import com.eveningoutpost.dexdrip.Services.MissedReadingService;
 import com.eveningoutpost.dexdrip.UtilityModels.AlertPlayer;
 import com.eveningoutpost.dexdrip.UtilityModels.BgGraphBuilder;
 import com.eveningoutpost.dexdrip.UtilityModels.Constants;
+import com.eveningoutpost.dexdrip.UtilityModels.Intents;
 import com.eveningoutpost.dexdrip.UtilityModels.Pref;
 import com.eveningoutpost.dexdrip.UtilityModels.PumpStatus;
 import com.eveningoutpost.dexdrip.stats.StatsResult;
@@ -75,6 +78,11 @@ public class BroadcastService extends Service {
 
     protected String TAG = this.getClass().getSimpleName();
     protected Map<String, BroadcastModel> broadcastEntities;
+
+    private BroadcastReceiver statusReceiver;
+
+    private String predictedBWP;
+    private String predictedIOB;
 
     /**
      *  The receiver listening {@link  ACTION_WATCH_COMMUNICATION_RECEIVER} action.
@@ -209,7 +217,6 @@ public class BroadcastService extends Service {
         }
     };
 
-
     @Override
     public IBinder onBind(Intent intent) {
         return null;
@@ -227,6 +234,25 @@ public class BroadcastService extends Service {
         registerReceiver(broadcastReceiver, new IntentFilter(ACTION_WATCH_COMMUNICATION_RECEIVER));
 
         JoH.startService(BroadcastService.class, Const.INTENT_FUNCTION_KEY, Const.CMD_START);
+
+        statusReceiver = new BroadcastReceiver() {
+            @Override
+            public void onReceive(Context context, Intent intent) {
+                final String bwp = intent.getStringExtra("bwp");
+                if (bwp != null) {
+                    predictedBWP = bwp;
+                } else {
+                    final String iob = intent.getStringExtra("iob");
+                    if (iob != null) {
+                        predictedIOB = iob;
+                    }
+                }
+            }
+        };
+
+        LocalBroadcastManager.getInstance(this).registerReceiver(statusReceiver,
+                new IntentFilter(Intents.HOME_STATUS_ACTION));
+
         super.onCreate();
     }
 
@@ -235,6 +261,12 @@ public class BroadcastService extends Service {
         UserError.Log.e(TAG, "killing service");
         broadcastEntities.clear();
         unregisterReceiver(broadcastReceiver);
+
+        try {
+            LocalBroadcastManager.getInstance(this).unregisterReceiver(statusReceiver);
+        } catch (Exception e) {
+            Log.e(TAG, "Exception unregistering broadcast receiver: " + e);
+        }
         super.onDestroy();
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Services/broadcastservice/BroadcastService.java
```diff
@@ -495,6 +527,9 @@ protected Bundle prepareBgBundle(BroadcastModel broadcastModel) {
                 bundle.putLong("treatment.timeStamp", treatment.timestamp);
             }
 
+            predictedIOB = "";
```
tolot27, line 530: IMHO this always sets the variables to an empty string which does not change till line 569. Maybe you should initialize the strings to `""` in line 84/85?

tolot27 (changes requested):

app/src/main/java/com/eveningoutpost/dexdrip/Services/broadcastservice/BroadcastService.java
bigdigital: This is a place where the variable resets in case if the iob data would not be updated in BgGraphBuilder.

bigdigital: But in case if settings.isDisplayGraph() is false, the predictedIOB and predictedBWP always would be empty. Maybe it would be better to transfer iob only if graph is enabled.

tolot27: The only location where `predictedIOB` and `predictedBWP` are set to likely non-empty value is in the `statusReceiver` `onReceive` event. The next use is `bundle.putString("predict.IOB", predictedIOB);` but right before the if condition these two variables will be emptied. Hence, en empty string is always passed to `bundle.putString`. That's my main concern. Did I oversee anything?

bigdigital: Yes. the variables are filled via statusReceiver by receiving the broadcast messages.  And those broadcast messages can  triggers by bgGraphBuilder.defaultLines(false)  function which calls addBgReadingValues(simple) and the last one function calls  Home.updateStatusLine("iob", df.format(iob.iob))  . So when the code reaches to the bundle.putString , the **iob** and **wpb** values would be filled with the valid data.

tolot27: Okay, many of the functions are annotated with `synchronized` but `sendBroadcast(intent)` broadcasts the given intent to all interested BroadcastReceivers [asynchronously](https://developer.android.com/reference/android/content/Context#sendBroadcast(android.content.Intent). How can you be sure that all operations are finished until the `bundle.putString` lines are reached?

bigdigital: You are absolutely right that there is no guarantee that broadcast msg will be received at the moment the bundle.putString lines are reached.  I thought about this and do not found any other way to make sure the predictedIOB  would be cleared.  The similar broadcast receiver for predictedIOB  is located in Home.java and it is reset to "" by onResume which is also not the best place to reset this info. But it is at least reset the value after you reopen the app.

tolot27: How about this:


```java
    private String predictedBWP = "";
    private String predictedIOB = "";

...

            public void onReceive(Context context, Intent intent) {
                final String bwp = intent.getStringExtra("bwp");
                if (bwp != null) {
                    predictedBWP = bwp;
                } else {
                    predictedBWP = "";            
                }
                final String iob = intent.getStringExtra("iob");
                if (iob != null) {
                    predictedIOB = iob;
                } else {
                    predictedIOB = "";            
                }
            }
```

and in `onBind` moving `predictedBWP = ""; predictedIOB = "";` after `bundle.putString`.

PS: I've moved out `iob` from the `bwp` else part because i thought that these values are set independently (in certain cases). Is this correct or is `iob` only set/changed if `bwp` is set/changed?

bigdigital: In the `BgGraphBuilder.java `you can see that `Home.updateStatusLine("bwp", bwp_update) `will triggers almost always, while the `prediction_enabled && simulation_enabled`. 
The `Home.updateStatusLine("iob", df.format(iob.iob)) `has more complex logic. So in most cases the bwp would trigger broadcast. But this will happen until the user did not switch off the  `simulation_enabled` option.

bigdigital: Ah, It looks like if use your proposed code, the iob would be always empty, because when BgGraphBuilder will send  **bwp** broadcast, it will clear the **iob** info ...

4RK4N:
so what exactly is blocking the merge of this PR ?

Navid200:
@bigdigital Your last post in this PR is a question.  Do you believe this PR is ready as is?

If this PR is merged, what else could WatchDrip need in the future that would require another change to xDrip?
Is it not possible to create a mechanism that would allow editing what Nightscout sends if needed in the future?
For example, is it not possible for WatchDrip to request a parameter and xDrip sends it on the next broadcast?  Then, if 2 months from now, WatchDrip needs another parameter, you will not need to open another PR.  You will just modify WatchDrip to request something that it did not previously.

Or, is that not practical?

If that is not practical, can you think of any other parameters that may become required in the future and include them right now in this same PR?

Please note none of my comments are needed to be answered and addressed.  I am just hoping to help the conversation flowing.

oakeley:
Please could this be added to xDrip? This would enable so much. Many thanks @bigdigital for proposing it and @4RK4N for seconding it

bigdigital:
> Do you believe this PR is ready as is?

The last time i tested it, it works fine. 

> If this PR is merged, what else could WatchDrip need in the future that would require another change to xDrip?

Maybe a small fix to transfer the dismissed snooze alarms action to the watch https://github.com/bigdigital/watchdrip/issues/4

> For example, is it not possible for WatchDrip to request a parameter and xDrip sends it on the next broadcast? Then, if 2 months from now, WatchDrip needs another parameter, you will not need to open another PR. You will just modify WatchDrip to request something that it did not previously.

This would be of course great, but as i know there is no such global store, where you can get all those values. For example, predicated IOB  and  BWP are not stored anywhere. And updates only by graph update.

4RK4N:
> > Do you believe this PR is ready as is?
> 
> The last time i tested it, it works fine.
> 
> > If this PR is merged, what else could WatchDrip need in the future that would require another change to xDrip?
> 
> Maybe a small fix to transfer the dismissed snooze alarms action to the watch [bigdigital/watchdrip#4](https://github.com/bigdigital/watchdrip/issues/4)
> 
> > For example, is it not possible for WatchDrip to request a parameter and xDrip sends it on the next broadcast? Then, if 2 months from now, WatchDrip needs another parameter, you will not need to open another PR. You will just modify WatchDrip to request something that it did not previously.
> 
> This would be of course great, but as i know there is no such global variable store, where you can get all those values. For example, predicated IOB and BWP are not stored anywhere. And updates only by graph update.

adding parsed iob cob and maybe tbr data from the external status line aaps sends into xdrip would be good. i have a hacky modification doing that right now. for a proper PR the ExternalStatusService.java would need a getCOB() added and maybe some testing on getTBR(), getIOB() and getAbsoluteBR() if they output the desired result.

many xdrip users also use aaps and in turn the xdrip predictions are not as good as aaps predictions. which is totally ok since that's not the primary purpose of xdrip. but to have the aaps external status data sent by the broadcastservice would help aaps users a lot.

app/src/main/java/com/eveningoutpost/dexdrip/Services/broadcastservice/BroadcastService.java
bigdigital: @tolot27 as i understand this pr was stuck because the "bwp" and "iob" broadcast data which is generated in BgGraphBuilder transfers with help of sendBroadcast finction which is runs asynchronously. Maybe to resolve this problem we can change  sendBroadcast in [updateStatusLine ](https://github.com/NightscoutFoundation/xDrip/blob/master/app/src/main/java/com/eveningoutpost/dexdrip/Home.java#L690) to [sendBroadcastSync](https://developer.android.com/reference/androidx/localbroadcastmanager/content/LocalBroadcastManager#sendBroadcastSync(android.content.Intent))  which should block the execution of the thread .

jamorham: What about using the keystore data like is done with `bwp_last_insulin` in `PebbleDisplayTrend` ? I could easily add more keystore fields if this doesn't include everything needed. Then you don't need to worry about asynchronous broadcasts.

bigdigital: Yep. I guess this should help to transfer data across different parts of the application.  It is enough to add bwp_update data (but maybe better to divide the current string inot two separate variables with the carb and insulin info). And a separate keystore for "iob" which is declared by this line ` Home.updateStatusLine("iob", df.format(iob.iob));`

bigdigital: Yes. Now works fine, added appropriate code changes

jamorham (approved):

PR #2295, opened 2022-08-25
Decommission "lower fuzzer"

Navid200:
There is no need for this setting any longer, which lowered Fuzzer to 37.5 seconds.
Fuzzer is now always 30 seconds.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
@@ -15,7 +15,6 @@ import android.view.View;
 import android.widget.Toast;
 
 import com.eveningoutpost.dexdrip.AddCalibration;
-import com.eveningoutpost.dexdrip.BuildConfig;
 import com.eveningoutpost.dexdrip.GcmActivity;
 import com.eveningoutpost.dexdrip.Home;
 import com.eveningoutpost.dexdrip.Models.APStatus;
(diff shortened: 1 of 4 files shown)


Navid200:
I am currently running this as my active xDrip.

tolot27 (approved):
Only one side note: please can you link the commit or source location containing the fixed fuzzer value of 30?

Navid200:
> Only one side note: please can you link the commit or source location containing the fixed fuzzer value of 30?

Sure,

This was the PR that added "Lower fuzzer" :
https://github.com/NightscoutFoundation/xDrip/pull/1760

And this is the current line that permanently sets the fuzzer to 30 seconds:. https://github.com/NightscoutFoundation/xDrip/blob/fd5ff00ac31b043a3f2ab833c3a33954e5f15057/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java#L92

jamorham:
@Navid200 are you happy with how the fuzzer is currently working in the latest/recent versions with the 30 second value?

Navid200:
> @Navid200 are you happy with how the fuzzer is currently working in the latest/recent versions with the 30 second value?

I cannot reproduce the problem the "Lower fuzzer" setting was meant to fix.  The problem occurred randomly about once a year.  I thought about attempting to recreate the problem by changing the local time of the phone.  But, I never did.
Therefore, I cannot tell you that the problem has been fixed or not.

But, the fact is that there is no need for the "Lower fuzzer" setting any longer considering it will in fact now raise the fuzzer instead of lowering it.
And I would like us to remove settings that are not needed any longer.

Thanks

jamorham (approved):

PR #2296, opened 2022-08-25
Alerts: fix custom sound handling

jamorham:
This PR is designed to fix the problem with custom sound selection for Alerts

Previously these were unselectable and would not function even if they could have been selected.

This PR is designed to supersede https://github.com/NightscoutFoundation/xDrip/pull/1986 and https://github.com/NightscoutFoundation/xDrip/pull/2285 and resolve issues as described here: https://github.com/NightscoutFoundation/xDrip/discussions/1983 and https://github.com/NightscoutFoundation/xDrip/discussions/2282

It gives basically the same treatment as was prototyped for custom sounds for `Reminders` which was done in this commit: https://github.com/NightscoutFoundation/xDrip/commit/9f3ebd1b2c7981a5dde3e56ca77bff4a0c4feced

The changes are:

- Selection to use wildcard mime type `audio/*`
- URI is stored instead of absolute path
- `takePersistableUriPermission()` is used to grant permission past reboot
- Utility version of `setMediaDataSource()` is used to handle different file opening mechanisms
- Strings are extracted from some prompts to enable localization

Because this touches `AlertPlayer` I have raised a PR so that additional testing can be done before merging with the nightly. 

The utility method should be able to handle old and new style uris/paths and will fallback to a default sound if, for example, the custom sound file is deleted or moved.

I have tested this on Android 11 and 12 but not on previous models. Android documentation implies that this should work down to Android 4 and overall this should be more the "official" way that these files should be handled and so should have a good chance to succeed. 

I think this is a cleaner way to handle things overall with fewer changes and making life easier for the user and fixing the original feature which was broken rather than removing it.

Testing on Android 6 and 8 would be helpful. Comments welcomed. Ideally I would like to merge this 1st Sept.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/AlertList.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/AlertList.java
@@ -253,23 +253,9 @@ public class AlertList extends ActivityWithMenu {
         listViewHigh.setAdapter(simpleAdapterHigh);
     }
 
-    private String shortPath(String path) {
+    private String shortPath(final String path) {
         try {
-            if (path != null) {
-                if (path.length() == 0) {
-                    return "xDrip Default";
-                }
-                Ringtone ringtone = RingtoneManager.getRingtone(mContext, Uri.parse(path));
-                if (ringtone != null) {
-                    return ringtone.getTitle(mContext);
-                } else {
-                    String[] segments = path.split("/");
-                    if (segments.length > 1) {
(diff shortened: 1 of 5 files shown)


jamorham:
Tested on Android 13 and working well.

tzachi-dar:
I have checked this on two phones and it seems to work well.
Code also looks OK, although I don't understand all the details here.

PR #2302, opened 2022-08-30
Bugfix - correcting highlighting glitch in Home dropdown menu

Lupum1001:
Fixing a visual glitch in the Home dropdown menu. The highlighting wasn't working correctly.
image

### Corrected behavior
image


tolot27 (approved):
LGTM, simple and straightforward.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -400,7 +400,6 @@ protected void onCreate(Bundle savedInstanceState) {
             binding.setHome(home);
             homeShelf.set("arrow_configurator", false);
             dialog.setContentView(binding.getRoot());
-            dialog.getWindow().setBackgroundDrawable(new ColorDrawable(Color.TRANSPARENT));
```
jamorham, line 403: What is the reason for this change?

Lupum1001: The line must've been getting overridden by the 'Toolbar' theme that I had to remove. 
When you long-press on the top of the Home screen you get that gray pop-up with the toggle switches on it. Disabling the theme resulted in a transparent background for that pop-up.


I tried to use 'git blame' to figure out which was newer, the theme or the transparency, but I don't think either's been touched since 2017.
Part of me doesn't entirely hate the no-background appearance, but the opacity of that black layer would need to be bumped up quite a bit if we went that route; the bleed-through is just too distracting otherwise.

jamorham: Can you describe what the screenshot above is showing, is this with this PR or something else? Do you mean that the toolbar theme affects the pop-up dialog because it inherits from the view context?

jamorham: It inherits the style because it uses the incoming view context via the listener `new Dialog(v.getContext());` I think it should be using the parent context as its displayed in the parent. The change to the toolbar style shouldn't affect the shelf dialog. I will change this.

jamorham: Please merge upstream master as I have changed the context for the shelf dialog

Lupum1001: I merged in the changes. It looks like you also removed the transparency line, so this PR just removes the theme from the toolbar now.

PR #2305, opened 2022-09-03
Remove debug smoothing from engineering mode

Navid200:
Please let's remove this setting from engineering mode, or please tell me why it can be confusing and should remain there.

Thanks


jamorham:
I think this is a useful feature so it doesn't need to be hidden by engineering mode overall but the issue is that its incompatible with using calibration plugins because it uses the same mechanism as that dual point display, so the preference text at least needs to mention that it shows smoothed data instead of plugin data.

Navid200:
> I think this is a useful feature so it doesn't need to be hidden by engineering mode overall but the issue is that its incompatible with using calibration plugins because it uses the same mechanism as that dual point display, so the preference text at least needs to mention that it shows smoothed data instead of plugin data.

I completely missed that.  Thanks for explaining.  Will update.

Navid200:
| Before | After |  
|--------|-------|    
| Screenshot_20220903-215857 | Screenshot_20220903-213243 |

PR #2361, opened 2022-09-30
Dexcom clarifications

Navid200:
There is no change to any functions.  The only changes are to what the user sees in order to avoid possible confusions for newcomers.  

The following shows the new items (hightlighted) and why.  
  
| Item | Why |  
| ----- | ----- |  
| Screenshot_20220930-162449  | A-  There are still posts asking if xDrip supports Dexcom 1 |  
| Screenshot_20220930-162543 |  A <br/>  B- When asked why you have disabled this, some say because it says G5; I'm using G6.  |  
| Screenshot_20220930-162615 | B |  
| Screenshot_20220930-162642 | A |  
| Screenshot_20220930-162704 |  |  
| Screenshot_20220930-162721 | |

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/SourceWizard.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/SourceWizard.java
@@ -41,7 +41,7 @@ public class SourceWizard {
     private Tree<Item> root = new Tree<>(new Item(gs(R.string.choose_data_source), gs(R.string.which_system_do_you_use)));
 
     {
-        Tree<Item> g5g6 = root.addChild(new Item("G4, G5 & G6", gs(R.string.which_type_of_device), R.drawable.g5_icon));
+        Tree<Item> g5g6 = root.addChild(new Item("G4, G5, G6 & Dex1", gs(R.string.which_type_of_device), R.drawable.g5_icon));
         {
             Tree<Item> g4 = g5g6.addChild(new Item("G4", gs(R.string.what_type_of_g4_bridge_device_do_you_use), R.drawable.g4_icon));
             {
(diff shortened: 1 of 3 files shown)


jbr7rr:
Why did you choose to use dex 1 and not Dexcom ONE as it is mostly refferenced to online and in the official branding by dexcom? dex 1 takes some thinking to link it to the Dexcom ONE imo.

Navid200:
> Why did you choose to use dex 1 

On a device with a very small screen, I don't want it to go to a second line.
But, I never tested it.
Also, there is no need to make it extremely clear!

tolot27:
> In other words, what would Dex1 be if not Dexcom One?

At first I thought you mean G4. I find Dex1 also slightly confusing. What about just "ONE"?

Navid200:
> Also, Dexcom or Dex is redundant because it applies to G5/G6 as well but is not mentioned there.

Dex being redundant is something that is true about the name of the product Dexcom One.  So, no, One will not do in my opinion.

My objective here is to be clear to those who use Dexcom One.  
If someone who uses G5 or G6 find this confusing, I don't see that as a problem as long as they can find G5 or G6 because that's what they need to find.

If someone who uses Dexcom One tells me that this is no better than status quo, I think we will need to change.

Anyway, I'd like to hear @jamorham about this.
Even if  I change this to what you suggest, it will be pointless if he doesn't like it.

Navid200:
@tolot27 I just understood what you meant I think:

Dexcom G5 &#8722;> G5
Dexcom G6 &#8722;> G6
Dexcom One &#8722;> One


I'm OK making any change.

@jamorham Any preferences on your part?

tolot27:
> @tolot27 I just understood what you meant I think:
> Dexcom One −> One

Exactly what I mean. 😄 Maybe you can use the capitalized variant ONE. That is maybe more recognizable because it matches the marketing letters.

jamorham (approved):

Navid200:
Thanks

PR #2380, opened 2022-10-13
First commit to use LTE.

tzachi-dar:
The long term goal is to allow xdrip to show the BG (for followers) even when no phone is near by. This should be done by following a nightscout web site.

It seems that the only library that allows a wear os watch to do http queries is cronet (see https://developer.android.com/guide/topics/connectivity/cronet). Using cronet requires that one will use 'com.google.android.gms:play-services-cronet:18.0.1' which in turn requires play-services-wearable to be updated to version 18.0.0 This forces the use of AndroidX.
And here we are with this change.
So, right now, it only goes to example.com and gets a page from there, but we have many  changes caused by this.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -14,5 +14,3 @@ org.gradle.configureondemand=true
 org.gradle.jvmargs=-Xmx4g -XX:MaxPermSize=2048m -XX:+UseParallelGC -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
 #
 
-android.useAndroidX=true
-android.enableJetifier=true
--- a/wear/gradle.properties
+++ b/wear/gradle.properties
@@ -1,0 +1,2 @@
+android.useAndroidX=true
+android.enableJetifier=true
--- a/wear/src/main/res/layout/activity_body_permission.xml
+++ b/wear/src/main/res/layout/activity_body_permission.xml
@@ -45,7 +45,7 @@
                 app:circle_color="#0086D4"
                 app:circle_radius="20dp" />
 
-            <android.support.v4.widget.Space
+            <androidx.legacy.widget.Space
                 android:layout_width="8dp"
                 android:layout_height="8dp" />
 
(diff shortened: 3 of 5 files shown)


tzachi-dar:
Obviously,  wear/src/main/java/com/eveningoutpost/dexdrip/BaseWatchFace.java is not ready for merge, but I would still like to get feedback if the path that I'm taking is reasonable. For example, does one see any other way but using cronet and AndroidX?

jamorham:
Doesn't wear-os just sync over wifi/lte and so it will pick up from the phone wherever it is?

tzachi-dar:
I'm afraid this does not work like this. When I have turned off BT on the phone, the sync stopped even when they have been on the same wifi netwrok.

jamorham:
It should work, I've used it like this before over LTE some years ago. See https://support.google.com/wearos/answer/6056862

tzachi-dar:
What do you mean by "it should work"? Do you mean that smartphone and watch should just synchronize?
From my experience on the galaxy watch 4, this does not work, and the text you have sent also says that this should work over wifi (and not LTE), although this does not work even when both are connected to the same wifi network.

jamorham:
I mean I've used a collector on wear with a LTE connection on Huawei Watch 2 and received the data on a distant handset via the cellular network. The operation was identical as if they had a bluetooth connection. I've observed the same on wifi. The wear os link to the companion app on the phone and then xDrip should be identical whether its connected via bluetooth/wifi/or mobile data. That was my experience using it but this was some time ago but its not an xDrip feature its a wear-os feature.

tzachi-dar:
After some more playing with this it indeed works transparently over LTE :-)
So, I can either drop this PR at all, or leave only the changes that move us to use AndroidX.
What do you recomend?

jwoglom:
I suspect upgrading to AndroidX would unblock adding a lot of other features for new wear os 3+ watches, e.g. tiles and custom (non-xDrip) watchface complications in the xDrip-provided watchfaces. I've looked into adding both and it seems they're not supported in the current wearable support libraries.

jamorham:
I think migrating wear to android x could be useful. From the change-set it looks much simpler than for the main app.
A new PR would be good for that but it would also be good to know the answers to the following:
1) How does android-x change the file size of the wear component?
2) How compatible will it be, will people with old watches have problems? - particularly with google play service versions? (this was an issue previously) 
3) Are there any other compatibility issues relating to it? How much feature testing has /can be done etc?

PR #2385, opened 2022-10-17
xDrip sync follower source - fix wear complication

eksime:
This fixes the re-occurrence of #303 when the hardware data source is set to xDrip+ Sync Follower.

It may be nicer to introduce a proper "source_info" for data received from Gcm further down the line - that change could be done in addition to this one, as this change still accounts for the source_info being `null` when the bgrecord is from an xDrip follower source.

This also allows the complication to function when there is no current sensor information.

### DIFF
--- a/wear/src/main/java/com/eveningoutpost/dexdrip/ListenerService.java
+++ b/wear/src/main/java/com/eveningoutpost/dexdrip/ListenerService.java
@@ -2343,95 +2343,103 @@ public class ListenerService extends WearableListenerService implements GoogleAp
                 Sensor.InitDb(context);//ensure database has already been initialized
             }
             Sensor sensor = Sensor.currentSensor();
-            if (sensor != null) {
-                for (DataMap entry : entries) {
-                    if (entry != null) {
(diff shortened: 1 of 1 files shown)


allengh:
I am having the same issue with xdrip+ as the Hardware Data Source is "**xDrip+ Sync Follower**" with Wear OS 2 on a Ticwatch S2. The xdrip+ watchfaces work but not the xdrip Prefs complication (it shows NULL NULL). I have searched on the Facebook group and also xdrip docs but adding a calibration solution simply does not work for me. I made sure that the xdrip Prefs in my smartwatch has location just in case but not helping.

Peleton11:
Hi! Hope I'm in the right place to get the complication working on pixel watch 2..

Pixel 6, Android 14
Pixel watch 2
xdrip+ sync follower

I can get the xdrip+ watchface setup but the complication on any of the standard watchfaces is saying NULL NULL. 

Would be great to find a fix if possible, as would prefer following daughter's levels as a complication rather than full watchface.

Any help much appreciated!

PR #2406, opened 2022-10-30
add 720min/12h to list of snooze values

savek-cc:
Needed when child is sleeping, alarms are on follower

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
@@ -99,7 +99,7 @@ public class SnoozeActivity extends ActivityWithMenu {
 
     static final long infiniteSnoozeValueInMinutes = 5256000;//10 years
     //static final int snoozeValues[] = new int []{5, 10, 15, 20, 25, 30, 35, 40, 45, 50, 55, 60, 75, 90, 105, 120, 150, 180, 240, 300, 360, 420, 480, 540, 600};
-    static final int snoozeValues[] = new int []{ 10, 15, 20, 30, 40, 50, 60, 75, 90, 120, 150, 180, 240, 300, 360, 420, 480, 540, 600};
+    static final int snoozeValues[] = new int []{ 10, 15, 20, 30, 40, 50, 60, 75, 90, 120, 150, 180, 240, 300, 360, 420, 480, 540, 600, 720};
 
     static int getSnoozeLocation(int time) {
         for (int i=0; i < snoozeValues.length; i++) {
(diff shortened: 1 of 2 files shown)


Navid200:
You can snooze for 10 hours, and that's not enough?   You want to snooze for 12 hours?

I mean if I have just injected insulin and have eaten and my blood glucose is going up, I know that it is going to come back down shortly when the insulin catches up with the carbs.  So, I would snooze for 1 hour.
I cannot imagine why I would ever snooze for 10 hours.

Would you please help me understand when this is required?

savek-cc:
small child goes to bed at 19:00. Phone stays in rooms, parents monitor from own mobile including Alarms - all evening and all nicht. Alarm in childs room would only scare/wake the child but not alert the parents (thus the two followers with alarms).
By snoozing 10h, that gets you to 05:00 - I still don't want that child to wake up scared by a loud low glucose alert at 5am.
12h snooze gets us from 19:00 to 07:00 - that again does work as 7am is "time to get up".
Currently, we're using the snooze 'till re-enable - and promptly forgot to do so on multiple ocasions. Leasing to "no alarms in kindergarden".
Hope that explains the need.

jamorham:
This sounds reasonable to me. On the same topic is there any smart automation that could be done such as an option to snooze alarms when charging if operating as a master or something like that so you don't have to remember to enable it?

savek-cc:
Snooze when charging as an option would also work. Also thought about re-enabling based on kindergarden wifi or something (think AAPS automations).

Navid200:
> This sounds reasonable to me. On the same topic is there any smart automation that could be done such as an option to snooze alarms when charging if operating as a master or something like that so you don't have to remember to enable it?

Ideally, there would be intelligence overseeing this such that if there was no connectivity between master and follower, the snooze on master would disable.  
Or, let's say the intelligence would ensure that at least the alert triggers on either the master or follower and avoid the erroneous unintended case where the alert would trigger on neither phone.

Allowing an alert to be snoozed for 10 hours, even though I now understand the concept and intent, still sounds like a very dangerous case with no safety if it is unintentionally used incorrectly.

savek-cc:
> Allowing an alert to be snoozed for 10 hours, even though I now understand the concept and intent, still sounds like a very dangerous case with no safety if it is unintentionally used incorrectly.

We already have "Snooze until you re-enable" - which is what you have to use if you actually want 12h today. So how is "snooze forever" any better? (As this actually does lead to forgetting to re-enable)

Navid200:
> So how is "snooze forever" any better?

It's not better.  I never suggested that

jamorham (approved):

PR #2412, opened 2022-11-01
fix ISO 8601 date string parsing if it contains milliseconds

tolot27:
This PR enhances #1879 and fixes the exception reported with comment https://github.com/NightscoutFoundation/xDrip/pull/1649/#discussion_r983252108. The exception is caused if the date string contains milliseconds but no trailing 'Z'.

With the second commit milliseconds are retained, if any.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/DateUtil.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/DateUtil.java
@@ -25,7 +25,7 @@ public class DateUtil {
 
     /**
      * Takes in an ISO date string of the following format:
-     * yyyy-MM-dd'T'HH:mm:ss.SSS(Z|+/-HH:mm)
+     * yyyy-MM-dd'T'HH:mm:ss.SSS(Z|+/-HH:mm) and parses it to a Java Date object.
      *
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Models/DateUtil.java
```diff
@@ -42,6 +42,21 @@ public static Date tolerantFromISODateString(String isoDateString) {
...
+     * Takes in an ISO date string of the following format:
+     * yyyy-MM-dd'T'HH:mm:ss.SSS(Z|+/-HH:mm) and parses it to a Java Date object.
+     *
+     * The milliseconds and time zone/offset values are optional.
+     *
+     * @param isoDateString the iso date string to parse
+     * @param omitMilliseconds whether milliseconds should be returned or ignored
+     * @return the date object
+     */
+    public static Date tolerantFromISODateString(String isoDateString, boolean omitMilliseconds) {
```
jamorham, line 55: The use of `omitMilliseconds` parameter isn't implemented here which makes this method signature confusing. It should fall back to the existing method when `false`

tolot27:
Will resolve the reviewer comment.

PR #2498, opened 2022-11-25
Add Support for MiaoMiao / Libre2 using OOP2 on wear app.

cr-ellis:
This makes it possible to use a wear OS watch as collector for MiaoMiao.  Tested using a Libre2 (EU) sensor.

In order for it to work an updated version of OOP2 needs to be installed on the watch. (As discussed with @tzachi-dar).

### DIFF
--- a/wear/src/main/AndroidManifest.xml
+++ b/wear/src/main/AndroidManifest.xml
@@ -226,6 +226,7 @@
         <receiver android:name=".NSEmulatorReceiver">
             <intent-filter>
                 <action android:name="com.eveningoutpost.dexdrip.NS_EMULATOR" />
+                <action android:name="com.eveningoutpost.dexdrip.OOP2_DECODE_FARM_RESULT" />
             </intent-filter>
         </receiver>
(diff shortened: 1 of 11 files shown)


tzachi-dar:
Thanks cr-ellis for this pr.
I'm a little busy this week, but hope to review this before the week end.

Tzachi

tzachi-dar:
Still reviewing this. Quite a big PR.

tzachi-dar:
Finaly got to review it more seriesly. The code that is written seems to work.
But there are more changes that I think should be taken to this PR in order to bring it to in par with the existing code in the app.
This includes taking changes from the files nfcreaderx.java and librealarmreciever.jaca.
This will probably needs taking also readingdata.class and LibreTrendUtil and LibreTrendPoint.
All this are needed in order to better filter packets with errors.

I'll try to take a look if this can be achieved in a reasnoble time next.

cr-ellis:
Appreciate the update.  Let me know if you need anything else from me.

tzachi-dar:
@cr-ellis can you please check #2654 and see that there is no regression there?

PR #2515, opened 2022-11-30
Highlight transmitter days for modified TX also

Navid200:
This has been tested by shifting the scale.
The three conditions for Firefly have been verified.  The three conditions for mod firefly have been verified.

I don't have G5 or old G6.  So, I cannot test those.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
@@ -2167,15 +2167,22 @@ public class Ob1G5CollectionService extends G5BaseService {
                 if (!battery_status.equals("OK"))
                     l.add(new StatusItem("Transmitter Status", battery_status, BAD));
             }
-            Highlight TX_dys_highlight; // Transmitter Days highlight
+            Highlight TX_dys_highlight = NORMAL; // Transmitter Days highlight
             final int TX_dys = DexTimeKeeper.getTransmitterAgeInDays(tx_id); // Transmitter days
-            if (vr != null && (FirmwareCapability.isTransmitterRawCapable(getTransmitterID())) || TX_dys < 69) {
-                // Transmitter days < 69 or G5 or old G6 or modified Firefly
-                TX_dys_highlight = NORMAL;
-            } else if (TX_dys < 100) { // Unmodified Firefly with 68 < Transmitter days < 100
-                TX_dys_highlight = NOTICE;
-            } else { // Unmodified Firefly with transmitter days > 99
-                TX_dys_highlight = BAD;
+            if (vr != null && FirmwareCapability.isTransmitterModified(getTransmitterID())) { // Modified Firefly
+                if (TX_dys < 149) { // Transmitter days < 149
+                } else if (TX_dys < 180) { // 148 < Transmitter days < 180
+                    TX_dys_highlight = NOTICE;
+                } else { // Transmitter days > 179
+                    TX_dys_highlight = BAD;
+                }
+            } else if (vr != null && FirmwareCapability.isTransmitterRawIncapable(getTransmitterID())) { // Unmodified Firefly
+                if (TX_dys < 69) { // Transmitter days < 69
+                } else if (TX_dys < 100) { // 68 < Transmitter days < 100
+                    TX_dys_highlight = NOTICE;
+                } else { // Transmitter days > 99
+                    TX_dys_highlight = BAD;
+                }
             }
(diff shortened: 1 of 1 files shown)


Navid200:
The transmitter days font color is normal by default.
When the number of days left for starting sensors becomes less than 30, the font color changes to brown.
When sensors cannot be started any longer, the font becomes red.

This is the existing scenario right now for regular transmitters.
The only change this PR makes is to add the same behavior for modified transmitters.

app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
```diff
@@ -2167,15 +2167,22 @@ public void run() {
...
+            Highlight TX_dys_highlight = NORMAL; // Transmitter Days highlight
             final int TX_dys = DexTimeKeeper.getTransmitterAgeInDays(tx_id); // Transmitter days
-            if (vr != null && (FirmwareCapability.isTransmitterRawCapable(getTransmitterID())) || TX_dys < 69) {
-                // Transmitter days < 69 or G5 or old G6 or modified Firefly
-                TX_dys_highlight = NORMAL;
-            } else if (TX_dys < 100) { // Unmodified Firefly with 68 < Transmitter days < 100
-                TX_dys_highlight = NOTICE;
-            } else { // Unmodified Firefly with transmitter days > 99
-                TX_dys_highlight = BAD;
+            if (vr != null && FirmwareCapability.isTransmitterModified(getTransmitterID())) { // Modified Firefly
```
tolot27, line 2172: IMHO a more readable version (because of fewer conditions):
```java
            if (vr != null) {
                if (FirmwareCapability.isTransmitterModified(getTransmitterID())) { // Modified Firefly
                    if (TX_dys > 179) { // Transmitter days 180+
                        TX_dys_highlight = BAD;
                    } else if (TX_dys > 148) { // Transmitter days > 148
                        TX_dys_highlight = NOTICE;
                    }
                } else if (FirmwareCapability.isTransmitterRawIncapable(getTransmitterID())) { // Unmodified Firefly
                    if (TX_dys > 99) { // Transmitter days 100+
                        TX_dys_highlight = BAD;
                    } else if (TX_dys > 68) { // Transmitter days > 68
                        TX_dys_highlight = NOTICE;
                    }
                }
```

tolot27 (changes requested):
Im unsure whether the warning thresholds correspond to the PR description. The unmodified transmitter can be started till day 99 (inclusive). If you write that the notice highlight will start if less than 30 days are left to start a new sensor, then it should start at day 71 (`> 70`) and not day 69 or day 151 (`> 150`) for the modified transmitter.

Otherwise, the description can be changed into "... will start if 30 days or fewer are left ...".

image

Navid200:
There are individuals who use a mod transmitter and start a sensor and after 2 hours, post and ask why it didn't start.  After checking, we see that they have too many days shown for transmitter days.  This may be avoided if the user sees a red item on the status page as those who use regular transmitters see.

This PR fixes that and makes mod transmitters behave the same way with respect to informing the user that there are too many days on the transmitter and a sensor cannot be started.

jamorham (approved):
Looks like it should work

jamorham:
@Navid200 do you want to incorporate the changes suggested by @tolot27 ?

Navid200:
> @Navid200 do you want to incorporate the changes suggested by @tolot27 ?

I may have misunderstood the suggestion.  What I understood was a discrepancy of 1 day between the PR and this description here.  I don't think that is something we need to worry about.

Navid200:
I edited the description to remove the discrepancy:
Screenshot 2023-05-02 133453

Navid200:
> > > @Navid200 do you want to incorporate the changes suggested by @tolot27 ?
> > 
> > 
> > I may have misunderstood the suggestion. What I understood was a discrepancy of 1 day between the PR and this description here. I don't think that is something we need to worry about.
> 
> I'm sure, jamorham meant the suggestions i made in [#2515 (comment)](https://github.com/NightscoutFoundation/xDrip/pull/2515#discussion_r1103776846)

That's definitely an improvement.  Thanks.  I will change the PR to take advantage.
Since you had also changed the numbers, that got my attention, and I overlooked the other change you had suggested.

I will change the conditional arrangement as you have suggested.  Thanks for that.

With reference to the numbers, the unmodified transmitter highlighting is already in the code.
I will review again.  My intent is to leave the unmodified behavior as is, and only add a change to mod devices behavior.  I will also change the numbers as yo have suggested if it keeps the behavior of the unmodified transmitters as is.

Thanks for the original suggestion and thanks for pointing out what JamOrHam's suggestion was.

Navid200:
@jamorham I have made the requested change.

jamorham (approved):

PR #2540, opened 2022-12-08
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ master ]
 
+permissions:
+  contents: read #  to fetch code (actions/checkout)
+
 jobs:
   build:
 


jamorham:
Can you explain why this PR is not spam and the overall rationale that this change is designed to mitigate? thanks

sashashura:
Because `push` events happen often and even in a short time-frame of unlikely compromise of a dependency it may gain write access to the repository: commits, artifacts, releases, etc.
Because running CI with the following write-all permissions when read-only is enough is simply wrong?
image

jamorham:
But why make a change like this via `unit_test.yaml` when that can be modified by a PR itself. Why not set the repository `Workflow permissions` to read only instead?

sashashura:
> But why make a change like this via unit_test.yaml when that can be modified by a PR itself.

The threat model is different. It is true that a workflow can be modified in the pull request itself. For example someone could remove unit tests to make it look nothing is broken. However all workflows triggered on `pull_request` from forks have read-only permissions. It is impossible to "hack" permissions by modifying the workflow in the pull-request itself.\

The threat in this case is a potential asynchronous compromise of a third party tool/dependency that is used by the unit tests. When a legit pull-request is accepted and the workflow is started on `push` event there is an opportunity for the tool/dependency to make malicious modification to the repository.

>Why not set the repository Workflow permissions to read only instead?

Yes, it is possible to change the defaults on Org or repository level. It is up to you to decide which way to choose. Some people prefer explicit way, others settings way. In any case as you can see from the screenshot in this repository neither way was chosen. Hopefully my "spam" pull request will result in the workflow running with read-only permissions whatever option you decide to take.

jamorham:
Thanks for bringing this issue to my attention. I'm not sure why read isn't the default for github token permissions.

In terms of your work, I think if you're going to raise a very large number of PRs with random repositories using automation then you should be very clear in your PR description what your purpose is in order to be transparent and not flagged as spam.

PR #2559, opened 2022-12-15
Try to fix the issue of login to google backup.

tzachi-dar:
This worked on my phone, still needs approval from more people.


tzachi-dar:
I have tried this with another phone (xiaomi). Before the fix the error was that I can not login.
After applying the fix it tells me that I need to give xDrip permission to use drive. But I can not do it.
The issue might be related to the fact that the account belongs to a boy who is less than 13 years old.
There seems to be some work arounds for the issue, but since he will be 13 in a few days, will repeat the test then.
I have also tried adding my account to the phone, but still got the same error.

To be continued...

Navid200:
If you use an official apk (not a self-compiled one), there should be no problem logging in.

tzachi-dar:
Thanks navid.
I'll verify this.

PR #2583, opened 2022-12-28
Add documentation link

kleenkanteen:
I think it would be helpful to people, found a good site I've been reading recently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
  <img align="right" src="Documentation/images/download-xdrip-plus-qr-code.png">
  Info page and APK download: https://jamorham.github.io/#xdrip-plus
 
+ Documentation: https://xdrip.readthedocs.io/en/latest/
+
 <img align="right" src="https://travis-ci.org/jamorham/xDrip-plus.svg?branch=master"><a align="right" title="Crowdin" target="_blank" href="https://crowdin.com/project/xdrip"><img align="right" src="https://badges.crowdin.net/xdrip/localized.svg"></a>
 
 ## Features


kleenkanteen:
@jamorham any thoughts? current github readme doesn't link to docs.

psonnera:
The issue is there is no single documentation link. There are 2 main sites. And the community isn't very involved in helping documenting.

Navid200:
In my opinion having a single documentation site is not as important as having reliable documentation sites even if there are multiple reliable documentation sites..

No matter what we do, there will be documentation created by people who are not part of this team.
Many of those become obsolete and the developer stops updating them.  yet, users keep seeing them.

An example symptom of that is almost everyone using Dexcom G5 has engineering mode enabled.  They are under the impression that engineering mode is required for hard resetting their transmitters.
That requirement was removed from xDrip some time in 2018.   But, the damage is done.  There are documentation sites that still show that requirement.  And there are users who still refer to those obsolete documents.

I agree that we need to have links from this repository to our documentation sites and it has been my plan to add them.
This will provide guidance to users as to which external documentation sites are approved by this repository.  Hence, hopefully, moving users away from obsolete documentation and giving credibility to the ones that are valid.

Another problem, in my opinion, is that we still have documentation in a folder inside this repository.
In my opinion this repository should contain only links to documentation.  There should be none here inside the repository.
The individual approving PRs should not have to also spend time approving documentation PRs.  He has enough responsibility already.  But, to remove the existing documentation from this repository, we need to first stop adding more, move the knowledge to documentation sites, and replace links from the internal docs to externals docs.

It is my plan to work on all of this. 

Those are all my personal views.

Issue #1786, opened 2021-07-20
Smart alert doesn't work


Qannes:
Despite having smart snoozing and smart alerting enabled, I get alerts when my son is going in the right direction, but is near an alert threshold.

I have added screenshots of the graph, the alert settings, eventlog and alert levels, in hope that it makes it easier to troubleshoot
![Screenshot_20210720-185130_xDrip+](https://user-images.githubusercontent.com/87724445/126390286-cdba73cb-acdd-4ea1-8c56-90320a18d205.jpg)
![Screenshot_20210720-185149_xDrip+](https://user-images.githubusercontent.com/87724445/126390290-5c97f4c6-33ba-433f-b881-408e2ee701ee.jpg)
![Screenshot_20210720-185333_xDrip+](https://user-images.githubusercontent.com/87724445/126390301-409fb9a1-fa55-461a-ad9c-93b20429d0fe.jpg)
![Uploading Screenshot_20210720-185812_xDrip+.jpg…]()

StephenBrown2:
Been having this issue for a long time as well, my wife's phone is a follower to mine and also alerts even when I'm headed in the right direction.

Navid200:
I'm sorry about the delay in responding.

Looking at the logs, I see four HIGH level alert triggers.  They are at 16:01, 16:40, 17:41 and 18:51.  The corresponding readings (again, from the logs) are 8.3, 11.6, 12.0 and 8.6 respectively.

The two at 16:01 and 16:40 are expected because at those times, the trend is up.

But, the other two, at 17:41 and 18:51, occur when the glucose is decreasing.  So, one would expect the alert should not trigger because of smart alerting being active.

I have cropped the main screen showing the readings:
![126390286-cdba73cb-acdd-4ea1-8c56-90320a18d205](https://user-images.githubusercontent.com/51497406/144663085-6f80c6d5-619c-4e94-8419-42b40f0e5295.jpg)
You can see two transient readings that cause xDrip, only a computer program, to see an upward trend.  I have highlighted them with white circles.
Those are the reason for the alerts.

We can look at the readings now after the fact and ask why it triggered then.  But, at the time, xDrip has no way of knowing what the next reading is going to be.

ace-fjeldsven:
No worries about the delay, all time spend on this, is volunteer time, and I am just glad that you are doing this in the first place.

[...]

PR #2612, opened 2023-01-08
Reduce smart alerting and smart snoozing thresholds for the High alerts

Navid200:
Partly fixes: https://github.com/NightscoutFoundation/xDrip/issues/1786

I am awaken in the middle of the night for a high alert after having woken up previously and having taken a corrective insulin.
Now, the glucose is actually dropping and while crossing the high alert threshold, the alert has gone off again.

This happens because of the extremely high thresholds for smart snoozing and smart alerting as Tzachi has explained here: https://github.com/NightscoutFoundation/xDrip/issues/1786#issuecomment-988308678  

This needs to be tested functionally and that's what I will be doing now.  But, I have opened this so that others can test it as well.
So: **Not tested yet**

I have backed off the thresholds for the high alerts a lot more than for the low alerts.  The reason is that the high alerts can be complemented by the persistent high alert.  If I am high and have snoozed the alert; but, the glucose is not dropping very fast, the persistent high alert can trigger letting me know that I have been high for too long.  So, I am not worried if in that situation, the high alert does not repeat.

But, there is no alert that would complement the low alerts.  So, even though I have also reduced the thresholds for the low alerts.  But, I have not reduced those as much.


Navid200:
Let me be the devil's advocate:
Could this change cause a problem?

If you don't have a persistent high alert, and you get a high alert and snooze it, you could remain high for a long time with your readings decreasing at a very low rate.
Right now, without this change, the rate has to be 4 times higher for the snooze to persist.

The solution will be to have distinct thresholds for smart alert versus smart snooze.
We will keep the smart snooze threshold as is.  We only reduce the smart alert threshold.
I can do that if needed.

tzachi-dar:
A few comments:

1) When I wrote this code, I was a father to a small boy, and I realy cared about every hypo. So the values there are quite hysterical. I wanted to change that for a long time, but as you can see the change is not easy.
I think that the change should be protected by a preference key because people might be depending on this.

2) The more complicated problem is what the change should be. Well, I guess that when going up, the suggested values seems reasonable, but we might also add a test for 15 minutes (3 points) with a value of 10 and do an or with that. (yes I mean or and not an and because the condition as they are right now are very sensitive to noise). 

3) The bigger question is what could be considered a reasonable pace of going down. I guess that this depends on how high the value is. So for example, I would look at:
if bg > 300 we need at least 10 every 15 minutes.
if bg > 200 we need at least 10 every 15 minutes.
else we need 7 every 15 minutes.

Now comes the question if people will be able to understand something like that ( I guess some will)...

Navid200:
@tzachi-dar I have changed the PR.  I am only changing the thresholds for the high alerts.  I am leaving the low alerts alone for now.
Please let's try this for the high alerts first.  If everything works out with no problems, we can then think about a more sophisticated approach for the low alerts.

jamorham:
This looks reasonable to me but I think that some documentation should be added here because this portion of code is problematic.  It needs a TODO added as I don't think this code behaves the same way for one minute readings. I can do this.

Navid200:
Done

jamorham (approved):

PR #2621, opened 2023-01-10
Add the US 14 days sensors with a SN that ends with E6.

tzachi-dar:
Still needs testing.

xdrip APK can be downloaded from: https://drive.google.com/file/d/1qlw9lSgDPYoF1dW0dbmfA3xjgpfCXp_w/view?usp=sharing

oop2 APK can be downloaded from:  https://drive.google.com/file/d/1vu3x4rvggb2k2rVlTc0kb1t4Yk_ULin5/view?usp=sharing

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/Bubble.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/Bubble.java
@@ -44,7 +44,7 @@ public class Bubble {
         if (!Pref.getBooleanDefaultFalse("external_blukon_algorithm")) {
             if (patchInfo != null && patchInfo.length > 0) {
                 int first = 0xff & patchInfo[0];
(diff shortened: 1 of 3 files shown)


psonnera:
Having a tester now, when should direct intent be used?

kleenkanteen:
What does the title mean? Which US sensors are you talking about, libre 2? Also does xDrip+ currently use OOP2?

gui-dos:
@kleenkanteen:

The title is not correct, indeed. It is the first byte of the sensor `patchInfo` which is `0xE6` for the newer US models instead of `0xE5`

psonnera:
@kleenkanteen L1 14days US require OOP2 for xDrip+. L2 US is not supported (only L2 EU). For the title, as @gui-dos says.
Tested functional. More testing welcome.

jamorham (approved):

PR #2622, opened 2023-01-10
This change changes the behavior when one is scanning a sensor that is not the current sensor.

tzachi-dar:
The old behavior was to close the old sensor in the case that a new sensor was scanned. The new behavior is this:
If the scan was a manual scan, we ignore the scan and return (this is the reason that the check has moved to the first possible position). 
If this is a reading from a device like miaomiao the old behavior is preserved (close the sensor). The reason for the difference is that if one is using a device, than he has already sticked the device to the sensor, and there is no going back.
But manual scan can be done on a new/old sensor in a mistake, and next the correct sensor will be read.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Models/SensorSanity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Models/SensorSanity.java
@@ -104,17 +104,20 @@ public class SensorSanity {
         PersistentStore.setString(PREF_LIBRE_SENSOR_UUID, "");
         PersistentStore.setString(PREF_LIBRE_SN, "");
     }
-    
     public static boolean checkLibreSensorChangeIfEnabled(final String sn) {
+        return checkLibreSensorChangeIfEnabled( sn, true);
+    }
+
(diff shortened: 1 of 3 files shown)


jamorham:
Can you better explain the problem that this is designed to solve? thanks

tzachi-dar:
The scenario is that my sons starts to wear the new sensor about 36 hours before he stops using the old sensor.
If he needs to manually scan, than (before the change) he had to make sure not to scan the wrong sensor. If he failed to do it the sensor would be closed and he would need to restart it and add new calibrations again.

With the proposed change, a scan of the wrong sensor will not get BG results but will not close the sensor.
One has to explicitly close the sensor (or use a device like miaomiao, in which a move is also explicit).

tzachi-dar:
I'm going on vacation, so will respond next week.

tzachi-dar:
Seems like I forgot all about this...

In any case, yes this is indeed a behavioral change (I was not aware of the possiblity to automaticly start a sensor).
So, I think that adding a preferance key is probably the way to go.
What do you think? Should this be enough or should I add anything else?

PR #2630, opened 2023-01-16
Falling rate, rising rate and units

Navid200:
This is what we have now:
Screenshot_20230115-205435

The PR changes "rising" to "rising rate", and "falling" to "falling rate".
And it appends /min to the units.  

These are what we will have after this PR:
Screenshot_20230115-204714
  
Screenshot_20230117-170942

On a small screen, the units wrap.  The following image shows an example tested with a virtual device.
Screenshot 2023-01-17 161122

### DIFF
--- a/app/src/main/res/values/arrays.xml
+++ b/app/src/main/res/values/arrays.xml
@@ -172,9 +172,9 @@
     </string-array>
 
     <string-array name="risingEntries">
-        <item>1 mg/dl/m (0.06 mmol/l/m)</item>
-        <item>2 mg/dl/m (0.11 mmol/l/m)</item>
-        <item>3 mg/dl/m (0.17 mmol/l/m)</item>
+        <item>1 mg/dl/min (0.06 mmol/l/min)</item>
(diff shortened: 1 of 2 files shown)


tolot27:
What does '/m' mean? Per minute? If yes, you should use 'min' because that's the official abbreviation. 'm' is the abbreviation of meter/metre.

Navid200:
@tolot27 
Yes, min is the abbreviation for minute.  But, this is a unit, not an abbreviation.
The unit of time in the SI system is s for second.
However, non-SI units accepted for use with the international system includes min for minute as you have suggested.

I didn't want to use m.  The only reason I have used it instead of min is that it may become too large for a small device.  
So, I am saying that I also am not happy with what I have!  Just wanted to explain why I have it.
I appreciate your comment.  It is 100% valid.

I am open to suggestions.

Navid200:
Never mind.  The summary will wrap when needed.  I will change to min.  Thanks

Navid200:
@tolot27 I have made the change.  Thanks for the suggestion.

jamorham (approved):

PR #2652, opened 2023-01-29
Limit high blood glucose values shown on the graph to 400 mg/dL

Navid200:
This is what a user has got with show Libre trend enabled.
326923318_582882383746798_5979460552736510366_n

This makes the variations on the curve look nonexistent obviously due to the significant dynamic range.  

We already limit readings to 400 here: https://github.com/NightscoutFoundation/xDrip/blob/b58ca21907247af952a8a933b0e93c2eb55ef3bf/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java#L1265-L1266

But, not everywhere.
This PR does the same to other sources.
It would make xDrip's behavior consistent.

I am currently running this.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/BgGraphBuilder.java
@@ -1246,35 +1246,19 @@ public class BgGraphBuilder {
                 }
 
                 if ((show_filtered) && (bgReading.filtered_calculated_value > 0) && (bgReading.filtered_calculated_value != bgReading.calculated_value)) {
-                    if (bgReading.filtered_calculated_value < 400) {
-                        filteredValues.add(new HPointValue((double) ((bgReading.timestamp - timeshift) / FUZZER), (float) unitized(bgReading.filtered_calculated_value)));
-                    } else {
-                        filteredValues.add(new HPointValue((double) ((bgReading.timestamp - timeshift) / FUZZER), (float) unitized(400)));
-                    }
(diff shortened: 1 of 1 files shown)


jamorham:
Instead of doing it conditionally like this use `Math.min(x,y)` to cap to a high value. Needs review from @tzachi-dar when ready as well.

Navid200:
@tzachi-dar I have made the change that jamorham suggested.  
Please review when you get a chance.
I am currently using this as my main xDrip.

jamorham:
You can do the math.min before the unitized I think to avoid doing additional conversions where one will be discarded.

tzachi-dar:
Generally looks good.
You can do here two more improvements:
1) Make a function that does Math.min((float) unitized(xxx), (float) unitized(400) and use it in all places.
2) Replace the 400 with a constant (for example MAX_SHOWN_BG). it is already 4 times in the code.

Navid200:
@tzachi-dar @jamorham I now do Math.min first to pick one of the two candidates.  Then, I run unitized only once on the result.  This was caught by jamorham and thanks for catching it.  

With respect to your suggestion, I can add a constant if necessary.  Is it?

But, are you OK with the overall concept?  This affects Libre raw.  Are you OK with us limiting what is shown on screen to 400 in that case?

tzachi-dar:
yes, we should not have values on the screen higher than 400. In most times this is just noise, and the 'damage' remains for  24 hours.

jamorham:
You can use `BgReading.BG_READING_MAXIMUM_VALUE` instead of 400 directly. That way if it changes no code changes will be needed.

Navid200:
Replaced 400 with the existing max value constant.  Compiled and is running on my phone.

jamorham (approved):
Looks like it should work

PR #2671, opened 2023-02-13
Fake data source tweak options

Navid200:
Screenshot_20230213-094435


jamorham:
Can you describe what this PR does? thanks

Navid200:
This PR lets a developer who needs to use the fake data source to adjust the frequency or amplitude or both.

The frequency can be tripled.
The amplitude can be tripled.
Running the original command (enable fake data source) resets both of those tweaks.

This way a developer can speed up his tests or access values that the default source does not reach without having to edit the code.  But, simply by issuing a voice command.

jamorham (approved):

Navid200:
Thanks

PR #2687, opened 2023-02-17
Rename Samsung Workarounds

Navid200:
We are now using this for phones other than Samsung.  For example, Xiaomi.

People disable the setting thinking that it should only be enabled if they have a Samsung phone.

The question comes up why buggy Samsung is shown on the status page, or in a log, when they are not using a Samsung.  
  

<br/>  

This PR will remove "Samsung" from the title and summary of the setting and from the logs to avoid the confusion.

This is compiled and I am running it right now.
Screenshot_20230217-195913


jamorham (approved):

Navid200:
Thanks

PR #2698, opened 2023-02-24
Cgm usage % in stats

krisz768:
For example, in Hungary, the insurance for cgm depends on the use of sensors. (90 days) It would be nice if it was easily available in the statistics page.


Navid200:
It is quite possible that I have misunderstood exactly what this is.  So, you can show the formula to clarify here in the description.

But, please note that some of us use snapshots of the statistics pages to create our reports for our doctors.
So, adding anything that is not necessary could clutter the page.

krisz768:
I know that there is capture rate in the prev data graph but that's only up to 14 day. I think that this is a key value in the statistics that is missing. There is no information about the use of sensors in the statistics, which matter for the data. (Aside from the absolute #bs value, but it is difficult to interpret)

(I agree with the page cluttering thing, this is why i think xdrip needs a new stat page with desciption, range ect. but this is different problem)

Navid200:
If I understand you correctly, you say that capture rate already exists and it is available to used for the extra status line; however, there is no capture rate there for more than 14 days.

So, why don't you make that your PR?
To add the option of adding capture rate for a longer period of time to the status line?

The statistics pages are used for creating reports for doctors.  I don't think I am the only who likes to do that.
My doctor doesn't need to see the capture rate of my CGM.

krisz768:
Because i don't think that's a good idea. With 14 day selected the previous data is already hardly readable. If i add 90 day option, it will be useless because it will be  unreadable. Basically the 90 day option only the cap will be useful/usable. This is why i decided to place it in the stats. But i can try it.

Navid200:
That sounds like a layout problem.
I hope we will not modify a page just because we cannot figure out how to add it to another page.

xDrip should have a way of producing reports for doctors.  Currently, the only way to do that is to use the statistics page.  Please don't modify them.

The extra status line, I suppose, is meant to be used for things that you need to see continuously.
Are you saying that you want to see the 14-day option on the extra status line and that leaves no room for also showing a 90-day version?
Do you see a reason anyone would need to see both continuously on screen?

benceszasz:
The CGM usage statistics is an important information and actually it is part of ATTD Standardized CGM metrics:
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6467165/
and the AGP report:
http://www.agpreport.org/agp/agpreports
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8991298/
There is recommended minimal data collection: 70%-80%

benceszasz:
@Navid200 
Using statistics and graphs for decisionmaking or any kind of analysis should only happen if you have enough data, this makes the CGM usage % an important factor. Some examples from the links:
_For reliable analysis of glycaemic indices, the recommended minimum system activity is 70% of the time_
_An estimated HbA1c (eA1C) can be calculated if adequate rtCGM/iCGM data (70% or 10 days of the 14 days of CGM data) are available._
_A minimum of 14 consecutive days of data with approximately 70% of possible CGM readings over those 14 days appears to generate a report that enables optimal analysis and decision-making; standard reporting and visualization of CGM data are important._

Navid200:
@benceszasz 

We are in agreement.  Capture rate is very important.  I never disputed that.

But, so are many other parameters and we cannot place them all on the statistics page.

That page is already cluttered.

benceszasz:
@Navid200 I was just trying to answer your question about why and how important CGM usage % is. And I find that any stat only makes sense if the usage rate is high enough. But of course, this does not mean that it is mandatory to put it on the stats page, no matter how important the information is :)

jamorham:
I am prepared to add this to statistics page but please provide before and after screenshots for UI change and detail testing which has been done. thanks.

Navid200:
@jamorham 
Would you consider making it programmable?
I will make the distribution page programmable so that the user can decide what it should and what it shouldn't include.
Then, if someone  wants the distribution page to show the capture rate, thewy can.  if someone wants the distribution page to show very high and very low, they can.
And if someone wants a very clean page to contains only a small number of parameters, they can have that.

krisz768:
What if we(i) make a stat page that user editable?

The current page is working but it's not great. 

We need (i think):
- Editable elements (visible/invisible, sequence)
- something that visually displays data for example a bar with red yellow green segments for each stat (the user can set the bar visibility and ranges for colors)
- description for each one. 
- etc

Navid200:
@krisz768 Thanks for being open.

The challenge would be to do this without making things too complicated.
Many users find xDrip too complicated to use.
We often tell users about how they can do something and their response is "I had no idea that possibility existed".

With respect to something that would visually show the results, you know that we already have a pie chart, right?

In general, I think we agree.  But, we nee JamOrHam's approval.  It will be awesome to have your help.

Navid200:
@krisz768 I have opened a discussion where I have suggested adding the options to add or remove items from the statistics page:
https://github.com/NightscoutFoundation/xDrip/discussions/2921

Let's see what @jamorham thinks about that first.

If the answer is positive, we can look at the existing settings page that currently exists for the extra status line as a starting point.
But, the page would only be accessible from the statistics page.  So, we will not be adding a new page to any existing settings page.  We will only add a fourth item to the existing statistics 3-dot menu.

Navid200:
@krisz768 I have been trying to add data binding to this for a few days so that we can control visibility of lines in the layout.   But, it is hard and it may take me a while.

If you like, why don't you fix take a screenshot as is and one screenshot after your PR and post as requested by our lead developer.

If this is merged and some day I figure out how to control visibility, I can add it to your capture rate as well.

Navid200:
The stats page is now editable.  The user can choose which elements to be present and which ones to exclude from the distribution page.  After we added this ability, it was my hope to add capture rate to the page as this PR intends to do.

We now have a new problem though.  xDrip now offers Companion app mode.  There are companion apps that provide readings at once every 3 minutes.  There are also companion apps that provide readings once a minute.
This means that we don't have a universal expected rate.  The expected rate can be any of the following:
- 24 X 12 for a sensor that provides one reading every 5 minutes
- 24 X 20 for a sensor that provides one reading every 
- 24 X 60 for a sensor that provides one reading every minute

But, it is not even that simple.  The Companion app mode does not capture all the readings.  
As an example, please see this:
589077170_25033165283008583_1675470553850072930_n
This is posted by an xDrip user on facebook.  They claim this is Linx in Companion app mode.  This is Linx: https://linx-sensor.com/pages/faq
The following is from FAQ:
Q: How often does the LINX CGM update glucose readings?
A: The LINX CGM provides real-time glucose readings every minute.

So, it proves one reading every minute.  But, the companion app mode does not get one reading every minute.
When I zoom into the screenshot of the xDrip main screen, I see 26 readings in the hour.  And I can see that the readings are not uniformly distributed.
This means that we don't have a clear expected number of readings in xDrip in general.
So, how can we calculate the percent capture rate?  

I opened a discussion so that we can go into detail about this and hopefully decide how we should deal with it: https://github.com/NightscoutFoundation/xDrip/discussions/4256

PR #2711, opened 2023-03-01
Silent sensor expiry alert

Navid200:
I finally got to test this and it was not silent.
I don't think you wanted this to make noise.

I have made it silent.

I am now running this and I will confirm that it is silent when it gets to the 9-hour and 2-hour alerts.


Navid200:
I see that I have overlooked a clever condition you have put in that only triggers when the screen is on.  
OK.  I close this.

PR #2726, opened 2023-03-11
Fix typos

fwolter:


fwolter:
I just saw it when understanding the code around insulin profiles, so I fixed it.

PR #2736, opened 2023-03-16
Remove old log access

Navid200:
This PR removes the option to access the old logs.

There are individuals who still post these logs.  We have to tell them to post the new logs.  Can we please get rid of the old ones so that everyone would use the same set of logs?  
This PR does that.


jamorham (commented):
You also need to remove from `menu_eventlog_activity.xml`

Navid200:
Thanks so much for catching this.

Navid200:
Will be ready in a few minutes.

Navid200:
It's now also removed from the current logs menu 
Screenshot_20230322-150920
as shown below.

Navid200:
It is compiled and am running it now.

jamorham (approved):

PR #2739, opened 2023-03-17
Wake workaround for all brands

Navid200:
Tested on Android 11 Motorola.

Does not trigger on this phone.  But, there may be other phones and other brands that would benefit.

If the artifact is experienced by anyone after this PR, they can disable the "Wake workarounds" setting.


jamorham (approved):
I'm nervous about this as some phones may get occasionally jitter but I think we can give it a try

PR #2743, opened 2023-03-20
Linear ascending volume

Navid200:
This PR modifies the behavior of the ascending volume profile only when the "Delay Ascending Volume" setting is disabled.  By default, this setting is enabled.

The following shows the profile using extra logs enabled after this PR.
Screenshot_20230320-111637

Currently, before this PR, the volume is increased this way:
0.666
0.333
0.3
0.3
0.3

That means, the volume, currently, does not reach 100% (High) in a single 5-minute cycle.  It only reaches 70% (Medium) in a single 5-minute cycle.  
Only if you don't snooze the alert and it keeps going, on the next reading, will it reach 100%.

Also, for the first 3 minutes, currently, the volume is set to 0.3.  On some phones, any volume setting below 0.3333 (like 0.3) is silent.

This change will make the volume increase linearly from start and reach the high level (100%) in one 5-minute cycle.

I am currently using this and it is perfect.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/UtilityModels/AlertPlayer.java
@@ -127,6 +127,10 @@ public class AlertPlayer {
     final static int MAX_VIBRATING_MINUTES = 2;
     final static int MAX_ASCENDING_MINUTES = 5;
 
+    // Ascending without delay profile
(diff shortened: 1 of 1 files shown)


Navid200:
Screenshot 2023-03-20 114856

Please note that the medium volume is 0.7 and high volume is 1.

Navid200:
The following graph shows how the volume changes every minute after an alert triggers, including if you leave the "Delay Ascending Volume" at the default disabled setting, if you have chosen the ascending volume profile..  
Screenshot 2023-04-03 114606

jamorham (approved):
You make a convincing argument

Navid200:
Thanks so so much

Issue #1778, opened 2021-07-14
G6 Timestamp Issue when newly activated.


joshbohmer:
When first activating a G6 transmitter xDrip incorrectly reports the timestamps as negative for the first few hours. The transmitters internal times starts from 0ms.

This prevents xDrip from generating any packets that need a timestamp (like backfill requests) from being generated for the first few hours after a new transmitter is activated 

Once xDrips timestamp goes positive <0 the errors go away.
<img width="500" src="https://user-images.githubusercontent.com/3621222/125596909-ad129196-1141-4964-81f2-6d0054fba118.jpg" />

tolot27:
While investigating the issue I discovered that the screenshot above shows slightly different values of "tx start:" (`transmitter_start_timestamp`). That is really strange because the transmitter start time is stored in the database.

jamorham:
I think this is probably quite simple. Its just that a transmitter timestamp before transmitter start is being requested.

I expect this will come from backfilling and simply capping the backfill window to the transmitter start (or session start) would resolve the issue. I will look at that.

I don't think its related to the other mentioned issues.

tolot27:
@jamorham And why are different time strings reported for "tx start"? The max difference is 12 seconds. That cannot be a rounding issue.

emp-00:
I have never seen such an error message. No hits in my logs for "TimeKeeper" and "comparison". Running Dexcom G6 with xDrip nightly 2021-05-21. Also never saw such messages (before or after sensor start) with other xDrip versions. Totally new to me.

joshbohmer:
@jamorham think you're correct about the backfill observation

The timestamp reaches 0 after 3 hours, so it makes sense.

@emp-00 You'll only ever see this for the first 3 hours after waking a transmitter wakeup & if xDrip requests backfill data (or anything else that calculates the timestamp before 0)

Navid200:
The limitation that backfill does not work until the start time is 3 hours in the past gets really bad with G7.  Instead of 3 hours, the limitation is 24 hours for G7.
![Screenshot_20240306-101114](https://github.com/NightscoutFoundation/xDrip/assets/51497406/7dae86a2-6153-49af-b384-40a14a6481c2)

I will try to condition the backfill action so that it will not attempt to backfill prior to the start time.

joshbohmer:
Interesting, I thought that was done already but clearly not.

Simplest method I would think will be to do MAX(0,start_timestamp)

[...]

PR #2746, opened 2023-03-24
Don't start G6 if transmitter age is not known

Navid200:
If the transmitter age is not know, currently, a sensor can still be started.
A start will not be allowed if there has been no contact.  But, with a new transmitter, even after a first contact, the transmitter age may still not be known

This PR will not allow the sensor to be started as long as the transmitter age is not a valid integer.
It also adds a log if a start is not allowed in addition to a toast message.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
@@ -54,6 +54,7 @@ public class StartNewSensor extends ActivityWithMenu {
     // private TimePicker tp;
     final Activity activity = this;
     Calendar ucalendar = Calendar.getInstance();
+    private int transmitterAgeInDays = DexTimeKeeper.getTransmitterAgeInDays(getTransmitterID());
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
@@ -101,8 +102,9 @@ public class StartNewSensor extends ActivityWithMenu {
 
         ucalendar = Calendar.getInstance();
         if (Ob1G5CollectionService.usingNativeMode()) {
-            if (!DexSyncKeeper.isReady(Pref.getString("dex_txid", "NULL"))) {
-                JoH.static_toast_long("Need to connect to transmitter once before we can start sensor");
+            if (!DexSyncKeeper.isReady(Pref.getString("dex_txid", "NULL")) || !(transmitterAgeInDays > -1)) {
+                JoH.static_toast_long("Need to connect to transmitter before we can start sensor");
+                UserError.Log.e(TAG, "Need to connect to transmitter before we can start sensor");
                 MegaStatus.startStatus(MegaStatus.G5_STATUS);
             } else {
                 startSensorOrAskForG6Code();   // If we're using native mode, don't bother asking about insertion time
@@ -179,7 +181,6 @@ public class StartNewSensor extends ActivityWithMenu {
         if (Ob1G5CollectionService.usingCollector() && Ob1G5StateMachine.usingG6()) {
             if (JoH.pratelimit("dex-stop-start", cap)) {
                 JoH.clearRatelimit("dex-stop-start");
-                val transmitterAgeInDays = DexTimeKeeper.getTransmitterAgeInDays(getTransmitterID());
                 val modified = FirmwareCapability.isTransmitterModified(getTransmitterID());
                 val endOfLife = transmitterAgeInDays >= ABSOLUTE_MAX_AGE_DAYS || (!modified && transmitterAgeInDays >= MAX_AGE_DAYS);
                 if (transmitterAgeInDays < MAX_AGE_DAYS - MONTH_WARNING_DAYS


Navid200:
So, let's say there is no proper connectivity and the user is allowed to start.  What's wrong with that you may ask?

Without this, a user will start sensor and will wait 2 hours (looking at the 2-hour counter, which in this case provides a false hope) and only then, may they realize that something is wrong and post in a forum and ask for help.

With this PR, they will not be able to start sensor.  So, hopefully, they will post and ask for help only after 20 minutes instead of 2 hours.

Navid200:
The following images show sample cases from the facebook support group.
I believe they at least occasionally occur when the start command enters the queue when the very first contact has been made with a brand new transmitter.

I have been able to avoid this myself by always waiting for full connectivity and the battery voltages and transmitter days showing on the system status page before submitting the start command.

By not allowing start to be submitted until the transmitter days has a valid value, I hope that the number of these trouble cases experienced by the users will decrease.


337295547_1606615116478080_2202796895831770355_n
336902646_963026018020336_3066166694820202667_n

jamorham:
As far as I know it will already refuse to start a session if the transmitter time is not known. Instead it will take you to the status page for the collector.

Navid200:
@jamorham I don't know how to communicate.  I don't know if you see the Gitter room messages.  We may have a crysis right now.
Would you please see these discussions and advise:  
https://github.com/NightscoutFoundation/xDrip/discussions/2750
https://github.com/NightscoutFoundation/xDrip/discussions/2747

Navid200:
@jamorham You can see the above two images.  None of them shows the transmitter time.  And both have allowed the start command to enter the queue.

jamorham:
The first image is from a very old version and the second one looks like to have included transmitter reset which clears all stored transmitter information.

Navid200:
@jamorham Would you please also have a look at this?
https://github.com/NightscoutFoundation/xDrip/discussions/2747

Navid200:
@jamorham How about this?
Screenshot_20230328-003746

Navid200:
@jamorham I wanted to prove to myself one way or the other.

So, I compiled the latest xDrip after having added a single line to create a log showing transmitter days when I start a sensor.
Then, I hard reset a resettable transmitter.  
I then cleared xDrip cache and data and uninstalled xDrip so that there would be no record of the particular transmitter on the phone.  
Then, I installed my version of xDrip (just one added line of a log).
After the first connection was made, I attempted to start.  xDrip allowed it.  Looking at the logs, I saw the log showing -1 for transmitter days.

From that experiment, I concluded that the conditional linked below does not consider the possibility that the returned value may be null.
https://github.com/NightscoutFoundation/xDrip/blob/3254cca8ac8827851b21c72613ea2468085c74dd/app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java#L185

That conditional was created by me.  You rewrote my PR to clean it up.  But, the conditional is not complete and that is my fault.  
This PR completes the conditional.  If transmitter days returns a negative value, meaning transmitter days is null, it will not allow user to start.  So, the linked conditional will not even be executed because the start will not be permitted.

Navid200:
It seems to me that with a brand new transmitter (or hard reset transmitter), it will take a while (several read cycles, not just one) after the first contact has been made for the correct number of days to be reported by the transmitter.

Navid200:
This may or may not be related to this: https://github.com/NightscoutFoundation/xDrip/issues/1778

All I know is that I personally have been able to avoid the failed first start on a brand new transmitter, consistently, by delaying the start to when the battery voltages are shown on the system status page.
This PR delays the start by a few read cycles.  My hope is that the delay will be enough to resolve the issue experienced frequently by newcomers to xDrip.

app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
```diff
@@ -101,8 +102,9 @@ private void sensorButtonClick() {
 
         ucalendar = Calendar.getInstance();
         if (Ob1G5CollectionService.usingNativeMode()) {
-            if (!DexSyncKeeper.isReady(Pref.getString("dex_txid", "NULL"))) {
-                JoH.static_toast_long("Need to connect to transmitter once before we can start sensor");
+            if (!DexSyncKeeper.isReady(Pref.getString("dex_txid", "NULL")) || !(transmitterAgeInDays > -1)) {
```
jamorham, line 105: This will work but its really confusing. Why not just do it as `|| transmitterAgeInDays == -1` ?

jamorham:
I think what you're saying is that it is possible to start a transmitter when the age is reported as -1 which is the default if unknown and that should be corrected. See the comment I made just on the logic readability.

app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
Navid200: This is due to my inexperience.   Can I be certain that null will always show as -1?
I wasn't sure about that?  I mean I have no ideas why null will turn into -1 and not for example -2 in another case!

Navid200: Will correct now in a few minutes.

Navid200: Testing

Navid200: Done

app/src/main/java/com/eveningoutpost/dexdrip/StartNewSensor.java
```diff
@@ -54,6 +54,7 @@ public class StartNewSensor extends ActivityWithMenu {
     // private TimePicker tp;
     final Activity activity = this;
     Calendar ucalendar = Calendar.getInstance();
+    private int transmitterAgeInDays = DexTimeKeeper.getTransmitterAgeInDays(getTransmitterID());
```
jamorham, line 57: I suggest this should be a private method returning the value so that it can be used in both places it is needed as it can be updated dynamically in the background after the class has been instantiated but before the user clicks anything.

Navid200: Should the private method be inside Home.java?
Or, should it be in StartNewSensor.java?

jamorham: private method inside this class. The use of method is just to avoid duplication but also to allow dynamic results as it can change in the background. 

```
private int getTransmitterAgeInDays() {
   return DexTimeKeeper.getTransmitterAgeInDays(getTransmitterID());
```

Then replace variable use by call to that method

Navid200: Thanks - Done

jamorham (approved):

jamorham (approved):

Navid200:
Thanks

jamorham:
see the use of local variable here. https://github.com/NightscoutFoundation/xDrip/commit/891adba8f2cebe97134c22016dfe0ab270d0369c Please check everything still working.

The idea is to ensure code works best if data states change while the UI is open (or if activity life cycle means variable isn't updated with the way you had this before) - So in the two places this data is used, pull it in via a method but use the single method to avoid duplicating the functionality in each of those two places and enable better maintainability. Where the data wont change, eg during a single codeblock being executed then pull it in to a local variable which also reduces duplication of code, makes it cleaner to read as the reader knows the data is not expected to change and also makes it more maintainable if we ever wanted to change how that variable is populated.

Navid200:
There is also an instance on line 107.
Is that OK as is?

Do you mean check that it works after you merge?
Or, do you mean I change my PR now and check?

Navid200:
I see what you mean.  
sensorButtonClick() is one of the two places.
And startSensorOrAskForG6Code() is the other.

Is it OK if I change my PR now after approval for testing?

Navid200:
Testing

jamorham:
Its already merged, please just confirm that my edits are working.

PR #2757, opened 2023-03-31
Don't ask user to start G6 before connectivity

Navid200:
After selecting G6 in source wizard, a red notification on the main screen asks user to start sensor or change settings.
Shortly after, a dialog appears asking user to start sensor.

This PR delays those until transmitter number of days resolves to a valid integer.
In the meantime, it shows the following instead.
Screenshot_20230331-004622

This PR complements this PR: https://github.com/NightscoutFoundation/xDrip/pull/2746
But, the two PRs are independent.  Meaning if you merge only one and hold off on the other for more info, there will be no issue.  It will still be a step forward,

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -159,6 +159,7 @@ import com.github.amlcurran.showcaseview.targets.ViewTarget;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.internal.bind.DateTypeAdapter;
+import static com.eveningoutpost.dexdrip.utils.DexCollectionType.DexcomG5;
 
 import java.io.ByteArrayOutputStream;
 import java.io.File;
@@ -2616,8 +2617,8 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
         }
 
         if (!isSensorActive) {
-            // Define a variable (notConnectedToG6Yet) that is only true if G6 has been selected and transmitter days is unknown.
-            boolean notConnectedToG6Yet = Pref.getBooleanDefaultFalse("using_g6") && DexTimeKeeper.getTransmitterAgeInDays(getTransmitterID()) == -1;
+            // Define a variable (notConnectedToG6Yet) that is only true if Native G6 is chosen, but, transmitter days is unknown.
+            boolean notConnectedToG6Yet = DexCollectionType.getDexCollectionType() == DexcomG5 && Pref.getBooleanDefaultFalse("ob1_g5_use_transmitter_alg") && Pref.getBooleanDefaultFalse("using_g6") && DexTimeKeeper.getTransmitterAgeInDays(getTransmitterID()) == -1;
             if (notConnectedToG6Yet) { // Only if G6 has been selected and transmitter days is unknown.
                 notificationText.setText(R.string.wait_to_connect);
             } else { // Only if G6 is not selected or G6 transmitter days is known.
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -4,6 +4,7 @@ import static android.Manifest.permission.WRITE_EXTERNAL_STORAGE;
 import static com.eveningoutpost.dexdrip.models.JoH.msSince;
 import static com.eveningoutpost.dexdrip.models.JoH.quietratelimit;
 import static com.eveningoutpost.dexdrip.models.JoH.tsl;
+import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.getTransmitterID;
 import static com.eveningoutpost.dexdrip.utilitymodels.ColorCache.X;
 import static com.eveningoutpost.dexdrip.utilitymodels.ColorCache.getCol;
 import static com.eveningoutpost.dexdrip.utilitymodels.Constants.DAY_IN_MS;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -2612,7 +2614,12 @@ private void updateCurrentBgInfoCommon(DexCollectionType collector, TextView not
         }
 
         if (!isSensorActive) {
-            notificationText.setText(R.string.now_start_your_sensor);
+            boolean G6_not_connected_yet = Pref.getBooleanDefaultFalse("using_g6") && DexTimeKeeper.getTransmitterAgeInDays(getTransmitterID()) == -1;
```
fwolter, line 2617: You could stick to the Java convention that variable names should be camelCase, i.e. start with a lower character.
```suggestion
            boolean g6NotConnectedYet = Pref.getBooleanDefaultFalse("using_g6") && DexTimeKeeper.getTransmitterAgeInDays(getTransmitterID()) == -1;
```

Navid200: My objective is to get this change added so that xDrip functionality improves for the inexperienced user.
I am more than happy to make any necessary change to get approval.

Therefore, if I am told that this is what's holding this PR up, I will change it.

Let me add that looking at this same file, I can see plenty of existing uppercase instances.  For example: https://github.com/NightscoutFoundation/xDrip/blob/35d548f08bee8289a3a18e9d825b3c8145135921/app/src/main/java/com/eveningoutpost/dexdrip/Home.java#L258

So, SHOWCASE_G5FIRMWARE is already there.  If we are going to change the code for some reason, we should be consistent.
If, we want to say, let's leave the existing cases alone.  But, let's follow this guideline for the new variables, I am happy to follow such a guideline if I can get an approval of it.  

So, I thank you for having a look and making the suggestion.  But, I am not going to make your requested change because I have no idea it is going to make any difference in what really matters in getting this approved or not.

fwolter: It just caught my eye when skimming through your PR. The constant in line 258 is fine by convention. https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html

Navid200: Thanks a lot for providing this guidance.
I will change to follow the standard.  I have no reason not to follow it.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -2628,7 +2635,7 @@ private void updateCurrentBgInfoCommon(DexCollectionType collector, TextView not
                     dialog = builder.create();
                     dialog.show();
                 } else {
-                    if (!Experience.gotData() && !QuickSettingsDialogs.isDialogShowing() && JoH.ratelimit("start-sensor_prompt", 20)) {
+                    if (!Experience.gotData() && !QuickSettingsDialogs.isDialogShowing() && JoH.ratelimit("start-sensor_prompt", 20) && !notConnectedToG6Yet) {
```
jamorham, line 2638: So your `!notConnectedToG6Yet` condition should come before the ratelimit as everything to the right of the ratelimit is not evaluated when the ratelimit has been exceeded. I believe in this case you would only want the ratelimit to restricting firing of the conditional block after all conditions have been met.

jamorham:
This doesn't seem to be conditional on using dex collection type so I think it would break for other collector types as the condition can never be met.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
Navid200: In other words, if the user has not selected G6 or if they have and it is already connected, the dialog will be shown as it did before this PR.
However, if G6 is selected and is not connected yet, this PR does not allow the dialog to be shown.

jamorham: Yes but if the user reaches the condition and the timekeeper condition is not met, but then the connection occurs in the background and would be true and the user then re-enters this conditional block again within the 20 second rate limit period, it will fail because the ratelimit has been triggered even though it otherwise would have passed if the check order is changed so the ratelimiting is last. The `&&` operator means the other clauses are not evaluated once one is false. Simply checking the ratelimit by calling its method triggers the time to be stored so that the condition will fail the next time you call it within the specified time period.

Navid200: I'm so sorry I don't follow you.
I keep comparing what xDrip does now to what I want it to do after this PR, to help me see that this PR is correct.

If ratelimit fails, xDrip does not show the dialog right now.  In that case, I don't want xDrip to show the dialog after this PR either.  So, I don't understand the dellima you are describing.

Navid200: Sorry!
I misunderstood the purpose of the bracket.  As soon as ratelimit executes, it is going to show the dialog.
Thanks for catching this.

Navid200: Done

Navid200: Tested

Navid200:
Now that the other PR is merged, should I edit this one to use the new method from it?
So that if some day it is modified, everything stays related and consistent?

Navid200:
Resolved conflict after refactoring to utilitymodels.

Tested.  Everything works fine.

Navid200:
Resolved conflicts after refactoring to models.

Tested.  Everything works fine.

Navid200:
Thanks for catching my error, and offering guidence.

jamorham (approved):
I think this should work and I believe you have tested it.

Navid200:
Yes, it is tested.  
I hope you don't mind.  But, we may need to tighten this.  But, we may not.  This is a first, and hopefully last, step.

PR #2773, opened 2023-04-07
Let the insulin types layout view grow dynamically

fwolter:
After adding a new insulin type the list overflows. This PR makes the list grow dynamically.

Before:
Screenshot_20230407-214228

After:
Screenshot_20230407-214615


jamorham:
I haven't tried this but shouldn't it be `wrap_content` instead of `match_parent` and is the scrollview active to allow this to work on smaller screens etc?

fwolter:
You're right, `wrap_content` makes more sense. I changed it. The content becomes scrollable on small screens. Tested it on an older physical device.

jamorham (approved):

Issue #1762, opened 2021-06-21
Give more flexibility to Other Alerts


Navid200:
**Current status**
There is only one override silent mode setting for all the alerts in this category.
There are only a few of these alerts that can have a specific sound file chosen.  The others inherit it from the other alerts.
There is no ability to enable or disable the vibrator for these alerts.
These alerts do not follow the chosen volume profile for all alerts.
The current override silent mode for the other alerts does not work for Android 10 or 11.

**Preferred Solution**
Please enhance the alerts categorized as "Other Alerts" to offer the missing features described above.

**Existing related open issues**
https://github.com/NightscoutFoundation/xDrip/issues/193
https://github.com/NightscoutFoundation/xDrip/issues/555

thedewi:
I would also like to control how often these repeat. "Forecast low" is the most useful alarm but it only repeats after 20 minutes, which is enough time for a "wait and see" to become "too late to prevent the hypo"

Navid200:
This isn't really an issue.
It is more of a discussion.

PR #2799, opened 2023-04-23
Clarify Silent mode vs Do not disturb

fwolter:
I had a deep dive why Other alerts/"Override Silent mode on these alerts" isn't working on my phone. Turns out the setting is only referring to the Do not disturb function.

This PR changes the text from "Override Silent mode on these alerts" to "Ignore Do not disturb" for the following reasons:

- The option has no effect when the phone's volume is 0 (I'd call that silent). Alarms won't come through if you set the tristate sound/vibrate/no sound button in the pull down menu to no sound. This is a different behavior to the glucose alerts. They do override the no sound setting. The renaming helps the user to differentiate between the behaviors and prevents confusion like here https://github.com/NightscoutFoundation/xDrip/issues/193#issuecomment-1098418384
- The suggested wording is compatible with the Android settings. See screenshot below.
- The current text doesn't fit onto the screen of a Galaxy S8+. See screenshot below.

It also adds a summary to the setting, explaining that it has no effect when using Notification Channels.

I tested the behavior with and without Notification Channels.

Old:
grafik

New:
grafik

Android wording:
grafik

### DIFF
--- a/app/src/main/res/values-de/strings-de.xml
+++ b/app/src/main/res/values-de/strings-de.xml
@@ -775,7 +775,8 @@
     <string name="rising_threshold">Steigungswarngrenze</string>
     <string name="set_sound_for_bg_alerts">Setze den Ton, der für Glukosealarme benutzt wird.</string>
     <string name="alert_sound">Alarmton</string>
-    <string name="override_silent_mode_these">Lautlos-Modus bei diesen Alarmen überschreiben</string>
+    <string name="override_silent_mode_these">"Nicht stören" ignorieren</string>
+    <string name="no_effect_when_using_notification_channels">Diese Einstellung hat keinen Effekt, wenn die Benachrichtigungskanal-Funktion aktiviert ist. Das Verhalten kann dann in den Android-Benachrichtigungseinstellungen geändert werden.</string>
     <string name="persistent_repeat_max">Maximal alle x Minuten wiederholen</string>
     <string name="momentum_indicates_low">Wenn der Trend anzeigt, dass eine Unterzuckerung bevorsteht</string>
     <string name="notify_on_low_battery_level">Benachrichtigen, wenn der Ladezustand niedrig wird.</string>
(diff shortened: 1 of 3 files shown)


Navid200:
Please note that I had a similar intent with this PR, which was turned down:
https://github.com/NightscoutFoundation/xDrip/pull/2079

Regardless, I would like to make a point about a statement you have made:
"Alarms won't come through if you set the tristate sound/vibrate/no sound button in the pull down menu to no sound".

That would be a serious shortcoming of an app that is meant to be useable as a medical app. even if not officially.
That's why we have volume profiles.
If you choose an appropriate volume profile, the alerts that respect volume profile will come through regardless of what you do to your volume.

Unfortunately, we have a few volumes that do not respect the volume profile, as summarized here: https://github.com/NightscoutFoundation/xDrip/issues/1762.
In my opinion, missed reading, forecast low and persistent high alerts should also respect the volume profiles.

Navid200:
Just to emphasize, this is not a matter of convenience.  it is in fact a matter of safely.  Please let me give an example.

A parent and child are sleep.  The parent's xDrip stops receiving readings.  Previously, the parent was listening to a message on the phone that happened to have a very high volume.  So, the parent lowered the volume on their phone and forgot to raise the volume back up.
Now, the parent's missed reading alert does not wake her up because the missed reading alert does not respect volume profiles.  So, it is not loud enough to wake up the parent.

The parent wakes up in the morning and finds out that she has had no readings following her child for the past 6 hours!
I do not continue the story.  you can imagine ...

fwolter:
Thanks for pointing me to the PR. I didn't search the closed PRs before filing this. I think the wording of the glucose alerts should be something like "Override silent mode _and_ DND" and the Other alert setting should only referring to DND, because it won't override a volume of 0.

@jamorham wrote "xDrip doesn't override DND mode" here https://github.com/NightscoutFoundation/xDrip/pull/2079#issuecomment-1108811141. The HEAD version I tested overrides DND in Other alerts, but not volume 0. The glucose alerts override volume 0 and DND.

PR #2832, opened 2023-05-15
Clarify notification

Navid200:
This is what the notification looks like before you expand it:
Screenshot_20230515-142329

It reads "Please enable or disable engineering_mode".
That can be misleading.

After the PR, it will read "Please disable Engineering Mode" as shown below.
Screenshot_20230520-212152  

The price to pay is that we now need two different notification methods, one for advising to disable and the other for advising to enable.


Navid200:
This is what the notification looks like (before being expanded) after this PR.
Screenshot_20230520-212152

jamorham (approved):

PR #2849, opened 2023-05-25
Add the option to smooth libre2 data based on 10 15 or 25 minutes.

tzachi-dar:
This PR was made by Keencave.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
@@ -140,7 +140,12 @@ public class LibreReceiver extends BroadcastReceiver {
                                 // period of 4.5 minutes to collect 5 readings
                                 if (!BgReading.last_within_millis(DexCollectionType.getCurrentDeduplicationPeriod())) {
                                     long smoothing_minutes = LibreReceiver.libreFilterLength(xdrip.getAppContext());
-                                    List<Libre2RawValue> smoothingValues = Libre2RawValue.weightedAverageInterval(smoothing_minutes);
+                                    long dataFetchInterval;
+                                    if ( smoothing_minutes == 25L )
+                                        dataFetchInterval = 20L;
+                                    else
+                                        dataFetchInterval = smoothing_minutes;
+                                    List<Libre2RawValue> smoothingValues = Libre2RawValue.weightedAverageInterval(dataFetchInterval);
                                     smoothingValues.add(currentRawValue);
                                     processValues(currentRawValue, smoothingValues, smoothing_minutes, context);
                                 }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
@@ -7,8 +7,10 @@ import static com.eveningoutpost.dexdrip.models.Libre2Sensor.Libre2Sensors;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.os.PowerManager;
+import android.preference.PreferenceManager;
 import android.text.format.DateFormat;
 
 import com.eveningoutpost.dexdrip.models.BgReading;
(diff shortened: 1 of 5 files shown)


jamorham:
Will this change have the effect of changing the smoothing window time to 25 minutes where before this change it would be 20 minutes?

Navid200:
Is the default set to what we currently have?
If not, would you please make it like that so that after merge, no one would experience an unexpected change in behavior?

For consistency, would you please always mark the default when you use an array?
You can see existing examples here:
https://github.com/NightscoutFoundation/xDrip/blob/2e4954ea741d7efdb575948794ff5586bb92e1d2/app/src/main/res/values/arrays.xml#L472
 and here:
https://github.com/NightscoutFoundation/xDrip/blob/2e4954ea741d7efdb575948794ff5586bb92e1d2/app/src/main/res/values/arrays.xml#L486

keencave:
@jamorham: Good point. In the code contain a principal descrepancy, A 25 min weighted average filter is applied to a 20 min time interval. The PR extends the 20 mins interval to 25 mins. This will add 5 additional values with 20%, 16%, 12%, 8% and 4% weight to the current results. To be save we should change the 25 to 20 minutes to avoid unexpected changes. 
@navid200: In fact no, see above.

keencave:
Two options: 
a) For 25 minutes use only 20 minutes data. Exact same results, bad code.
b) Change 25 minutes to 20 minutes, slightly changed results.
I would prefer a)

jamorham:
@keencave I don't fully understand what the PR is doing but could you add a 20 minute option in addition to the 25 minute and make that the default so there is no functional change for users on upgrade (until they select a different time period) - I may not fully understand this PR so make your best judgement but as a policy we like to have new code not cause unexpected behavior changes to existing users who may update and have no idea this feature has happened.

keencave:
@jamorham There was a coding error in the inital implementation. We work with a 20 min time interval of data but apply a moving average filter for 25 mins data length. The results are valid but not what was mathematical intended. Lets state the code is running since 4 years without problems.
To fix that I will replace the 25 min option with 20 min as the default. For that a 25 min average filter will be used to have 100% backward compatibilty. For 10/15 mins we use a 10/15 mins average filter length as intended.

keencave:
Remark: We talk about time intervals of data which are filtered. Delays are the result of the filter length. As longer these are the longer the delay is. It is hard to estimate/calculate them as this is system theory. I am not an expert in that. We should adress the topic in terms of filter length.

Navid200:
If this PR is merged, and an existing Libre user updates and leaves the setting at default, will they experience any difference with respect to delay or anything else?

keencave:
@navid200: For this version they will have a different behavior. Very limited difference but different. Let me change the code to exactly reproduce the current results as the default option.

Navid200:
Thanks a lot for that.
Would you have an objection to marking the default setting?

I started doing this a while ago and suggest we all do it.

Imagine a user changes a setting a few times.  Then, they decide to go back to default.  How would they know what the default was?  
I have used the following to mark the default.
>

I am happy to use something else if there is a preference.  But, I suggest that we all make it a habit to mark the default.

keencave:
All changes done. Added the default '>' indication. The "25" minutes wording is untouched. It was used the last years to explain the slow reaction time for community users.
@tzachi-dar will apply the changes to the PR soon.

tzachi-dar:
I have updated the PR with latest changes.

Navid200:
Thanks

psonnera:
I'll try this weekend.

app/src/main/java/com/eveningoutpost/dexdrip/LibreReceiver.java
```diff
@@ -137,9 +139,15 @@ public void run() {
                                 Log.v(TAG, "got bg reading: from sensor:" + currentRawValue.serial + " rawValue:" + currentRawValue.glucose + " at:" + currentRawValue.timestamp);
                                 // period of 4.5 minutes to collect 5 readings
                                 if (!BgReading.last_within_millis(DexCollectionType.getCurrentDeduplicationPeriod())) {
-                                    List<Libre2RawValue> smoothingValues = Libre2RawValue.last20Minutes();
+                                    long smoothing_minutes = LibreReceiver.libreFilterLength(xdrip.getAppContext());
```
jamorham, line 142: You should replace with `Pref.getStringToInt("libre_filter_length", 25)` or do the same in `libreFilterLength()` if you expect to need to reuse it. There is no need to manage shared preference and context yourself as this is already encapsulated inside the `Pref` helper class.

tzachi-dar:
Updated the PR. Keencave, can you retest?

keencave:
Will do it tomorrow.

keencave:
Retest done successfull. Works like intended.

jamorham (approved):

PR #2853, opened 2023-05-26
CareLink Follower  Updates: Multi follow, correct device type, TZ correction

benceszasz:
Multi follow
Implement new CareLink multiple follow APIs, integrate new APIs. New CareLink Follow parameter: CareLink Patient user, which must be provided only if care partner with multiple patients are used. For care partner with single patient the patient is automatically determined using the patient list API.

Correct device type
Device type may be wrong in CareLink after moving from standalone CGM to 7xxG pumps. This causes an error in data retrieval since standalone CGM and 7xxG pumps use different API for data retrieval. To address this issue the recent CareLink data upload infos for patient accounts are checked and the actual device type is extracted from data upload infos (device used for uploading data).

Timezone determination correction
In CareLink data TZ may be missing, which was previously extracted from Last SG's datetime. However Last SG is sometimes empty and TZ cannot be determined. If Last SG is not provided in CareLink data, then Last Alarm's datetime is used to get correct TZ.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/carelinkfollow/CareLinkDataProcessor.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/carelinkfollow/CareLinkDataProcessor.java
@@ -234,8 +234,8 @@ public class CareLinkDataProcessor {
             // Only Guardian Connect, NGP has all in notifications
             if (recentData.isGM() && recentData.lastAlarm != null) {
                 //Add notification from alarm
-                if (recentData.lastAlarm.datetime != null && recentData.lastAlarm.kind != null)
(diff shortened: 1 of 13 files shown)


Navid200:
Thanks you so much

Navid200:
@benceszasz Thanks so so much for doing this.
Is this ready?

Navid200:
I mean is this ready to be merged?

benceszasz:
@Navid200 yes, every commit is included, it is ready to be merged.
I think I have tested every possible combination: EU/US | patient/care partner | one patient/multiple patients | patient provided/no patient provided | good CareLink account / mixed up CareLink account (device type is wrong in CareLink after changing from standalone CGM to 780G pump)

jamorham (approved):
thanks for your work on this

PR #2856, opened 2023-05-26
Added inactive days button to the G6 status page

Navid200:
There are individuals who order extra G6 transmitters so that they have a replacement ready.  The result is that they are always using older transmitters than they have to.
At times, we see transmitters that still have many days left, yet, the battery dies.
But, to confirm that the transmitter is old, we have to ask them to enable engineering mode.

This PR will let us see what we need with no need to enable engineering mode.
After this PR, inactive days is shown below transmitter days as shown below.
Screenshot_20230526-163850


Navid200:
Please consider this scenario:

Someone posts on facebook asking for advice because their G6 has stopped working.
We go through everything and we suspect the problem is battery.
Considering the battery hardly ever dies before 110 days unless you have put the transmitter on shelf for a very long time, it help to confirm that before deciding what the next step should be.

If the battery is dead when the transmitter has been sitting on the shelf for a very long time, it is pointless to contact Dexcom.
If the battery is dead when the inactive days is only a month or two, it makes sense to contact Dexcom because this would be a very unlikely scenario of battery dying too early.

We always tell users that engineering mode is only for testing.
But, this isn't really testing xDrip.  This is verifying what the number of inactive days is.
Having it visible on the status page by default removes the need for us to ask the user to enable engineering mode.

Please reconsider.
I hate to say this.  But, if you insist, I am OK adding a button to show it only if the user taps on the transmitter days.

Navid200:
I have added a blank entry at the bottom of the status page to serve as a button.
Placing the button on the existing transmitter days line would locate the button adjacent to the existing button that updates the battery voltages.  As the lines are very close, it would be difficult for someone who does not have thin fingers to press the button they intend to press.

This is what the status page look like after this PR.
Screenshot_20230914-220116

Navid200:
I have tested this on an old Android 8 phone, on an Android 11 Motorola Stylus and on An Android 13 Pixel 6a.
They all show the inactive days for about 10 seconds.

Navid200:
The blank row now is really blank.
Screenshot_20230923-092638

Tapping it, will replace the empty row with inactive days.
Screenshot_20230923-092730

Is there any chance the timer could interfere with the collector wake?

Navid200:
The timer is about 50 seconds if you stay on the status page.
If you leave the status page, the counter stops.  When you return to the status page, inactive days will still be visible until the count finishes.

Navid200:
I just had an idea.  I am going to this differently for actually a slightly different purpose.

PR #2897, opened 2023-06-19
Report current IoB value in the /pebble endpoint.

bart2:
This PR implements IoB value reporting in the /pebble endpoint. It is using existing `ioBForGraph_new` method to get a series of IoB values and then finds one close to the current timestamp.

I've also removed `number` parameter from `ioBForGraph_new`, as it was unused.

With this change watchface developers will be able to display IoB values reported by xDrip+.

Emulator based testing:

When IoB data is not available, `iob` field reports "unknown":
Screenshot_20230618_212827_unknown_iob
When IoB data is available, current value is reported:
Screenshot_20230618_213038_iob_reported

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/Treatments.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/Treatments.java
@@ -978,7 +978,7 @@ public class Treatments extends Model {
     }
 
     // NEW NEW NEW
-    public static List<Iob> ioBForGraph_new(int number, long startTime) {
+    public static List<Iob> ioBForGraph_new(long startTime) {
 
        // Log.d(TAG, "Processing iobforgraph2: main  ");
         JoH.benchmark_method_start();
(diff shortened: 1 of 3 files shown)


jamorham:
Please describe the rationale for changes, changes made and testing which has been done. thanks

bart2:
@jamorham, sure!

Main rationale for this change is to address the existing TODO in the codebase, to actually report IoB value in the /pebble endpoint, instead of always returning 0.

Example use case for that change is adding IoB value in the xDrip watch face for Gamin smartwatches: https://apps.garmin.com/en-US/apps/8fab3746-f56f-4b41-b3c7-5f4aaeaed6c9.

This is a fairly small change, reusing existing IoB calculation method `ioBForGraph_new` in the new `getCurrentIoB` method which is then called when /pebble endpoint is accessed. To approximate the current IoB, `getCurrentIoB` searches the IoB frames produced by `ioBForGraph_new` for ones that near the current timestamp (+/- 5 mins) and returns the IoB value attached to that timestamp.

As for testing, I only tested in the Android simulator so far. I'd like to get feedback whether the current approach to this change is reasonable before I start testing it on an actual phone.

Hope it helps!

jamorham (approved):

PR #2928, opened 2023-07-05
Fetch IoB from Omnipod app

bart2:
This PR implements IoB fetching from Omnipod App running on the same device, through the notifications subsystem. The idea was initially discussed in https://github.com/NightscoutFoundation/xDrip/discussions/2728.

The IoB value is exposed through the /pebble API endpoint and can then be used by certain smartwatches.

Changes made:
- `UiBasedCollector` was already doing something similar for BG reading, so I extended that class.
  - Used `PersistentStore` to save the current IoB value as well as timestamp when the write to the store occurred.
  - On the read path, recorded write timestamp is compared with the current time and `null` is returned if the write happened more than 5 minutes ago.
- Added two settings:
  - `enable_iob_in_api_endpoint` controls the previous feature implemented in https://github.com/NightscoutFoundation/xDrip/pull/2897. I figured it might be useful to have a killswitch for that functionality in case it breaks existing smartwatch users.
  - `fetch_iob_from_omnipod_app`: this is the main setting flag for the new Omnipod IoB fetching feature.
    - If set to false, the graph based IoB calculation is used.
    - If set to true, Omnipod App IoB is used.
- Added simple unit test for the IoB text parsing method.

# Testing
I'm running a dev build of xDrip+ on my main phone that is also running the Omnipod App. Here we can see the current IoB reported in the notifications area by the Omnipod App:
Screenshot_20230704_112646_One UI Home
And the same IoB value reported in the /pebble endpoint:
Screenshot_20230704_112723_Chrome


jamorham:
This looks like a well crafted and presented PR for a useful feature. 
In typical engineering style I will ignore everything that is right about it and focus only on the areas of concern and where I think changes should be made:

- I think this should be made more generic so that the feature is for receiving IoB from another app. So the wording and storage names, preferences and variables should be refactored to make it generic. Eg, that the Omnipod app is just the first app supported for this feature.
- Instead of constructing your own facilities to have a persistent record which is timed out and invalidated, instead look at the `Persist` helper class and see how things like `Persist.StringTimeout` are used. This will do everything for you and I have just extended it to support the `Double` type with `Persist.DoubleTimeout`  - This will return the double you store until the timeout expires and then will return null. It cannot store null though but your code doesn't look like it needs that.
- Your parsing doesn't look like it will support mutli-language or different locale number formatting. You should update your initial match and regular expression handling to be able to handle multiple match strings (via list or array) beyond just `IOB:` and consider allowing the regular expression to use either `,` or `.` in a number so that you can support European decimal formatting. There is a helper method `tolerantParseDouble()` which you can reuse to assist with final parsing. More exotic languages may require significant extra effort, but these described changes will significantly improve compatibility and ability for the same code to be reused for different languages and different apps or app versions.

Navid200:
There are too many settings in xDrip already.
Would you please add new settings only if absolutely necessary?

You say you have added the first switch just in case it kills something.  Haven't you tested this?

Would you please explain why there is a need for the second switch?

bart2:
Hey, @Navid200 :)

I did test the new feature on my phone and it worked fine. What I got concerned about, are different smartwatch apps that might be out there, relying on the "iob" key to always return 0.0.

Testing the integration of that feature on multiple different smartwatches would be quite difficult for me (both due to the investment required as well as time commitment), so I figured a killswitch might be a valuable option in case users come complaining their smartwatch app stopped working.

Let me know what do you think. Thanks!

Navid200:
@bart2 Thanks for explaining

Just so that you know where I am coning from, I spend a lot of time helping people who use xDrip.
The majority of the problems are wrong settings.  It takes time to go over all the settings that could break a Dexcom G6 being used by xDrip.  And after going over all the settings, often, I find that they have a wrong setting.

There are too many settings in xDrip and many of them are added by a developer who decided to add a kill switch.  Now, that developer is long gone and every time I open a PR to remove such a setting, I am told that this will be confusing, or how can you be sure that this will not break an already working setup?
So, we end up with more and more kill switches that I can never remove!

All I can do is to remind developers to please "only add settings if there is absolutely no alternative way to accomplish the same objective".
Thanks for reading.

jamorham (approved):

gdc27health:
How as an end user of a WearOS device can I access the IoB.  Is it possible for this to be written to a ranged value complication?

gitanO0:
I'm now running an Omnipod 5 and have enabled the two settings for IOB... one for the web service and the other that gets the IOB from the Omnipod notification "companion app".  The web service is working properly and reporting the correct IOB but xdrip main UI does not show any IOB on the graph or as text at the top.  Is this functionality not implemented or am I missing a setting somewhere?

PR #2931, opened 2023-07-07
Remove (retire) the OB1 setting

Navid200:
I would like to remove this setting because we ask everyone to enable it anyway.
It has been tested for years.  Please let's retire the setting.

This is the only menu that has changed.  The very first setting on the page is gone.
Screenshot_20230707-110356

I am running this now as my active collector working fine.
Unfortunately, I cannot test the wear portion as I don't have a wear OS watch.


Navid200:
Picked up a G6 session in progress fine.
Started a G6 sensor fine.

jamorham:
I appreciate what you are trying to do here and I think it is about time that this option is hidden because turning OB1 off will almost certainly break things for users so they should never do it. 

This is a UI issue and the changeset you've proposed while technically "cleaner" has the potential to cause problems if there are any errors in it and we don't have an easy way to test all the edge cases that might occur. I can see what I think is at least one error just reading through it so I think we should take a less risky approach.

Basically I would plan to simply remove the setting from the UI and force it to always be true by just setting that setting to true every time. (Probably via the migrations class)

Navid200:
Thanks.  Will work on this now.  This PR may temporarily close.

PR #2966, opened 2023-07-26
No more accidental issues!

Navid200:
Before:
Screenshot 2023-07-25 201650  
  
**Edit: This is not valid anymore.  Please see the post after the next commit to see the latest.**
After:  
Screenshot 2023-07-25 201831

This PR removes the ability to open a new issue as you can see in the above images.
However, all the existing issues will still be there and visible to all.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,41 +1,7 @@
 ---
-name: "Please do not open an issue unless asked to do so by a developer in a discussion (see below)"
-about: Please include a link to the discussion in the first post.
+name: "Please do NOT open an issue!"
+about: 
 label: status-triage
 ---
-
-### Please don't open an issue if you haven't already consulted with the developers
-- Please go back and use Discussions (on the previous page) to present your case if you haven't already.
-- If you have consulted with the developers and were asked to open an issue, please proceed.
-- Please also search the existing issues for similar problems before opening a new one.
-- Please do not prefix you issue title with BUG: or LIBRE: for instance. The issue maintainer will tag every issue with appropriate labels.
-- Add screenshots only if necessary, e.g. write the version number instead of adding a screenshot of the system status page.
-
-### Subject of the issue
-Link to the discussion thread:  
-A clear and concise description of what the bug or request is.
-Example: The BG graph shows negative data points between 3 and 5 am, which is incorrect and can only be an artefact.
-
-### Your environment
-- What version of xDrip, e.g. 2020.11.27 or 2011270800 or "Nightly build 27th Nov 2020". The version can be found at the system status page next to "Code:".
-- What type of hardware data source and how do you get your data there (e.g. G6 or Libre with bluecon bridging device, Nightscout Follower, etc.)
-- Is your issue specific to a device like and Android phone or a smartwatch? If yes, please provide the exact device name and type as well as the OS version (e.g. Samsung Galaxy S7 running Android 10.1).
-- All settings related to the issue, e.g. activated auto calibration.
-
-### Expected behavior
-A clear and concise description of what you expected to happen.
-Example: BG values are always positive and should be displayed in line with the previous and following values, regardless of the time of day.
-
-### Actual behavior
-A clear and concise description of what actual happens.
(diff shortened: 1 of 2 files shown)


Navid200:
Users will still be able to create an issue from a discussion when instructed by a developer.

Navid200:
Developers will still be able to open issues.

paravoid:
As a drive-by contributor, and with respect to all of your work, I'd recommend against this. Not being able to open new issues is against the ethos and common practice of open source & community development, and feels a bit... alienating to me. Honestly, the already existing practice of resolving valid issues with "please open a discussion" has already felt a bit hostile to me and personally discouraged me from reporting bugs (or opening up "discussions").

It'd be helpful to understand the problem you are trying to solve. Is the consideration around the volume of bug reports, the backlog of it all, users being confused or something else? If the concern is about the volume of stale requests, have you considered something like e.g. https://docs.github.com/en/actions/managing-issues-and-pull-requests/closing-inactive-issues to just close inactive issues instead?

Navid200:
@paravoid 
The person who has to close the issues that should never have been opened has to spend time that can be spent elsewhere.
So, yes, the problem mainly is the volume.

Do you know that 
- on some Android 10 phones, if you rotate the phone to Landscape and then, return to portrait mode, the screen remains in Landscape!?  
- we have a database corruption issue?  
- how many redundant lines of code exists in xDrip because of WearOS?  
- the missed reading alert does not use volume profile, so you could just sleep through it?  
- there is no easy (safe) way to snooze an alert while driving if you don't have a companion watch?  
- the smart snooze and smart alert are not perfect;  so, we get alerts when we shouldn't?
- we also support a version of Nightsocut after Heroku ended their free service? 

I know all of those and more because I have gone through all the issues and categorized them.  
The number of open issues was approaching 1000.  There are currently 133 open issues.
Having 1000 open issues will not allow contributors to see what the real problems are that needs their attention.
I don't want them to be closed just because they have been inactive.  They are still problems that need to be resolved.
Can you help us with any of those?

We also have a lot of issues that have been addressed and resolved.  So, things are working.  This PR will streamline the process more.

We don't need a new issue stating "I started a sensor and after the 2-hour warm-up it is not working".  
That's not an issue.  That's really a question.  
We spend a lot of time addressing those anyway.   

A user experiencing problems, should start by opening a discussion.  I go through all discussions.  I either address them myself or ask someone with more expertise to address it.
If/when needed, some of these are turned into issues.
I am not saying there are no issues.  That's not the purpose of this PR.  I am saying GitHub is not set up to give 100% flexibility for the management of a free project.  
This PR let's us change the behavior forced by GitHub that does not serve our needs.

Your opinion is important.  It will be taken into consideration before this PR is rejected or approved. 
This team is an awesome team.  What is unique about the active members is that they do what needs to be done.  We need more members.  Do you want to help?  Can you dedicate ongoing time to this project?

xDrip has a lot of users worldwide.  But, it has only a few contributors.  We need to make life manageable for those few serving the large community.

Navid200:
Whether people open issues or discussions, I spend time on them.  I am not complaining about that.
The time I spend here, I spend because I want to, not because I am doing anyone any favors.

We decided to ask everyone to use discussions as a starting point.
This decision was made quite a while ago.  And it has not been revoked.

This PR addresses the issue that our mentioned decision is not executed perfectly by everyone.

I can add a few lines to thank people for coming to the repository to provide feedback.

Navid200:
This is what you will see, after this PR, clicking on new issue.
Screenshot 2023-08-08 124352

Another change is that currently, when you click on the button, you will be taken to discussions and you will need to click on new discussion to open one.
After this PR, clicking on the button is the same as clicking on new discussion.  So, you can right away decide the category and start typing.

jamorham:
@Navid200 can you explain in words also what this PR does and how it works? thanks

Navid200:
If someone clicks on "New issue", they will be taken to a page with a note and a button on the right.

This is the note:
`Thanks for taking the time to provide your input, which is very important. Please open a discussion (link on the right). A team member will respond, and if needed, create a corresponding issue. This may be unlike what you have experienced in other repositories. But, please have no doubt that we want to fix bugs and add more capabilities. Thanks for your help.`

The button takes the user to the page that they can also go to be going to discussions and clicking on "New discussion".




Developers can bypass this and open an issue directly.  I have already explained this in a different channel.

Navid200:
@jamorham Please merge this

Navid200:
There is no need for this PR. 
I can convert an issue that is a question to a discussion myself.

PR #2998, opened 2023-08-12
Use decimal value separator depending on locale ("." or ",") on EditTextActivity

gusakovgiorgi:
When creating a new alarm and entering a decimal threshold value some European languages (Ukrainian, for example) languages use `,` for fractions. However, the xDrip keypad does not allow you to use `,`. In this case, if you enter 4.3 it'll be parsed as 4,0. To avoid that we have to use a decimal line separator depending on the current locale.

Additionally, when `alertThreshold.setKeyListener(DigitsKeyListener.getInstance("0123456789" + decimalValueSeparator)); `is applied there is no need to set input type because DigitsKeyListener.getInstance modifies that anyway.

Part of the documentation for `setKeyListener`

> Note that this method has significant and subtle interactions with soft keyboards and other input method: see KeyListener.getInputType() for important details. Calling this method will replace the current content type of the text view with the content type returned by the key listener

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/EditAlertActivity.java
@@ -49,6 +49,7 @@ import com.eveningoutpost.dexdrip.watch.thinjam.BlueJayEntry;
 import com.eveningoutpost.dexdrip.wearintegration.WatchUpdaterService;
 
 import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
@@ -214,9 +215,8 @@ public class EditAlertActivity extends ActivityWithMenu {
         doMgdl = (prefs.getString("units", "mgdl").compareTo("mgdl") == 0);
 
         if(!doMgdl) {
-            alertThreshold.setInputType(InputType.TYPE_CLASS_NUMBER);
-            alertThreshold.setInputType(InputType.TYPE_NUMBER_FLAG_DECIMAL);
-            alertThreshold.setKeyListener(DigitsKeyListener.getInstance(false,true));
+            char decimalValueSeparator = DecimalFormatSymbols.getInstance().getDecimalSeparator();
+            alertThreshold.setKeyListener(DigitsKeyListener.getInstance("0123456789" + decimalValueSeparator));
         }
 
         uuid = getExtra(savedInstanceState, "uuid", null);


Navid200:
Can I recreate the problem by choosing the Ukrainian language and then do what?

gusakovgiorgi:
> Can I recreate the problem by choosing the Ukrainian language and then doing what?

Then go to the alert list -> edit or add a new alarm.
In the threshold enter 4.5 (units should be set to mg/dL)
Press save.
Value will be saved as 4.0

gusakovgiorgi:
https://github.com/NightscoutFoundation/xDrip/assets/9167854/386973c0-f99c-4a1c-9551-646853ed1c56

Navid200:
That is expected behavior.
If you choose mg/dL, the blood glucose parameter will be an integer.  Your readings will be around 110.

If you want an alert at 4.4, you need to choose mmol/L.  Then, the parameter will be a fractional with one decimal point.

gusakovgiorgi:
> That is expected behavior. If you choose mg/dL, the blood glucose parameter will be an integer. Your readings will be around 110.
> 
> If you want an alert at 4.4, you need to choose mmol/L. Then, the parameter will be a fractional with one decimal point.

I've chosen mmol/L.

gusakovgiorgi:
> That is expected behavior. If you choose mg/dL, the blood glucose parameter will be an integer. Your readings will be around 110.
> 
> If you want an alert at 4.4, you need to choose mmol/L. Then, the parameter will be a fractional with one decimal point.

See the video above, please.

Navid200:
> See the video above, please.

I did.
Screenshot 2023-08-12 123954

gusakovgiorgi:
> > See the video above, please.
> 
> I did. Screenshot 2023-08-12 123954

If you watch further I'll see that I changed it to mmol/L and reproduced the issue.

Navid200:
I changed the language in my xDrip.
My unit is already mmol/L.
I went to look at the alerts and can see that the alert threshold uses , instead of . to identify the fraction.
I changed the fraction leaving the , in place and it worked fine.

I then replaced , with . and I can see that the fractional part is dropped.
But, why would someone who uses the Ukrainian language in which , is used to represent fraction use . for that?

I don't understand why anyone would have a problem with xDrip as is?

gusakovgiorgi:
> I changed the language in my xDrip. My unit is already mmol/L. I went to look at the alerts and can see that the alert threshold uses , instead of . to identify the fraction. I changed the fraction leaving the , in place and it worked fine.
> 
Yes, if you leave the comma then it works. If you delete, then you can enter only "." and the fractional part will be dropped.
Also, if you create a new alarm then there won't be any predefined value with the ",". So you have to enter it from scratch.
> I then replaced , with . and I can see that the fractional part is dropped. But, why would someone who uses the Ukrainian language in which , is used to represent fraction use . for that?
> 
You can't use "," in this case, because the current implementation is locale independent and allows entering only dots from your keyboard.
> I don't understand why anyone would have a problem with xDrip as is?

Navid200:
OK, I can see the problem now.
I wish I hadn't had to work on getting the information out.

I would present this differently if I were you.  The very first sentence in your description reads "When editing ...".  And that word (edit) in this case is very confusing.
I would start the sentence with "When creating a new alert ...".
And then, all the conversation we had so far would become unnecessary because your description would go right to the heart of the problem.
The Ukrainian language uses , for fractions.  However, the xDrip keypad does not allow you to use ",".
That' the problem that needs to be resolved.

gusakovgiorgi:
> 

You're right. The description might've been more elaborated. Sorry for taking your time.

Navid200:
That's OK.  That's my job.  You don't need my approval though.  You need someone else's approval.  I would like us all to do what we can to help save his time.

Navid200:
Thanks for reporting the problem and suggesting a fix.  
Let's wait for a review to see if this is the best solution or not.

jamorham:
This is a perfectly reasonable and useful change. The problem is that this particular activity is in a very old part of the code and in other places we have had to be careful with parsing numeric values containing separates other than `.` because they are not supported natively by java. There are also no unit test for this part of the code. 
I previously created a simplistic utility method of `tolerantParseDouble()` which we then use to wrap parsing of decimals in many places. I would have to check that this change doesn't affect parsing and that these code paths are safe. I haven't done that yet.

gusakovgiorgi:
> This is a perfectly reasonable and useful change. The problem is that this particular activity is in a very old part of the code and in other places we have had to be careful with parsing numeric values containing separates other than `.` because they are not supported natively by java. There are also no unit test for this part of the code.
> I previously created a simplistic utility method of `tolerantParseDouble()` which we then use to wrap parsing of decimals in many places. I would have to check that this change doesn't affect parsing and that these code paths are safe. I haven't done that yet.

Thank you for the insights!  The code that I used here is just for enabling entering decimal separator to the `EditText`. Currently, the value is still saved as `,` for some locales so this part didn't change.

Navid200:
@jamorham Is there anything I can do (tests) to help move this forward?

Navid200:
@jamorham Please tell me how to test what concerns you.

Or, please merge this and if we discover issues, we will redact and fix then.  After all, it is a Nightly.

Thanks

Teute0815:
As a german we've got the same problem. If adding "comma" as a numeric value is a problem, there might be another fix:
Main Menu / "Glucose Levels" require a seperator as well, when using mmol.
In those dialogs the dot as seperator works perfectly fine with german keybord. Maybe there is something different underlying, that could be easily applied to alarm levels.

Navid200:
If you attempt to enter a threshold for the persistent high alert, do you experience the same problem?

Navid200:
@gusakovgiorgi Thanks so much for opening this PR and answering all my questions.
I opened another PR using everything I learnt from you here and addresses the concerns of the lead developer: https://github.com/NightscoutFoundation/xDrip/pull/3819
It has now been merged.  

Thanks to you, it should now be fixed as of the 2025.03.12 release.
Thanks a lot

PR #3027, opened 2023-08-18
Remove the fallback setting

Navid200:
This PR retires a setting.
The setting is Fallback to xDrip.  The setting is permanently set to false.

I am running this now in a G6 in native mode as a test.
Unfortunately, I don't have a G5 transmitter to test.

This setting, as is, can be dangerous.  Someone who uses this setting enabled could get a sudden change in readings because their glucose changes suddenly, or because there is a fallback.  How would they know the cause of the sudden change?

We should either use native mode or not.  There should be no switching back and forth on the fly.  Therefore, there should be no need to this setting.


jamorham:
I agree with this concept but will have to check the change is safe.

jamorham (approved):

PR #3066, opened 2023-09-05
Development Guide

bart2:
I think a development guide might be useful for new people thinking about contributing to the project. Knowing about the older JDK requirement would save me quite some time when I first tried to set up Android Studio to be able to build xDrip.

Also, see a recent question related to this subject: https://github.com/NightscoutFoundation/xDrip/discussions/2926

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,6 +17,8 @@ Patches are welcomed! To ensure the best all round results they need be a good f
 
 The best way to get patches accepted is to discuss your ideas with project maintainers using [Discussions](https://github.com/NightscoutFoundation/xDrip/discussions) prior to implementation and discuss how to structure things so that they can fit well within the project and also the rationale for the change.
 
+Check the [development guide](DEVELOPMENT_GUIDE.md) on how to setup your dev environment.
+
 #### Copyright of submitted contributions
 
 Any submitted patches and pull requests must either be the original work of the submitter or be available under a suitable open source license which is compatible with inclusion in a GPL licensed project.
--- a/DEVELOPMENT_GUIDE.md
+++ b/DEVELOPMENT_GUIDE.md
@@ -1,0 +1,19 @@
+# xDrip Development Guide
+The following steps describe how to setup a development environment required to build xDrip.
+- Fork https://github.com/NightscoutFoundation/xDrip in GitHub.
+- Download, install and launch the latest Android Studio from https://developer.android.com/studio.
+- In the startup menu, click `Get from VCS`.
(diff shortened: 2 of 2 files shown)


bart2:
@Navid200, I think that makes sense. I don't mind documentation being in a wiki, as long as it is accessible somewhere (and we could link to it from https://github.com/NightscoutFoundation/xDrip/blob/master/CONTRIBUTING.md) :)

@JamOrHam, would it be possible to setup wiki permissions for @Navid200? I'd be happy to contribute to it as well.

Navid200:
@bart2 Just to clarify.

1- In my opinion, it helps if documentation is outside the repository to minimize the time jamorham has to spend on it so that he can spend the limited time he has on the technical front, where we really need him.

2- This would take control completely out of his hands though as far as which documentation he considers acceptable.
The solution to that is that the Wiki as well as the README.md, which he has complete control on can have pointers to external documentation.
This allows him to let users know which external sites he approves of.
So, if an external documentation becomes outdated, all he has to do is to change the pointer in the wiki or Readme to point to a more up-to-date external documentation.  
He can decide to point to multiple sites or a single site or whatever.

3- The wiki contains some minimal documentation.  If he really likes to directly be in control of some particular document, he can put it in the wiki.

4- jamroham already has documentation outside the repository in github.io.  He can also maintain some documentation there and point to it from the wiki again if he likes to be in direct control of some material literally.

All of this gives the repository owner effectively full control on documentation promoted by this repository.

This is my opinion and I am happy to be corrected.

The Readme.md needs a PR and I can open such if this is acceptable.
However, I am not aware of a reasonable way to open a PR to modify the wiki.

jamorham:
Any development guide should probably include which versions of Android studio have been tested to be okay as there are often introduced breaking changes/regressions with later versions that could trip people up. Likewise the lombok plugin is required and not mentioned here.

bart2:
@jamorham, yeah including Android Studio version(s) in the development guide is a good idea.

I was not aware of the lombok plugin. I was able to build and test xDrip+ without it. I think it further highlights a need for an official development guide.

Where would you see it hosted though? Does wiki seem a reasonable place for it? Or should we keep it in the code repo itself (for example this PR)?

jamorham:
I was thinking of bundling the lombok plugin in the source tree as since google broke it it needs to be loaded separately but I am not sure whether the one from the jetbrains repository has now been updated to work. It was previously integrated as a standard part of IDEA.

Navid200:
This is what I have:
https://navid200.github.io/xDrip/docs/dev/dev.html 

And I will update the Android Studio page as we move forward.  
I will also add more pages if needed.

PR #3071, opened 2023-09-06
Remove Load/Save Settings from the 3-dot menu

Navid200:
To save or load settings, it makes sense to do it by first going to settings.
If you do that, you will see a section titled "Copying settings".
Under it, you can copy or load settings using either the QR code or creating/loading the file.

What currently exists under the 3-dot menu is redundant.
Can we please remove it to help clean up?

This is what it will look like after this PR:
Screenshot_20230906-175335

This has been tested.

### DIFF
--- a/app/src/main/res/menu/menu_home.xml
+++ b/app/src/main/res/menu/menu_home.xml
@@ -57,14 +57,6 @@
                 android:orderInCategory="1"
                 android:onClick="exportCSVasSiDiary"
                 android:showAsAction="never"/>
-
-            <item android:id="@+id/importsettings"
-                android:title="@string/load_save_settings_on_sdcard"
-                android:checkable="false"
-                android:onClick="settingsSDcardExport"
-                android:visible="true"
-                android:orderInCategory="1"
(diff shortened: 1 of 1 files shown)


jamorham:
This menu deals with import export of data from xDrip of which settings are an important part. I don't really see why to remove the feature from a deep menu.

Navid200:
There are too many settings in xDrip.
It is so bad that when we do no need a new setting, I try to modify my PR not to add a new setting.  Because there are so many settings making it hard for a newcomer to find out what is what.

So, when I see the same setting available in more than one place, I cannot help wondering why.
If one needs to save or load settings, they can do it by going to settings.  To me, that is the intuitive thing anyone would do.
So, comparing the two, I suggest, we keep the one in settings and remove this one to clean up.
We have a long way to go in this front.  This is a small step.
Who is going to complain?  I mean how hard is it to tap on settings to copy the settings?

Navid200:
This is a minor small change to make the user interface consistent removing redundant settings.
But, it has not been considered.  So, there is no point keeping it open.

Issue #3082, opened 2023-09-11
Limit Google Drive daily backup to WiFi


Navid200:
### Discussed in https://github.com/NightscoutFoundation/xDrip/discussions/3007

<div type='discussions-op-text'>

<sup>Originally posted by **Navid200** August 14, 2023</sup>
Can we have an optional setting added to the Google Drive page to not upload the automatic daily backup using data and do it only when there is WiFi?</div>

PR #3083, opened 2023-09-12
Google Drive daily backup mobile data option

Navid200:
This allows user to limit daily backups to WiFi and exclude mobile data.

Fixes: https://github.com/NightscoutFoundation/xDrip/issues/3082


jamorham:
Will this default to not allowing backup via mobile data?

Navid200:
@jamorham Thanks for looking at this and thanks for catching that.
It is now unchecked by default so we only use WiFi by default.
Only if user checks this, will it also upload using mobile data.

Thanks

jamorham:
Would having the preference disabled by default change the default behavior though, so if someone installs a new version then it will stop backing up over mobile data where as it prior to this change by default it would have done?

I don't think changing the UI checked flag is the way to set the default value to true because as far as I know the value is only written to preference store on change.

Navid200:
@jamorham 
When I look at cloud_storage_api_use_mobile, I see nothing other than what I have done here.
Would you please tell me why the same concern does not exist for that?

jamorham:
To answer your question: Because it defaults to true and is handled via a preferences xml rather than a databinding activity.

There are a couple of different ways you can handle this issue depending of course on whether you think the existing behavior should be maintained which in this case I think it probably should.

Either invert the logic of the preference so that you can have it defaulting to false but that is a bit backward or when you read it in Backup you can default it to true if not set and the in the UI component you can initialize it if not set. You can use `Pref.isPreferenceSet()` to check if its set and if not set it to true. 

This needs to happen before the screen starts rendering as I don't think the prefsview is observable. To keep it clean you want encapsulate that behavior and so I think create a static public method in `Backup.java` to do this and then call it from the view model initialization block where currently it is already doing `clear()` 

Otherwise we could extend the prefsview to perhaps understand defaults but that is potentially a bigger change but the main reason you're having this problem is it doesn't support unset booleans being true.

One other way to achieve the same is to actually define your setting inside the regular preferences as then it will go through an initialization stage but as far as I'm aware (without checking) all of these backup settings are handled within the backup selector activity and there is no place in the preferences xml for related settings. But I'm just giving you an overview of the constraints and various possible solutions.

Navid200:
Thanks for the guidance.
I chose the first option.  I hope I didn't misunderstand.

Navid200:
Screenshot 2023-09-13 184830

app/src/main/res/layout/activity_backup_picker.xml
```diff
@@ -5,7 +5,7 @@
 
         <variable
             name="vm"
-            type="com.eveningoutpost.dexdrip.cloud.backup.BackupActivity.ViewModel" />
+            type="com.eveningoutpost.dexdrip.cloud.backup.BackupActivity.ViewModel" />Changed default to true
```
jamorham, line 8: Is this supposed to be here?

Navid200: No, sorry about that

app/src/main/res/layout/activity_backup_picker.xml
```diff
@@ -103,6 +103,18 @@
                 app:checked="@={prefs[`backup-automatic-enabled`]}"
                 app:showIfTrueAnimatedFastOff="@{vm.showAuto}" />
 
+            <CheckBox
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:checked="true"
```
jamorham, line 109: I think this checked = true could interfere with the one which follows the preference

Navid200: What do you mean by "the one which follows the preference"?

Navid200: I thought that's how I can have it enabled by default.

Navid200: Do you mean "False" in this?
Pref.getBooleanDefaultFalse(PREF_AUTO_BACKUP_WIFI)

Navid200:
Can I do this then?

jamorham:
Well I think this is getting very confusing now. The whole point of you inverting the logic was so that you could default the preference to being true. Am I to understand it that you want the checkbox to show checked when the preference is false? If so you can often invert the logic on display but this is a two way binding so I'm not sure about that. Surely from my previous comment you just remove it being checked and let it follow the preference which should default to false and the checkbox should be unchecked by default representing false?

Navid200:
No,
The preference is true by default because I thought that was required.
To make it true, it is Use WiFi only instead of use mobile.

Navid200:
If someone doesn't mind uploading using mobile, they need to disable the new setting.  If they leave it at default, it will only upload when there is WiFi.  If there is no WiFi, it will not upload the daily backup.

Navid200:
Let's say I have a service that charges a lot for mobile data and I want to minimize mobile data use.
After this PR, I will have to leave the new setting WiFi only enabled, which is enabled by default.

Navid200:
I created a new method that sets the preferences default to true.

jamorham:
To be clear, I would prefer not to change the existing default behavior. If I understand this correctly (and there have been changes and possibly double negatives) then merging this will cause anyone who isn't using wifi to cease having backups occur. I would prefer that the default remains as it is and when users are setting up the backup they can make it wifi only if they want.

Navid200:
I misunderstood.  Thanks.  I now understand you don't want me to change the default to not use mobile data.

My database backup size, on Google Drive, is 35MB.
That's several orders of magnitude greater than the data size used by xDrip uploading readings to Nightscout.

35MB x 30 > 1GB

There are individuals who subscribe to cell phone services with limited data amount.
I thought if someone installs xDrip and enables daily backup and find out at the end of the month that they have to pay a lot, they may be disappointed in xDrip.
That's why I would like to change the default.

Looking at the other side of the coin, if you approve this PR, I can see someone can install xDrip and enable daily backups but miss the clear note "WiFi only".  In 2 months, their phone breaks.  They buy a new phone and go to Google drive and see no backup because they never had access to WiFi over the past 2 months.
That's a disaster I suppose your preference will avoid.

I see both of those as bad scenarios we should avoid.
Will you consider the following workaround?

I will add a notification that comes up when someone enables daily backups.  It will say something like "xDrip by default will use WiFi only for daily backups.  If you want to also use mobile data for daily backups, please disable the setting `WiFi only`".
This will be a notification that will need to be dismissed.

Will you consider that?

Navid200:
If the only way to move forward with this is to have mobile data enabled by default, can I add a notification when daily backup is enabled to inform the user that by default mobile data will be used?

app/src/main/res/layout/activity_backup_picker.xml
```diff
@@ -103,13 +107,26 @@
...
+                android:dependency="backup-automatic-enabled"
+                android:layout_gravity="center"
+                android:gravity="center"
+                android:layout_marginBottom="20dp"
+                android:text="@string/daily_backup_even_mobile"
+                app:checked="@={prefs[`backup-automatic-mobile`]}"
```
jamorham, line 119: I still don't think this will work properly because even though you've changed the logic to default to true, the prefsview I don't think will handle that and will default to false and without a significant rewrite of that component which carries its own risks I don't think its easily solvable. I could take a look at that though as it would be better if we were not constrained by that. For other activities we rely on the XML settings to set the defaults.

For things as they are, the only alternative options are I think you either need to set the setting to true if it isn't currently set so that new users get this defaulted true setting or invert your logic so that it is something like "Back up over WiFi only" which then defaults to false and then you don't need to worry about presetting it to true.

I will take a look at PrefsView and get back to you on whether its easily solvable that way.

Navid200:
Mobile data is used by default.
The new setting (WiFi only) is unchecked by default.
Screenshot 2023-09-30 152654

jamorham:
See https://github.com/NightscoutFoundation/xDrip/commit/af3a008255800be4b4fce08b5609b2edf5cd255c this should allow you to specify like ```prefs[`backup-automatic-mobile:true`]``` to default a prefs view item to true if it has never been set.

app/src/main/res/layout/activity_backup_picker.xml
```diff
@@ -111,11 +111,12 @@
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 android:dependency="backup-automatic-enabled"
+                android:checked="true"
```
jamorham, line 114: this shouldn't be specified as it should be handled by the prefsview

Navid200:
Should I squash the commits?

jamorham (approved):

PR #3087, opened 2023-09-17
Remove setting: Unbond G5 before each read

Navid200:
Is there any scenario that requires this setting to be enabled?
If not, please let's retire it.

Screenshot_20230917-101348

Screenshot_20230917-101223


jamorham (approved):
Yes it doesn't appear to be used by the OB1 collector

Issue #1765, opened 2021-06-24
Update Nightscout REST API to v3


tolot27:
As described in https://github.com/NightscoutFoundation/xDrip/issues/475#issuecomment-850774431, we should support the new Nightscout REST API to track changes and deletions better.

List of tasks to implement:
- [ ] determine Nightscout REST API version
- [ ] show API version in Nightscout Uploader and Follower status page
- [ ] create a new class for new API
- [ ] create a Factory class to instantiate the correct Nightscout communication class
- [ ] implement new commands like corrected sgv (#1090) or changed/deleted treatments (#1403)
- [ ] merge Uploader and Follower source code to avoid similar problems like #1807 addresses
- [ ] support uploading of activity
- [ ] support for the new Transmitter ID and Sensor Code fields (#1624)
- [ ] incorporate it into a sync framework

The requests which be used for the new REST API are described at [HTTP Methods](https://restfulapi.net/http-methods).

tolot27:
@sulkaharo API v3 UPDATE does not support an eventType or note for svg entries. Does it matter?

Any further hints?

tolot27:
The requests which be used for the new REST API are described at [HTTP Methods](https://restfulapi.net/http-methods)

Icke12:
> As described in https://github.com/NightscoutFoundation/xDrip/issues/475#issuecomment-850774431, we should support the new Nightscout REST API to track changes and deletions better.
> 
> List of tasks to implement:
> - [ ] determine Nightscout REST API version
> - [ ] show API version in Nightscout Uploader and Follower status page
> - [ ] create a new class for new API
> - [ ] create a Factory class to instantiate the correct Nightscout communication class
> - [ ] implement new commands like corrected sgv (#1090) or changed/deleted treatments (#1403)
> - [ ] merge Uploader and Follower source code to avoid similar problems like #1807 addresses
> - [ ] support uploading of activity
> - [ ] support for the new Transmitter ID and Sensor Code fields (#1624)
> - [ ] incorporate it into a sync framework
> 
> The requests which be used for the new REST API are described at [HTTP Methods](https://restfulapi.net/http-methods).

tolot27:
@Icke12 You just quoted my description. What do you want to say?

Navid200:
@sulkaharo 
Would you please tell us what the need is for updating to V3?  
What do we lose if we don't update?
What do we gain if we update?

Navid200:
I have been told the following two statements.
[...]

PR #3089, opened 2023-09-17
Support subpath in url for Nightscout follower

jeroennijhof:
This will add support for subpaths in the Nightscout follower url.

Currently when you have a Nightscout instance running at i.e. https://domain.com/subpath/ it will strip the /subpath/ from the url which is not correct. By using relative paths for the @GET methods subpaths are working.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollow.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollow.java
@@ -61,10 +61,10 @@ public class NightscoutFollow {
                 "User-Agent: xDrip+ " + BuildConfig.VERSION_NAME,
         })
 
-        @GET("/api/v1/entries.json")
+        @GET("api/v1/entries.json")
         Call<List<Entry>> getEntries(@Header("api-secret") String secret, @Query("count") int count, @Query("rr") String rr);
 
-        @GET("/api/v1/treatments")
+        @GET("api/v1/treatments")
         Call<ResponseBody> getTreatments(@Header("api-secret") String secret);
     }
 


Navid200:
Is there any Nightscout server that cannot produce a hostname without a subpath?

jeroennijhof:
I'm not sure what you mean but the library used, retrofit, makes sure the baseurl will always end with a slash so it will be safe to use the relative path.

Navid200:
Nightscout needs a server.  
If you use Heroku, or Google Cloud or anything else I am aware of, the hostname will be something like thgis:

user.heroku.com

or 

user.chickenkiller.com 

for example.


I am asking why would anyone have a subpath?  
If everyone can create a hostname that looks like user.heroku.com, which has no subpath, why would we need to worry about a subpath?

jeroennijhof:
Nightscout needs a server indeed, but since it's open source you can decide yourself how you host it right? I'm not really sure why you want to restrict choices?

Navid200:
The reason xDrip still exists and is supported is that a few volunteers have spent a lot of time on it and continue to support it.
It makes no sense to support something that is not absolutely needed considering the limited resources we have.
We have open issues because we don't have enough developers to fix them.
Some come here and open a PR and after it is merged, we never see them again.  Then, something breaks, and we have to reverse engineer to figure out what is causing the problem.
We cannot approve every single PR without asking questions.

Is there a reason you cannot set up your path as user.domain.com without a subpath?  
If you can, you will not need this PR, right?  Then, it is not a necessity.  
I could be wrong about this.  But, you are not helping me see why.

psonnera:
Whilst this change, if tested correctly, doesn't look bad at all, I believe the community would benefit having more information on the Nightscout deployment method and use case in question. Subpath probably allows reusing an existing DNS for either more apps or Nightscout instances. @jeroennijhof is that the case?

jeroennijhof:
@psonnera that's correct. We currently run 5 Nightscout instances and since we already had a ssl certificate, domain and vps it's pretty easy and cheap to run it all on one vm. Sure I can set it up with subdomains but that means I need to buy 5 ssl certificates to be more secure. I don't like to expose the api secrets on http. And since I run into this I figured there will be more people out there wanting to reuse their VM, DNS and SSL.

@Navid200 thank you for your explanation I really understand the pain since I run a couple of o.s. projects myself so if you want to close the PR, that's fine.

Navid200:
No, I don't want to close the PR.
I ask questions in some PRs.  But, I am not the one who makes the last call.  Having had this conversation, my hope is that it addresses some questions that jamorham would have had to ask anyway.  So, my hope is that asking these questions helps him.

Thanks

Let's now wait for his response.

jamorham:
What are the pro's and cons of merging this? Can it break any current user's settings?

Navid200:
I am embarrassed to say I jumped in and asked questions without looking at the literal changes.  Just 2 slashes?!
I'm sorry.

We just need to see test results. 

And I believe the last empty line is something that is needed.

jamorham:
A single character is enough cause serious problems. This change maps the API to a relative instead of absolute root url. There is a very strong chance that it works but I would like to ensure the pros and cons have been tested to avoid us breaking any existing installations that might be affected by this. I would like to hear from @jeroennijhof if they think there could be an unintended consequences for existing users with this change?

jeroennijhof:
@jamorham the only way this change could have any unintended issues for people when they already have a subpath in the Nightscout url i.e. with the api full path. Then the url will become invalid as in /api/v1/api/v1/entries or sorts. But since the majority have Nightscout running under it's own domain name I think it's not a really big issue, but I'm not sure.

app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollow.java
```diff
@@ -212,4 +212,4 @@ public static void resetInstance() {
         UserError.Log.d(TAG, "Instance reset");
         CollectionServiceStarter.restartCollectionServiceBackground();
     }
-}
```
tolot27, line 215: Please do not remove the EOF line break!

tolot27:
> @jamorham the only way this change could have any unintended issues for people when they already have a subpath in the Nightscout url i.e. with the api full path. Then the url will become invalid as in /api/v1/api/v1/entries or sorts. 

Can this be detected and removed from the URL setting?

A unit test would be great to test against this potential problem. Afterwards a fix can be implemented, i. e. by removing the API subpath. This is at least necessary for #1765.

PR #3094, opened 2023-09-21
No update available toast message

Navid200:
If you check for updates when there is no update available, xDrip provides no information.
The following question occasionally comes up.
Does "Check for updated version" work?

This PR shows a toast message clarifying that there is no update.
Screenshot_20230921-121759

It only does that when checking manually.
If you have enabled automatic update check, there will be no toast message everyday.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/UpdateActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/UpdateActivity.java
@@ -173,8 +173,8 @@ public class UpdateActivity extends BaseAppCompatActivity {
                                     }
                                 } else {
                                     Log.i(TAG, "Our current version is the most recent: " + versionnumber + " vs " + newversion);
-                                    if ((callingMethodName).equals("checkForUpdate")) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/UpdateActivity.java
```diff
@@ -80,6 +80,9 @@ public class UpdateActivity extends BaseAppCompatActivity {
     private static String CHECKSUM = "";
 
     public static void checkForAnUpdate(final Context context) {
+        StackTraceElement[] stacktrace = Thread.currentThread().getStackTrace();
```
jamorham, line 83: This isn't a good way to achieve this when you can simply add an alternate method signature. I will do this and you can use that to add the UI element.

jamorham:
See the change I made in: https://github.com/NightscoutFoundation/xDrip/commit/ff0ebe11e3ca37b0a8f998a4c4a9def4b3ca4c6b which should help you to cleanly achieve this.

Navid200:
Thanks

jamorham (approved):

PR #3111, opened 2023-10-03
Allow to start even if the number of days is -1 for a mod TX

Navid200:
This will bypass the check for transmitter days for starting a sensor for mod transmitters.


Navid200:
This is not the best solution.
This is the right first step: https://github.com/NightscoutFoundation/xDrip/pull/3113

PR #3121, opened 2023-10-08
CareLink reCAPTCHA login using WebView

benceszasz:
New browser (webview) based login and token refresh in the background for CareLink Follow to handle the new CareLink login page with reCAPTCHA


benceszasz:
I have tested it in with all kinds of accounts in every region and it is working fine. My private version has been used by quite a few people for more than a week, based on which I have made corrections and fine-tunings.
I know, that there is still room for improvements and fine-tuning, but without this browser based login it is not possible to use CareLink Follower datasource at all.

jamorham (approved):

jamorham:
@benceszasz many thanks

PR #3122, opened 2023-10-08
Selective activation time update

Navid200:
After a hard reset of a mod transmitter, the read time stamp becomes a very large value.  I am not sure.  But, this could be because of a format mismanagement of a negative number, turning a very small negative value into a very large positive value (6214 days or 17 years)!

The activation time calculated from that goes back to 2006.  When this is written into the persistent store, resulting transmitter days value becomes invalid (unknown).

This PR avoids writing into the persistent store an activation time that is so far back in the past.

Tests show that the lock situation reported by users (https://github.com/NightscoutFoundation/xDrip/discussions/3078) does not happen any longer after this PR.


Navid200:
Will open another to address this better.

PR #3123, opened 2023-10-09
Block incorrect raw timestamp

Navid200:
After a hard reset, the raw timestamp is incorrect until after sensor start.
With the incorrect timestamp, the corresponding activation time is incorrect resulting in the transmitter days parameter to be reported as -1.
As long as transmitter days is -1, xDrip refuses to start sensor.
This creates a catch-22 situation.

This PR does not allow an incorrect raw timestamp to be used to update the transmitter age.
Tests show that the catch-22 situation is avoided after this PR.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/g5model/Ob1G5StateMachine.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/g5model/Ob1G5StateMachine.java
@@ -130,6 +130,7 @@ public class Ob1G5StateMachine {
     private static volatile BgReading lastGlucoseBgReading;
     private static volatile AuthRequestTxMessage lastAuthPacket;
     private static volatile boolean backup_loaded = false;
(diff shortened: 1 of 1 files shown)


jamorham:
Can you elaborate on what sort of raw timestamp is being seen during this condition? I'm not sure if this is the best way to deal with this without knowing what the condition is.

Navid200:
The raw timestamp sent after a hard reset is 6214 days (7 years).

Navid200:
Let me be the devil's advocate:

Let's say someone solders wires to the battery contacts and wires them to a power supply such that in 7 years the transmitter is still running.
At that point, after this PR, the raw timestamp will be switched to 0.  But, the native timestamp will show 6214.
But, as soon as transmitter days goes over 180 days, my understanding is that the transmitter will refuse to start a session.
I don't see the point of such an experiment.  
Regardless, in 7 years, I don't think G6 will still be as popular as it is today.

jamorham (approved):

PR #3133, opened 2023-10-16
Clarification of "Stop sensor" menu

Navid200:
When you tap on stop sensor, you get a menu that offers you to reset all calibrations even if you are using a standard G6.
In that case, calibrations are all submitted to the transmitter and xDrip has no way of resetting them.  Therefore, the option could cause confusion.  

This PR removes that option for standard G6.
Screenshot_20231016-160636
<br/>  

It also adds a warning that you cannot simply restart the sensor for any Firefly G6.
Screenshot_20231016-160700
<br/>  
  
There are cases where people who have been using Libre for some time and have just switched to G6 stop sensor and expect to just start it again.  After they post in a forum and ask, they are told that they cannot just restart a G6 sensor as they could with other sensors.  
<br/>  
  
Is there a way to add a conditional to an xml file so that I don't have to create a new xml file as I have done here?


jamorham:
I think the best thing to do is if I update the Stop Sensor activity to use databinding then this sort of thing is easy. Don't create an additional layout file...

Navid200:
Data binding is used now.

This is the confirmation menu you will get if you tap on stop sensor from the main menu if calibrations cannot be reset by xDrip.
Screenshot_20231225-155700

This is what you will see if you tap on stop sensor on the confirmation menu and calibrations cannot be reset.
Screenshot_20231225-155835

This is what you will see if you tap on stop sensor on the confirmation menu and you are using a G7.
Screenshot_20231225-160334

Navid200:
Changes made to the layout file:

1- Wrapping with layout tags to enable data binding
2- Including Import ms file so that we can access the variable resettableCals for removing the reset cal button conditionally
3- Removed redundant Stop Sensor title.
4- Added a conditional to reset cal button.
5- Added an on-click listener to the stop sensor button.
6- Removed redundant formatting lines.
7- Automatic code formatting (the reason for most of the highlighted differences)

Navid200:
This is tested with G6 Firefly, G7 and fake data source.

jamorham:
You shouldn't be modifying the `MicroStatus` interface for stop sensor as they are separate responsibilities. Instead see I have just added a `ViewModel` child class in the `StopSensor` activity which is where you can put your logic and reach out to where you need to to get application state. If you find yourself repeating logic tests then create a new utility class for those and reference those methods from wherever you need to. You will need to adapt this PR to incorporate my changes and add back the ones which I haven't included.

Navid200:
This is the reason I would like to make it clear to those who consider stopping G7, in xDrip, that there is no way back.
Screenshot 2024-01-16 144640

jamorham (approved):

PR #3158, opened 2023-10-27
Reduce alpha of AOD Widget on low display brightness

Der-Schubi:
This tones down the alpha of the Always On Display Widget on low display brightness. Below a display brightness of 60 the alpha is gradually reduced from 1.0 down to 0.4 for brightness 0.


Der-Schubi:
Another one for Always On Display @Navid200 ;)

ah7675:
Thank you for keep improving the experience.
Is it possible to make this optional? Or let user adjust slope?
The new alpha value looks too dim on my phone, 0.8 was perfect on my Pixel8.

Der-Schubi:
> Thank you for keep improving the experience. Is it possible to make this optional? Or let user adjust slope? The new alpha value looks too dim on my phone, 0.8 was perfect on my Pixel8.

I was thinking of making it adjustable. Just wanted to put it up here and see what everyone is thinking.
I had a much simpler version of that in my own code for over a year now, I just switched from 0.8 to 0.4 below a brightness of 10. This worked good for my Pixel 5.

Navid200:
What is the status of this?

Considering AOD now has a dedicated settings page, one (or two) more settings should not break the bank.

PR #3167, opened 2023-11-05
Fix Bluetooth connectivity problems

simelis:
Remove use_auto_connect as it's value never changes.
Reset always_scan after successful connection or manual collector restart.
Restart bluetooth once BleDisconnectedException with status 133 is caught.
Restart bluetooth more frequently on CONNECT_NOW failures.
Add aggressive_bluetooth_restart advanced setting.

Related discussion here: https://github.com/NightscoutFoundation/xDrip/discussions/3165

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -307,7 +307,7 @@ dependencies {
     }
     implementation 'com.github.jamorham:amazfitcommunication:master-SNAPSHOT'
     implementation 'io.reactivex:rxjava:1.3.3'
-    implementation "com.polidea.rxandroidble2:rxandroidble:1.12.1"
+    implementation "com.polidea.rxandroidble2:rxandroidble:1.17.2"
     implementation 'com.google.guava:guava:24.1-android'
     implementation 'com.embarkmobile:zxing-android-minimal:2.0.0@aar'
     implementation 'com.embarkmobile:zxing-android-integration:2.0.0@aar'
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
@@ -1340,18 +1340,20 @@ public class Ob1G5CollectionService extends G5BaseService {
         // TODO under what circumstances should we change state or do something here?
         UserError.Log.d(TAG, "Connection Disconnected/Failed: " + throwable);
 
-        if (throwable instanceof BleDisconnectedException && genericBluetoothWatchdog())
-        {
-            int status = ((BleDisconnectedException) throwable).state;
-            UserError.Log.d(TAG, "Received BleDisconnectedException, status = " + status);
-            // Bluetooth gets stuck for some devices with error 'Disconnected from MAC='XX:XX:XX:XX:XX:XX' with status 133 (GATT_ERROR)'
-            if (status == 133) {
-                if (JoH.ratelimit("BleDisconnectedException", 300)) {
-                    UserError.Log.e(TAG, "Initiating bluetooth reset due to BleDisconnectedException with status 133");
-                    JoH.niceRestartBluetooth(xdrip.getAppContext());
-                }
-                else{
-                    UserError.Log.d(TAG, "BleDisconnectedException rate limited");
+        boolean aggressive_bluetooth_restart = Pref.getBoolean("aggressive_bluetooth_restart", true);
+        UserError.Log.d(TAG, "aggressive_bluetooth_restart: " + aggressive_bluetooth_restart);
+        if(aggressive_bluetooth_restart == true) {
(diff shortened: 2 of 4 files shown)


Navid200:
xDrip already has too many settings.
New settings should only be added if/when absolutely necessary in my opinion.

The Bluetooth watchdog is only a last resort.  When it triggers, it could disconnect anything and everything.  I have experienced my Android Auto GPS stopping while driving when it triggered.

Navid200:
When you say the auto connect value never changes, are you sure this is not related to the trust auto connect on the Bluetooth settings page under less common settings?

simelis:
> xDrip already has too many settings. New settings should only be added if/when absolutely necessary in my opinion.
> 
> The Bluetooth watchdog is only a last resort. When it triggers, it could disconnect anything and everything. I have experienced my Android Auto GPS stopping while driving when it triggered.

I'm aware and perfectly fine by Bluetooth disconnecting everything, if that is needed to keep the connection with the sensor alive. This is child's phone which doesn't run any critical programs or services other than related to diabetes management, xDrip being the most critical of them all. I know there are people who have a separate phone dedicated for xDrip/AAPS, and I'm sure that on time BG data is also a big priority for them.

I agree about xDrip having too much settings, especially when a lot of them are only relevant for older sensors like g5, which is not very clear for new users. Should we put some effort in regrouping settings or hiding those, which are not relevant for the current data source?

Regarding this PR, would you suggest that the same Bluetooth watchdog setting could be used for this case?

Navid200:
@simelis I am one of the volunteers.  I am not the owner of this repository who is the lead developer.

There are phones that have trouble.  Xiaomi is one.  Samsung sometimes is one.  
The challenge for us is to try to make as many people as we can happy.  

I have removed some of the old settings that were not needed.  I will continue to do that.
Just because a setting includes G5 in the title does not mean that is not needed for G6 or Dexcom One.

If we change a setting in xDrip to help Nokia, we have to make sure it will not hurt Samsung or Pixel or Motorola or ...

Please wait for @jamorham for answers to your questions.

simelis:
> When you say the auto connect value never changes, are you sure this is not related to the trust auto connect on the Bluetooth settings page under less common settings?

It's a private variable, which has a default value of false and is not set anywhere, the app compiles just fine after I removed that variable and its usages in handleWakeup(). I don't see any way it could be changed anywhere else. The trust auto connect setting has a separate getTrustAutoConnect() method, which is used only when specialPairingWorkaround() is used. So no, as far as I understand, it is not related to this setting, it's hard for me to find out what is the purpose of that variable at all.

simelis:
> Just because a setting includes G5 in the title does not mean that is not needed for G6 or Dexcom One.

Sure, I don't trust the names, I always look it up in code. The settings inside Bluetooth section is quite a mess though.

g5_bluetooth_watchdog setting, for example, is not used together with generic bluetooth_watchdog anywhere in code, so they could be merged.

bluetooth_watchdog_timer, close_gatt_on_ble_disconnect, bluetooth_use_scan, bluetooth_excessive_wakelocks, bluetooth_frequent_reset, use_transmiter_pl_bluetooth, use_rfduino_bluetooth, pref_dex_collection_polling, always_discover_services, blukon_unbonding are used only in DexCollectionService and are not relevant for all other implementations.

bluetooth_allow_background_scans is used only by JamBaseBluetoothSequencer, that is, BlueJay, InPen, LeFun and MiBand services.

It's not clear to me where Companion Bluetooth setting is actually used, I can trace it to UiBasedCollector, that is, manual entry? Not sure how it's related.

It would make much sense to hide these settings when they are not relevant.

Navid200:
Some settings are used for Libre as well and not just Dexcom.
UI based is companion app: https://navid200.github.io/xDrip/docs/Follow/CompanionApp.html

jamorham:
Which handsets and android versions has this been tested with?

simelis:
> Which handsets and android versions has this been tested with?

@jamorham I'm testing this using Nokia G21 on android 13 for about 4 days now. The aggressive Bluetooth restart kicks in about 4-5 times a day for me, but the data loss is getting fixed in 10 to 15 minutes without any user action, and that makes me quite happy with the results ;)

simelis:
> Considering this is for Nokia, I suggest we make sure that existing solutions don't work first.
> 
> Do you mind having a look at the following and giving the suggestions a try? https://dontkillmyapp.com/hmd-global

Hi, first of all, thanks for your work supporting this awesome tool!
We've started using AAPS and connectivity problems with Nokia remained not acceptable for looping even with the code changes I made, so we just moved to Google Pixel which seems to be working quite well. 
I have looked into dontkillmyapp while we were still on Nokia, but everything mentioned there is quite old and were not relevant for our phone.
I would say we can close this PR without merging, as the connectivity problems of that phone were not resolved using my changes, so there's no big need for a merge. Thanks again and keep up the good work!

PR #3172, opened 2023-11-06
New G7 Firmware

Navid200:
Before:
Screenshot_20231106-161830

After:
Screenshot_20231106-162712

Canada


jamorham:
I think from your screenshot this is probably the case, but please confirm that the firmware is working okay?

PR #3157, opened 2023-10-26
Add Afrezza, Apidra, Levemir, Basaglar and Tresiba profiles

TheConen:
This is NightscoutFoundation#1388 with the merge conflicts resolved.

Since #1388 seems to be dead, I resolved the merge conflicts and created this PR.

PR #3177, opened 2023-11-07
Fix insulin profile list not showing all insulin profiles

TheConen:
As @Navid200 mentioned in #3157, the current insulin profile list has a fixed size and does not show all insulin profiles.

This PR fixes that by changing the fixed vertical size to a dynamic size based on the content.

Before (Notice the missing FIASP which should be at the top:
280917582-8fa0d862-27ee-4fc7-a555-94b31b13aca3

After (depending on the screen size/resolution you need to scroll down to see the buttons):
grafik
grafik


Navid200:
There is a (user) workaround.

If the user goes to Android settings and enables the setting that allows the orientation to automatically change between landscape and portrait based on gravity, rotating the phone to landscape mode will allow the user to scroll through the list.

jamorham:
Hi, thanks for this. Notes about this PR:
1) Please integrate the latest changes on master as some of them touch this and there are new support libraries via androidx so please check compatibility.
2) Please explain the process that this PR uses to fix the problem - I believe you are simply adjusting the layout to use wrap content so that the scrollview contains the full size of the data. But please confirm that is what is happening.
3) Please remove the changes that have been made to other files. I assume that there are no functional changes to `initiallexicon.txt`  and `insulin_profiles.txt` and so please remove these files otherwise document the changes and provide a rationale for them that can be reviewed. json files are text files and there is wider ide compatibility using the txt extension afaik.

TheConen:
1.) Done. Tested it on my OnePlus Nord 2 5G and several devices in Android Studio Emulator. Worked fine on all devices. 

2.) The previous layout used two nested ScrollViews, with the inner scroll view being a fixed height of 440dp. This caused problems depending on the screen resolution. I changed this to one ScrollView containing all the content and matching the height to wrap the content, dynamically adjusting to the screen resolution.

3.) The correct file extension for JSON files is `.json` (see RFC 8259 https://datatracker.ietf.org/doc/html/rfc8259). If you use a different file extension, you are losing out on all the support your IDE gives you, like linting, syntax highlighting, auto completion etc. and you are conveying wrong information. JSON is more than just text - it has rules and a syntax that it needs to adhere to. XML and YAML (or all the Java code) are "just" text files, too - but your manifest file is still named `AndroidManifest.xml` and not `AndroidManifest.txt`. (Fun fact: there is quite a debate about the correct file extension for YAML being `.yml` or `.yaml` - according to the official YAML website, it is `.yaml` https://yaml.org/faq.html)

jamorham:
You make a good argument regarding json file extension. I may be quoting how things were in 2016 with android studio.
Please instead just confirm that, other than the file extension change, the only changes are to whitespace formatting and there are no functional changes to those files and that you have checked that the lexicon file is still being processed correctly?

TheConen:
There are no changes to the actual content of the file. I just changed the file extension and did an auto-format (Ctrl+Alt+L) with Android Studio.

From what I can tell, everything works fine. `initiallexicon` and `insulin-profiles` are both read correctly (I checked with the Debugger).

jamorham (approved):

PR #3183, opened 2023-11-13
Remove setting Authenticate before each read

Navid200:
This PR permanently enables the setting "Authenticate before each read" on the G5/G6/Dex1 Debug Settings page, and removes the setting.  
There are individuals who disable it and wonder why they have trouble.

This image shows what the settings page looks after:
Screenshot_20231113-142044

Selecting different Libre options, the logs show the exception the same as selecting G6.
Screenshot_20231113-141719

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/IdempotentMigrations.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/IdempotentMigrations.java
@@ -147,7 +147,7 @@ public class IdempotentMigrations {
         Pref.setBoolean("use_ob1_g5_collector_service", true);
         Pref.setBoolean("ob1_g5_fallback_to_xdrip", false);
         Pref.setBoolean("always_unbond_G5", false);
-        Pref.setBoolean("authentificate_before_reading", true);
+        Pref.setBoolean("always_get_new_keys", true);
     }
 
 }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/IdempotentMigrations.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/IdempotentMigrations.java
@@ -147,6 +147,7 @@ public class IdempotentMigrations {
         Pref.setBoolean("use_ob1_g5_collector_service", true);
         Pref.setBoolean("ob1_g5_fallback_to_xdrip", false);
         Pref.setBoolean("always_unbond_G5", false);
+        Pref.setBoolean("authentificate_before_reading", true);
     }
 
 }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
@@ -2493,7 +2493,7 @@ public class Preferences extends BasePreferenceActivity implements SearchPrefere
            //  removePreferenceFromCategory("use_ob1_g5_collector_service", "ob1_options");
            //  removePreferenceFromCategory("ob1_g5_fallback_to_xdrip", "ob1_options");
            //  removePreferenceFromCategory("always_unbond_G5", "ob1_options");
-           //  removePreferenceFromCategory("authentificate_before_reading", "ob1_options");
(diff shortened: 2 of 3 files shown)


jamorham:
So there are a number of problems with this PR. I will go through them below:
1) This doesn't affect the correct key which is `always_get_new_keys` yours is targeting the string value which is incorrect.
2) I don't believe this will have any influence on the OB1 collector and I think will apply only to the defunct non-ob1 collector for G5 so I don't think there could be operational problems for users regardless of this setting. But tell me if you think that is incorrect?

Having said that I think it is okay to remove this setting as it is needless clutter for something users shouldn't change and applies to an obsolete collector.

Navid200:
Thanks,

I will fix it.
Thanks for the review.

Navid200:
I'm not clear about your direction.  You say:

> Having said that I think it is okay to remove this setting  

That's what I wanted to do.  To remove the setting.
If this PR removes the setting, please merge it. 
If it doesn't, would you please explain a little bit more what I need to change?

Navid200:
Are you saying I should leave IdempotentMigrations alone and not change it because this setting doesn't really exist?
If yes, I understand and will remove that change to that file if you confirm.

jamorham:
Hi, please re-read my previous comment. Point 1 explains why the PR wont do you you intend it to do and Point 2 explains that I don't believe there are currently any users affected by the problem you mention as the rationale - but I am not against removing the setting as it is not needed but this PR doesn't do that properly so please re-check your work.

Navid200:
@jamorham I have a screenshot, in the very first post, that shows that this PR removes the setting, which is the intent.

Navid200:
I did recheck my work.  I just did again.
Searching for always_get_new_keys, before this PR, I get 11 matches in 6 files.

7 of the 11 are wear.
The remaining 4 are the following:
Screenshot 2023-12-05 090224
Screenshot 2023-12-05 090352
Screenshot 2023-12-05 090432
Screenshot 2023-12-05 090525

In this PR, I am deleting the first, which removes the setting.
And I am forcing it to be true using the migration method.

Please tell me what I am overlooking.

Navid200:
In these PRs, where I remove a setting that is not needed any longer, I follow your approach here as you directed me:
https://github.com/NightscoutFoundation/xDrip/commit/23ccbd14d94f0a9c19c23011b4820557812eedb2

Please tell me what I have done here that is not the same.  Or what I have not done that is needed.

app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/IdempotentMigrations.java
```diff
@@ -147,6 +147,7 @@ private static void legacySettingsFix() {
         Pref.setBoolean("use_ob1_g5_collector_service", true);
         Pref.setBoolean("ob1_g5_fallback_to_xdrip", false);
         Pref.setBoolean("always_unbond_G5", false);
+        Pref.setBoolean("authentificate_before_reading", true);
```
jamorham, line 150: The key here is wrong

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -2493,6 +2493,7 @@ private void removeLegacyPreferences() {
            //  removePreferenceFromCategory("use_ob1_g5_collector_service", "ob1_options");
            //  removePreferenceFromCategory("ob1_g5_fallback_to_xdrip", "ob1_options");
            //  removePreferenceFromCategory("always_unbond_G5", "ob1_options");
+           //  removePreferenceFromCategory("authentificate_before_reading", "ob1_options");
```
jamorham, line 2496: the key here is wrong

app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/IdempotentMigrations.java
Navid200: How did I miss that?!  
Thanks

Navid200:
Corrected the keys.  Thanks for catching that.
Screenshot_20231205-114245

jamorham (approved):

PR #3195, opened 2023-11-18
Update G7 warmup and lifetime

Navid200:
When starting a G7, the warmup time shows as 2 hours.  When, the correct value is 30 minutes.
And the device is shown to expire in 10 days.  But, in fact, it will expire in 10.5 days.

This PR updates both of those.

The expiry has been verified already.
The warmup will be tested in 10 days.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/g5model/FirmwareCapability.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/g5model/FirmwareCapability.java
@@ -79,6 +79,10 @@ public class FirmwareCapability {
         return isFirmwareRawCapable(version); // hang off this for now as they are currently the same
     }
 
+    static boolean isG7Firmware(final String version) {
+        return KNOWN_ALT_FIRMWARES.contains(version);
+    }
+
     public static boolean isTransmitterPredictiveCapable(final String tx_id) {
         return isG6Firmware(getRawFirmwareVersionString(tx_id));
     }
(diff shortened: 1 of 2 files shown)


jamorham:
Looks like it should work, let me know when tested.

gui-dos:
You can get the exact G7 `warmupLegth` in seconds by using the Tx/Rx opcode `0x52`.

With the Dexcom G7 I am getting 27,  the ’5406’ in the 15-byte packet `52 00 c0d70d00 5406 02010404 ff 0c00`

0x654 = 1620
1620 / 60 = 27.0

Here is my Swift code:
https://github.com/gui-dos/DiaBLE/blob/abd7633bce33d1989d29fc7678a73c1c590d058e/DiaBLE/Dexcom.swift#L522

Navid200:
@gui-dos Are you suggesting that the warmup time is going to vary from G7 to G7?

gui-dos:
@Navid200

No, I tested three G7s and all of them had that 27 minutes warmup time: that packet can be sniffed during the sensor activation.

I have read that other users experience a warm-up time slightly less than 30 minutes, and this is proven confirmation ;-)

Navid200:
I have tested this.  
The warmup period is correct now.

If we want to choose a value other than 30 minutes, I would choose 35 rather than 27.

When someone feels that they are low and look at their CGM and see that it says warmup ends in 5 minutes, they may decide to wait 5 minutes to get a reading instead of using their blood glucose meter.
If it turns out that they do not get a reading when we promise, it would be unacceptable in my opinion.

Therefore, to be safe, I strongly advise against using 27.

jamorham (approved):

PR #3199, opened 2023-11-20
No stop sensor for G7

Navid200:
There is no circumstance under which anyone would need to stop a G7 session in progress.

If you disagree, please explain.  Why would anyone ever need to stop a G7 session?

Otherwise, please let's remove the (redundant) option.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
@@ -101,8 +101,10 @@ public class NavDrawerBuilder {
                         }
                     }
                 }
-                this.nav_drawer_options.add(context.getString(R.string.stop_sensor));
-                this.nav_drawer_intents.add(new Intent(context, StopSensor.class));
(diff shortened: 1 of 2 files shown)


Navid200:
What if I just don't want to receive readings any more for whatever reason?
Just stop collection in xDrip.  Then, you will have no more readings from the session.  There is no reason to stop the session just because you don't want to receive any more readings.

What if I open my next device and it happens to have an identical pairing code to my current device?
Then, you are, going to have a problem with establishing connectivity even if you stop the current session.  Stopping the session on your current device is not going to change anything.
You will need to move the old device out of the reception area of your phone.  As soon as you do that, you will be able to use your new device.  And if you still have a session in progress on it, it will make no difference.

<br/>  
  
Please tell me why we need to be able to stop a session on a G7.

Navid200:
Why do we need to remove this option from xDrip.  xDrip is working fine with it as is.

Someone is going to tap on it and after, ask, how can I undo this?!

Please, there is no reason whatsoever anyone would ever need this option.  Please let's make xDrip cleaner and simpler by removing unneeded options.

Navid200:
Dexcom app also offers to stop the session.
Even though that still makes no sense, I close this PR.

PR #3202, opened 2023-11-22
Allow history to show all readings even for 1 reading per minute

Navid200:
To recreate the problem reported here, https://github.com/NightscoutFoundation/xDrip/discussions/3190, I created a test version in which, I increased the period to 10 minutes (1 reading per 10 minutes rate).
Then, history I had generated with a 5 minute period (1 reading per 5 minutes), was shown only covering half a day everyday.

This PR increases the max rate to 1 reading per minute for history.

Fixes: https://github.com/NightscoutFoundation/xDrip/discussions/3190

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/BGHistory.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/BGHistory.java
@@ -50,6 +50,8 @@ public class BGHistory extends ActivityWithMenu {
     private int noDays = 1;
     private SharedPreferences prefs;
     private TextView statisticsTextView;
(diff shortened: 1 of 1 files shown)


Navid200:
I cannot test this because I don't have a source that has a 1 reading per minute rate.
But, I verified that I can see the problem by going the other way and intentionally changed the code temporarily to reduce the allowed rate to less than 1 reading per 5 minutes and I could see the error as reported.

If you tell me how I can test it at 1 per minute, I will.
Otherwise, please merge and let it be tested as a Nightly.

jamorham (approved):
So it is possible to retrieve sample period via `DexCollectionType.getSamplePeriod()` but I'm thinking there could be problems if collection type is changed etc so for this table view which isn't used that frequently it is fine to request more rows than we might always need.

PR #3210, opened 2023-11-26
CareLink Follower - CarePartner app authentication

benceszasz:
**Issue**
Auth token is only valid for 1-3 hours, if it cannot refreshed for exmaple due to lack of internet connection, a new login must be carried out manually.

**Solution**
CarePartner app authentication is valid for 1 week, thus even if xdrip is unable to connect to CareLink for a few days (for example due to the lack of internet connection), it will be able to renew the login (refresh the token) automatically in the background after it can connect to CareLink again without needing to login again manually using the browser. The user account restriction applies to this authentication process as well: **a different dedicated follower account must be used in every app, otherwise the previous auth token will be revoked.**

**Credits**
CarePartner app authentication process is based on the impressive work of **@palmarci** in extracting the entire authentication process of the mobile app.

**Limitations**
When the same account is logged in from another xdrip or official mobile application the previous login will be closed (authetication token is revoked). A **different dedicated follower account must be used in every official and xdrip app**, otherwise the previous auth token will be revoked.

**Testing**


palmarci:
Great work as always! Thanks for this

jamorham (approved):
Thanks for this. If strings are stable, where they are hard coded you could extract these to string values for localization in future.

jamorham:
Yes a later PR is fine. I have already approved this one so don't change it.

benceszasz:
Ok, thanks! I will start to extract the hard coded texts and put them into the translatable strings and create a new PR.

PR #3216, opened 2023-11-30
CareLink Follower - Optimized HTTP client, translatable texts, server communication error

benceszasz:
**Optimized HTTP client usage**
Reuse same HTTP client in every call.

**Handling server communication error**
Sometimes either at the initial background HTTP request or later at the login page CareLink server resets the connection. This seems to happen if there was a recent request or multiple login attempts from the same IP or just because maybe CareLink is overloaded. An error detection logic and alert dialog was added for this issue.

**Translatable texts**
Messages, popup texts were moved to translatable strings.


jamorham (approved):
Thanks for this

PR #3218, opened 2023-12-01
Include sensor start time in local app broadcast

Jackenmen:
This can be useful to auto-record sensor change time in the app that receives the bundle whenever such change happens.

See https://github.com/nightscout/AndroidAPS/pull/3092 for an example of usage.


jamorham (approved):
Seems reasonable, looks safe.

PR #3220, opened 2023-12-02
G7 cleanup

Navid200:
Correcting some notes to show G7 when G7 is being used.
Adding G7 to the menu items that correspond to G7 as well.


Navid200:
Screenshot_20231202-162601

Screenshot_20231202-162917

Navid200:
Screenshot_20231206-094356

This is tested.

jamorham:
Overall I think this is useful but I don't really like the logic too much because I think you're going to potentially have problems in the transition between when the sensor is active before and after firmware information has been read. Can you modify `isDeviceG7()` to simply check for 4 digit `tx_id` ?

Navid200:
Can I open another PR to change "G6/Dex1 Support" to an array allowing 3 options (G5, G6/Dex1, and G7)?
Then, instead of looking for the number of characters in the transmitter ID, we can look at that setting.

I will test transitions and everything.
After that is merged, this PR will become simpler.

Navid200:
Consider that in the future, we may have other devices that use a 4-digit transmitter ID.

Navid200:
I did the change.  
If you approve adding a specific selector for G7, I will then replace these with conditionals on the selector.
But, for now, this should be good.

Navid200:
Screenshot_20231202-162917

Navid200:
Screenshot_20231217-224608

Navid200:

Navid200:
Tested with a G7.
Looks good.
Everything looks as shown in the screenshots above.

jamorham (approved):
Looks like it should work other than if AAPS is fussy about the data source name but if so hopefully that can be resolved.

app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutBatteryDevice.java
```diff
@@ -143,6 +145,9 @@ boolean alwaysSendBattery() {
         @Override
         String getDeviceName() {
             if (Ob1G5StateMachine.usingG6()) {
+                if (shortTxId()) { // If using G7
+                    return "G7 Device";
```
DiBochev, line 149: change to "G7 Transmitter";

Navid200: G7 transmitter and sensor are together.  There is no reason to call it a transmitter any longer.  It is valid to call it a device to avoid calling it a transmitter.

PR #3225, opened 2023-12-07
Add a 12-hour notification to sensor expiry

Navid200:
12 hours has a significance for G7.
It's useful for someone who likes to start their sensor about the same time every 12 hours considering that the sensor runtime is 10 days + 12 hours.

After adding 12 hours, a 9-hour option seems not the best option.  So, I moved it to 6 hours.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/alert/SensorExpiry.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/alert/SensorExpiry.java
@@ -27,7 +27,8 @@ public class SensorExpiry extends BaseAlert {
     private static final long[] THRESHOLDS = {
             // need to be in ascending order so first hit is first applicable to avoid multiple triggers
             Constants.HOUR_IN_MS * 2,
-            Constants.HOUR_IN_MS * 9,
+            Constants.HOUR_IN_MS * 6,
+            Constants.HOUR_IN_MS * 12,
             Constants.HOUR_IN_MS * 24,
     };
 


jamorham:
I'm not sure I understand what you're getting at with regards to someone starting the sensor at the same time if there is 12 hour at the end of the session?

I think I didn't choose 12 hours as a time block the first time around was to maybe try to avoid things like the alert triggering in the middle of the night, but maybe that wouldn't work anyway. It is supposed to react somewhat to the user doing things and alert after it becomes due.

Navid200:
I am putting together a page in the guide explaining the different options one has for using the 12 hour grace period.  But, this PR is not about that.

Right now, a G7 user will get a notification if their screen is unlocked at 9 hours before sensor expiry.

Please let me explain one of the three options for using the 12 hour grace period.
The user always inserts their sensor at about 9:00 am.
When they do that, the sensor expiry is in 10.5 days, which will be at 9:00 pm.  
If this is merged, in 10 days, they will get a notification at 9:00 am.  So, it reminds them to insert their next sensor.

As long as this is not merged, they will get a notification when they have been 3 hours late at 12:00 pm, 9 hours to sensor expiry.

Navid200:
My understanding is that this is not an alert, it is a notification.  The alert only makes a sound if the screen is unlocked.
So, it should make no sound if one is sleep.

Navid200:
The scenario I have explained allows a new sensor to work for 12 hours before the user starts using it.  So, it is like pre soaking for 12 hours.  In this scenario, the user uses 9 sensors in 90 days.
But, the sensors always have a 12-hour overlap.

jamorham (approved):

PR #3229, opened 2023-12-09
Add IPv6 support for link local address.

tzachi-dar:
This is achieved by adding the interface name to the IP addresses.

Fix the host addresses returned for LocalLink and sitelocal ipv6 addresses. The address returned for them should
be in the format IPv6-address%scope_id.

See https://docs.oracle.com/javase/8/docs/api/java/net/Inet6Address.html

Still needs more testing.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/Mdns.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/Mdns.java
@@ -23,9 +23,14 @@ import com.eveningoutpost.dexdrip.xdrip;
 import com.github.amlcurran.showcaseview.ShowcaseView;
 import com.github.amlcurran.showcaseview.targets.ViewTarget;
 
+import java.net.Inet4Address;
+import java.net.Inet6Address;
 import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.net.SocketException;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
+import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


jamorham:
Let me know if you think it is ready for merge? Will it break any existing behavior?

tzachi-dar:
It should not break any existing behavior as the change is limited to ipv6 with link local addresses which don't work before the change.
Still under testing. So far all is working well. Will give it a few more days and then we can merge.

jamorham:
Ok, just let me know when ready. thanks

app/src/main/java/com/eveningoutpost/dexdrip/utils/Mdns.java
```diff
@@ -294,15 +331,40 @@ public void onResolveFailed(NsdServiceInfo serviceInfo, int errorCode) {
...
 
                 final InetAddress host = serviceInfo.getHost();
-                final String address = host.getHostAddress();
-                UserError.Log.d(TAG, serviceInfo.getServiceName() + " Resolved address = " + address);
+                final String address = getFixedHostAddress(host);
+                UserError.Log.d(TAG, serviceInfo.getServiceName() + " Resolved address = " + address );
                 final String short_name = shortenName(serviceInfo.getServiceName().toLowerCase());
                 if (!address.contains(":") || (iplookup.get(short_name) == null) || (JoH.msSince(iplookup.get(short_name).received) > 60000)) {
-                    iplookup.put(short_name, new LookUpInfo(address, JoH.tsl(), serviceInfo));
+                    iplookup.put(short_name, new LookUpInfo(address, JoH.tsl(), serviceInfo ));
```
tolot27, line 367: There is no need to add a space before the last two brackets, also not two lines above.

tolot27: It looks like still no common code formatting is established in this project and many parts of the code is inconsistently formatted (i.e. function `getFixedHostAddress`).

jamorham:
@tzachi-dar ok for merge?

tzachi-dar:
Sure, you can merge this.

PR #3237, opened 2023-12-12
G7 sensor 

Navid200:


Navid200:
If some day a G8 comes out, we will not need to revisit this.

Navid200:
I am running this for my active G7 now.

Navid200:
That does not stop the sensor.  It only recognizes the sensor to have stopped.

PR #3245, opened 2023-12-18
Incorrect G7 note to start sensor

Navid200:
Right after first connectivity to a G7 device after transmitter days becomes known, xDrip shows the following note before full connectivity has been established:

*Now start your sensor or check Settings*

If xDrip knows transmitter days for a G7, it means that the sensor has already started.  There is never a need to start a G7 sensor except for troubleshooting if xDrip fails to automatically start.
Under normal circumstances, you never need to tap on start sensor for a G7.

This PR continues to show the following note instead until full connectivity after which warmup period is shown.

*Verify settings and wait for connectivity.*


jamorham:
As mentioned in another PR, I will expose `shortTxId()` so you can use that instead of repeating the logic.

Navid200:
G6 will be verified before tomorrow.

Navid200:
I will verify this in 4 days.

Navid200:
I have tested this now. 
It does what it is supposed to.  The correct note is shown on the main screen for G7.

Navid200:
I'm using shortTxId now.  Is anything else missing in this?

PR #3246, opened 2023-12-18
G7 Classic Status page cleanup

Navid200:
There are three items on the classic status page that are not useful for G7.
The reason is that the G7 Bluetooth name is not related to the transmitter ID as it was with G6.

How many people are tapping on forget device thinking that they are unpairing the transmitter?
I did!

If some day, we can figure out a workaround, we can bring them back.
But, for now, please approve removing them.

Screenshot_20231218-174712


jamorham:
So in principle this is a good change I think but I would like you to use some better helper methods to declutter things and avoid duplication.

I will make a change for `shortTxId()` to be public static, so just use that instead of repeating the check in your own code.

Additionally I will include support for `showIfTrue` in to that layout, so you can do: app:showIfTrue="@{ms.using_g7()}" to avoid manually handling the visibility.

Navid200:

Navid200:
Tested.  Looks OK.
The static page looks as shown above.

Navid200:
I have made the change you asked for and I have tested it.
This is ready for merge.

app/src/main/java/com/eveningoutpost/dexdrip/ui/MicroStatusImpl.java
```diff
@@ -28,12 +31,20 @@ public String gs(String id) {
...
     public boolean xmitterBattery() {
         return DexCollectionType.usesClassicTransmitterBattery();
     }
+
+    @Override
+    public boolean using_g7() { // True if we are using G7
+        if (DexCollectionType.getDexCollectionType() == DexCollectionType.DexcomG5 && Pref.getBooleanDefaultFalse("using_g6") && shortTxId()) { // If using G7
+            return true;
+        }
+        return false;
```
DiBochev, line 48: return DexCollectionType.getDexCollectionType() == DexCollectionType.DexcomG5 && Pref.getBooleanDefaultFalse("using_g6") && shortTxId()

app/src/main/java/com/eveningoutpost/dexdrip/ui/MicroStatus.java
```diff
@@ -12,4 +12,6 @@ public interface MicroStatus {
 
     boolean xmitterBattery();
 
+    boolean using_g7(); // True when using G7
```
DiBochev, line 15: not java style method naming. Java style method naming is camel case
also use java doc for "True when using G7"

app/src/main/java/com/eveningoutpost/dexdrip/ui/MicroStatusImpl.java
```diff
@@ -28,12 +31,20 @@ public String gs(String id) {
     }
 
     @Override
-    public boolean bluetooth() {
-        return DexCollectionType.hasBluetooth();
+    public boolean bluetooth() { // Dexcom with Bluetooth except G7
+        return DexCollectionType.hasBluetooth() && !using_g7();
```
DiBochev, line 35: not sure if you are braking logic for other sensors with the extra check

Navid200: I don't know what you mean by that.  Please explain.

app/src/main/java/com/eveningoutpost/dexdrip/ui/MicroStatusImpl.java
Navid200: Thanks a lot.  Fixing it in a minute.

app/src/main/java/com/eveningoutpost/dexdrip/ui/MicroStatusImpl.java
Navid200: I see what you mean now.  
First, let me thank you for spending the time to review.

Please let me explain this concern.
There are only two possibilities under all circumstances.  Either we are using G7 or not.

If we are not using G7, what is being anded is true.  Anding true with anything changes nothing.  Therefore, nothing changes with respect to other sensors.

If we are using G7, the returned value becomes false, which removes three lines from the status page.

Navid200:
Tested again after the syntax correction.

app/src/main/java/com/eveningoutpost/dexdrip/ui/MicroStatusImpl.java
Navid200: Wow!  You have a point.  There is a third case of the value being unknown.  I need to make sure if the value is unknown, it doesn't break anything.  Thanks so much.

Navid200: @DiBochev Is it OK now?

Navid200:
I'm now using ShortTxId.  Is anything else needed?

app/src/main/java/com/eveningoutpost/dexdrip/ui/MicroStatusImpl.java
```diff
@@ -28,12 +31,21 @@ public String gs(String id) {
...
     public boolean xmitterBattery() {
         return DexCollectionType.usesClassicTransmitterBattery();
     }
+
+    @Override
+    public boolean usingG7() {
+        try { // True if we are using G7
+            return DexCollectionType.getDexCollectionType() == DexCollectionType.DexcomG5 && Pref.getBooleanDefaultFalse("using_g6") && shortTxId();
+        } catch (Exception e) { // If there are any unknowns, show that we are not using G7
+            return false;
```
DiBochev, line 48: just a suggestion maybe it is good idea to add debug log here

Navid200:
We now have a best collector option of of G7.
So, to avoid code duplication, I am using that.

I have tested this after this last change with both G7 and G6.  There is no change to the classic status page of G6.
For G7, please see the images below.

| Before | After |  
| -------- | ---- |  
| Screenshot_20240221-095004 | Screenshot_20240221-095115 |

Navid200:
The option to forget device does not work for G7.  And that is why it is important to remove it because everyone is not aware that it does not work:

Navid200:
@jamorham 
Please merge this also so that people don't think that forget device works for G7.

jamorham (approved):

PR #3262, opened 2023-12-30
Broadcast start time from G7

Navid200:
**PLEASE DON'T REVIEW**
**Please review 3263 instead**
<br/>  
  
This PR changes what is broadcasted, for sensor start time, only if a known G7 firmware is being used to the time coming from G7, which is the only time that matters as far as when the device will stop working.

I have not tested this because I cannot compile AAPS since we use a different version of gradle.

I will update if I can test.


Navid200:

Navid200:
I think this is a better approach: https://github.com/NightscoutFoundation/xDrip/pull/3263

PR #3280, opened 2024-01-07
README update to remove references to G4 and G5

Navid200:
**Outdated - Please skip this first post**  
  
---  
  
This PR has added a link to my guide on the main page.
I would also like to add a link to Patrick's.  However, I am not sure what his plans are.  It can be added later.

It also adds a link to the release notes.

It has made minor tweaks to clean things up and update to what we do today as opposed to may be a few years ago.
This is based on my understanding and I can edit it of course if wrong.  
  
---

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,10 @@
 # Nightscout xDrip+
+> Enhanced personal research version of xDrip
 
-## Guides and Documentation
-* Info page: https://jamorham.github.io/#xdrip-plus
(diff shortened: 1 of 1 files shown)


Navid200:
@jamorham This is really a benign PR.  It only updates the README file.

jamorham:
Yes we can merge this. We do still support those devices its just that they are no longer manufactured so no-one will be using them as far as I know.

jamorham (approved):

PR #3283, opened 2024-01-08
Plugin Calibration: fix getting intercept

aptash:
In case of using plugin for calibrations, bgReading.filtered_calculated_value was using the plugin's slope but not the plugin's intercept.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/BgReading.java
@@ -610,7 +610,7 @@ public class BgReading extends Model implements ShareUploadableBg {
                     Log.d(TAG, "USING CALIBRATION PLUGIN AS PRIMARY!!!");
                     if (plugin.isCalibrationSane(pcalibration)) {
                         bgReading.calculated_value = (pcalibration.slope * bgReading.age_adjusted_raw_value) + pcalibration.intercept;
-                        bgReading.filtered_calculated_value = (pcalibration.slope * bgReading.ageAdjustedFiltered()) + calibration.intercept;
+                        bgReading.filtered_calculated_value = (pcalibration.slope * bgReading.ageAdjustedFiltered()) + pcalibration.intercept;
                     } else {
                         UserError.Log.wtf(TAG, "Calibration plugin failed intercept sanity check: " + pcalibration.toS());
                         Home.toaststaticnext("Calibration plugin failed intercept sanity check");


Navid200:
We should not encourage using the plugin.
It will be nice to hear from the old timers why the plugin was originally created.

Currently, more and more CGM manufacturers offer factory calibration where the devices are already calibrated and do not need calibrations.
If such a device needs calibration, the manufacturer's scheme should be used for calibrating the device.

I wonder if a long timer ago when calibration was needed as a daily routine, we added plugins to improve accuracy.  But, at that time, calibration did not invalidate anything.
Now, if you use a plugin, you are effectively ignoring what the manufacturer has done to improve accuracy.

aptash:
As I said in the [discussion](https://github.com/NightscoutFoundation/xDrip/discussions/3174#discussioncomment-7785510):

> In our case Sibionics constantly shows 2-3 mmol below a glucose meter.

But we don't constantly calibrate it. Usually once in low end and once in high end. That's enough. It could be re-checked again in a few days.

There are some people who don't use calibration for Sibionics. But their sensors show data close to glucometers.

Navid200:
xDrip offers to calibrate raw readings.
xDrip should never offer to calibrate readings coming from a manufacturer's app.

Navid200:
xDrip offered to calibrate Dexcom G5.  Dexcom G5 transmitted raw values in addition to final values.
So, when xDrip offered to calibrate G5, it was only offered if you chose to use G5 in non-native mode meaning you chose to use the raw values.

Or, you could choose to use xDrip in native mode.  In that case, calibrations would be sent to the transmitter.  So, the transmitter did the calibrations.
So, xDrip never offered to calibrate final readings coming from a G5.

Libre 1 was not even a CGM.  It missed the C in the CGM.
There were manufacturers that created devices that would be attached to Libre 1, like miaomiao, that scanned it and transmitted readings.  
I don't believe Libre 1 readings were officially approved to be used for dosing.
Therefore, calibrating it would not have been as bad as calibrating readings coming from a Decom G6.

A CGM that is in the market today that is approved to be used for dosing, should not be calibrated.
If we do that with xDrip, we will be crossing some lines that should not be crossed.

jamorham:
On the face of it, this fix looks correct but I expect it has been overlooked because the filtered value is not really used very much. Can you explain what problem you are trying to solve with this? @aptash

aptash:
Thanks a lot for your attention!

@jamorham sometime ago I created the discussion https://github.com/NightscoutFoundation/xDrip/discussions/3174 and currently I'm trying to solve these problems in case of using a plugin for calibration: 
- backfilled data are sent as raw-data to aaps
- xDrip sends not calibrated data to Nightscout

During the debugging these problems (with NS and sending backfilled data to aaps) I noticed a bug and decided to create this PR.

I know the opinion about calibration and factory calibrations but anyway half of people (we made a poll) are using calibration for Sibionics CGM just to correct the level of data.

image

jamorham:
@aptash I would deal with those other issues separately. So just to confirm this is just from looking at the code and isn't intended to fix any known issue?

aptash:
> So just to confirm this is just from looking at the code and isn't intended to fix any known issue?

Yes. Exactly.

jamorham (approved):

PR #3295, opened 2024-01-16
No battery options for G7

Navid200:
Since we don't want to offer the option to customize the battery warning level for G7, this PR removes the setting.

The title of the category has been updated to exclude G7.

It has been tested with G6 and G7.


Navid200:
Using the new best collector G7 setting.

The logic is negative.  But, that's OK.  
It is very unlikely if a new Dexcom device comes along in the future, we would want to allow user to edit the battery trigger levels anyway.

Tested after the latest update with G7 and G6 and looks good.
The following shows what you see for a G7.

| Before | After |  
| ------- | ------ |  
| Screenshot_20240221-102938 | Screenshot_20240221-102855 |

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -1731,6 +1732,12 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
                 }
             }
 
+            if (getBestCollectorHardwareName().equals("G7")) { // Remove battery options, from G5/G6/Dex1/G7 Debug Settings, if we are using G7 or One+
```
jamorham, line 1735: Really good idea to wrap this with a try{} catch{} pair and log at wtf level if an exception is generated, otherwise future changes to preferences structure can cause crashes.

Navid200: Sure, thanks.  It will be done.

Navid200: Done

jamorham (approved):

Navid200:
Thanks

PR #3297, opened 2024-01-19
Ascending volume to medium instead of maximum

Navid200:
Edit (April 11, 2024):  
This is now about more than just adding 3 subpages.  It now also adds the new setting to allow ascending volume profile to ascend to medium instead of high.  
  
---  
  
This PR creates 3 subpages on the Glucose Alert Settings page.
There are already many settings on the page.  I would like to add one more.  There will be too many settings.
Grouping the settings such that settings that are strongly tied together on the same subpage would allow us to reduce the number of settings while allowing the user to easily find the settings they need.

| Before | After |  
| ------- | ------ |  
| Screenshot_20240119-152722 | Screenshot_20240119-152408 <br/><br/> <br/>  <br/>  <br/>  <br/>  <br/>  <br/>  <br/>  <br/>    |


Navid200:
In this image, I have circled the settings that are being grouped together.
Screenshot_20240119-152722
  
The third one has only one setting in it.  But, that's the one that will have a second setting after/if this PR is merged.

Navid200:
If we create too many sub pages (too much hierarchy), it could make things worse as far as helping user find what they are looking for.  So, we should not overdo it.  

Here, I am only placing settings on a sub page that should not be needed by anyone on a daily basis.  These are settings that most users spend some time tweaking until they are happy and hopefully after that, they never touch them again.

Secondly, the settings that are placed on the same subpage are either dependent on each other, or are strongly related to each other.

Navid200:
And these are the three sub pages.
Screenshot_20240215-000604  
  
Screenshot_20240215-000631  
  
Screenshot_20240215-000652

Navid200:
I have added the option to ascend to medium instead of maximum if the ascending volume profile is selected.
By default, this new setting is disabled resulting in the ascending volume profile acting as it did before.

Enabling the new setting does not change the slope.  So, the ascending speed is not change.  Only the value is clamped at the medium level if the new setting is enabled.

Navid200:
I have tested this by monitoring the detail logs and verifying that the volume goes through the expected steps in both cases when the setting is enabled or not.

Navid200:
The request for this has been made by someone who may occasionally take a while to walk to the phone to snooze the alert.
In the meantime, the ascending volume has reached the max volume annoying everyone else.
They suggested that it may be better if it ascended to medium instead.

When I planned this, I could have taken two different approaches.  
I could have changed the max to medium, which would have lowered the slope.  I intentionally avoided that because I did not want to slow down the ascension.  So, I decided to just clamp it.

jamorham (approved):
Seems reasonable

PR #3303, opened 2024-01-21
Last connected Dex status page English only

Navid200:
Please lets have this in English only the same as everything else on the same page as well as all the logs, so that the support staff can do their job without having to constantly look for a translator.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/JoH.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/JoH.java
@@ -780,6 +780,33 @@ public class JoH {
         return qs((double) t, 0) + " " + unit;
     }
 
+    public static String niceTimeScalarEn(long t) { // English only - Please do NOT translate
+        String unit = "second";
+        t = t / 1000;
+        if (t != 1) unit = "seconds";
+        if (t > 59) {
+            unit = "minute";
+            t = t / 60;
+            if (t != 1) unit = "minutes";
+            if (t > 59) {
+                unit = "hour";
+                t = t / 60;
+                if (t != 1) unit = "hours";
+                if (t > 24) {
+                    unit = "day";
+                    t = t / 24;
+                    if (t != 1) unit = "days";
+                    if (t > 28) {
+                        unit = "week";
+                        t = t / 7;
+                        if (t != 1) unit = "weeks";
+                    }
+                }
+            }
+        }
+        return qs((double) t, 0) + " " + unit;
+    }
+
     public static String niceTimeScalar(double t, int digits) {
         String unit = xdrip.getAppContext().getString(R.string.unit_second);
         t = t / 1000;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
@@ -18,6 +18,7 @@ import static com.eveningoutpost.dexdrip.g5model.Ob1G5StateMachine.usingAlt;
 import static com.eveningoutpost.dexdrip.models.JoH.emptyString;
 import static com.eveningoutpost.dexdrip.models.JoH.msSince;
 import static com.eveningoutpost.dexdrip.models.JoH.niceTimeScalar;
+import static com.eveningoutpost.dexdrip.models.JoH.niceTimeScalarEn;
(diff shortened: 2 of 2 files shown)


Navid200:
This is an example of what we see in the facebook group:
420175018_10163676423846515_6389554714124573967_n

So, imagine you see 27 and something that is a unit in another language.  Now, is that 27 seconds or 27 minutes.
Last connected 27 seconds ago is great.  Last connected 27 minutes ago is bad.

Navid200:
The following shows my test result after this for the same language I believe.
Screenshot_20240120-225024

Navid200:
I don't care as much about when the battery weas last checked or the number of days on the sensor.
But, last connected, please let me have it in English.

Navid200:
It's not just the time it takes to look for translations.  It could lead to mistakes also.

app/src/main/java/com/eveningoutpost/dexdrip/models/JoH.java
```diff
@@ -780,6 +780,33 @@ public static String niceTimeScalar(long t) {
...
+            if (t != 1) unit = "minutes";
+            if (t > 59) {
+                unit = "hour";
+                t = t / 60;
+                if (t != 1) unit = "hours";
+                if (t > 24) {
+                    unit = "day";
+                    t = t / 24;
+                    if (t != 1) unit = "days";
+                    if (t > 28) {
```
DiBochev, line 799: why t > 28?! should it be t > 7 
Please check if this is an error

DiBochev:
Just an idea to optimize the code:
`public static String niceTimeScalarEn2(long t) { // English only - Please do NOT translate
        String unit = "second";
        t = t / 1000;
        if (t > 59) {
            unit = "minute";
            t = t / 60;
            if (t > 59) {
                unit = "hour";
                t = t / 60;
                if (t > 24) {
                    unit = "day";
                    t = t / 24;
                    if (t > 28) {
                        unit = "week";
                        t = t / 7;
                    }
                }
            }
        }
        if (t != 1) unit += "s";
.....`

app/src/main/java/com/eveningoutpost/dexdrip/models/JoH.java
Navid200: Even though it may look like I am adding this code, I am not.
I have copied an existing code, removed translation strings from it, and replaced them with English text.

Your questions should be posed with respect to the original existing code here: 
https://github.com/NightscoutFoundation/xDrip/blob/8da5e5d094478e266011f60ac4041042e0178754/app/src/main/java/com/eveningoutpost/dexdrip/models/JoH.java#L783-L809  

If your questions and suggestions are accepted, the original code as well as my copy can be modified.
If your suggestions are not accepted for the original code, I don't see why they would only apply to my copy.

Navid200: Thanks

The lead developer will see your review and my response to it.  Hopefully he can make a decision.

Thanks to your comment, I realized that I needed to clarify that I am just copying an existing code and am not creating code from scratch.  So, I appreciate your review.

Navid200:
@jamorham The translation work that translated time units is here: https://github.com/NightscoutFoundation/xDrip/commit/32a9cba255b9cf31486123f61744a7e408659fe8
It was added in 2018.
I doubt that it was done just to translate the time units on the status page.  

I didn't want to undo it because it is used everywhere.  I don't want to change the time units to English everywhere.  I only want to do that on the status page.
So, I copied the code that has all the strings in it and added En to it to reflect English and changed all the string references to English text and I used it only for the status page.

If you want me to do this differently, please tell me.  But, parameterizing this will be a lot of work.  is it necessary?  I will do it if that is the only way.

Navid200:
@jamorham This takes time:
433438067_10228921122345633_5257209364848274046_n
  
I would like to help people.  But, this is not helping me.
Would you please review this PR?

Navid200:
@jamorham The issue this PR will address is really slowing me down as a customer support staff.

Would you please have a look?

Navid200:
If this is unacceptable, which would be a shame because I then wish to know how come we don't translate the logs.   I think we don't because you would like to be able to see the logs in English.  For the same reason, I need to be able to see the value of last connected in English.

If you accept the intent, but don't like how I have done it, please tell me what you don't like.

If you don't like the intent, can I change the color of the text if it is 5 minutes or greater?  But, this will eventually become a mess if one reading a minute some time in the future becomes the norm.
So, I really hope you accept the intent.

PR #3318, opened 2024-01-27
Correct warmup time even before firmware is known

Navid200:
I relied on firmware, in this PR https://github.com/NightscoutFoundation/xDrip/pull/3195, to show the correct warmup time for a G7.
Today, I noticed that it still showed 2 hours for warmup right after connecting to G7.  
Screenshot_20240126-093424

After one more read cycle, it showed the correct value.
I suspect the reason is that Firmware is not known right away after connectivity.

This PR changes the condition to rely on what the user has chosen instead of relying on the firmware.

I will be testing this in 10 days.


Navid200:
Did not connect in time.  I will try to test again in 10 days.
Screenshot_20240205-103140

Navid200:
Looks good:
Screenshot_20240215-084451

Der-Schubi:
I can confirm, works fine!

Navid200:
Redundant!

PR #3319, opened 2024-01-27
Offer to stop a G7 in engineering mode only

Navid200:
We have already started to see posts of people asking for help who have stopped a G7.  
Under normal circumstances, there is no reason to ever stop a G7.

This PR will not show the stop option in the main menu if we are using a G7 unless engineering mode is enabled.

I have tested this with a G7 and verified that enabling engineering mode shows the stop option and disabling engineering mode removes the stop option.
I have tested it with a G6 and confirmed that the stop sensor option is available regardless of engineering mode.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
@@ -103,7 +103,7 @@ public class NavDrawerBuilder {
                         }
                     }
                 }
-                if (getBestCollectorHardwareName() != "G7" || Pref.getBooleanDefaultFalse("engineering_mode")) { // If we are using G7, offer the stop sensor option in engineering mode only
+                if (!getBestCollectorHardwareName().equals("G7") || Pref.getBooleanDefaultFalse("engineering_mode")) { // If we are using G7, offer the stop sensor option in engineering mode only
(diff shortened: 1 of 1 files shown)


Navid200:
This is an example of a post that worries me about xDrip offering to stop a G7.  
<br/>  

Screenshot 2024-01-27 085008

app/src/main/java/com/eveningoutpost/dexdrip/NavDrawerBuilder.java
```diff
@@ -101,8 +103,10 @@ public NavDrawerBuilder(final Context context) {
                         }
                     }
                 }
-                this.nav_drawer_options.add(context.getString(R.string.stop_sensor));
-                this.nav_drawer_intents.add(new Intent(context, StopSensor.class));
+                if (getBestCollectorHardwareName() != "G7" || Pref.getBooleanDefaultFalse("engineering_mode")) { // If we are using G7, offer the stop sensor option in engineering mode only
```
jamorham, line 106: You should use `equals()` for string comparisons as otherwise it is comparing the object reference rather than the content.

Navid200: Thanks for the guidance.
Done

Navid200:
Tested again after correcting the syntax.
Tested with G7 and G6.

jamorham (approved):

Navid200:
Thanks

PR #3333, opened 2024-02-10
Add SecurityException to IOException (needed for android 14).

tzachi-dar:
People have reported that they are getting SecurityException when scanning a libre2 device.

Looking into the code it seems that our code is correct, but we just need to catch this exceptions and retry.


jamorham:
I'm not sure under which circumstances a security exception will clear on a retry but there are lots of inconsistencies in android 14 so we can try this and see if it helps.

jamorham (approved):

Issue #1331, opened 2020-05-05
AAPS Extra statusline is broken


FreDiabetics:
Every time xDrip receives that my loop is paused 
of currently waiting. It cuts the statusline in xDrip. 

The last chars are missing if 'loop deactivated' is shown.

Is there anybody else with this issue? 

-  **Loop deactivated: the last 7 characters are missing _-(spaces included)_**
![Screenshot_20200506-003253_xDrip+](https://user-images.githubusercontent.com/45721291/81122922-3d6e5880-8f32-11ea-9f34-ed812ba116e6.jpg)

-  **Loop activated**
![Screenshot_20200506-003915_xDrip+](https://user-images.githubusercontent.com/45721291/81122966-54ad4600-8f32-11ea-95e0-fd39460743c1.jpg)

wagnefrede:
Nobody wanna help?!?!

Navid200:
@FreDiabetics Would you please ask using the support channels to see if anyone else is experiencing the same behavior?
https://www.facebook.com/groups/xDripG5
https://gitter.im/jamorham/xDrip-plus
Thanks

Navid200:
@tolot27 Is this a layout issue?

tolot27:
> @tolot27 Is this a layout issue?

Yes, it is.

Navid200:
@FreDiabetics What is the complete word if it did not have 7 characters missing?

olorinmaia:
I have found a solution and making a PR to xDrip. Tested OK with latest AAPS with localization that has most letters.
Need to increase allowed text length from 40 to 70 to be sure. 
Example of the longest statusline from AAPS I could produce which is 64 letters, 70 should be fine.
Uzavretý okruh deaktivovaný
0,40 u/h 0,42U(0,43|-0,00) -0,14 29g

Navid200:
This should be fixed as of the February 20, 2024 release.  If not, please let me know.

Thanks @olorinmaia for fixing it.

PR #3348, opened 2024-02-18
Increase max text length for AAPS statusline

olorinmaia:
Solves https://github.com/NightscoutFoundation/xDrip/issues/1331

This change make sure all languages get all info through from AAPS to xDrip+ statusline without being cut short. 

Example of the longest statusline from AAPS I could produce which is 64 letters, 70 should be fine.
Uzavretý okruh deaktivovaný
0,40 u/h 0,42U(0,43|-0,00) -0,14 29g

Before:
image

After:
image


Navid200:
No way!  You fixed it already?!
Well done!
Thanks

jamorham (approved):

Issue #3243, opened 2023-12-17
Remove duplicate Norwegian language in Crowdin and xDrip+


olorinmaia:
I'm language coordinator for "Norwegian" and "Norwegian Bokmal" in AAPS and xDrip+ in Crowdin.

We had a similar issue in AAPS with duplicate Norwegian language entries in Crowdin. The correct language to translate is "Norwegian bokmål", as this is the default system language in Android for Norwegian phones. There is only one written language in Norway that is broadly used. I think Norwegian was added to crowdin by mistake, like with AAPS.

**Thus "Norwegian" https://crowdin.com/project/xdrip/no (45% translated,) can be removed/deleted in Crowdin and corresponding language choice in xDrip+ can also be removed.**

The choice for Norwegian bokmål in xDrip+ can be renamed from "Bokmal" to "Norsk bokmål" in language settings which should point to https://crowdin.com/project/xdrip/nb (100% translated in crowdin now :) )

Referring to conversation with Mathis Walter @ crowdin :)

![image](https://github.com/NightscoutFoundation/xDrip/assets/114103483/bcfceb2c-c1f6-41f6-8296-799dd17aad2c)

Navid200:
Someone has added the language that you want to delete.
Why are they wrong and someone else is right?
What would be the problem if you allow this language to stay?  It will not stop anyone from using the language you want to keep, does it?

Navid200:
Is this what you want to be deleted?
![Screenshot 2023-12-17 174314](https://github.com/NightscoutFoundation/xDrip/assets/51497406/378d5fad-05f2-4263-b293-0f25830c0a33)

<br/>  

If yes, you can see there have been updates to it on October 8, only 2 months ago.
I don't think we (xDrip) should get in the middle of ethnic disagreements.  I prefer that we find a solution that does not involve deleting someone else's hard work.

Navid200:
Would you accept the following as a solution?

We change Bokmal to Norsk (Bokmal)?  And we change Norsk to Norsk (Nynorsk)?

And this is conditional on the other parties agreement.  So, we need to reach out to the others and get their agreement as well before we go ahead.  But, would you accept this?

olorinmaia:
[...]

PR #3349, opened 2024-02-18
Remove duplicate Norwegian language and add legacy settings update to transfer to correct language

olorinmaia:
This is my attempt to fix first part of https://github.com/NightscoutFoundation/xDrip/issues/3243

This will remove "Bokmal"-option in Language-preferences and change so that "Norsk" points to "nb"-language in Crowdin (100% translated). It will also transfer users who have forced "no"-language over to "nb"-language.

I have built and verified that this works as expected locally. Please advise if there is something I haven't thought of.

What I tested: Forced language = "no" before updating. Then update to this PR and verified that I was moved to the correct "nb"-translations for Norwegian.

Second part removing files from /res folders and "no"-language from Crowdin I need some assistance from @jamorham or  @tolot27


jamorham (approved):
Thanks for doing this

Issue #1196, opened 2020-01-22
Treatment syncs between master and follower (and viceversa)


Foxandxss:
I have a master xDrip with miaomiao uploading to nightscout and also a follower that gets the data from nightscout.

If any of those two creates a new treatment, say 5 units of insulin, I can see the bolus on the phone that was created and also at nightscout. The other phone says there is no IOB and the green curve at the bottom is not displayed.

It is bad because I am the follower and I am the one that enters the boluses and my GF is always asking me: "Hey, do I have insulin?" She should be able to know as well.

Not sure if a bug, a missing feature or I am doing something wrong.

Thanks

PS: Adding a treatment from nightscout seems to appear on the phones. I removed it afterwards but xDrip didn't notice.

jamorham:
One of the NightScout Follower options is to also download treatments. Check that this is enabled on both devices? I would expect it to work but if not I would say it is better to use the native xDrip+ sync feature to synchronize between the handsets.

Foxandxss:
It is enabled in both devices. If not, adding directly at nightscout wont update the phone and it does. It is just from master to follower (or viceversa) via xDrip. Native sync is supposed to be enabled. I say supposed because it is a complicated thing.

tolot27:
@jamorham Please assign this issue to me.

Navid200:
@Foxandxss Is this still happening with the latest xDrip Nightly?

tolot27:
This problem is not solved because #1403 is still a draft PR.

Cagier:
Does anyone know if there is an earlier version of xDrip+ that does not have this issue? I would happily downgrade the follower phone to overcome this issue as we cannot see the treatments from school otherwise.

Navid200:
As of the release on February 24, 2024, there is a new switch in the "Upload Treatments" submenu titled "Skip items from xDrip" enabled by default.
If you disable this switch, treatments uploaded by another xDrip will also be downloaded.

Please tell us if you experience any issues or have any questions.

Thanks for reporting this.

PR #3357, opened 2024-02-22
Download treatments originated from xDrip as well

Navid200:
We have two open issues related to the transfer of treatments to and from followers:
https://github.com/NightscoutFoundation/xDrip/issues/735
https://github.com/NightscoutFoundation/xDrip/issues/1196

This PR creates a workaround:

This PR will allow a follower to download treatments uploaded to Nightscout by the master.
It also allows the master to download treatments uploaded by the follower.

1- In order for the follower to be able to upload to Nightscout, `Settings` &#8722;> `Cloud Upload` &#8722;> `Nightscout Sync` &#8722;> `Enabled` must be set.
Also, `Settings` &#8722;> `Cloud Upload` &#8722;> `Nightscout Sync` &#8722;> `Extra Options` &#8722;> `Upload treatments` must be enabled.

2- For the master to download treatments, `Settings` &#8722;> `Cloud Upload` &#8722;> `Nightscout Sync` &#8722;> `Download treatments` must be enabled.  
<br/>    
  
---  
  
#### **Tests**  
I have done a test with two test phones set up as a master and a follower confirming that treatments can be uploaded from either phone and the other phone would download it as well after this PR.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutTreatments.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutTreatments.java
@@ -130,7 +130,7 @@ public class NightscoutTreatments {
                     Treatments existing = Treatments.byuuid(nightscout_id);
                     if (existing == null)
                         existing = Treatments.byuuid(uuid);
-                    if ((existing == null) && (!from_xdrip)) {
+                    if (existing == null) {
                         // check for close timestamp duplicates perhaps
                         existing = Treatments.byTimestamp(timestamp, 60000);
                         if (!((existing != null) && (JoH.roundDouble(existing.insulin, 2) == JoH.roundDouble(insulin, 2))


Navid200:
This may be considered a conflict as how the follower needs to be set.  It's hardware data source must be set to Nightscout follower.  But, in order for it to be able to upload treatments to Nightscout, Nightscout sync also needs to be enabled.

One may wonder that the readings downloaded by the follower when then be uploaded again to Nightsocut.
My understanding is that Mongo will address such collisions.
So, there will be no issue on Nightscout.

If you ask me the following question, I cannot answer and will need to investigate if needed:
Will the follower use data to download and then again use data to upload the same data back to Nightscout?  is that not a waste of data? 

If this indeed is a concern, we will need to add a conditional to only upload treatments from a follower and not readings as well.

jamorham:
I would suggest making it a preference switch for advanced option with defaulting to off so it can be merged for wider testing but I would think there could likely be a good reason for that duplicate checking being in there.

Navid200:
That makes perfect sense.  Thanks

Navid200:
Tapping on download treatment takes you to a new submenu.  There, you have the old download treatment checkbox.  But, in addition to that, you will also see a new switch called skip items from xDrip.  This new switch is enabled by default, which makes xDrip behave as it did before.

If you disable this switch, xDrip will also download treatments that have been uploaded to Nightscout by another xDrip.

I hope this is acceptable.
<br/>  
  
---  
   
#### **Tests**  
I have tested again two test phones acting as master and follower.  Leaving the new switch enabled (default), xDrip behaves as before and does not download treatments from the other phone.
Disabling the new switch, either phone downloads treatments uploaded by the other phone.

Navid200:
I have added the ability to set/unset the new setting in the follower menu as well.

| Before | After |  
| ------- | ------ |  
| Screenshot_20240223-130418 | Screenshot_20240223-125925 |

Navid200:
And this is the follower submenu:
Screenshot_20240223-130650

jamorham (approved):
looks good

Issue #266, opened 2018-01-12
Upload of "eventType: Sensor Start" to the Nightscout treatments works only to 1st NS site 


bordeb:
When starting new dex sensor it seems that, if there is more than one Nightscout sites (REST-API) in the Base URL settings (space delimited), the upload of the "eventType: Sensor Start" works only for the 1st (first) NS site in a row. All other NS sites are missing this data. I tried to confirm this with fresh setup of multiple NS sites and clean install of the xDrip+ app and can confirm this issue. I also tried to reorder the NS sites under Base URL settings and it is always only the first one in a row that receives the while all others are missing the "eventType:" "Sensor Start" data.
xDrip + app ver latest stable: 2017-11-13 
NS ver latest: 0.10.2-release-20171201

jamorham:
Can you try this again in the latest nightly? March 6th.. let me know if I have resolved it. Thanks

bordeb:
Thank you @jamorham ! At the moment the applied sensor is almost new (3 days from application) so I will have to wait to put new one to test this but I will report it here for sure. 
Is there any way to "emulate" "fake" some virtual sensor replacement just for test with some other spare phone? I could do that easily if I could only now how to achieve such "testing virtual environment"?

bordeb:
Hi @jamorham , I have just tried with nightly 6th March and it seems that the issue is still not resolved. The behavior that I can see is still the same.

bordeb:
HI! I just want to shed some attention to this old  issue. I am still experiencing this issue. With dexcom as well as with libre sensors.

Navid200:
@bordeb Is the problem still there with more recent releases of xDrip and Nightscout?

Navid200:
Closing due to inactivity

jonprogers:
This issue is still happening as reported here: https://github.com/NightscoutFoundation/xDrip/discussions/2299#discussioncomment-3842822 and here https://github.com/NightscoutFoundation/xDrip/discussions/2358#discussioncomment-3939664

@Navid200 asked me to transfer the information to this thread.

Where there are two space separated URLs specified as the Base URL in xDrip settings;

- glucose values (type sgv) are uploaded to both nightscout instances
- CGM calibration events are uploaded to both nightscout instances
- other treatments and events (for instance I've experienced it with:- carbs, insulin, CGM sensor start/stop) only go to the first URL specified

[...]

PR #3362, opened 2024-02-25
Upload treatments to multiple Nightscout servers

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/266

We have had an open issue (see above link) for 6 years where sensor start/stop or treatments are only uploaded to the first server if we have more than one.

This PR fixes that.
I have tested entering treatments and starting and stopping and deleting treatments.  they all are reflected on both servers after this PR.

However, I must say that I am not closing the queue.  I have commented it out.
I am sorry I am not sure the negative side effect of that if any!
So, please review this taking my inexperience into account.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutUploader.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutUploader.java
@@ -807,12 +807,12 @@ public class NightscoutUploader {
                                     if (!r.isSuccessful()) {
                                         throw new UploaderException(r.message(), r.code());
                                     } else {
-                                        up.completed(THIS_QUEUE);
+                                        //up.completed(THIS_QUEUE); // We may need this if there are more URLs
                                         Log.d(TAG, "Success for RESTAPI treatment delete: " + up.reference_uuid + " _id: " + this_id);
                                     }
                                 } else {
                                     Log.wtf(TAG, "Couldn't find a reference _id for uuid: " + up.reference_uuid + " got: " + this_id);
-                                    up.completed(THIS_QUEUE); // don't retry
+                                    // up.completed(THIS_QUEUE); // don't retry // We may need this if there are more URLs
                                 }
                             }
                         } else {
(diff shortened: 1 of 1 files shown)


jamorham:
I haven't fully reviewed these code paths but this looks like it would create infinitely repeating uploads

Navid200:
We have a main (outside) loop that goes over all URLs here: https://github.com/NightscoutFoundation/xDrip/blob/6b152b08aeb82b09e02b03dd93e0c77ee8ba94ef/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutUploader.java#L467  
<br/>  

Every time we go through this loop (every URL), this line is executed:  https://github.com/NightscoutFoundation/xDrip/blob/6b152b08aeb82b09e02b03dd93e0c77ee8ba94ef/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutUploader.java#L496 
<br/>  
  
That line uploads readings, calibrations and treatments.  But, only for treatments, there is a loop created here: https://github.com/NightscoutFoundation/xDrip/blob/6b152b08aeb82b09e02b03dd93e0c77ee8ba94ef/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutUploader.java#L769  
<br/>  
  
Therefore, this loop is inside the first loop I pointed out.  
In other words, for every URL, we have to go through this loop for uploading treatments.  
  
But, there are several instances inside this internal loop that upon fail or success, the queue is completed like this as an example:  https://github.com/NightscoutFoundation/xDrip/blob/6b152b08aeb82b09e02b03dd93e0c77ee8ba94ef/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutUploader.java#L810  
<br/>  
  
The problem is that when the time comes for the outside loop to increment to the next URL, this queue is not recreated.
That's why the treatments are only uploaded to the first URL.  
  
I removed the que terminations.  When testing, I have not noticed an infinite upload.

I admit that I have not resolved this gracefully.  But, I want this resolved so that we can close one more issue.
How can I proceed to do this right?

Navid200:
I am going to open another PR with better comments and description.

Navid200:
I have opened another PR to address the same issue.  This time, I am closing (confirming) the upload queue:  https://github.com/NightscoutFoundation/xDrip/pull/3630

PR #3366, opened 2024-02-26
Show download treatment for nsfollow only

Navid200:
In this recent PR, I replaced a setting with a screen:
https://github.com/NightscoutFoundation/xDrip/pull/3357

A setting that is only supposed to be shown on the hardware data source tab for a particular data source is managed in Preferences.java.
I failed to replace the setting with the new screen there.  As a result, now download treatment is shown regardless of what hardware data source you choose.   Even if you disable collection, it will be shown.  
Please see the following image.  
Screenshot_20240226-093216
  
It shows settings for G7.  You can see the Download Treatment option circled.  That should not be visible.  It should only be visible if Nighscout follower is chosen.  
<br/>  

This PR:
- Removes a typo underscore from the key name of the new screen.
- Replaces the setting key with the screen key in Preferences.java so that it is only shown when nsfollow is chosen.  

I have tested this.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
@@ -1256,7 +1256,7 @@ public class Preferences extends BasePreferenceActivity implements SearchPrefere
             }
 
 
-            final Preference nsFollowDownload = findPreference("nsfollow_download_treatments");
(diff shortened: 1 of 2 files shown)


Navid200:
I effectively broke xDrip.  So, it definitely was my fault.
This PR is my attempt to fix it.

jamorham (approved):
Testing testing and more testing...

PR #1686, opened 2021-03-28
Libre2 sync raw

jordipbou:
Adds synchronization of Libre2 raw values with xDrip+ Sync service. 

No options added as it uses Less Common Settings > Advanced settings for Libre 2 < show Raw values in Graph to decide if sync raw values (on master) and to show them on follower.

For implementing it a new message type has been added "l2r" to xDrip+ sync sender and receiver.

Issue #2680, opened 2023-02-15
Libre one minute testing


jamorham:
Over time we have been progressing to be able to support 1 minute readings. This work isn't complete but it is at the stage where users can test it out.

This currently only works with patched app data source but may be extended to other data sources as required.

To test this feature out:

1 - Make sure to be using engineering mode https://github.com/NightscoutFoundation/xDrip/wiki/Engineering-Mode
2 - Make sure you are using the patched app data source
3 - Go in to xDrip main settings and search for `one minute`
4 - Enable the one minute checkbox - be sure to read the associated warning
5 - Now you need to either force stop and re-open the xDrip app or reboot your phone to effect the change. The same is needed if you want to disable the feature again later.

Please observe how the app performs. Does everything work as you would expect?

Things to check:

- Do alarms work correctly?
- Does predicted low feature work correctly?
- Do trend arrows and deltas show correctly?
- Do you get missed readings?
- If you broadcast or upload the data from xDrip elsewhere what happens to these receivers?
- Try with and without the `graph smoothing` preference switch enabled.

It is expected that this change will in its current form cause some problems. Please itemize anything you notice, with screenshots (if appropriate) The idea of user testing is to identify the best way to resolve these issues. 

For example there will likely be a smoothing emitter which can receive data at any interval and then emit to services that can only accept 5 minute readings at that rate. This is work in progress still.

PhilTU:
I have noticed an apparent "duplication" of data points on the chart which I think arises from backfilling missed data overlapping recorded data points. In reality it is backfilled data points with different values to the actual data points in close proximity, the times are different so it is not truly duplicating, but on first glance at some zooms it looks like there are two lines.
![Screenshot_20230220-062330](https://user-images.githubusercontent.com/76240903/220065790-5924ce69-db12-4c66-a903-35aac8bd8805.png)
![Screenshot_20230220-075703](https://user-images.githubusercontent.com/76240903/220065797-c3c32630-989d-4646-b0d2-6db765d78c5b.png)


The backfilling looks to generate data points from the sensor history, as the latter is a 15 minute value so presumably some interpolation is used. Apologies if my interpretation is flawed.

PhilTU:
Just to add that in general this feature is working well, I share data to a MiBand via Watchdrip+ and that has worked too. Noisy data, alarms and trend arrows have so far behaved as expected. I am hazy about the backfilling feature (regardless of the time base).

JimZam:
[...]

PR #3371, opened 2024-02-28
Libre2 sync raw

jodli:
Following up on #1686.
I've fixed the commit history and solved the review requests from the original PR.

Tell me what you think and thanks for this awesome work! 💝


jamorham:
Can you explain what problem this is intended to solve? Many thanks

jodli:
From the original authors PR:

> Adds synchronization of Libre2 raw values with xDrip+ Sync service.
> No options added as it uses Less Common Settings > Advanced settings for Libre 2 < show Raw values in Graph to decide if sync raw values (on master) and to show them on follower.
> For implementing it a new message type has been added "l2r" to xDrip+ sync sender and receiver.


Basically it adds support for syncing the raw BG measurements (the small white dots) from the Libre 2/3 which transmits a new value every minute. So far it only syncs a new value every 5 minutes to the followers.

jodli:
> Ok thanks. Which data source is this to be used with? Do you know what the interactions are with the support for 1 minute readings in xDrip?

So far I only know the Libre 2 and 3 sensors that support sending new values every minute, and the data source for these has to be set to [Libre2 (patched App)](https://xdrip.readthedocs.io/en/latest/install/libre2patch/). My son is using [Juggluco]((https://xdrip.readthedocs.io/en/latest/install/libre2patch/?h=libre#juggluco)) to share the data to xDrip.

---

I've also found the [Smoothing](https://xdrip.readthedocs.io/en/latest/install/libre2patch/?h=libre#smoothing) article in the docs.
> Smoothing adds a delay in xDrip+ BG from raw data. The delay is about half the smoothing period. Set the delay accordingly to the sensor behavior for your safety. If you don't use a DIY closed loop system with xDrip+ providing glucose values, set it to 10 minutes.

This is exactly the case I'm experiencing: It only syncs a smoothed value every 5 minutes to xDriip followers.

---

> My concern this that this will generate a lot of wake ups for network traffic which will impact on battery life. The code should also at a minimum respect the follower power saving with something like a conditional of `!Pref.getBooleanDefaultFalse("plus_follower_save_power")`

I did not know about that setting. Sure, I will add it to the code.

jamorham:
Does the data source you are using support 1 minute readings? I'm not sure from your message if it does or not. I think it is important to test this with one minute readings. I thought that would disable the value smoothing. Graph smoothing is something else which doesn't affect the live data value.

jodli:
Yes, the libre sensors do send a value every minute.

You can see how the graph looks on my son's xDrip:


You can also see the smoothing here where it's following the small white dots which represent the raw values.

jodli:
> @jodli Have you tried with the one minute readings feature enabled? see here: #2680

I'm not really sure what that change is about. On the device receiving the data from Juggluco we already have a raw value every 1 minute. The only thing missing here is to syncing those values to the followers.

Navid200:
> I'm not really sure what that change is about. On the device receiving the data from Juggluco we already have a raw value every 1 minute. The only thing missing here is to syncing those values to the followers.

Wait!
You want to make all these changes to the xDrip code and you are not even going to use xDrip as a collector?  You just want to use it as a follower?

jodli:
I use both actually. My son has the CGM and is using xDrip as the collector. Then he's syncing that data to his followers (me and my wife).

I'm making these changes to be able to see all values my son also sees. 😆

Navid200:
@jodli I think that's the point about the suggestion.

xDrip already provides multiple methods for follower schemes like Nightscout ot xDrip sync.
I don't have personal experience.  But, if you set up your son's xDrip to receive readings once a minute, you can then try to use Nightscout or xDrip sync to follow him.
I hope that would work.
And if it doesn't that may be where we should focus on improving xDrip rather than adding yet another different follow scheme.

jodli:
Now I'm confused. 
This does not add another follower scheme. It's utilizing the "xDrip+ Sync Follower" Data Source.

To summarize:


On my xDrip I don't see the raw values every minute (like he does as shown [here](https://github.com/NightscoutFoundation/xDrip/pull/3371#issuecomment-1979774603)), but a smoothed value every 5 minutes.

jamorham:
@jodli Mainly I am asking is because as far as I know the libre raw data handling is quite hacky and the support for one minute readings is intended to supersede this which is why I am asking if this solves the problem overall or whether there are still things missing.

jodli:
Hey @jamorham! I've tested the 1 minute reading support from #2680 and it works for the Libre (patched App) data source, although it's refusing a few readings, because of a too close timestamp (59-60 seconds between 2 readings).

This results in a weird discrepancy in the 1 minute readings between Juggluco and xDrip:



This does however not solve my problem, because the 1 minute readings, even though they are in xDrip, are not synced to the followers (using the xDrip+ sync follower data source).

jodli:
Hey @jamorham 

I've cleaned up the conditionals about when to sync raw values to followers. I hope this makes more sense now.

Thanks for taking another look!

PhilTU:
> > Ok thanks. Which data source is this to be used with? Do you know what the interactions are with the support for 1 minute readings in xDrip?
> 
> So far I only know the Libre 2 and 3 sensors that support sending new values every minute, and the data source for these has to be set to [Libre2 (patched App)](https://xdrip.readthedocs.io/en/latest/install/libre2patch/). My son is using [Juggluco](https://xdrip.readthedocs.io/en/latest/install/libre2patch/?h=libre#juggluco) to share the data to xDrip.
> 

Libre 2 sensors can also have Libre Bluetooth as the Hardware data source giving one minute values.

The confusion arises here because the alternative app Juggluco is master (has control of the sensor) and feeds data into xDrip+ by emulating the patched app. The diagram showing the Libre 3 feeding into the child's xDrip+ as a CGM input is not quite reality. This is why navid suggested it was a follower scheme, because xDrip+ is following juggluco via the patched app interface.

jamorham:
I wanted to give an update of where I see the roadmap for this feature. Recently google discontinued support for the messaging services we were using. A new system is in place and in live use but there are some usage limits on it and so I don't really want to add new message types at the moment. 

I have been working on a much higher capacity system which looks like it will work and also is designed specifically for xDrip usage patterns so will give a lot more flexibility in terms of what we can do. I'm currently testing this but it isn't ready for roll-out yet. Tentatively I would expect it to take another month.

What I would also expect is that we can add full support for 1 minute readings (although I am hesitant to wake the device up every minute and so granularity might be 2 minutes by default perhaps) - I also expect to be able to support more background syncing of data that doesn't need to be real-time. This should allow us to be more open about syncing certain types of data because we wont have to weigh up the load on phone/battery so much as a consideration.

PR #3380, opened 2024-03-09
Missed reading alert on by default

Navid200:
The missed reading alert is disabled by default.  Anyone who needs it must enable it.

Whether we set the setting to on or off by default, a group of users need to change it to their preference.
If we set the default to enabled, the group that needs to change it is the group who is annoyed by the missed reading alert.
If we set the default to disabled, the group that needs to change it is the group that could face a disaster if they don't!
That would be a parent waking up in the morning and finding out that they have had no reading from their child for hours.

Please also note that when we face the database corruption, the sensor could stop working.  If the missed reading alert is disabled, the user will have no idea until they check.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/MissedReadingActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/MissedReadingActivity.java
@@ -80,7 +80,7 @@ public class MissedReadingActivity extends ActivityWithMenu {
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(getApplicationContext());
         int startMinutes = prefs.getInt("missed_readings_start", 0);
         int endMinutes = prefs.getInt("missed_readings_end", 0);
-        boolean enableAlert = prefs.getBoolean("bg_missed_alerts",false);
+        boolean enableAlert = prefs.getBoolean("bg_missed_alerts",true);
         boolean allDay = prefs.getBoolean("missed_readings_all_day",true);
         boolean enableReraise = prefs.getBoolean("bg_missed_alerts_enable_alerts_reraise",false);
         
--- a/app/src/main/res/layout/activity_missed_readings.xml
+++ b/app/src/main/res/layout/activity_missed_readings.xml
@@ -25,6 +25,7 @@
 
                         <CheckBox
                             android:id="@+id/missed_reading_enable_alert"
+                            android:checked="true"
                             android:layout_width="wrap_content"
                             android:layout_height="wrap_content"
                             android:text="@string/enable_missed_reading_alert"


Navid200:
**Test:**  
  
I installed this on a phone with no xDrip on it.  The missed reading alert was enabled.

Navid200:
Considering the two scenarios I have described in my first post, I believe a safer approach is to have the alert enabled by default to avoid a possible disaster as opposed to avoid an annoyance.

Navid200:
**Test2:** 

I uninstalled xDrip.  Then, installed the latest Nightly before this PR.  I did not open the missed reading alert page.
I then installed this PR over it.  Checking the missed reading alert, the alert is enabled.

This means anyone who is currently using xDrip who has never visited the missed reading alert page will have the alert automatically enabled after they update their xDrip to this PR.

I then disabled the missed reading alert.
I then, installed this PR again over it.  Checking the missed reading alert, I can see that it is still disabled.
This means that if an xDrip user intentionally disables the alert, installing this version over will not enable the alert.

jamorham:
I don't really want to enable this by default, it is annoying because if someone is out of range of their phone then the phone starts alarming.

Navid200:
@jamorham Then, they can disable it.

There are thousands of people who are not even aware that this alert exists.  This is a medical app.  Please allow it to act like one.  No one is obligated to keep it active.  

Are you saying if this PR is approved, it will not allow one to permanently disable the setting?

Navid200:
I have this alert enabled.  I need to know I have no readings when I am sleep.  I don't want to find out when I wake up that I have had no readings for hours.

You are correct that if I walk away from the phone for longer than 30 minutes, this alert will trigger.
The phone has no means of discovering if I am near or not.  So, I accept this is a limitation of this alert.

We have two options.  We can leave this disabled y default or enable it by default.  Each one of those two options has a pro and a con.  The con for leaving it disabled could cause a disaster that I personally don't want  to be responsible for if I can do something about it and this is what I am doing about t.

The con for enabling it is that if one walks away from their phone, this alert will trigger.
Then, they can disable it if they choose to.  Then, if the next day they wake up and find out that they had no readings for 8 hours and got no alert, we did our part.  It was their choice that caused the problem.

Navid200:
People ask is there a way for me to know that I have had no readings for a while?
When I tell them about the missed reading alert, they thank me.
I will keep doing that.
My worry is those who are not aware and suffer and never ask.

Navid200:
https://navid200.github.io/xDrip/docs/Alerts/MissedReading.html

Navid200:
@jamorham The missed reading alert has a checkbox that says `all day`.
If you disable it, it will give you the option of choosing a time period.  So, you can enable it only for when you are sleeping knowing that you won't walk away from the phone, unless you have the habit of sleep walking, in which case an alert will be a blessing as well.  :o)  

Would you please reconsider this?

Navid200:
There are people who actually find it quite useful when the user walks away from their phone!
<br/>  

Screenshot 2024-07-07 171024

jamorham:
The missed reading alert is disabled by default intentionally. I don't think it should be enabled by default. Users who want to use it can enable it.

Navid200:
Very well.  Thanks for the response.

I will try to highlight this in the guide so that hopefully no one will ever be surprised when it is too late.

PR #3414, opened 2024-03-31
Broadcast correct G7 sensor start time

Navid200:

This PR changes what is broadcasted, for sensor start time, only if a known G7 firmware is being used.
The G7 firmware is true either when using a G7 or One+.

What is then broadcasted is the start time from the device instead of the start time of the xDrip session.

I have not tested this because I cannot compile AAPS since we use a different version of gradle.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/BroadcastGlucose.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/BroadcastGlucose.java
@@ -2,6 +2,7 @@ package com.eveningoutpost.dexdrip.utilitymodels;
 
 import static com.eveningoutpost.dexdrip.models.JoH.dateTimeText;
 import static com.eveningoutpost.dexdrip.models.JoH.msSince;
+import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.getTransmitterID;
 import static com.eveningoutpost.dexdrip.utilitymodels.Constants.MINUTE_IN_MS;
 import static com.eveningoutpost.dexdrip.utilitymodels.Unitized.usingMgDl;
 
@@ -10,6 +11,8 @@ import android.os.Bundle;
 
 import com.eveningoutpost.dexdrip.BestGlucose;
 import com.eveningoutpost.dexdrip.BuildConfig;
+import com.eveningoutpost.dexdrip.g5model.DexSessionKeeper;
+import com.eveningoutpost.dexdrip.g5model.FirmwareCapability;
 import com.eveningoutpost.dexdrip.models.BgReading;
 import com.eveningoutpost.dexdrip.models.Calibration;
 import com.eveningoutpost.dexdrip.models.JoH;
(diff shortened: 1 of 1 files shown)


jamorham:
I don't think you're checking the collection type here. What if was previously dex and then changes to libre? Wont you in that case continue to broadcast the previous transmitters start time? You need to also check the collector type first as part of the logic to decide if your value from the session time keeper can be valid. You probably should also exclude it if it returns at the default value (0) because otherwise you are hanging off the assumption that having firmware version information means the session keeper value is also valid.

Navid200:
I'm sorry about that.  Fixing it now.
Many thanks

Navid200:
Done

app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/BroadcastGlucose.java
```diff
@@ -147,7 +151,13 @@ public static void sendLocalBroadcast(final BgReading bgReading) {
                 }
 
                 bundle.putInt(Intents.EXTRA_SENSOR_BATTERY, BridgeBattery.getBestBridgeBattery());
-                bundle.putLong(Intents.EXTRA_SENSOR_STARTED_AT, sensor.started_at);
+                if (getBestCollectorHardwareName().equals("G7")) {// If we are using G7 or One+
+                    if (FirmwareCapability.isDeviceG7(getTransmitterID())) { // Only if there is connectivity
+                        bundle.putLong(Intents.EXTRA_SENSOR_STARTED_AT, DexSessionKeeper.getStart());
```
jamorham, line 156: I think you should read  `DexSessionKeeper.getStart()` in to a local variable and check it is > 0 as part of the overall logic because otherwise you are assuming it is valid. Also shouldn't all these conditions be part of a single condition and if not met then use the existing behavior. Here it looks like you would omit the value if there isn't data. Is that what you want or do you want to emit the sensor start time in this case?

Navid200: We have another problem, which is the sensor start time is not valid for G7.  You were not happy with my approach that addressed both issues here: https://github.com/NightscoutFoundation/xDrip/pull/3263

This PR is only addressing the broadcast for G7 without addressing the incorrect sensor start time.

When someone is using G7 and they change the transmitter ID in order to switch to their new G7, before connectivity is established, the sensor start time should not be broadcasted because it is incorrect.
That's why if we are using G7, but there is no connectivity, which is the case between two sensors for a few read cycles, this PR will not make any broadcast.

Navid200: I don't understand your first statement.
If we are using G7 and if we know what the firmware is, how could Dex start time be incorrect considering that is what we show on the system status page.

jamorham: It is fine if you don't think we should broadcast anything about the session start time for g7 until we are sure. That is why I asked if it was intended.

Your code assumes that `getStart()` will have valid data. I believe that as things are at the moment this will be the case if we have data in `FirmwareCapability` but I don't believe they are actually linked and if something changed in the future, for example not reading the session start time right away (which is unlikely but possible as we might have to read firmware version first) then this new code will break.

That is why I am suggesting that you also check that `getStart()` returns >0 as part of your condition before broadcasting data because the whole purpose of this is not to broadcast invalid data. If the session keeper is not initialized yet then you will get 0 as the default value I believe. Best to check >0 in case I ever make -1 the default as it is for some other things.

Navid200: Understood.  Makes perfect sense.  Thanks.  Please let me fix it.

Navid200: Done

jgslade:
I have tested the original of this PR over the life of my previous sensor and starting my new one. I did experience an issue where when starting my new sensor it kept trying to read from the old one and got mismatching errors for about 20 mins. But other than that everything is looking good, no duplicated data, and the correct start time was sent to AAPS.

jamorham (approved):

PR #3426, opened 2024-04-08
Companion app Dexcom One+

Navid200:
https://github.com/NightscoutFoundation/xDrip/discussions/3425


jamorham:
Let me know if this is ready for merge

Navid200:
I cannot test it myself because I don't have a One+.

But, can we not merge this and let the user who asked for it test it?
The change cannot break anything else, can it?  If it doesn't work, we can remove it.

jamorham:
Are you confident on the package id?

Navid200:
Sending you a message on Matrix (Gitter).

Navid200:
Yes, as confident as I can be.

PR #3436, opened 2024-04-22
Broadcast start time from transmitter for G6 as well

Navid200:
We changed the broadcasted start time for G7 to what G7 reports instead of the xDrip's local start time.
So far, we have had positive feedback from users and no complaints.

However, we now have one similar complaint about G6.

This PR does the exact same thing for Firefly G6 as well.


Navid200:
It may seem there is redundancy with 3 new variables.
My concern was if we are using G6 and everything is working and then, we change the transmitter ID to that of a G7, Until we establish connectivity to G7, we are using G7 and we have had connectivity to G6.  I didn't want this to result in making any broadcast.

That's why there is a variable that is set only if we are using G6 and we are connected to G6.
And there is a variable that is set if we are using G7 and we are connected to G7.

jamorham (approved):
Thanks for explaining your design here and inline intention comments

PR #3437, opened 2024-04-25
Graph smoothing settings reorganization

Navid200:
We have two settings on two different pages that both are related to graph smoothing.
One enables or disables it.
The other shows unsmoothed values as well when enabled.

This PR puts both of these settings on the same page and removes show unsmoothed from the other misc. options page.


Navid200:
The setting on the display settings page is replaced with a button that takes you to a new page.
Screenshot_20240424-205656  
  
And this is the new page: 
Screenshot_20240424-205939

jamorham (approved):

PR #3457, opened 2024-05-11
G6 and G7 persistent authentication logs cleanup

Navid200:
This type of question has been coming up for years.  
<br/>  
  
Screenshot 2024-05-10 224115

This PR will show the logs when detail logs are enabled.
But, under normal circumstances, the two logs will not be shown.  
They neither help the users now me as I consider them expected.  Please correct me if I am wrong.
Otherwise, please merge.


jamorham:
Yes its okay to quiet this log message

jamorham (approved):

Issue #3219, opened 2023-12-02
Vehicle mode does not stop after bluetooth is disconnected, causing 'speak readings' to continue.


Gferrar123:
### Discussed in https://github.com/NightscoutFoundation/xDrip/discussions/3217

<div type='discussions-op-text'>

<sup>Originally posted by **Gferrar123** November 30, 2023</sup>
Vehicle mode does not stop after bluetooth is disconnected, causing 'speak readings' to continue.
Running Version 2023.11.14 
Samsung Galaxy 23 with Android 14
Dexcom G6
I have recreated it with two different bluetooth pairings and the issue has been confirmed by @Navid200 
Any fix would be great, but I would love to see a vehicle mode trigger based on Android Auto running, rather than a specific bluetooth connection.</div>

FriedelPas:
The same here with the latest Samsung System update that installed One UI 6.0 and Android 14 on a Samsung Galaxy S21 FE 5G
Running version 2023.07.16

Navid200:
Please note that you can disable the ongoing speak readings, after leaving the car, by disabling "Vehicle mode" under xDrip+ Motion tracking.

This is not a solution to the problem.  But, it is a way to disable the behavior until this is addressed.

FriedelPas:
Yes, I know and have done that, what indicated me more clear that this was the reason. But I was thinking that adding extra input could help adressing the problem.
BTW, after the update of samsung, the phone said that xDrip was developped for an older version of the system and possible could cause problems. I not noticed problems at that time, but when first time driving with car the problems started. Until now it is the only issue that I had with the update.

Would be nice when problem solved, but also now xDrip stay the game changer in treathing my diabetes very well! Thanks to xDrip I can have values more then 90% in range and so getting an Hemoglobine A1C of 5.1%, so I'm very happy with this app.

Kaapeli64:
I've been using Vehicle mode happily for years, today I got Android 14 update on my OnePlus Nord 2 5G and it no longer exits the Vehicle mode. In the morning I still had "Vehicle mode: Enabled" followed with "Vehicle mode: Disabled" prints in the log, but now after the update I only see "Vehicle mode: Enabled" prints one after one, every time I enter the car. xDrip no longer detects the exits.

Navid200:
Please try the xDrip release of May 12, 2024 or any after that and if the problem persists, please report it here.

Kaapeli64:
[...]

PR #3458, opened 2024-05-11
fix bluetooth device disconnection not being correctly detected

richard-dennehy:
This should fix https://github.com/NightscoutFoundation/xDrip/issues/3219

The underlying issue is that xDrip is expecting disconnected bluetooth devices to go straight from `STATE_CONNECTED` to `STATE_DISCONNECTED`, but as of Android 14, there's a transition to `STATE_DISCONNECTING` in between.

This doesn't seem to be a new state, and I'm not sure why or when this changed - I can't find any Android changelogs that mention this.

I've made a minimal change to the condition, and I've tested it locally and it works on my device.

Screenshot_20240511-152008


Navid200:
I don't see anything wrong with this. I will test either today or tomorrow.
Except my phone is running Android 15 beta now.  Hopefully it will work on that as well.
Thanks for doing this.

jamorham:
This looks reasonable and should work. I expect some older implementations just never reported the disconnecting state which is why it wasn't picked up earlier in testing.

jamorham (approved):

PR #3461, opened 2024-05-13
Wake up for G7 once a minute

Navid200:
A G7 could shift the time stamp off the one-per-5-minute grid as a result of a disconnect.
xDrip uses the timestamp and wakes up 5 minutes after.  However, considering the timestamp may not be on the correct grid, xDrip may wake up when G7  does not transmit.

There are many reports of users having disconnect issues.

This PR changes the wake schedule from once every 5 minutes to every minute for a G7 or One+ only.
My concern is if battery consumption may go up.
My hope is that the impact on battery will be negligible.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/g5model/DexSyncKeeper.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/g5model/DexSyncKeeper.java
@@ -58,7 +58,6 @@ public class DexSyncKeeper {
     // anticipate next wake up from time
     // -1 means we don't know anything
     static long anticipate(final String transmitterId, final long now) {
-
         final long last = PersistentStore.getLong(DEX_SYNC_STORE + transmitterId);
         if (last < OLDEST_POSSIBLE) {
             return -1;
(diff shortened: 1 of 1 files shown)


TWilhoWI (commented):
Seems fine for me.

Navid200:
This clearly raised the power consumption for me.  We cannot do this.
I wanted to avoid adding complicated logic.  But, we have no choice.  We cannot raise the xDrip power consumption like that.
I will need to monitor the times of the incoming readings and keep a tally for the last two.  If they are ever apart by a value less than 5 minutes, only then, should we scan once a minute.
This will avoid scanning once a minute when you walk away from your phone and should keep the battery consumption effectively unchanged.

I will need time to come up with the change.

Navid200:
I have an idea.  I should have an update in a few hours.

Navid200:
This is getting a little bit complicated!
There has been a suggestion that it is xDrip that messes up the timestamps.
The suggestion is that when G7 starts transmitting once a minute, it just transmits the same value and time over and over again until the time comes for the next reading in 5 minutes.  But, xDrip clearly shows timestamps that are off grid.

I cannot fix it if I am not 100% sure what's going on.
I need more time sorry.

Navid200:
@TWilhoWI This has been fixed by our lead developer.  it is being tested now.
It should come out as a Nightly hopefully some time this week.

TWilhoWI:
Perfect! Please do keep me posted!

Navid200:
There is no reason for this PR any longer.
The problem has been fixed by JamOrHam in any release at or after May 22, 2024.
Please update if you are using xDrip to collect from G7.

PR #3464, opened 2024-05-15
Make X axis scaling consistent

trallen:
At present, when predictions are added to the BG line chart, the chart changes its scale.

For example, when displaying 3 hours of data, and a prediction event is added (adding a bolus, for example), the chart will now display ~6 hours (depending on the decay of the insulin) in the same view. This re-scales the chart, which can be confusing if you have locked the time period.

This commit changes this behavior, but only when "Locked Time Period Always Used" is on.

If "Locked Time Period Always Used" is set, the existing X axis scale will remain unchanged when predictions are made, instead adding the prediction offscreen to the right of the current BG value. (That is, you will see the start of the prediction, but will have to scroll right to see the full prediction.)

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -2296,8 +2296,12 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
                 break;
         }
 
-        float ideal_hours_to_show = DEFAULT_CHART_HOURS + bgGraphBuilder.getPredictivehours();
-        // always show at least the ideal number of hours if locked or auto
+        // always show at least the ideal number of hours
+        float ideal_hours_to_show = DEFAULT_CHART_HOURS;
+        // ... and rescale to accommodate predictions if not locked
+        if (! homeShelf.get("time_locked_always")) {
+            ideal_hours_to_show += bgGraphBuilder.getPredictivehours();
+        }
         float hours_to_show =  exactHoursSpecified ? hours : Math.max(hours, ideal_hours_to_show);
 
         UserError.Log.d(TAG, "VIEWPORT " + source + " moveviewport in setHours: asked " + hours + " vs auto " + ideal_hours_to_show + " = " + hours_to_show + " full chart width: " + bgGraphBuilder.hoursShownOnChart());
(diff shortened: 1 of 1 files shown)


Navid200:
Thanks for the PR.

Everything I state is my opinion.  You should not take it as an official xDrip statement.  I am not making decisions about what is approved or not.  So, if I suggest that you do something, it is only my opinion.  :o)  

Because this affects what is seen on the screen, it may help with the review if you show before and after screenshots.  I know that may be a lot of work.  But, if I was the one who had to decide to approve or reject this, I would need to see it.

One of the major concerns is how something changes xDrip.  Just because we improve something, it may not be enough to go for it.  If every user has got used to it the way it is, changing it may not be a very good idea.  Again, that is why it will be useful to be able to see it to decide how big a change is it.

For example, I am curious exactly what it looks like and how much of the curve in the future it still shows.
If nothing is shown in the future, I wonder if we can shot by one hour to shows one hour less in the past so that we can show one hour more in the future when this kicks in.  Just wondering.

Thanks so much for the fast implementation.  Thanks for your help.  Thanks a lot.

jonprogers:
@trallen Thanks for the prod to have a look at this. Nothing you propose here would have any negative impact on me.

trallen:
> Because this affects what is seen on the screen, it may help with the review if you show before and after screenshots. I know that may be a lot of work. But, if I was the one who had to decide to approve or reject this, I would need to see it.

Thanks, I had not thought of this.

This is the "before" screen -- no predictive info is loaded:

Screenshot_20240516-125856_xDrip+

In the current master, this is what the screen looks like when Predictive Simulations are turned on:

Screenshot_20240516-130111_xDrip+

This PR changes the display behaviour when predictions are available:

Screenshot_20240516-130748_xDrip+

In particular, the slope is maintained as-is. (Well, technically, it's not -- the Y axis does change a little, since the prediction drops below 2 at some point, so instead of a Y range of ~2 - ~12, both screenshots with predictive simulations have a Y range of 0 - ~12. This is outside the scope of this PR, and I'll be looking into this behaviour in more detail later.)

Lastly, to show how predictions are reached via scrolling, I'm trying a small video; hopefully it'll work:

https://github.com/NightscoutFoundation/xDrip/assets/10269952/3e4c4662-d996-4150-bb16-6affea1ad8b3

Navid200:
This looks good to me.

Navid200:
@trallen I'm sorry I had overlooked your previous contributions.
Thanks for your service then and now.

trallen:
> @trallen I'm sorry I had overlooked your previous contributions. Thanks for your service then and now.

Happy to help with such an awesome project! :-)

jamorham:
This looks like it should work and would be a positive change. @trallen how long have you been running this to confirm the change is robust? The viewport changes can sometimes be tricky.

trallen:
I've been running it since I submitted the PR, and have not seen any issues yet. Thanks for taking a look!

jamorham (approved):

Navid200:
@trallen Thanks

PR #3476, opened 2024-05-23
Auto start G7 even if the session was stopped recently

Navid200:
There is a condition that may not allow a G7 to auto start for 2 hours.
I have shown how to recreate the problem here: 
https://github.com/NightscoutFoundation/xDrip/discussions/3475  

I then, repeated the procedure for recreating the problem with this PR and it auto started right away.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -12,6 +12,7 @@ import static com.eveningoutpost.dexdrip.utilitymodels.Constants.DAY_IN_MS;
 import static com.eveningoutpost.dexdrip.utilitymodels.Constants.HOUR_IN_MS;
 import static com.eveningoutpost.dexdrip.utilitymodels.Constants.MINUTE_IN_MS;
 import static com.eveningoutpost.dexdrip.utilitymodels.Constants.SECOND_IN_MS;
+import static com.eveningoutpost.dexdrip.utils.DexCollectionType.getBestCollectorHardwareName;
 import static com.eveningoutpost.dexdrip.xdrip.gs;
 
 import android.Manifest;
@@ -2623,7 +2624,8 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
         boolean isSensorActive = Sensor.isActive();
 
         // automagically start an xDrip sensor session if G5 transmitter already has active sensor
-        if (!isSensorActive && Ob1G5CollectionService.isG5SensorStarted() && !Sensor.stoppedRecently()) {
+        // There will be no auto start unless the sensor has not been stopped withing the last 2 hours, or unless it is G7 or One+.
+        if (!isSensorActive && Ob1G5CollectionService.isG5SensorStarted() && (!Sensor.stoppedRecently() || getBestCollectorHardwareName().equals("G7"))) {
             JoH.static_toast_long(getString(R.string.auto_starting_sensor));
             Sensor.create(tsl() - HOUR_IN_MS * 3);
             isSensorActive = Sensor.isActive();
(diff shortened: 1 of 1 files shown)


Navid200:
Please think of someone who has been using G6 with xDrip.
They want to switch to their first G7.
They stop their G6 session.  Then, they change the transmitter ID to the pairing code of their G7.  They will have no reading for 2 hours.

Navid200:
I stopped a G6 session.  
Then, changed the transmitter ID to the pairing code of a G7 I had inserted about 10 hours ago.
After connectivity, the following is what I see.

Screenshot_20240523-183218
Screenshot_20240523-183244
Screenshot_20240523-183321
<br/>  
  
And when I tap on start sensor, this is what I get:  
Screenshot_20240523-183338
<br/>  
  
But, it does not auto start.  I expect that it will eventually start 2 hours after I stopped the G6.  I will wait and report.

Navid200:
And exactly 2 hours after having stopped the last session, the readings started.

We need this PR because otherwise, under some circumstances, users will have to wait 2 hours before getting readings.

jamorham:
I don't think this is the right way to resolve the issue as I think this may cause the sensor to auto start when it should be stopped under some conditions. I'm looking at an alternative method.

Navid200:
We agreed there was no reason to stop a G7.
If a user is done using a G7, they can change the pairing code to something else, change hardware data source to something else like disable collection, ...

There is no benefit gained by stopping a G7 considering there is nothing we can do with a G7 other than get readings from it if it is still sending them.  

Therefore, this is the best way to deal with it.
Would you please merge this for now.  When you come up with the new scheme, please add it then.

I hate to ask you to put the start option back in because I wanted to remove it myself.
But, if you really are not going to merge this, we have to put the start back in.

Navid200:
If a user chooses to enter the pairing code of a G7 to establish connectivity to it, why would a session starting be a problem?  

What would be the advantage of being connected to a G7 if receiving readings from it is unacceptable?
It would be with a G6 because you can start multiple sensors with the same transmitter.  While you are removing the old sensor one and inserting a new one, you wouldn't want a sensor to start until you are ready.  So, the 2-hour delay in that case makes sense.
But, we cannot have any of that with G7 anyway.  We cannot start another sensor on it.

cassafrass387:
Hope this gets merged soon because I just had this happen as well from what I can tell. Switched from G6 to G7, had to update the app and now I've got readings on my t:slim but none on xDrip as of yet. If you'd like logs just lemme know.

Navid200:
@cassafrass387 I am so sorry about this.
I also hope that this is merged soon.

You can install the previous Nightly: https://github.com/NightscoutFoundation/xDrip/releases/tag/2024.05.15
That will allow you to start a session.  And you will have readings right away.
Then, you can install the latest Nightly again.

Navid200:
@jamorham I suspect that the current delay of 2 hours originates from the time of G5.
Because one would stop a session.  Then, start their next G5 sensor using their t:slim pump and in 2 hours (G5 warmup time) they would have readings on the pump and xDrip would auto start and they would also have readings in xDrip.

G6 warmup time is also 2 hours.
Now, one may want to suggest that we change the delay from 2 hours to 30 minutes got G7.
The problem is that G7 is a completely different concept.  The two consecutive sessions never run on the same device in case of G7.  In fact, you can insert your new G7 while you are still using your previous one taking advantage of the 12 hour grace period so that when your old device stops working, you will have readings from the new device right away with no need to wait 30 minutes.

I perfectly understand that the concept of session is outdated now considering the session is not running on xDrip, but on G7.  
So, I respectfully ask you to consider this:
Please merge this so that users can get readings right away.
Then, we can address this exactly the way you like and update xDrip.  If you accept this request, the xDrip users will not have to wait for us to come up with a perfect solution and they can right away have readings as soon as they establish connectivity.

Navid200:
To avoid further confusion imposed on the users, I will change the G7 guide to warn users about this and ask them not to update to the latest Nightly.

Navid200:
This is another person experiencing the issue this PR fixes: https://github.com/NightscoutFoundation/xDrip/discussions/3489

Navid200:
This has now been fixed (tested) in the latest Nightly: https://github.com/NightscoutFoundation/xDrip/releases/tag/2024.05.28

Therefore, there is no need for this PR.

Thanks for the fixes

PR #3483, opened 2024-05-24
Android 15 compatibility

Navid200:
Attempting to install the latest Nightly (May 22, 2024) on Pixel 6a running Android 15 beta 2, installation is blocked by this message:
Screenshot_20240524-164802
  
This PR makes it possible to install even though you will need to tap on "install anyway" on a warning you will get similar to what is shown here:
https://navid200.github.io/xDrip/docs/FAQ/UnsafeAppBlocked.html.  
  
I am currently using this to collect from G7 and upload to Nightscout.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -89,10 +89,10 @@ android {
 
     defaultConfig {
         applicationId "com.eveningoutpost.dexdrip"
-        minSdkVersion 23
+        minSdkVersion 24
         // increasing target SDK version can cause compatibility issues with Android 7+
         //noinspection ExpiredTargetSdkVersion
-        targetSdkVersion 23
+        targetSdkVersion 24
         // change versionCode only when downgrade should be prevented
         // eg, when data structures are incompatible
         versionCode 1603091400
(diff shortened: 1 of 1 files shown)


Navid200:
The problem is also explained in a discussion here: https://github.com/NightscoutFoundation/xDrip/discussions/3477

Navid200:
Working fine.  No issues so far

Navid200:
Screenshot_20240525-094814

Navid200:
I am using this with no issues.

Navid200:
Works fine as xDrip sync master and Tidepool master as well.

jamorham:
So to merge this we must be mindful of the following:
https://developer.android.com/about/versions/nougat/android-7.0-changes
I can do this on a subsequent nightly to differentiate it and I'm hopeful that some recent changes mean that we should be okay.

Navid200:
Wow!
Thanks for that.

I will go through that list and try what is in xDrip that may be related.
If I find anything that behaves unexpectedly, I will report here.

jamorham (approved):

PR #3486, opened 2024-05-25
Rename to urgent low soon alert

Navid200:
I have 2 concerns with what we have now.

1- There are individuals who see the word "Forecast" and assume that this is a simulation.
2- There is no way to know if this alert takes into account the insulin on board or the carbs on board, which it doesn't.

This PR renames a few phrases in order to clarify that this is not a simulation and as such also does not take into account insulin on board or carbs on board.

The following image shows what it will look like in English.  
  
**Edit: This image is outdated.  Please see the post below for the update.**
333817523-7cafa340-bbb6-4f6e-845f-9d91ef21e946  
  
The title "Urgent Low Soon Alert" is used by Dexcom: 
Screenshot 2024-05-25 122421  
  
Those who have already used the Dexcom app may find it easier to understand what the alert is if they have seen the title before.  

All in all, my intent is to help xDrip users understand this very important and useful alert better.


Navid200:
Screenshot_20240525-180625

olorinmaia:
Excellent explanation and will make this alert more understandable for most users.

jamorham:
I think what you're suggesting here is copying the design of a manufactures app which I do not think is a good idea especially as we're using a different mechanism for this type of alert. I also find the wording used in the proposed description quite confusing. I'm not against updating it but I don't think this is the right change to make. 

Also we could take in to account the other aspects IOB/carb simulation for the alarm which was the original intention when I implemented that feature but the existing trend based algorithm appeared to be very effective without needing any user input which is why it remains that way so far.

Navid200:
I suppose when the time comes that this alert will take insulin and carbs also into account, if a user does not enable xDrip predictive simulations, like me, the alert will behave exactly as it does now.
Anyway, that is something that will happen in the future.  Who knows when?  Let's worry about that day when that day comes.

The only reason I want to rename this is that it is not clear what it does.
Let me think about an alternative set of phrases.

Navid200:
I changed the title back to what it was.

I am editing the summary to state that carbs on board and insulin on board are not taken into account.

When the day comes that this alert changes, if that day comes, to take those into account, I will update the summary then to validate it.

jamorham (approved):

PR #3499, opened 2024-05-30
Sensor expiry note on home screen wrap

Navid200:
The following images show  what the sensor expiry looks like on May 28 and May 29, 2024 releases respectively.
Screenshot_20240529-223455
Screenshot_20240529-220259

The change of the Android target has affected the behavior of linear layout and caused the sensor expiry note to wrap.
I am sorry I missed this change in my tests of Android 15 upgrade.

This issue has been reported here: https://github.com/NightscoutFoundation/xDrip/discussions/3495  

My take is that Android 6 gave no space to an item if it was invisible.  But, it seems Android 7 dedicates space to it even if it is invisible.
I am not sure what the purpose of collector status is on the home shelf.  This PR gives 99% of the space to sensor expiry and 1% to collector status.
The result is that as long as the collector status is invisible (empty), it will be effectively taking no space and the final result will be as it was before upgrade to Android 15.

### DIFF
--- a/app/src/main/res/layout/activity_home.xml
+++ b/app/src/main/res/layout/activity_home.xml
@@ -540,21 +540,19 @@
 
                 <TextView
                     android:id="@+id/nanoStatusText"
-                    android:layout_width="0dp"
+                    android:layout_width="wrap_content"
                     android:layout_marginTop="2dp"
                     android:layout_marginStart="5dp"
                     android:layout_height="wrap_content"
-                    android:layout_weight="1"
                     android:text="@{nano.color_watch, default=`Nano Status`}"
                     app:showIfTrue="@{nano.color_watch.length() > 0}"/>
 
                 <TextView
                     android:id="@+id/ExpiryStatusText"
-                    android:layout_width="0dp"
+                    android:layout_width="wrap_content"
                     android:layout_marginTop="2dp"
                     android:layout_marginStart="5dp"
                     android:layout_height="wrap_content"
-                    android:layout_weight="99"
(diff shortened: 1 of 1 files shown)


Navid200:
This is the result after this PR:
Screenshot_20240530-114243

jamorham:
I think this will break the collector status which is where you see things like Sensor Expired etc. Is the problem actual something to do with the layout_width being 0dp? You can prototype the change by making the android:text field not observable and just replacing with the placeholder text while you debug the layout.

Navid200:
This isn't changing the collector status.  It's only changing how much room on screen is allocated to it.
I have tested this.

jamorham:
So how does it look when both are displayed?

Navid200:
OK,

I now understand.
I am embarassed to say I don 't know hot to show anything as collector status on home screen.

How can I do that so that I can properly test this?

jamorham:
See my comment 7 mins ago which explains how to prototype that for testing

Navid200:
Under normal operating conditions, when will both of these two fields be occupied?
Warmup?

Navid200:
Screenshot_20240601-123634 Screenshot_20240601-130007  
  
The following shows a test where I have forced the collector status to also be visible at the same time as sensor expiry is.
Screenshot_20240601-130739

Navid200:
Thanks for the review and catching my error.

Navid200:
This is what this PR shows after just starting a new G7.
Screenshot_20240602-072150

jamorham (approved):

PR #3506, opened 2024-05-31
No More references to G5

Navid200:
This PR removes any references to G5 with a more general title or a title without G5.
Keeping G5 in the titles gives the impression that the app is very old.

This PR only changes the names.  There is no change to any functionality and users can continue to use G5 as they did before.


Navid200:
These are the updated menus.  
  
Screenshot_20240531-174753  Screenshot_20240531-174820  Screenshot_20240531-180826  
  
Screenshot_20240531-174942
  
Screenshot_20240531-182257

jamorham (approved):
Thanks

PR #3515, opened 2024-06-03
Remove G5 from status page header

Navid200:
This is what the status page will look like after this:  
Screenshot_20240603-075232  Outdated - please see next post
 
I forgot about this one in my previous PR: https://github.com/NightscoutFoundation/xDrip/pull/3506

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/MegaStatus.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/MegaStatus.java
@@ -119,7 +119,7 @@ public class MegaStatus extends ActivityWithMenu {
     }
 
     private static final String G4_STATUS = "BT Device";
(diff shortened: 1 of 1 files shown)


Navid200:
Screenshot_20240603-084816

jamorham:
I don't want to make changes that could potentially include trademarks unless absolutely necessary. I'm fine with removing G5 though.

Navid200:
These headings have been fluid ever since One came along.  Since then, there have been even more.
And there will be more coming soon.
So, my hope is that we can use a heading that remains intact as new devices come along.
So, I hope you accept the update I have made to this PR to address your concern.

If not, please let me know and I will just remove G5.

jamorham (approved):
Yes this is fine now

PR #3520, opened 2024-06-05
migrate Espresso test to Android X

tolot27:
During compilation of the debug app missing libraries are shown in `HomeEspressoTest.java` because this test was not migrated (74a4f02) to the androidx libraries. This was also mentioned in #3255.
This PR migrates this Espresso test to androidx. Still not all tests are passing, but many. Upgrading the libraries to newer ones will break even the first test. But this is out of scope for this PR.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -372,7 +372,6 @@ dependencies {
     androidTestImplementation 'androidx.test.ext:junit:1.1.5', {
         exclude group: 'com.android.support', module: 'support-annotations'
     }
-    androidTestImplementation 'androidx.test:core:1.2.0'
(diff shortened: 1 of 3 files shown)


tolot27:
By the way, the test is still failing. Will this address with another PR.

jamorham:
Am I right in understanding that this test now will run but then fails?

tolot27:
> Am I right in understanding that this test now will run but then fails?

Yes, the Espresso tests are is running but most of the time at least two of the four tests fail because the GUI interaction is not correct (mostly caused by incorrect scrolling). This can be fixed in other PRs. My main intend was to fix the build errors and do some cleanup.

jamorham (approved):

PR #3526, opened 2024-06-09
Session stop notification for G7

Navid200:
When a G7 comes close to 10.5 days of life, 2 different things can happen:  

1- User switches xDrip to their new G7 before the current session expires.
In that case, this PR changes nothing.  
  
2- User waits for the session on their current G7 to end while still connected to it in xDrip.  
In that case, this PR will end the xDrip session and issues a notification the same as what it would do for a G6.  

Without this PR, xDrip never issues a notification.  
I have been in this situation when the session has expired and I had no idea for a while until I got a missed reading alert.  
<br/>  
  
---  
  
**Test**  
I installed this on a test phone.  I established connectivity to my active G7.  
After getting readings for a few cycles, I switched to an expired G7.  The xDrip session was successfully ended and a notification was received.  
  
I can test this in a real scenario where the session really expires in 4 days.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/g5model/CalibrationState.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/g5model/CalibrationState.java
@@ -58,7 +58,7 @@ public enum CalibrationState {
 
     private static final SparseArray<CalibrationState> lookup = new SparseArray<>();
     private static final ImmutableSet<CalibrationState> failed = ImmutableSet.of(SensorFailed, SensorFailed2, SensorFailed3, SensorFailed4, SensorFailed5, SensorFailed6, SensorFailedStart);
-    private static final ImmutableSet<CalibrationState> stopped = ImmutableSet.of(Stopped, Ended, SensorFailed, SensorFailed2, SensorFailed3, SensorFailed4, SensorFailed5, SensorFailed6, SensorFailedStart, SensorStopped);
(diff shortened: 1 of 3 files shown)


Navid200:
When a G7 session ends, the sensor status changes to expired.
But, xDrip does not consider this as stopped.
This PR changes that so that expired is also considered as stopped.
The result is that xDrip will notify the user that the session has ended.

Navid200:
Just in case this comes up:
I know that we have a sensor expiry alert:
https://navid200.github.io/xDrip/docs/Sensor-Expiry.html
This PR is not related though.

Our sensor expiry alerts are quiet for a very good reason.  We don't want to wake up someone to tell them that their sensor will expire in 2 hours.
But, when the sensor just expires, that is a completely different situation.  If someone doesn't know that their sensor has stopped, things could get bad.  
And I suppose that's why when a G6 sensor stops, we have a notification, which is not quiet.
This PR allows the exact same notification to also notify us if a G7 has just expired.

Navid200:
Screenshot_20240610-133705  Screenshot_20240610-133641

jamorham:
Yes I've been testing the same here, this is a needed change I think as was missing previously.

jamorham (approved):

Navid200:
Even though the notification is provided, I don't think this really stops the session.  
Screenshot_20240613-111332
  


After starting the next session, the classic status page, still shows the previous start time, 6 days ago.

Screenshot_20240613-112111

But, the Dex status page shows the real start time, 3 hours ago.
Screenshot_20240613-112128

PR #3529, opened 2024-06-12
Data table crash

Navid200:
Install xDrip (not over a previous install).
Do not set up a hardware data source.
Enable data tables.  
Tap on data table.  xDrip will crash.  

It crashes when calling parseDataForStats.
And what causes the crash is the for loop in that method.  

I have extended the existing try and catch to also include the call to the parseDataForStats method.

The same test shows no crash anymore after this PR.  It shows an empty data table.


Navid200:
Tested with Android 9

jamorham (approved):
Makes sense thanks

PR #3530, opened 2024-06-12
Retire old G5 settings

Navid200:
This PR removes 2 settings and the category that only contains the two.
The settings are titled:
Scan constantly
Force Dex to UI thread

Considering we have retired Android 4, 5 and 6, am I wrong to think that no one should need these two settings any longer?  
 
Screenshot_20240612-130522
  
The following image shows what the Debug settings page will look like for G7 after this PR.  
It will look exactly the same for a G6 with the addition of the battery options at the bottom also present.  

Screenshot_20240612-122431


jamorham (approved):
These settings are used by an old old collector which no-one should be using anymore or they will experience many problems. Those settings were experimental when they were produced. You also can remove the findPreference() lookup for these in Preferences.java as well as those are no longer needed either.

PR #3533, opened 2024-06-13
Micro tech LinX companion app

Navid200:
This PR adds a new sensor to the companion app options.
I don't have this sensor myself.  

https://github.com/NightscoutFoundation/xDrip/discussions/3532

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/UiBasedCollector.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/UiBasedCollector.java
@@ -91,6 +91,7 @@ public class UiBasedCollector extends NotificationListenerService {
         coOptedPackages.add("com.medtronic.diabetes.simplera.eu");
         coOptedPackages.add("com.senseonics.gen12androidapp");
         coOptedPackages.add("com.senseonics.androidapp");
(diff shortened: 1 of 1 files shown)


Stealthwick:
I have this sensor. I have tested it and it works.

yangyang-test:
The XDrip app can normally retrieve LinX data on Samsung, Xiaomi, OPPO, and VIVO smartphones, but it does not work on HONOR phones.

Navid200:
@yangyang-test Is that statement related to this PR?

Navid200:
@yangyang-test xDrip cannot retrieve data from LinX.  This PR will add the ability to use the companion app mode, which is a workaround.

If you are testing this PR and are finding a particular phone does not work, does that phone show a new notification for every new reading on the official Linx app?

If we can get support from the manufacturer, we can add a collector for Linx.  If we succeed, xDrip will be able to show readings from Linx with no need to have the official app installed.

The disadvantage of the companion app mode is that if either the companion app or xDrip malfunctions, the system will fail.  So, there are two points of failure instead of just one.

jamorham (approved):

Bennani12:
IMG-20240626-WA0007
It work for same hours in Redmi note 8 then stoped completely

PR #3539, opened 2024-06-14
US server setting in the Dex share follow menu

Navid200:
This setting is only available in the Dex share upload menu.
But, it is required to be set correctly for a Dex share follower to operate as well.  
This PR adds the US server setting to the conditional Dex share follower menu.  
The menu before and after this PR is shown below.  
  
| Before | After |  
| ------- | ------ |  
| Screenshot_20240614-172942 | Screenshot_20240614-173252 |


Navid200:
This setting already exists.
xDrip as is can be used as a Dex share follower.

The only problem is that for someone who has never used xDrip and is setting it up as a Dex share follower, there is no way to know that they need to go to a different menu to choose the correct setting for the server.
This PR shows the same existing setting also right after the only other two important settings required, user ID and password.

jamorham (approved):
This looks sensible

PR #3540, opened 2024-06-15
Battery optimization prompt clarification

Navid200:
When a switch has a visual representation, it is best to have a unique summary specifying what the switch does when enabled.  
As an example, please have a look at the following screenshots of the speak reading switch.  
Screenshot_20240615-074535 Screenshot_20240615-074553  
The summary says "it will speak each new reading" regardless of if the checkbox is enabled or not.  
This tells the user what the checkbox does if/when enabled.  This is a reasonable expectation from the user to consider the summary to explain what the switch does when enabled.  
  
If the switch did not have a visual representation, it would make sense to use the summary to tell the user what tapping on the switch will do.  
But, when the switch already has a visual representation, making the summary change depending on the switch condition makes it a confusing setup.  
<br/>  
  
The following shows what the battery optimization prompt switch does now before this PR.  
Screenshot_20240615-074836  Screenshot_20240615-074903  
Look at the summary for when the checkbox is disabled.  It says "recommended".  What if the user confuses that with us telling them that we recommend that they enable the switch?  
<br/>  
<br/>  
  
---  
#### **This PR**  
This PR changes the summary to one statement regardless of if the checkbox is enabled or not as shown below.  
Screenshot_20240615-075026 Screenshot_20240615-075043  
  
Now, we are using the summary to tell the user what the checkbox does when enabled.  And the checkbox representation (checkmark present or not) tells the user if the checkbox has been set or not.  Exactly the same as the speak reading checkbox and almost every other checkbox in xDrip.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -621,7 +621,7 @@
     <string name="short_speak_readings_shortcut">Show a shortcut in the ☰ menu to toggle Speak Readings</string>
     <string name="speak_readings_shortcut">Speak Readings Shortcut</string>
     <string name="battery_optimization_off">On Android 6+ ensure battery optimization is switched off (recommended).</string>
-    <string name="battery_optimization_on">Don\'t keep asking about battery optimization (not recommended).</string>
+    <string name="battery_optimization_on">Will not keep asking about battery optimization when enabled (not recommended).</string>
     <string name="battery_optimization_prompt">Battery Optimization prompt</string>
     <string name="advanced_bluetooth_settings">Advanced Bluetooth Settings</string>
     <string name="bluetooth_settings">Bluetooth Settings</string>
--- a/app/src/main/res/xml/pref_advanced_settings.xml
+++ b/app/src/main/res/xml/pref_advanced_settings.xml
@@ -1494,8 +1494,7 @@
                 <CheckBoxPreference
                     android:defaultValue="false"
                     android:key="requested_ignore_battery_optimizations_new"
-                    android:summaryOff="@string/battery_optimization_off"
-                    android:summaryOn="@string/battery_optimization_on"
+                    android:summary="@string/battery_optimization_on"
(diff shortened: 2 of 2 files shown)


asavageiv:
Shouldn't it be called "No Battery Optimization Prompt"?

asavageiv:
Good change regardless.

Navid200:
@asavageiv In my opinion negative switches should not be used anywhere.
If we were creating xDrip from scratch, I would suggest to change this switch to one that would enable the prompt and set it to be enabled by default.

But, making that change now is problematic.
If we change it now, we will also have to deal with what happens if someone decides to go back to a release prior to this pr and that will make the pr more complicated.

I am open to suggestions.
I don't want anyone to think we have made a change that we really haven't.
So, I didn't want to make too many changes.

I agree with what you are saying.  But, I am not sure if it will help someone who sees this for the first time or not.  Because the question remains even if we set the title to "No optimization prompt".  A user may wonder if I want no optimization, do I enable the switch or do I disable it.

We need a convention that we consistently follow.  For example, no negative switches ever.

asavageiv:
That makes sense. I agree with using positive settings and it would really be separate from this change. Can we keep it in the model layer as it is now but invert it in the UI only?

Navid200:
> Can we keep it in the model layer as it is now but invert it in the UI only?

You will need an even number of inversions or else, you will be flipping the logic.

jamorham (approved):
Ok this is fine. Another can of worms is should we give the user this option at all? I'm thinking lots of stuff will break with battery optimization left enabled and I can't think of a use case where this switch would ever be used. This was probably added back when Android 6 was brand new.

asavageiv:
Without analytics we can't really know how much it's used, but I'd be fine with it being removed.

Navid200:
I was about to open a PR to remove this setting and I saw this.  
<br/>  
  
Screenshot 2024-06-27 102512
  
<br/>  
I think we need to keep it.

PR #3543, opened 2024-06-16
Fix WearOS complication not working with G7

### LINKED ISSUE #303: Complication displays permanently null null ()

nielsvanvelzen:
This should fix #303. The G7 does not report as "G7 Native" like the G5/G6 does. Instead it just reports as "G7" causing the ListenerService abomination to fail when saving new values.

Related: #3220 is the pr introducing this change

### DIFF
--- a/wear/src/main/java/com/eveningoutpost/dexdrip/ListenerService.java
+++ b/wear/src/main/java/com/eveningoutpost/dexdrip/ListenerService.java
@@ -2418,7 +2418,7 @@ public class ListenerService extends WearableListenerService implements GoogleAp
                                     changed = true;
                                     bgData.save();
                                 } else {
-                                    if (bgData.source_info != null && (bgData.source_info.contains("Native") || bgData.source_info.contains("Follow"))) {
+                                    if (bgData.source_info != null && (bgData.source_info.contains("Native") || bgData.source_info.contains("Follow") || bgData.source_info.contains("G7"))) {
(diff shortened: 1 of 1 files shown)


Navid200:
Do you know the purpose of that conditional?  I don't.
I have no experience with WearOS?  
Why is non-native mode excluded?  I mean what if someone is using a G5 in non-native mode?  Why is that not qualified to be shown with WearOS?

I can see that your proposed change may fix a problem.  But, I'd like to know the reason for a line before changing it.
I mean if there is no reason to have that restriction, another solution would be to remove it altogether.  Otherwise, we may possibly have to make this fix every time a new device is added to xDrip.

jamorham:
@nielsvanvelzen Can you confirm that you have tested this change and it is working? thanks

nielsvanvelzen:
Unfortunately it's almost impossible to test the wear app without rebuilding the mobile app too. The change is easy enough to verify the working by looking at the changes from the linked PR, the logs from the linked issue and the changes in this one.

jamorham:
Ok we can try this out

nielsvanvelzen:
Updated to the latest nightly and I can confirm everything is working again!

Issue #3398, opened 2024-03-22
xDrip can show an overlap when switching hardware data source


Navid200:
I have been able to repeatedly cause this as explained here.

1- Set xDrip to collect from a G6 in a simulator.
2- Change the hardware data source to Nightscout follower from a different source.

![Screenshot_20240322-152426](https://github.com/NightscoutFoundation/xDrip/assets/51497406/2a747510-d786-48d6-b22b-bf4b6cf66886)
  
The overlap occurs when switching from a G6 collector to Nightscout follower.  It does not when switching from Nightscout follower to G6 collector.
  
---  
  
I think xDrip should only backfill for a time period that has no readings.
I'm not sure.  But, I suspect that xDrip will avoid backfilling if there is existing data for the exact time stamps.  But, I believe it should avoid it even if the timestamps are different as long as there are readings in the same time period.

minsartp:
I have the same issues when switching from Dexcom One to One+: 
In my case, it went as follows: I put on the new (and my first!) One+ Sensor yesterday around 21:54. I STOPPED my old (One) sensor this morning at around 7:51 (it still had +-24 hours left). I immediately changed the Transmitter ID to start the new sensor, which effectively happend at around 7:57. I immediately saw all backfilled data in xDrip and AAPS. The earliest value that was backfilled is from 22:22 yesterday, which is probably the very first value after the warmup period. The new and old values are not exactly in sync (old: every 5 minutes starting at xx:01 -> xx:06 -> xx:11 , ...) (new: every 5 minutes starting at xx:02 -> xx:07 -> xx:13 , ...)
![Screenshot_20240629_075913_xDrip+](https://github.com/NightscoutFoundation/xDrip/assets/47970734/9cf5db10-9d2c-4ed5-80aa-4ee59422cafb)

Navid200:
I just did another test.  This time, I switched from fake data source to Nightscout follower.  And there is an overlap.  
![Screenshot_20240629-121550](https://github.com/NightscoutFoundation/xDrip/assets/51497406/39496494-cdfe-4351-912a-364a88dc352b)  ![Screenshot_20240629-122221](https://github.com/NightscoutFoundation/xDrip/assets/51497406/d8f8f5fd-3f3f-4093-a53d-9ae29add08f8)

Navid200:
**G6 > stop > G7 right away**  
  
[...]

PR #3557, opened 2024-07-03
G7 backfill overlap fix

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/3398 
 
**How to Recreate the problem**  
You will need a G7 session in progress and a test phone.  
On the test phone, uninstall xDrip and install the latest Nightly (June 28, 2024).
If you have a G6 session in progress, establish connectivity to it.  If you don't set up fake data source.  
After establishing connectivity to your G6, wait for another read cycle so that it backfills the past 3 hours.  

Now, stop the session.  It is important to stop the xDrip session.  There is no reason to stop the real session on G6.  

After having stopped the xDrip session, change the transmitter ID to your G7 session in progress.  
After establishing connectivity to G7, wait for another read cycle so that the backfill is completed.  You will see that you will get a 24-hour backfill even though you have had 3 hours of readings from G6.  So, you will end up with an overlap.  
  
 **Tests**  
All tests were run on an Android 11 test phone.  

1- Establishing connectivity to a G7 after a fresh install results in a 24-hour backfill.  
2- Establishing connectivity to a G6 after a fresh install results in a 3-hour backfill.
3- Establishing connectivity to a G7 after having stopped a G6 session, a backfill takes place to fill only the empty time after the last G6 reading.  In other words, it does not cause an overlap.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/g5model/Ob1G5StateMachine.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/g5model/Ob1G5StateMachine.java
@@ -134,6 +134,7 @@ public class Ob1G5StateMachine {
     private static volatile AuthRequestTxMessage lastAuthPacket;
     private static volatile boolean backup_loaded = false;
     private static final int OLDEST_RAW = 300 * 24 * 60 * 60; // 300 days
+    private static long relAutoSessionStartTime = HOUR_IN_MS * 3;
 
     public static long maxBackfillPeriod_MS = 0;
 
@@ -1063,12 +1064,6 @@ public class Ob1G5StateMachine {
     }
 
     private static void backFillIfNeeded(Ob1G5CollectionService parent, RxBleConnection connection) {
-        if (!Sensor.isActive() && Ob1G5CollectionService.isG5SensorStarted() && getBestCollectorHardwareName().equals("G7")) { // If we are using a G7 coming from a G6
-            // and continue beyond this point without an xDrip session, we can end up with a backfill overlapping the last G6 session.
-            // So, let's start a session now.
-            JoH.static_toast_long(xdrip.gs(R.string.auto_starting_sensor));
-            Sensor.create(tsl() - HOUR_IN_MS * 24);
-        }
         final Sensor sensor = Sensor.currentSensor();
         if (sensor == null) {
             UserError.Log.e(TAG, "Cannot process backfill evaluation as no active sensor");
@@ -1557,14 +1552,19 @@ public class Ob1G5StateMachine {
     }
 
     private static void checkAndActivateSensor() {
-        // automagically start an xDrip sensor session if  transmitter already has active sensor
+        // automagically start an xDrip sensor session if transmitter already has active sensor
         if (!Sensor.isActive() && Ob1G5CollectionService.isG5SensorStarted() && (!Sensor.stoppedRecently() || shortTxId())) {
             JoH.static_toast_long(xdrip.gs(R.string.auto_starting_sensor));
-            // TODO possibly here we want to look at last sensor stop time and not backtrack before that
-            Sensor.create(tsl() - HOUR_IN_MS * 3);
-            if (shortTxId()) { // If we are using G7
-                Sensor.create(tsl() - HOUR_IN_MS * 24);
+            if (shortTxId()) relAutoSessionStartTime = HOUR_IN_MS * 24; // If we are using a G7
+            final List<BgReading> last = BgReading.latest(1); // Last reading
+            if ((last != null) && (last.size() > 0)) { // Have we had a reading?
+                final long now = JoH.tsl();
+                final long since = now - last.get(0).timestamp; // Time since last reading
+                if (since < relAutoSessionStartTime) { // If the last reading was less than 3 hours ago, or if we are using G7 and the last reading was less than 24 hours ago
+                    relAutoSessionStartTime = since; // We will start the new session starting from the last reading.
+                }
             }
+            Sensor.create(tsl() - relAutoSessionStartTime);
         }
     }
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/g5model/Ob1G5StateMachine.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/g5model/Ob1G5StateMachine.java
@@ -92,6 +92,7 @@ import static com.eveningoutpost.dexdrip.utilitymodels.Constants.DAY_IN_MS;
 import static com.eveningoutpost.dexdrip.utilitymodels.Constants.HOUR_IN_MS;
 import static com.eveningoutpost.dexdrip.utilitymodels.Constants.MINUTE_IN_MS;
 import static com.eveningoutpost.dexdrip.utilitymodels.Constants.SECOND_IN_MS;
+import static com.eveningoutpost.dexdrip.utils.DexCollectionType.getBestCollectorHardwareName;
 import static com.eveningoutpost.dexdrip.utils.bt.Helper.getStatusName;
 
 
@@ -1062,6 +1063,12 @@ public class Ob1G5StateMachine {
     }
 
     private static void backFillIfNeeded(Ob1G5CollectionService parent, RxBleConnection connection) {
+        if (!Sensor.isActive() && Ob1G5CollectionService.isG5SensorStarted() && getBestCollectorHardwareName().equals("G7")) { // If we are using a G7 coming from a G6
+            // and continue beyond this point without an xDrip session, we can end up with a backfill overlapping the last G6 session.
+            // So, let's start a session now.
+            JoH.static_toast_long(xdrip.gs(R.string.auto_starting_sensor));
+            Sensor.create(tsl() - HOUR_IN_MS * 24);
+        }
         final Sensor sensor = Sensor.currentSensor();
         if (sensor == null) {
             UserError.Log.e(TAG, "Cannot process backfill evaluation as no active sensor");


Navid200:
I thought I had solved the problem.  But, further tests showed failure.
So, I have taken a different approach, being executing your ToDo comment.  That's what this PR does.

jamorham:
I'm not sure whether this actually look at the last sensor stop time and not backtrack before that rather than looking at when readings have been received?

Navid200:
I don't understand your point.
If the last reading is from a sensor that has not stopped yet, we still should not overlap it.
If that last reading is from Nightscout follow, and we are now switching to become a collector, even though the sensor we were following with Nightscout has not stopped yet, we should not backfill over the readings.  

If we switch from a G6 transmitter to another, and the first one has not stopped yet, we should not overlap its readings.

Did I misunderstand your point?

Navid200:
The overlap actually uploads to Nightscout and AAPS.  It is really bad.
Can you give me any scenario where you would like xDrip to backfill over a previous reading?

Navid200:
There are times a G7 user chooses to switch to their next G7 before their current 10.5 days are completed.
In that case, the previous sensor never stops and they start using the new sensor.
We still should not backfill over it if the new sensor started a few hours before they switch to it.

jamorham:
If the sensor is stopped then there should not be any readings for the period while the sensor is stopped. Here we are activating the sensor and I'm thinking we should cap how far back we backtrack the start time of this sensor to not before the end of the previous sensor session. That is what my todo says. 

If you look only at the readings then you potentially are overlapping sensor periods which I think is the cause of the problem overall.  So you can use `Sensor.lastStopped()` to see if there is a previous sensor and `stopped_at` to find the time it stopped and use that probably with `Math.max()` to cap the new sensor start time so it can't be before the previous one. Indeed this could even be a method added to `Sensor` to allow it to be fed any start time but it then will apply the cap based on any previous sensor and always start it safely which would probably be cleaner overall.

Navid200:
If the previous session has never stopped in xDrip, your suggestion will fail to avoid a backfill overlap.

This PR will avoid a backfill overlap under all possible conditions.

Navid200:
> 
> If you look only at the readings then you potentially are overlapping sensor periods which I think is the cause of the problem overall. 

I respectfully disagree.
Sensor periods overlap.  This is nothing new.  People have been presoaking G6 for a very long time.
What's new is that Dexcom has offered a 12-hour grace period.  Therefore, there are many G7 users who presoak every sensor by 12 hours.  So, the sensor periods overlap.  This is not a problem.  This is a fact of life and xDrip can cope with it easily.  

I didn't want to make this solution only for G7 because there have been reports of overlap even for G6.
So, I took a fundamental approach and deals with the cause of the problem at its source.
If we have a reading, we have already uploaded it to Nightscout and AAPS.  There is absolutely no justification for overlapping a different reading with it, which would be rewriting history.  So, we  avoid it.  If we have a reading, we never backfill over it.

Can you tell me how this is ever going to allow an overlap?

I have shown the different scenarios that can cause an overlap here: https://github.com/NightscoutFoundation/xDrip/issues/3398  

This PR avoids all of those.  
Would you please tell me under what condition this PR can cause a problem?

Navid200:
Let's say the user started their current G7 at 11:00 am 10 days ago.
Tonight, at 11:00 pm, their session will stop.
But, at 10:00 pm, they decide to go to bed and in order to go to bed, they switch to their new sensor, which they inserted this morning at 8:00 am, right then at 10:00 pm. 
This is 1 hour before the current session will expire.
When they switch, there will never be a session stop for the current sensor.  

If I use last stopped, it will find the session that stopped 10 days ago.  And there will be an overlap.  
This PR sees that the last reading was at 9:55 pm tonight.  So, it only backfills starting from then avoiding any possibility of overlap.

Would you please tell me what I am overlooking?

Navid200:
I am sorry that I misunderstood your TODO comment.
I can put that back in.
 
Regardless, I believe this is the best solution.  

Please give me a scenario that you believe this PR will fail, and I will try to test exactly the scenario you are concerned about.  Either I will learn that I am missing something and will change this PR back to draft, or I can show you that the scenario works fine after this PR and there should be no concern.

jamorham:
So I think the situation is that with G7 and possibly with G6 we get that the xDrip sensor session is not tied to the actual sensor (eg if you change transmitter id) - This was not the original intention of the Sensor class but that dates back to G4 and where we did the algorithm internally. In that case you are correct that it is the readings we should be looking for. 

The only problem I can see is if we somehow got readings from elsewhere that were not yet synced (via follow perhaps) but I think that is very much an edge case where the user would have to try hard to break things and then they would likely only incorrectly fill in any open gaps in the data.

Navid200:
Yes, I agree 100%.

Gaps should be avoided.  
If a gap is caused accidentally, having the Google Drive daily backup enabled may offer a possible solution.  In that case, the user can restore their latest daily backup;  then, use their current G7 to backfill the past 24 hours filling all gaps.
But, the best solution is to avoid causing gaps, which will not be backfilled over after this is merged.

Navid200:
Please, please, don't ask for a switch!    
:o)

jamorham (approved):

Navid200:
Thanks

I will make an announcement on facebook after merge.

Navid200:
I'm not sure if this is working as intended.  I still see overlaps uploaded to Nightscout and also in xDrip history.
This is from 2 days ago using the latest Nightly:
Screenshot 2024-08-25 103103


I will be watching very carefully on my next sensor change, in 8 days, and report here.

Navid200:
I'm sorry I don't think this was a perfect solution.
A problem I have is that after I test a PR and submit it, I will have to switch back to the latest official Nightly so  that I can see what everyone else sees.
Most developers can continue to use their PR.  I cannot.  If I do, I will be isolating myself from what most users use.  I am supposed to provide a link between the user community and the developers.  So, I cannot use my PR for a long time.  There may be a conflict between me as a developer and me as a support staff.  I am sorry.

I will have to open a new PR to address this and I am so sorry.

Der-Schubi:
Hi @Navid200 
I'm on a build from 12th of November. I started a new G7 while the old one had 2 hours or so left. And I kept the old sensor active until it expired. Then I changed the code and connected the new one.
After first connection with the new sensor xDrip backfilled all data til sensor start, leading to overlapping values.

Navid200:
@Der-Schubi Yes, I have experienced this as well.
I believe this PR only addressed going from G6 to G7, but not going from G7 to G7.
It has been my plan to fix it.

Thanks

Navid200:
@Der-Schubi Can you show the logs for when the backfill was made?  I expect it would have happened some time around 12:30.

Der-Schubi:
That was two days ago, but I'll have a look if there are logs left.

Der-Schubi:
`5 ^ 2025-01-09 12:34:39 ^ Ob1G5StateMachine ^ Saving battery data: Status: OK / VoltageA: 295 / VoltageB: 255 / Resistance: 7680 / Run Time: 0 / Temperature: 0
2 ^ 2025-01-09 12:34:39 ^ Ob1G5StateMachine ^ Store: BatteryRX dbg: 22002701FF00001E0000000001
2 ^ 2025-01-09 12:34:39 ^ G5CollectionService ^ BatteryInfoTx dbg: 222004
6 ^ 2025-01-09 12:30:56 ^ Snoozed Alert ^ Snoozed until: 2025-01-09 14:30:56
2 ^ 2025-01-09 12:30:56 ^ AlertPlayer ^ SnoozeOnNotificationDismissService called source = bg_alerts shown for: 12 seconds
6 ^ 2025-01-09 12:30:44 ^ Alerting ^ BG HIGH ALERT: 12,4  (@12:30)
2 ^ 2025-01-09 12:30:44 ^ jamorham JoH ^ setMediaDataSource from uri failed: uri = content://settings/system/notification_sound
java.io.FileNotFoundException: Direct file access no longer supported; ringtone playback is available through android.media.Ringtone
    android.database.DatabaseUtils.readExceptionWithFileNotFoundExceptionFromParcel(DatabaseUtils.java:164)
    android.content.ContentProviderProxy.openTypedAssetFile(ContentProviderNative.java:814)
    android.content.ContentResolver.openTypedAssetFileDescriptor(ContentResolver.java:2043)
    android.content.ContentResolver.openAssetFileDescriptor(ContentResolver.java:1858)
    android.content.ContentResolver.openFileDescriptor(ContentResolver.java:1689)
    android.content.ContentResolver.openFileDescriptor(ContentResolver.java:1635)
    com.eveningoutpost.dexdrip.models.JoH.setMediaDataSource(JoH.java:1178)
    com.eveningoutpost.dexdrip.utilitymodels.AlertPlayer.playFile(AlertPlayer.java:369)
    com.eveningoutpost.dexdrip.utilitymodels.AlertPlayer.VibrateNotifyMakeNoise(AlertPlayer.java:579)
    com.eveningoutpost.dexdrip.utilitymodels.AlertPlayer.ClockTick(AlertPlayer.java:324)
    com.eveningoutpost.dexdrip.utilitymodels.Notifications.FileBasedNotifications(Notifications.java:249)
    com.eveningoutpost.dexdrip.utilitymodels.Notifications.notificationSetter(Notifications.java:329)
    com.eveningoutpost.dexdrip.utilitymodels.Notifications.onHandleIntent(Notifications.java:142)
    android.app.IntentService$ServiceHandler.handleMessage(IntentService.java:77)
    android.os.Handler.dispatchMessage(Handler.java:109)
    android.os.Looper.loopOnce(Looper.java:232)
    android.os.Looper.loop(Looper.java:317)
    android.os.HandlerThread.run(HandlerThread.java:85)

2 ^ 2025-01-09 12:29:40 ^ Ob1G5StateMachine ^ Store: VersionRX dbg: 5200C0D70D00540600020404FF0C00
2 ^ 2025-01-09 12:29:40 ^ Ob1G5StateMachine ^ Wrote version request: 2
2 ^ 2025-01-09 12:29:40 ^ Ob1G5StateMachine ^ Cannot process backfill evaluation as no active sensor
6 ^ 2025-01-09 12:29:40 ^ Ob1G5CollectionService ^ Native Sensor is now Started: OK 1.6 hours
2 ^ 2025-01-09 12:26:56 ^ Ob1G5StateMachine ^ Store: VersionRX dbg: 4A0020C06A002A34000032454141B8979FD36100
2 ^ 2025-01-09 12:26:56 ^ Ob1G5StateMachine ^ Wrote version request: 1
3 ^ 2025-01-09 12:26:56 ^ Ob1G5StateMachine ^ New BgReading was null in processGlucoseRxMessage!
1 ^ 2025-01-09 12:26:56 ^ BgReading ^ No sensor, ignoring this bg reading
2 ^ 2025-01-09 12:26:56 ^ DexTimeKeeper ^ Refusing to update stored timestamp due to excessive processing latency: 19071ms
2 ^ 2025-01-09 12:26:56 ^ Ob1G5StateMachine ^ Cannot process backfill evaluation as no active sensor
2 ^ 2025-01-09 12:26:56 ^ Ob1G5CollectionService ^ onReceive UPDATE Name DXCMAz Value C9:8A:29:89:C6:3B Bond state 12 Paired bs:  Paired was  Pairing
5 ^ 2025-01-09 12:26:50 ^ Ob1G5CollectionService ^ Prompting user to notice pairing request with sound - On Android 8+ you have to manually pair when requested
2 ^ 2025-01-09 12:26:49 ^ Ob1G5CollectionService ^ onReceive UPDATE Name DXCMAz Value C9:8A:29:89:C6:3B Bond state 11 Pairing bs:  Pairing was  Unpaired
2 ^ 2025-01-09 12:26:49 ^ DexTimeKeeper ^ Clearing stored timing sync information for: 5949
2 ^ 2025-01-09 12:26:34 ^ DexTimeKeeper ^ No valid timestamp stored for transmitter: 5949
2 ^ 2025-01-09 12:26:06 ^ Ob1G5StateMachine ^ Cannot process backfill evaluation as no active sensor
6 ^ 2025-01-09 12:26:06 ^ Ob1G5CollectionService ^ Native Sensor is now Stopped: Sensor Expired
6 ^ 2025-01-09 12:26:06 ^ SENSOR ^ Sensor stopped at 2025-01-09 12:26:06
5 ^ 2025-01-09 12:26:06 ^ Ob1G5CollectionService ^ Marking sensor session as stopped`

PR #3558, opened 2024-07-05
Delay G6 start prompt by 5 more minutes!

Navid200:
We delayed the G6 start prompt until transmitter days were known here: https://github.com/NightscoutFoundation/xDrip/pull/2757
That change was merged on June 24, 2023.  
Since then, we have had 3 stable releases.  Yet, there are still reports of G6 sensors not starting on new transmitters and start commands getting stuck in the queue.  

As soon as the transmitter days parameter is known, we show the sensor start prompt.  When that happens, if you look at the status page, you will still see hunting transmitter for 5 more minutes.

Please let's delay the prompt until hunting transmitter is not shown any longer.  That effectively, means 5 more minutes of delay.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -70,6 +70,7 @@ import android.widget.ImageButton;
 import android.widget.TextView;
 import android.widget.Toast;
 
+import com.eveningoutpost.dexdrip.g5model.DexSyncKeeper;
 import com.eveningoutpost.dexdrip.g5model.DexTimeKeeper;
 import com.eveningoutpost.dexdrip.g5model.Ob1G5StateMachine;
 import com.eveningoutpost.dexdrip.g5model.SensorDays;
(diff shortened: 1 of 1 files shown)


Navid200:
What does this effectively change from the user's perspective?  

1- After entering the transmitter ID for a G6 transmitter, a red note is shown at the top of the main screen telling you to "verify settings or wait for connectivity".  This note is now shown for 5 more minutes before changing to "Now start sensor ...".  It could also show indefinitely if there is a transmitter connectivity issue.  
  
2- When you establish connectivity to a G6 transmitter, a "Start Sensor" prompt comes on screen asking you to start.  This prompt is shown 5 minutes later than before, or never as long as there is no synch established with the transmitter.

Navid200:
Without this change, we already have a conditional delay on allowing user to start a sensor here: 
https://github.com/NightscoutFoundation/xDrip/pull/2746  

But, that seems like a contradiction.  If we tell user that they can start and then, tell them they cannot start until there is connectivity, it seems like a bug.
But, my suspicion is that in some cases, users have been allowed to proceed with starting.  

This change will make everything consistent.  If we are not going to allow start, we will also not ask the user to go ahead and start.  
And I hope this will reduce the number of issues dues to the start command getting stuck in the queue.

Navid200:
**How to verify the UI contradiction**
  
Install the latest Nightly (June 28, 2024) on a test phone after an uninstall.  Remove all Dex devices from the list of previously connected devices.  

Set up xDrip to connect to a G6 transmitter with no session in progress.  
As soon as you approve a pair request, go to the main screen.  You will see the start prompt asking you to start.  
Screenshot_20240705-104524  
  
Go ahead and start right away.  You will see a note that you cannot start a sensor until there is connectivity with the transmitter.  
Screenshot_20240705-104540

jamorham (approved):
Looks sensible

PR #3572, opened 2024-07-16
Disable "Restart Sensor" when incompatible

Navid200:
This PR will disable `Restart Sensor` for devices that cannot be restarted.  
  
Without this, an inexperienced user may enable the setting and when the time comes for the next sensor, it may take up to half an hour for their restart attempt to fail and be obvious that something may be wrong.  
  
I have also seen a case where the failure would happen in just a few minutes, but it would attempt another restart.
The user had posted on facebook asking for help.
It took quite a while to realize that they had enabled restart sensor for a new G6.  

I cannot think of any case that would be at a disadvantage after including this.  Can you think of a scenario where this PR would disable restart sensor where it would actually be possible?  

G5 and very old G6 devices are the only ones that can benefit from enabling restart sensor.  This PR does not change that.  They will still be able to enable, and use, restart.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/g5model/Ob1G5StateMachine.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/g5model/Ob1G5StateMachine.java
@@ -1725,6 +1725,12 @@ public class Ob1G5StateMachine {
                     UserError.Log.uel(TAG, "Dex Native mode enabled.  For your device, non-native mode is either not possible or not recommended.");
                     JoH.showNotification("Enabled Native", "Native mode enabled", null, Constants.G6_DEFAULTS_MESSAGE, false, true, false);
                 }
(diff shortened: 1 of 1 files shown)


Navid200:
**Testing**  

I have enabled restart sensing before this and it remains enabled for a G6.
I then used this PR and after 10 minutes, the setting  was automatically disabled.  

Unfortunately, I don't have a G5 or very old G6 to test.

jamorham:
I think this is okay but you might want to use a higher log level like `ueh` or `wtf` to ensure that the event is very visible in the logs. It should be a one time only thing and informing the user of a forced change so I think it is good to make it very obviousl

Navid200:
Done!

Navid200:
I made the change you asked for.

jamorham (approved):

Navid200:
Thanks so much for the review.

I know that reviewing someone else's work is a boring job.  The person who does the work enjoys the creativity aspect.  The reviewer is looking at his/her work.

It is a hard, challenging and tiring work.  Thanks for doing it.

PR #3575, opened 2024-07-19
Retire allow OB1 initiate bonding

Navid200:
This PR will remove another setting from the Dex Settings page.  

I cannot think of a scenario when one may need to disable this.  Of course there is a case where you don't intend to connect to a device.  But, why would you enter the transmitter ID if you don't intend to connect.  I mean you can avoid xDrip attempting to connect by just disabling collector.  

Considering there is an alternative way to accomplish the same thing, I hope you allow this setting to be removed.
A considerable percentage of the requests for help is related to wrong settings on this page.
Even though we have a guide that shows the correct settings, incorrect settings still appear on facebook requests for help.
So, I wish to minimize the number of settings on this page.  

Is there a scenario in which we may need to disable this setting?

Thanks

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
@@ -1785,7 +1785,7 @@ public class Ob1G5CollectionService extends G5BaseService {
 
 
     private boolean getInitiateBondingFlag() {
-        return Pref.getBoolean("ob1_initiate_bonding_flag", true);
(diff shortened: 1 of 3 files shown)


Navid200:
This is what the page will look like after this for a G6.  
  
Screenshot_20240718-202632

jamorham:
I don't think you full understand how this works. Either end of a bluetooth connection can initiate the bonding. There is designed to be some adaptability in the code for this although much of it is moot after google changed bluetooth bonding behavior during covid. I do agree with removing the setting and forcing it to true but the code paths should remain in case we need to bring back that flexibility.

jamorham (approved):

PR #3579, opened 2024-07-20
fix missing timestamp on ongoing BG notification

richard-dennehy:
Fixes the issue described in https://github.com/NightscoutFoundation/xDrip/discussions/3578

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Notifications.java
@@ -651,6 +651,8 @@ public class Notifications extends IntentService {
         if (lastReading != null) {
 
             b.setWhen(lastReading.timestamp);
+            b.setShowWhen(true);
+
             final SpannableString deltaString = new SpannableString("Delta: " + ((dg != null) ? (dg.spannableString(dg.unitized_delta + (dg.from_plugin ? " "+context.getString(R.string.p_in_circle) : "")))
                     : bgGraphBuilder.unitizedDeltaString(true, true)));
 


jamorham:
Thanks for this. I believe this could work but also there was at one point a problem where every update to the persistent notification would cause a notification on Android Wear. I am not sure whether enabling this could cause that to reoccur.

richard-dennehy:
I don't have an Android Wear device to test this on, so I can't confirm either way, but it looks to me like the existing code [here](https://github.com/NightscoutFoundation/xDrip/blob/9ad1d595a2673f9b950cd75717fbae3e8a2c1cc8/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Notifications.java#L717) is intended to work around the potential issue you're describing

Navid200:
@PhilTU By any chance, will you be able to use your WearOS device, and a test phone, to do a test if I provide a test xDrip version to you?  
I'm sorry for putting you on the spot.  Please don't hesitate to decline.

PhilTU:
> @PhilTU By any chance, will you be able to use your WearOS device, and a test phone, to do a test if I provide a test xDrip version to you? 

Could I use the built in data generator rather than a sensor? If so, yes.

Navid200:
> Could I use the built in data generator rather than a sensor? If so, yes.

Yes, of course: https://navid200.github.io/xDrip/docs/FakeDataSource.html 

Do you mind sending me an email so that I can provide the test xDrip unless you can compile it yourself?
xDrip dot Navid at gmail dot com

Navid200:
@PhilTU This is the test xDrip: 
https://drive.google.com/file/d/1B2EuOGpxyeb80yU31vyLmZY7KJNOp0EQ/view?usp=sharing 
Please let me know when you have downloaded it.  

With respect to the tests, please kindly confirm the following: 
1- On the phone, pulling down from the top of the screen to expose the ongoing notification should show the time in addition to the value. 
2- On the watch, there should not be a new notification once every 5 minutes.  

Thanks

PhilTU:
Download and installed. Noted re version / signature. Awaiting arrival of mock data...

PhilTU:
Confirm that notifications now include relative time stamp. WearOS tomorrow....

PhilTU:
WEAROS configured and showing delta time display as expected. No notifications received as yet (of any sort) certainly not every 5 minutes as per item 2.

DSC_2375

PhilTU:
I do have frequent event log entries on the phone (one or two per minute)

"AlertPlayer

13:15

setting next alert to 1 minute from now (no problem right now, but needs a fix someplace else)"

This started after a low alert...

Screenshot_20240806-132730

Navid200:
@PhilTU Thanks so much for the tests.

I have seen that log (... but needs a fix somewhere else).  But, I have never got to work on it yet.  I will figure it out some time.

Thanks so so much for running the tests.

jamorham (approved):

richard-dennehy:
Thanks, all 👍

PR #3589, opened 2024-07-24
Adding more strings for translation in crowdin

olorinmaia:
Following up on https://github.com/NightscoutFoundation/xDrip/pull/3252

Built app and installed on test phone to verify that all new strings appear in the right place. 

There are still some strings to be added for translation but I stop here so there won't be to much to review.


Navid200:
Thanks for spending the time on this.  

I am sorry I have a problem with this.  Looking at this, I have no way of knowing exactly where they are used.  

Years ago, a string related to time was changed to be translated.  One of the places it is used is on the status page.
So, the Dex status page showing how long ago the last connection was is translated now.
I see a status page a user posts because they need help and it is in a language I don't speak.
I have no idea how long ago the last connection was.  So, I have to translate it back to English.  I provide support to xDrip users and I have to spend time and possibly make a mistake translating a phrase that shows up on a status page.  

I attempted to fix this and you can see it here:  https://github.com/NightscoutFoundation/xDrip/pull/3303  
But, it was never even reviewed.  So, I eventually closed it.  It is hard to change something after it has been released to users.  

We should never translate a phrase that can appear either in the logs or on a status page. 
Can you confirm that none of these is ever going to be used either in the logs or on a status page?

Thanks

olorinmaia:
> We should never translate a phrase that can appear either in the logs or on a status page.
> Can you confirm that none of these is ever going to be used either in the logs or on a status page?

I agree with you on this and I can confirm at least of my best knowledge that these new strings shouldn't affect status page or logs. 

I was thinking of adding strings to translate various sensor statuses (like warming up, stopped, sensor error etc), but I'm afraid they also then will appear in status page so I left that out for know. I will look closer at that later, as I need to test with active sensor on my self to be sure it doesn't interfere with status page. If that is possible I will make a dedicated PR for that.

This PR covers mainly various options in settings, that will help users that doesn't understand english well to get them translated in local language.

Navid200:
Great, thanks

Navid200:
I'm back!

Do you see this comment?
https://github.com/NightscoutFoundation/xDrip/pull/3566#issuecomment-2232980390  

After phrases are added to the list of translated ones, it will become harder to make any corrections to them.  
So, before adding a phrase, we have to make sure we are happy with it.  
I can give you a list of items here that I have concerns about.  
Just one example:

Screenshot 2024-07-25 001708
These are volume profiles.  We can call them all volumes.  Like high volume, medium volume and ascending volume.  But, we have not.  We have called them high, medium and ascending volume.  This is not extremely clean.  There is room for improvement.  If we translate this, we probably never touch them.  Not because they are perfect.  But, just because it becomes harder.  

You have chosen to translate BlueJay GTS settings.  I don't believe you can buy a GTS watch in any language other than English.  Please correct me if I am wrong.  So, if the user can use the watch looking at it everyday and not need translation, why would he/she need the settings inside xDrip related to GTS that he only needs to adjust once every blue moon to a different language?  Do you really think this will help anyone?

You have translated the names of the graphic arrows.  But, those are effectively names.  There should be no need to translate names.  

Now, going back to my original concern, I can create a new setting and choose to use an existing string.  How do you know where all of these strings have been used?  Have you done a search on every one of them to make sure none is used in logs or status pages?

olorinmaia:
Thanks for your comment / review @Navid200  :)

> After phrases are added to the list of translated ones, it will become harder to make any corrections to them. So, before adding a phrase, we have to make sure we are happy with it. I can give you a list of items here that I have concerns about. Just one example:
> 
> Screenshot 2024-07-25 001708 These are volume profiles. We can call them all volumes. Like high volume, medium volume and ascending volume. But, we have not. We have called them high, medium and ascending volume. This is not extremely clean. There is room for improvement. If we translate this, we probably never touch them. Not because they are perfect. But, just because it becomes harder.
> 

I agree when first adding new strings to translate it's a good time to adjust anything that is unclear or obviously should be improved. Your example could be changed as you suggest while we are adding this as it makes it more clean.

If there is any other strings you think should be improved I'm open for suggestions, for me nothing else sticks out. I can adjust PR according to that or other strings that needs adjusting, before getting pushed for translation.

> You have chosen to translate BlueJay GTS settings. I don't believe you can buy a GTS watch in any language other than English. Please correct me if I am wrong. So, if the user can use the watch looking at it everyday and not need translation, why would he/she need the settings inside xDrip related to GTS that he only needs to adjust once every blue moon to a different language? Do you really think this will help anyone?
> 
I myself got a bluejay GTS, and if the user doesn't understand english very well having the settings translated is helpful, because the UI in Bluejay GTS is only icons, no text. If more people mean that this isn't worth adding I can remove it from PR.

> You have translated the names of the graphic arrows. But, those are effectively names. There should be no need to translate names.
> 
I thought it made sense to translate, as names describe what they do, why should there suddenly inside an option appear non-translated options? An app like xDrip should have close to 100% translated text for all content (except for maybe system status page), as it helps the users to navigate and understand the app when all text is in their native language.

Adding strings for translation can also discover in some cases (like we've seen in AAPS), that over time strings become outdated, more people will review when they translate strings, and give feedback to devs that the original string is due for an update. Then a simple PR can be made to update those strings to make the app even more precise.

> Now, going back to my original concern, I can create a new setting and choose to use an existing string. How do you know where all of these strings have been used? Have you done a search on every one of them to make sure none is used in logs or status pages?
I haven't changed any existing string as far as I know, perhaps I don't understand what you mean also? :)

Navid200:
If someone's family name is Brown, you wouldn't translate that when talking to them in a different language, would you?  

JamOrHam is someone's avatar.  It is a name.  There is no reason for the word jam referring to preserved fruit to exist in xDrip and it does not.  When you see Jam, it is a short for JamOrHam.  And as such it would make no sense to translate it for the same reason you wouldn't translate Bluetooth when talking about technology.  
<br/>  

We are not translating the logs.  
Are you sure that none of the items you are adding to be translated is used in logs or status pages?

olorinmaia:
Bluetooth in Norwegian stays the same f. ex., but maybe not in other languages, its up to the translators at crowdin to judge for their own language, and i trust they do a good job in that matter. 

I agree i should leave out 'Jam' for the strings to be translated as it could be hard to know that this is a nickname of a person and shouldnt be translated 😅

The most important thing is that users of given language understand the app as well as possible. Its also an option, if you want to use an other language than standard english. 

Log entries are generated seperately as far as i know, and ive never seem strings being placed in these, i havent put strings for translation into these.

olorinmaia:
Closing as some of these strings are included in https://github.com/NightscoutFoundation/xDrip/pull/3716

Will add another one later and leave out bluejay and the other strings @Navid200 suggest shouldn't be translated

PR #3593, opened 2024-07-26
Scan QR code from image file

Navid200:
There are times a QR code scan is not possible.
This is an example: https://github.com/NightscoutFoundation/xDrip/discussions/2960 

This PR adds a new capability allowing the user to scan the QR code from a saved image file.  
<br/> 

---  
  
**Tests**  
Android 8, 9, 11 and 15 have been tested.  
The OB1 debug QR code, a saved settings QR code from xDrip, and the G7 QR code have been used for tests.  
  
---  
  
If you have any comments for improving the new strings, please do not hesitate to tell me.  
  
If you see anything in the code that you don't like, please tell me.  

Thanks

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -39,7 +39,7 @@
     <string name="pref_header_cloud_storage">Cloud Storage</string>
     <string name="auto_configure_title">Auto configure</string>
     <string name="prefs_auto_config_summary">Auto configure using a barcode.</string>
-    <string name="auto_config_cam_title">Scanner</string>
+    <string name="auto_config_cam_title">Camera</string>
     <string name="auto_config_cam_summary">Scan a QR code with your mobile (recommended)</string>
     <string name="auto_config_image_title">Image file</string>
     <string name="auto_config_image_summary">You can save a QR code as an image file (such as .png or .jpeg) on your mobile. Then, use this option to scan the QR code from the saved file.</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -39,6 +39,10 @@
     <string name="pref_header_cloud_storage">Cloud Storage</string>
     <string name="auto_configure_title">Auto configure</string>
     <string name="prefs_auto_config_summary">Auto configure using a barcode.</string>
+    <string name="auto_config_cam_title">Scanner</string>
+    <string name="auto_config_cam_summary">Scan a QR code with your mobile (recommended)</string>
+    <string name="auto_config_image_title">Image file</string>
+    <string name="auto_config_image_summary">You can save a QR code as an image file (such as .png or .jpeg) on your mobile. Then, use this option to scan the QR code from the saved file.</string>
     <string name="scan_share2_barcode">Scan Share Barcode</string>
     <string name="pref_share2_scan_barcode_summary">Or scan the barcode on the share receiver</string>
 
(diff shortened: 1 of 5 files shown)


Navid200:
After tis PR, tapping on auto config, will bring you to a new page shown below.  
Screenshot_20240726-181754
<br/>  

Tapping on the button at the top does exactly what auto config button does now.  
Tapping on the button at the bottom lets user go through the storage and choose an image file.  
If the image file has the right format, it will be imported.  if it doesn't appropriate logs will be issued.

Navid200:
Tapping on the button at the bottom on the new page to scan from an image file, the user will see a page like this:  
<br/>  
<br/>  
  
Screenshot_20240726-182730

Navid200:
I just tested this using a Pixel (Android 7) virtual machine and it worked to confirm that it works on Android 7 as well.

app/src/main/res/values/strings.xml
```diff
@@ -39,6 +39,10 @@
     <string name="pref_header_cloud_storage">Cloud Storage</string>
     <string name="auto_configure_title">Auto configure</string>
     <string name="prefs_auto_config_summary">Auto configure using a barcode.</string>
+    <string name="auto_config_cam_title">Scanner</string>
+    <string name="auto_config_cam_summary">Scan a QR code with your mobile (recommended)</string>
```
jamorham, line 43: shouldn't you mention "camera" here?

app/src/main/res/xml/pref_data_sync.xml
```diff
@@ -8,9 +8,24 @@
...
-            android:key="auto_configure"
+            android:key="auto_config_top"
             android:summary="@string/prefs_auto_config_summary"
-            android:title="@string/auto_configure_title" />
+            android:title="@string/auto_configure_title">
+            <PreferenceScreen
+                android:key="auto_configure"
+                android:summary="@string/auto_config_cam_summary"
+                android:title="@string/auto_config_cam_title" />
+            <Preference
```
jamorham, line 18: Can't we just have these one underneath each other in the same way as other preferences?

app/src/main/res/values/strings.xml
Navid200: I spent some time with ChatGPT for these.  But, I see your point.
The problem is if I focus on the meaning of the sentence, saying scan with your camera, it may be taken as take a camera, not your phone, out of your pocket and use it to scan.

What if I change that to the following?

Scan a QR code with your mobile camera

app/src/main/res/xml/pref_data_sync.xml
Navid200: Definitely!  I can do that.

The only reason I did this was because I wanted to give more prominence to the existing scheme. I mean if user can use camera, they should never use the second option.
 That's the only reason I did this.

app/src/main/res/values/strings.xml
Navid200: Sorry, I even confused myself.  I will think about this and improve the strings.

Navid200: How about this?

Title: Scanner (camera)
Summary: Scan a QR code with your mobile camera

app/src/main/res/xml/pref_data_sync.xml
Navid200: I have made both of those changes.

This is what it looks like now:
Screenshot_20240802-131859

asavageiv: I'll probably keep the QR code on my storage somewhere and almost always use that instead of Camera. That way I don't have to hunt for the code every time I need to reconfigure my app for G7

jamorham (approved):

PR #3596, opened 2024-07-27
Don't translate Last Connected

Navid200:
This is what is shown on the Dex status page when a language other than English is selected.

Screenshot_20240727-141047
  
As you can see some items are translated and some aren't.  
We don't translate anything in the logs.  For the same reason, we also should not translate items on the status page that are needed for helping someone troubleshoot.  
  
The last connected item is very critical.
If last connected is less than 5 minutes, it is a good sign.  if it is not, there could be a connectivity issue.  When there is a problem, this parameter is one of the first I need to see when helping someone troubleshoot.  
But, if it is translated, I will have to translate it back.  
  
This PR changes the behavior such that last connected on the Dex status page is not translated as you can see here:  
Screenshot_20240727-142102
  
---  
  
My first attempt at this was very primitive: https://github.com/NightscoutFoundation/xDrip/pull/3303  
I have now added an optional parameter to the method.  
All the existing calls go to the method without the new parameter resulting in translation as they did before this PR.  
Only one call to the method, on the Dex status page, sets the new parameter to true forcing the language to be English.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/JoH.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/JoH.java
@@ -751,27 +751,31 @@ public class JoH {
         return niceTimeScalar(-msSince(t));
     }
 
-    // temporary
     public static String niceTimeScalar(long t) {
-        String unit = xdrip.getAppContext().getString(R.string.unit_second);
+        return niceTimeScalar(t, false); // Calling the method with forceEn set to false so that the chosen language is used.
+    }
+
+    // temporary
+    public static String niceTimeScalar(long t, boolean forceEn) { // English will be used regardless if forceEn is true
+        String unit = forceEn? "second" : xdrip.getAppContext().getString(R.string.unit_second);
         t = t / 1000;
-        if (t != 1) unit = xdrip.getAppContext().getString(R.string.unit_seconds);
+        if (t != 1) unit = forceEn? "seconds" : xdrip.getAppContext().getString(R.string.unit_seconds);
         if (t > 59) {
-            unit = xdrip.getAppContext().getString(R.string.unit_minute);
+            unit = forceEn? "minute" : xdrip.getAppContext().getString(R.string.unit_minute);
             t = t / 60;
-            if (t != 1) unit = xdrip.getAppContext().getString(R.string.unit_minutes);
+            if (t != 1) unit = forceEn? "minutes" : xdrip.getAppContext().getString(R.string.unit_minutes);
             if (t > 59) {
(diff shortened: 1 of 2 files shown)


jamorham:
I think instead we should think about how the status page can be localized to make it easier for non-english speakers to understand what is there but also so it can be reverted to english for support purposes.

Navid200:
Please look at the following image:
352771051-85ab4636-2ce0-4363-b5e4-91e86cc7e4b7

You can see that on the sensor status line, the phrase "sensor status" is not translated.  But, the word after 11.2 is translated.  So, on the same line, we have some words in English and one in another language.
Likewise, if you look at the last connected line, you can see that last connected itself is not translated.  But, the time unit is translated.
And the word "ago" after it is in English as well.  

Considering we have words on the same lines in different languages, this cannot be intentional.  
This is the PR that caused this: https://github.com/NightscoutFoundation/xDrip/pull/646  
If you look at the PR conversation, you see that the focus is on reminders.  I don't believe either of you at the time considered that this would also impact the Dexcom status page.  

The only change I am making in this PR is to make the Dex status page consistent with respect to translation.  
  
---  
  
**Helping those who don't speak English**
If you believe we should translate the status page in order to help those who don't speak English, can I say that the way to address that would be to add a button to the page that when long pressed would translate everything consistently (not just 3 words) to the other language and keep it like that for 10 seconds so that the user can take a screenshot.  
  
In my opinion, however, that is not the right approach.
We already have individuals who speak English as well as the language they translate to in Crowdin.  Those individuals can be the link between us and those who need help.
There are facebook groups that are in other languages.  Those who are experienced with xDrip and provide support in such groups usually speak English.  They work with us.  They post in out discussions and bring up issues and make requests.
They can help those who need help with troubleshooting.
In my opinion, we cannot work in silos independently.  We need individuals who act as links between English and other languages as they already do in Crowdin.  We need similar individuals for support.  And they already are there.

Please reconsider.

jamorham:
My view is that we should focus on what is best for the users and so I think the goals should be:

1) To have a plan to localize more of what we can regarding technical status information to help users understand what the app is doing.
2) To make it easy to request support and especially from english speakers.

To this end I propose to implement a screenshot button in the status page which when pressed will generate a screenshot in English and then offer to save/share that file to make it easier for the user to find. If  they want to screenshot in their native language then they can just use the regular phone based screenshot button combo.

Navid200:
OK
Thanks a lot. 
I will do that

jamorham:
I already have this all implemented which I did before to check feasibility of the concept.

Navid200:
OK
I will not work on this
Thanks

Issue #1864, opened 2021-10-29
Give option to save Error and Event List log file locally


simonpickering:
It would be useful to be able to save the Error and Event List (3 dot menu > "View Events Log") data to a local file (as well as the current option of being able to email it [somewhere, but who knows where? :)]).

This would make it easier to share the log data in a searchable form (currently there are lots of screengrabs attached to issues) as well as letting users do their own debugging by comparing the log data and source code

tolot27:
You can use [DB Browser for SQLite](https://sqlitebrowser.org) read the sqlite db contained in the exported zip. UserErrors is the table of interest. You can export it to CSV as well.

Indeed, this is just for experienced users feasible.

Navid200:
The following are from developer conversations related to this request.

1- There should be an option to save the logs to a local zip file.

2- A note should be added to inform the user that there isn't much point uploading the logs before communicating with developers.  We currently ask users to use Discussions to communicate with developers.

3- There should be an emphasized warning, before save/upload,  that logs can contain sensitive information - not limited to glucose values, transmitter ids, web upload/nightscout endpoints/passwords etc and that users should only share logs privately with core developers and not post them publicly etc etc.

Navid200:
Starting with the August 2, 2024 release, you can save logs locally.

PR #3604, opened 2024-08-01
Save logs locally

Navid200:
This PR adds the ability to save logs locally in the mobile storage.  

Fixes: https://github.com/NightscoutFoundation/xDrip/issues/1864 

There are times one may need to save their logs for their own reference.
There are times one may want to submit logs to someone other than the lead developer.
There has been a request by someone who used a full Android watch claiming that the font size was so small on their screen that they could see nothing in the logs.  

The output is saved in the /Download/xDrip-export folder, which is where the settings file is saved. 
The same as the settings file, the output overwrites the previous export.  

The new method packLogs packages the logs.  The content used to be inside the SendFeedBack file.  The new method is in EventLogActivity file that calls both SendFeedBack and the new SendLogs methods.  This way, the same pack method is used for both options to minimize code duplications.  

This is what the Event Logs page looks like after this with a new button at the bottom for saving the logs instead of sending the,  
Screenshot_20240731-195828
  
The send feedback page shows a new warning at the top, which is also shown on the save logs page.
Screenshot_20240731-200049
Other than this new warning, nothing has changed on the send feedback page.  
  
The new save logs page is shown below.  
Screenshot_20240731-200301
  
The warning is translated so that users can understand better.  And considering the warning is for users, not for developers, it makes sense to translate it.
The logs themselves remain in English as they should be.    
  
Tests:
Tested with Android 8, 9, 11 and 15.
For testing the existing function of sending the logs to you, I did send logs about an hour ago.  There has been a slight change to the format of the logs that you receive.  Please look at what you got and tell me if it is broken.  

If there is anything you don't like, please tell me.

Thanks

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -703,6 +703,12 @@
                 <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
         </activity>
+        <activity
+            android:name=".utilitymodels.SaveLogs"
+            android:configChanges="orientation|screenSize"
+            android:label="@string/title_activity_save_logs"
+            android:noHistory="true">
+        </activity>
         <activity android:name=".deposit.DepositActivity" />
         <activity android:name=".utils.AndroidBarcode" />
         <activity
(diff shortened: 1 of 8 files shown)


Navid200:
I just tested this on Android 7 using a virtual machine in Android Studio and it works fine.  
Screenshot 2024-08-01 092052

Navid200:
The new save logs button appears between the two existing buttons as shown below when we are not at the top of the list.  
Screenshot_20240801_232547

Navid200:
After a successful save, a log shows where the log file has been saved.  
There is no toast intentionally.  A toast can be missed and it is really annoying when you notice a toast but are not fast enough reading it.
In this case, the log is perfect because after the save log button is tapped, we are automatically taken back to the event logs screen where we can see the log showing where the file ahs been saved.

Navid200:
The log file is not zipped as it is not being emailed.  I didn't think it would be a very large file.

jamorham (approved):

PR #3607, opened 2024-08-03
Add Cubot to the wake workaround list

Navid200:
Should we add this manufacturer to the list that can benefit from wake workaround?  
That's what this PR does.
Over the years, one of the phones that has been reported as having connectivity problems has been this.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/framework/BuggySamsung.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/framework/BuggySamsung.java
@@ -71,7 +71,8 @@ public class BuggySamsung {
                 || Build.MANUFACTURER.toLowerCase().contains("xiaomi")
                 || Build.MANUFACTURER.toLowerCase().contains("oneplus")    // experimental test
                 || Build.MANUFACTURER.toLowerCase().contains("oppo")      // experimental test
-                || Build.MANUFACTURER.toLowerCase().contains("huawei");      // experimental test
+                || Build.MANUFACTURER.toLowerCase().contains("huawei")      // experimental test
+                || Build.MANUFACTURER.toLowerCase().contains("cubot");
     }
 
 }


Navid200:
This is what xDrip sees with respect to the manufacturer name:
453301055_10229032790518249_5840595279795064752_n

  
The user confirmed that he occasionally sees the red slow wake note on the status page.

jamorham:
I'm not sure about this as I've used the King Kong mini and all I had to do was use its mediatek duraspeed app/setting to remove xDrip from its power saving restrictions. xDrip should pop-up and prompt the user on this model. Possibly things could have changed since then of course.

Navid200:
I thought duraspeed was for unihertz, not Cubot.  I will ask the owner to comment.

Anyway, I added this workaround to everything considering it is just an option.  it only kicks in if needed.  And you approved and merged.  But, later  you removed it because some devices were showing malfunction in display.
That tells me that if we exclude the devices that malfunction, we should be able to add others if it may help.  Do you believe Cubot will show visual distortion on screen?

TWilhoWI:
> I'm not sure about this as I've used the King Kong mini and all I had to do was use its mediatek duraspeed app/setting to remove xDrip from its power saving restrictions. xDrip should pop-up and prompt the user on this model. Possibly things could have changed since then of course.

Hi,

I did disabled Dura speed completely because I thought it would interfere.

Navid200:
@jamorham Do I misunderstand this?

I use Pixel.  Since the manufacturer of my phone is not on the list in xDrip, if I enable or disable wake workaround, my xDrip behavior will not change.
However, since Samsung is on the list, if a Samsung user changes wake workaround, the behavior of their xDrip changes.  
We have added that phone to the list so that they can benefit from wake workaround.  However, if enabling wake workaround is going to makes things worse for them, they can disable wake workaround.  

Is that not correct?

Why would it matter if we add a phone manufacturer to the list that does not need the workaround?  Could that actually stop those phones?
We added Xiaomi.  We have added several phones.  Why not add Cubot?

We have a user with a realme phone showing slowest wake. Can we not add that too?

jamorham:
The workaround breaks some other functionality of the phone so we should avoid it unless we are sure that it really is needed.  Are we sure it is still having slow wakeups even with duraspeed disabled?

Issue #3569, opened 2024-07-15
Incorrect sensor expiry alert "time remaining"


Navid200:
xDrip version: 2024.06.28    
Collector Dex G7
Sensor expiry shows sensor will expire in 5 hours.  But, it should say in 6 hours.  Or, it will show sensor will expire in 1 hour, but, it should show in 2 hours.  
  
The following screenshot on the left shows that the sensor will expire at 8:17 pm.
The one on the right shows the note associated with the alert, which says the sensor will expire in one hours at 6:19.  
The sensor really expired at 8:20 pm.  
So, the alert notification is incorrect by one hour.  

![Screenshot_20240714-182826](https://github.com/user-attachments/assets/bddb8996-e757-4176-97b6-fb8bb6f7e6bc) ![Screenshot_20240714-182843](https://github.com/user-attachments/assets/4672ba1b-6a84-4636-8a9b-f50e2112a80f)

PR #3610, opened 2024-08-04
Sensor expiry alert resolution increased to 1 decimal point

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/3569
Sensor expiry alert shows "Sensor will expire in 1 hour" even if 1.99 hours are still l eft to expiry as shown here: https://github.com/NightscoutFoundation/xDrip/issues/3569.  
  
I can speculate that the intent was for the alert to show 24 hours left and 12 hours left, and 6 hours left and 2 hours left.
But, because we want this to be a silent alert, we have made it wait for the screen to be unlocked or the charger to be disconnected.

As a result, when 2 hours are left to expiry, the alert does not trigger.  it triggers when the user unlocks the phone after that time.
But, the format chosen can only show integer values.

The 24-hour value shows 23 hours are left.
The 12 hour alert shows 11 hours are left.
It is really bad when the 2-hour alert shows 1 hour left.  
  
This PR rounds down the value to 1 decimal point instead of dropping the fractional part completely.  
  
---  
  
**Not tested yet**  
I have tested a fictitious case and verified functionality.
However, I will test this in real action in 10 days with a G7.  
  
Please let me know if I have overlooked anything.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/alert/SensorExpiry.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/alert/SensorExpiry.java
@@ -3,6 +3,7 @@ package com.eveningoutpost.dexdrip.alert;
 import static com.eveningoutpost.dexdrip.models.JoH.cancelNotification;
 import static com.eveningoutpost.dexdrip.models.JoH.niceTimeScalar;
 import static com.eveningoutpost.dexdrip.models.JoH.niceTimeScalarNatural;
+import static com.eveningoutpost.dexdrip.models.JoH.niceTimeScalarNaturalp1;
 import static com.eveningoutpost.dexdrip.models.JoH.showNotification;
 import static com.eveningoutpost.dexdrip.models.JoH.tsl;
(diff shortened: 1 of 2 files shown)


Navid200:
I have tested this with fake data source.  Please see the results in the following two images.  Please look at the time, the sensor expiry note at the bottom and the sensor expiry alert flag content.  
You can see that the sensor expiry alert flag now shows 1 decimal point.  
Screenshot_20240804-004914 Screenshot_20240804-004934

jamorham (approved):

Navid200:
There is still something wrong:
Screenshot_20240812-211628
  
There seems to be a 0.4-hour discrepancy.  That is 60 x 0.4 = 24 minutes.

I will investigate.

Navid200:
This one looks much better:
Screenshot_20240813-101519
 
I may just be looking at the rounding effect to 1 decimal point.  Let's see what the 2-hour notification looks like.

Navid200:
This looks pretty good to me.  
Screenshot_20240813-190657  Screenshot_20240813-190712

PR #3620, opened 2024-08-12
Do not assign a y position to notes

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/1299  
<br/>  
  
---  
  
**Recreate the problem**  
Install the latest Nightly (https://github.com/NightscoutFoundation/xDrip/releases/tag/2024.08.07).  
Disable predictive simulations.  
Set the BG units to mmol/L.
Enter an insulin treatment.
Tap on it and add a note.  You should see something like this.  
Screenshot_20240812-175734
  
Change the BG unit setting to mg/dL.  Now, you will see something like this.  
Screenshot_20240812-180000
  
You can see the y axis range has been expanded to very low values.  Now, delete the treatment.  You will see the scale going back to normal as shown below.
Screenshot_20240812-180204  
 
Enter a treatment again.  Add a note to it.  You should see something like this:  
Screenshot_20240812-180331  
Change the BG unit to mmol/L.  You will see this:  
Screenshot_20240812-180522  
You can see that the y axis range is now unreasonably extended on the positive side.  
 <br/>  
  
---   
  
**After this PR**  
The note will not have a y position value applied to them.  
Let's repeat the second test.  This is what you will see after entering a treatment and adding a note to it.  
Screenshot_20240812-181019  
Now, if you change the BG unit, you will see this.  
Screenshot_20240812-181152  
You can see that the y axis range remains reasonable.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -3519,7 +3519,7 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
 
             if (treatment_text.length() > 0) {
                 // display snackbar of the snackbar
-                final View.OnClickListener mOnClickListener = v -> Home.startHomeWithExtra(xdrip.getAppContext(), Home.CREATE_TREATMENT_NOTE, Long.toString(timestamp), Double.toString(position));
+                final View.OnClickListener mOnClickListener = v -> Home.startHomeWithExtra(xdrip.getAppContext(), Home.CREATE_TREATMENT_NOTE, Long.toString(timestamp), "-1"); // Let's not enter a y position to avoid having to worry about BG units
                 Home.snackBar(R.string.add_note, getString(R.string.added) + ":    " + treatment_text, mOnClickListener, mActivity);
             }
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/BgGraphBuilder.java
@@ -2370,7 +2370,7 @@ public class BgGraphBuilder {
                     final View.OnClickListener mOnClickListener = new View.OnClickListener() {
                         @Override
                         public void onClick(View v) {
-                            Home.startHomeWithExtra(xdrip.getAppContext(), Home.CREATE_TREATMENT_NOTE, time.toString(), Double.toString(ypos));
+                            Home.startHomeWithExtra(xdrip.getAppContext(), Home.CREATE_TREATMENT_NOTE, time.toString(), "-1"); // Let's not enter a y position to avoid having to worry about the BG units
                         }
                     };
                     Home.snackBar(R.string.add_note, message, mOnClickListener, callerActivity);


Navid200:
What gave me the idea that this was the solution was the existing note feature.
You can add a note to a treatment.  You can also just place a note.
If you look at the code, you will see that the existing code for placing a note, which does not exhibit this problem, does not assign a y position to the note by setting it to -1.

Seeing that and knowing that note does not have the issue, I tried it with treatment note and tests showed that removing the y position from treatment note solved the problem for treatment note as well.

jamorham:
I'm not too sure about this because it looks like you're removing the facility of notes being placed at particular points on the Y axis when the problem is that this position is encoded in the current units and if those change then the note positions become invalid.

I feel like a better solution would be to detect notes which are outside of the currently selected unit range and scale those appropriately or to convert the Y positions in the database during a units change. I think the first option is likely preferable, as for example mmol would always be under 24 and mgdl always above it.

Navid200:
Thanks for the review.

But, why does this problem never occur for a note?  It only occurs for a note attached to a treatment.
The only difference is that a note is set up with -1 as the parameter.
https://github.com/NightscoutFoundation/xDrip/blob/aee9315b237d735e58434cccecdd19ef5c397a4d/app/src/main/java/com/eveningoutpost/dexdrip/Home.java#L3494  

Screenshot 2024-09-27 105338

When I saw that this problem only occurred when I attach a note to a treatment, but does not when I just enter a note, I decided to see what the difference was.
The only difference is that we enter a y position for a treatment attached to a treatment.  While we leave it blank when we only create a note.

Navid200:
@jamorham Please see my post above.  This PR only does what xDrip does already for notes.  It's already been in xDrip for years and tested.  Why can we not use it for notes attached to treatments as well?

Navid200:
@jamorham I can do what you have suggested.

But,  will you do me a favor?   Would you please tell me that you are aware that note (not treatment note) already does what this PR suggests and knowing that, you still prefer that I do what you have suggested?
I would just like to make sure you are aware that we are already doing what this PR suggests and we have been doing it for years.

Thanks

Navid200:
Installed the latest Nightly, 2024.09.27, on a test phone.  Set the units to mg/dL.  Added a note.
Screenshot_20240930-170953
<br/>  

Added a treatment.  
Screenshot_20240930-171517
<br/>  

Changed the unit to mmol/L.  The y axis range remains unchanged.  
Screenshot_20240930-171545
<br/>  
Changed the unit back to mg/dL.  Tapped on the treatment and added a note (Test2) to it.
Screenshot_20240930-171641
<br/>  
Changed the unit to mmol/L.  
Screenshot_20240930-171715
<br/>  
  
You can see that the note remains where it was.  But, the treatment is now at a very high value causing the y axis range to expand.

Navid200:
> I'm not too sure about this because it looks like you're removing the facility of notes being placed at particular points on the Y axis when the problem is that this position is encoded in the current units and if those change then the note positions become invalid.  
  
The y position is defined by the treatment.  The note added to the treatment overrides that position in xDrip as is.  This PR avoids impacting the position of the treatment when adding a note to it.

Navid200:
@jamorham Why should we allow the position of a treatment to be changed just because a note has been added to it?
This PR places the note where the treatment is.  If the unit is changed then, the treatment moves accordingly as it did before there was a note attached to it.  The note has no choice but to move with it.

Would you please reconsider?

PR #3627, opened 2024-08-15
Disable Nightscout upload back to data source

Navid200:
Consider xDrip used as a Nightscout follower.
If the user accidentally sets up xDrip to upload to the same Nightscout site, xDrip will accept.
The only reason this does not cause a disaster is that Mongo fortunately issues a non-destructive error message.  
But, the upload is a waste of bandwidth.  This PR disables Nightscout upload in that scenario.  

The log that is issued is shown in the following image at the top.  
Screenshot_20240814-204811

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Constants.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Constants.java
@@ -61,6 +61,7 @@ public class Constants {
     public static final int ZXING_CAM_REQ_CODE = 49374;
     public static final int ZXING_FILE_REQ_CODE = 49375; // This is created by just incrementing the existing camera scan code from the zxing package
     public static final int SENSORY_EXPIRY_NOTIFICATION_ID = 2003;
+    public static final int DISABLED_NS_UPLOAD_NOTIFICATION_ID = 2004;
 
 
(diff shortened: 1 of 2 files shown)


Navid200:
I'm sorry for the auto format.  The indentations were so confusing!  

I have added a notification so that this cannot happen without the user knowing.
This is what the notifiacation looks like:
Screenshot_20240815-082334 Screenshot_20240815-082407 
Screenshot_20240815-082502

Navid200:
This is really all of the change to NightscoutUploader other than the two imports:
Screenshot 2024-08-15 084037

Navid200:
If someone wants to upload treatments to Nightscout both from the master and follower, this may be the only way.  
I have to do some tests on this.  So, I may have overlooked that scenario.

Navid200:
Even though this is good for not allowing a user to unintentionally upload to a Nightscout host that they are downloading from, it would disable a required feature.

A Nightscout follower cannot upload treatments.  The only way to do that is to set up xDrip to also be an uploader.  
Since Mongodb tolerates collisions (overwrites), this is not an issue.  
I wish there was an easy way to avoid uploading the redundant readings in this case.  But, other than adding the ability to upload treatments to Nightscout follower, I don't see how.

So, I am closing this.

PR #3634, opened 2024-08-19
No more confusing BG codes

Navid200:
These codes were added a very long time ago.
https://github.com/StephenBlackWasAlreadyTaken/xDrip/blob/bd43dad2b9cf25d9e6af13813acbaf262f5893c2/app/src/main/java/com/eveningoutpost/dexdrip/ImportedLibraries/dexcom/Constants.java#L132-L141  
The oldest existing record I can find is from almost 10 years ago.  But, they may have even been added before then.  

A lot has changed since then.  Are we sure any devices today send such low values as codes?  
So, we show a value of 13 as LOW, but a value of 12 as ?RF.
Neither 12 nor 13mg/dL makes any sense.  Both are unreasonable values for blood glucose.  

But, questions about them have come up over the years.  And there is no answer as to what they mean.  Therefore, they are confusing.

This PR removes the codes and shows LOW instead.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/BgGraphBuilder.java
@@ -2190,29 +2190,8 @@ public class BgGraphBuilder {
                 df.setMinimumFractionDigits(1);
                 return df.format(mmolConvert(value));
             }
-        } else if (value > 12) {
-            return "LOW";
         } else {
-            switch ((int) value) {
-                case 0:
-                    return "??0";
-                case 1:
-                    return "?SN";
-                case 2:
-                    return "??2";
-                case 3:
-                    return "?NA";
-                case 5:
-                    return "?NC";
-                case 6:
-                    return "?CD";
-                case 9:
-                    return "?AD";
-                case 12:
-                    return "?RF";
-                default:
-                    return "???";
-            }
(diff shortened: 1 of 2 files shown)


Navid200:
This is just one example of posts by users confused by what the code shown on the main screen means.

https://github.com/NightscoutFoundation/xDrip/discussions/3623

jamorham:
These are error codes that are sent, I don't think they represent actual glucose numbers which is why we don't display them like that.

Navid200:
OK, thanks  

But, are we sure newer devices still send these codes?  This was added to xDrip at the time of G4.  

And we show these codes even if the data source is not Dexcom.  Does that make sense?

Navid200:
This is a post in facebook today.  
<br/>  
<br/>


Screenshot 2024-09-26 080750  


I understand that Dexcom devices used to send these codes at the time of G4.  Do we know that they still send them now?

Assuming you prefer to keep this behavior because we don't know how to verify if newer devices send them, shouldn't we limit this behavior to Dexcom sensors only?

Navid200:
I accidentally pulled a G7 off my skin.
Look what it shows:
Screenshot_20241012-193725

That is 5, which  means sensor out of cal:
Screenshot 2024-10-12 194155


So, it seems like G7 still sends these codes.
I still wonder if we should limit showing these codes to the case of xDrip being a collector from Dexcom.

But, it is clear I was wrong and the codes are still being sent.
So, I close this PR.

PR #3635, opened 2024-08-20
Receive glucose data from Syai Tag App.

syai-dev:
Receive glucose data from Syai Tag App.

Screenshot_20240820-175444_AAPS

Screenshot_20240820-175705_xDrip+

Screenshot_20240820-175602_xDrip+


jamorham:
Thanks for contributing this but can you explain why you cannot instead use the existing mechanism described here: https://github.com/NightscoutFoundation/xDrip/blob/master/Documentation/technical/Incoming_Glucose_Broadcast.md ?

syai-dev:
Using a common channel, some users feedback can not distinguish between more than two data sources, in order to facilitate the user to distinguish the source of information to find our device type.

jamorham:
Can you explain more what you mean by this?

syai-dev:
> Can you explain more what you mean by this?

Users can directly choose our "Syai Tag" brand

Navid200:
Syai Tag is a sensor, right?

Can we not get readings directly from the sensor into xDrip with no need to the additional Syai Tag app?

Navid200:
If it is a requirement for the app to run on the same phone for this to work, we can add it as a companion app option.  Then, the outcome will be the same and consistent with all the other apps that can be a companion to xDrip.

I would prefer to have only one app as my CGM to simplify the setup.  But, if that cannot be done, we already have a companion app mode in xDrip.

PR #3652, opened 2024-08-31
Some cleanup

Navid200:
Just some cleanup


app/src/main/java/com/eveningoutpost/dexdrip/g5model/SensorDays.java
EasternPA, line 1: Is a 30-minute warmup period needed on the new if/then inserted at 112? Public statements seem to indicate it still applies.

From https://www.reddit.com/r/diabetes_t2/comments/1f1nrqv/dexcom_stelo/

image

jamorham (approved):

Navid200:
Thanks for the review and approval

PR #3663, opened 2024-09-14
Dex Share Follow customizable wake to reduce overall delay 

### LINKED ISSUE #1849: Dexcom Share: Ability to Specify Seconds Until Polling Retry or Allow Double Poll Retry
### Is your feature request related to a problem? Please describe.
This feature request is a result of my experience monitoring my T1D son.  He has a Google Pixel 3a as the collector device with him at all times and wears a Dexcom G6.  His endocrinologist wants data shared via Dexcom Clarity, so his Dexcom app receives the readings from the sensor and then his, mine and my wife's phones receive those readings from Dexcom Share as followers.  When he is in low cellular signal areas such as school there is a delay in his phone uploading the readings to the Dex Share servers.  That, in turn, causes our xdrip apps to unsuccessfully poll for the BG reading after the 5 minute interval (since it's not on the Dex Share server yet), then it waits a hard-coded 20 seconds before retrying to poll for for the BG reading again.  30-40% of the time the reading is still not there on the Dex Share server yet.  At that point the app gives up and sets the next polling interval back to 5 minutes.  For 30-40% of the time, my xdrip is 5-10 minutes behind.  I realize that I can go to the Classic Status Page and press the "Restart Collector" button to force a re-poll.  I also work at his school which does not have a nurse, so I am his.  It would be quite helpful to have the readings I'm looking at be timely without needed extra interaction to make it so.

### Describe the solution you'd like
I would like a new option in the menu to either 1) allow me to specify the number of seconds to wait for a Dex Share polling retry or 2) allow me to have a double retry for Dex Share with the first occurring after 20 seconds and if that is unsuccessful then another after 40 seconds.

### Describe alternatives you've considered
The alternative is to manually force it to poll after the polling retry fails by going to the Classic Status Page and pressing the "Restart Collector" button.

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/1849  

This PR adds the ability to add an optional delay to the Dex share follow polling time.  
This becomes necessary if the setup has unavoidable delay on the source side (pump/internet/ ...) as explained in this open issue: https://github.com/NightscoutFoundation/xDrip/issues/1849.  Without this, xDrip polls only twice, 5 minutes after the last time stamp and 5 minutes and 20 seconds after the last time stamp.  After that, if xDrip does not get a new reading, it will go to sleep for 5 minutes.  

Adding a small delay to polling could avoid an extra 5-minute extra delay in readings.  
  
The following image shows the Settings page after this PR, where you can see the new Dex Share Follow delay setting.
Screenshot_20240914-091039  
Tapping on that will lead you to the following page where you can choose one of several options for the delay.  
Screenshot_20240914-104706

<br/>  
  
**Testing**  
My main phone uploads to Nightscout.  
I have 2 test phones.  
Test phone 1 is a Nightscout follower as well as a Dex Share master.  It gets readings from my main Nightscout site.    
Test phone 2 is a Dex Share follower getting readings from test phone 2.

By adjusting the Nightscout follower delay on test phone 1, I can adjust the delay Dex share master (same phone) experiences.  
  
The following image shows the status page of test phone 2 under the following conditions:
Nightscout follow delay set to 40 seconds 
Dex Share follow delay set to 0.
Screenshot_20240914-101414
You can see that the delay from Last BG time to next poll time is 10 minutes.  This is because the poll at 5 minutes and 5 minutes and 20 seconds have failed and xDrip has gone to sleep for 5 minutes after which the total delay is 10 minutes.  
  
The following image shows the status page of test phone 2 with the Dex Share follow setting set to 40 seconds.  
Screenshot_20240914-103022
  
Comparing Last BG time and next poll time, you can see a delay of 40 seconds.  So, by intentionally delaying the reading by 40 seconds, we have avoided a 5-minute extra delay.  
  
I have tried all possible delays for test phone 1 and choosing enough delay for test phone 2, I have been able to avoid the extra 5-minute delay.  
  
The following image shows the status page of test phone 2 with the Dex Share Follow delay set to 2 minutes.
Screenshot_20240914-104028
You can see an effective delay of 7 minutes, which is an extra delay of 2 minutes over the default 5 minutes.


Navid200:
**Why did I change Anticipate?**

In a previous PR (https://github.com/NightscoutFoundation/xDrip/pull/3179), I added the ability to add a delay to Nightscout follow.  I did that by changing Anticipate.  
But, Anticipate is used by other methods like Dex share follow.  

Now, I have made a change to do it right.  
Nothing inside Anticipate is specific to Nightscout or Dex Share any longer.
Anticipate now has an optional parameter, lag.
The calling method can set this optional parameter to a non-zero value to delay the next poll time.  

This way, the different methods can continue to use Anticipate without interfering with each other.

Navid200:
We made a similar change to the Nightscout follower previously: https://github.com/NightscoutFoundation/xDrip/pull/3179 

This PR does the same for the Dex share follower.
The Anticipate method has been changed now to make it a more modular approach, which I believe how I should have done it when I submitted the other PR for improving Nightscout follow.

I have tested both Nightscout follow and Dex share follow after this PR to make sure both function as expected.

app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollowService.java
```diff
@@ -58,6 +58,7 @@ public class NightscoutFollowService extends ForegroundService {
     private static volatile Treatments lastTreatment;
     private static volatile long lastTreatmentTime = 0;
     private static volatile long treatmentReceivedDelay = 0;
+    private static long lag = Constants.SECOND_IN_MS * Pref.getStringToInt("nsfollow_lag", 0); // User can choose a wake delay with a 0 default.
```
jamorham, line 61: By setting this in a static field it will not get updated if the user changes the preference value while the class remains loaded.

app/src/main/java/com/eveningoutpost/dexdrip/cgm/sharefollow/ShareFollowService.java
```diff
@@ -57,6 +57,7 @@ public class ShareFollowService extends ForegroundService {
     private static volatile long lastBgTime;
 
     private static ShareFollowDownload downloader;
+    private static long lag = Constants.SECOND_IN_MS * Pref.getStringToInt("dex_share_follow_lag", 0); // User can choose a wake delay with a 0 default.
```
jamorham, line 60: Same comment here about storing this statically.

app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/utils/Anticipate.java
```diff
@@ -20,9 +17,11 @@ public class Anticipate {
      * If last + period and a bit >= now, ask again after last + period and grace
      */
 
-    public static long next(long now, final long lastTimeStamp, final long period, final long grace) {
-        final long lag = Constants.SECOND_IN_MS * Pref.getStringToInt("nsfollow_lag", 0); // User can choose a wake delay with a 0 default.
```
jamorham, line 24: Removing this from inside Anticipate is probably a good idea

Navid200:
I will look into this.

My intent here was to keep everything working with Nightscout as Nightscout and Dex share use the same Anticipate method.  So, I just brought the parameter out but kept everything unchanged.
I will look at this carefully to see if we can change it as you like.

Navid200:
> Can't we just remove the lag handling from inside Anticipate and then add the lag on to the anticipated timestamp from the calling class rather than passing it as a parameter? Would that work?

If I understand you correctly, we will then need to add the lag handling to multiple methods, Nightscout and Dex share and anything else that also uses Anticipate.  Any change we may need to make to the mechanism will have to be applied multiple times to different sections.
I think it is best to have the mechanism inside Anticipate and used by all methods that need it.

Navid200:
I'm sorry about the mistake.
I confirmed that indeed the delay was not being adjusted by the setting.
I changed the variable and made it non-static and confirmed with a test that the delay is adjusted by the setting now.

Navid200:
This is ready for a review again now.   I believe I have addressed your concern and thanks for the review and advice.

Navid200:
@ahaarrestad Do you see anything here that you don't like?

ahaarrestad:
@Navid200 I just found that in the code and it looks really clean. I am more of a backend programmer, so UI fixes like this are great. No comments except thumbs up from me

Navid200:
@ahaarrestad Thanks for the review.

Is there anything I can do to make it easier for developers like you?  Have I done anything that may have hindered it?

ahaarrestad:
@Navid200 The cleanup looks good to me. No real duplicate code and the code is documented. But @jamorham always has the final say on the changes :-)

Navid200:
@ahaarrestad Thanks
I was asking a general question not specific to this PR and more about how you feel in general about this repository and if I as a collaborator can do anything to make it easier for you.  I'm sorry it's OK if you don't want to respond.  
Thanks a lot for your contributions in the past and present.  Please keep up the good work if/when you can.

ahaarrestad:
@Navid200 I have mostly contributed by writing tests and some bugfixes and code cleanup. From my daily work I have found that tests makes it easier to rewrite code and get a stable product. xDrip+ is an old codebase and the guidelines in the Contributor.md file are good. I always try to follow the codestyle in the code to make it easier to review and reduce the chance of introducing bugs

Navid200:
@jamorham Would you please merge this?
You already reviewed it a long time ago.  
If there is something missing, please tell me.

app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/utils/Anticipate.java
Navid200: I took this comment as approval.  If I have misunderstood, please let me know.

app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollowService.java
Navid200: Thanks
I corrected it.  It's not static anymore.

Navid200:
Please merge this too.  It will close a request we have had for years

Navid200:
> Can't we just remove the lag handling from inside Anticipate and then add the lag on to the anticipated timestamp from the calling class rather than passing it as a parameter? Would that work?

I think I finally understand what you are saying!
Yes, I can do that:
I will remove the new lag parameter that I had added to anticipate.
The line that calls anticipate can add a value to the timestamp parameter to accomplish the same objective.  So, I will do that for Nightscout and Dex Share calling lines.
I will redo my tests and reopen.

Thanks

Navid200:
The lag is not handled outside Anticipate.
I have tested this for Nightscout follower and it works fine.
I cannot test Dex share because it seems there have been changes that do not allow me to upload to Dex share from xDrip.

Navid200:
It doesn't look like there is any interest in this functionality any longer based on the lack of response to my request for testers.
So, let's close this.

Issue #3641, opened 2024-08-25
Lock Screen Notification Text Color Inverted in Dark Mode


nickb24:
### Subject of the issue
Link to the discussion thread:  https://github.com/NightscoutFoundation/xDrip/discussions/3617
Every time a new reading comes in with the screen off and you turn the screen on, the xDrip ongoing notification colors are inverted.

### Your environment
- 20240807
- G6 and Nightscout Follower for testing
- Android 14

### Expected behavior
In dark mode, text should be white on the xDrip notifications at all times.

### Actual behavior

https://github.com/user-attachments/assets/90b0ebe6-ab2d-4fb8-a7e0-e90e45c9b9f5

It has to do with the specific color you set in the Android theme settings. I had mine set to blue, and with that setting when setting the target sdk to 24 it causes the bug. sdk 23 or 29 don't have it. I was able to reproduce on an emulator as seen in the video. The bug is weird though since as you can see in the video, if you go past the lock screen and pull down the notification once, then it fixes the color. It stays correct until the notification refreshes after 5 minutes (when the next glucose reading comes in).

### Steps to reproduce the behavior:
See video above.

jamorham:
You mention target sdk level 29, do you know about 25, 26, 27, 28 ? The reason I ask is that we would plan to incrementally advance targetsdk levels because there are associated issues with each level - some of which we may not be able to overcome. So if this problem doesn't exist on the next target sdk level than that would be a convenient fix compared to anything more drastic.

jamorham:
Ok thanks, that is useful but changing targetSdk level to 29 is a gigantic change in terms of app behavior so I am hoping to find a resolution which doesn't involve doing that.

Navid200:
> changing targetSdk level to 29 is a gigantic change  

I agree.  

When we retired Android 6 a few months ago, there were complaints even though Android 6 is really old.

If we change the target to 29, we will be retiring 7, 8 and 9.   If we do that now, we will be denying access to xDrip to a significant population worldwide.  It is too soon for that now.
The next step should be retiring Android 7 and I hope we can delay that as much as possible.

Navid200:
I see.  Thanks for correcting me.

jamorham:
The most significant thing about changing the targetSdk level is that it opts the app in to new behaviors and restrictions.

Navid200:
If we had a fully functional alpha channel, many would be using it and the Nightly channel would then be what it in my opinion should really be to test things like this.

[...]

PR #3675, opened 2024-09-20
Make sensor expiry notifications customizable to the minute

DmitriyShepelev:
Implements idea introduced in https://github.com/NightscoutFoundation/xDrip/discussions/3666.

Note that the current way of hard coding notifications at 24, 12, 6, and 2 hours before sensory expiry is preserved if a custom value isn't supplied.

Tested locally using phone and smartwatch.


DmitriyShepelev:
@Navid200 Kindly requesting your review. 😊

Navid200:
I can only give you my opinion.  Only the approval of the repository owner matters as far as what is merged and what is not.

In general, I would never include translations in a PR.  It makes your PR look  bigger than it really is.  And translations should be added by translators in Crowdin, not be developers in a PR.  

But, my main concern with this is that I don't think we should have alerts for sensor expiry.  We should instead have reminders.  
I don't believe anyone should wait for their sensor to expire before they insert the next one.  They can if they choose to.  But, no one should feel that it is necessary to wait for a sensor to stop before starting the next one.

To expand on that, imagine if something comes up for you and you don't get to start your sensor at the time you would have liked.  Let's say by the time you take care of the unexpected, you end up starting your sensor 4 hours later than you would have liked.  
There is no reason to start your next sensor 4 hours late just because you started the current one 4 hours late.
Four hours before your current sensor stops, you can start your next so that you go back to your preferred schedule.  No one can complain if you do that.
In this scenario, if you have an alert that goes off with respect to when the sensor is going to expire, it will actually go off 4 hours too late.

That's why reminders, multiple of them, instead of one alert will let you plan your day such that you get to start the sensor exactly when you need to.  

I agree that if a sensor expires, that requires an alert.  If I forgot that my sensor was expiring and I let it expire, I need an alert.  That alert (notification) already exists.  
You pointed out, in the discussion, that instead of that alert going off exactly when the sensor expires, it should go off may be 10-15 minutes before.
I buy that.  I am happy to work on that with you to improve xDrip.  But, that is not giving the user the option to define when it should go off.  I don't see a reason for that.

In my opinion, we can modify the existing sensor alert (not the existing sensor expiry notification) such that in addition to the timings that it currently has, it would have an additional notification at 15 minutes to expiry and we can set only that one not to be silent.  
Would that be acceptable to you?

We will still need to get the lead developer to review and approve.  I can not promise approval.  But, that change makes sense to me at least.
That's why I am asking your opinion.

wirbel-at-vdr-portal:
"To expand on that, imagine if something comes up for you and you don't get to start your sensor at the time you would have liked. Let's say by the time you take care of the unexpected, you end up starting your sensor 4 hours later than you would have liked."

That's why i would like to see that the 12hr/24hr sensor expiry time would be more clearly understandable for non-english users. They're always confused by the AM versus PM difference (which one of these is before or after lunch..), instead of 24hrs as they use.

That would add up to an 12hr+ issue in case of misunderstanding.

DmitriyShepelev:
> In this scenario, if you have an alert that goes off with respect to when the sensor is going to expire, it will actually go off 4 hours too late.

In this case, the user can just change the minutes before sensor expiry alert to go off 4 hours earlier than previously.

> That's why reminders, multiple of them, instead of one alert will let you plan your day such that you get to start the sensor exactly when you need to.

I don't find this functionality desirable, but my change preserves it for those who do.

My belief is that users should have the freedom to maximally customize their diabetes management, with minimal developer input.

@jamorham May you please review my PR?

Navid200:
As a diabetic, I am with you 100%.  I want to have absolute freedom.

As a developer, I am aware that 90% of those who have made changes to xDrip have disappeared and are no where to be found.
Look at the number of open issues in this repository.  Many of them are related to things that used to work, but don't any more.

Look at this: https://github.com/NightscoutFoundation/xDrip/issues/3641
You would think that this should just be a piece of cake to fix.  But, it's not.  

Any change made to xDrip, in my opinion, should have a very good reason addressing a serious shortcoming.

We have had a request to allow the 12-hour alert to become non silent as an option.
It is on my to-do list to open a submission for that.
It may be possible to make that single portion of the alert further customizable to also allow the time to be adjusted.

PR #3704, opened 2024-10-08
Persistent High Alert Threshold

Navid200:
This allows you to adjust your persistent high alert threshold without changing the readings colors on the screen or the statistics range limits.


Navid200:
| Before | After |  
| ------- | ------ |  
| Screenshot_20241008-112742 | Screenshot_20241008-112029 |

Navid200:
**Submenus**  

Screenshot_20241008-112055  
Screenshot_20241008-112112  
Screenshot_20241008-112127  
Screenshot_20241008-112143

Navid200:
There are only 2 new settings.  But, there were already many settings on the Extra Alerts page.  Instead of adding 2 more settings to it, I have created 4 submenus and the 2 new settings have been added to the Persistent High Alert submenu.  

I have renamed some of the existing settings now that they are on submenus.  As as example when the sound options for the persistent high alert and forecast low alert were on the same page, we need identifying names for them.  Now, each of those is on a submenu for the corresponding alert.  They don't need detail names any longer as the title of the submenu page identifies everything on that page.

Navid200:
By default, Threshold: High Value is enabled.  Therefore, on a fresh install or after an update, the persistent high alert will use the High Value as its threshold. 

This has been tested and is ready for a review.

Navid200:
When Threshold: High Value is enabled (default), the Threshold setting is disabled.  So, the value is dimmed and tapping on it will do nothing.  

When Threshold: High Value is disabled, the value shown as the summary on Threshold is bold and you can tap on it to edit it.  
Screenshot_20241008-112243

app/src/main/res/values/strings.xml
```diff
@@ -436,15 +438,15 @@
     <string name="new_version_date_colon">"New Version date: "</string>
     <string name="old_version_date_colon">"Old Version date: "</string>
     <string name="speak_your_note_text">Speak your note text</string>
-    <string name="alarm_if_above_high_value">Alarm if above high value</string>
+    <string name="alarm_if_above_high_value">Notify if above threshold and not falling for long</string>
```
jamorham, line 441: I don't think this english is clear what this means

Navid200: How about this instead?  

*Notify if above threshold for longer than the specified time below*

Navid200: I have changed the string to something that I hope is more clear.

Thanks for the review

jamorham (approved):

Navid200:
Thanks

Navid200:
While testing this before submitting the PR, I noticed something:

If you change the units and go right away to the persistent high alert without first going to the xDrip main screen, you will see the threshold shown without any units:
Screenshot_20241012-202526  
  
If you go to the main screen first, and then go to the persistent high alert page, the threshold unit will be shown correctly.  

I tested and noticed that the speak alert page behaved exactly the same way with respect to the "When changed by" item.  

So, I thought the reason we have never done anything about this could be that no one would ever need to keep changing their units back and forth.  After one changes their unit, they will leave it alone for a very long time may be forever.  So, this is not something that many would experience.

However, if you are testing a new feature, you may be doing this just to verify, which is exactly how I noticed it.

I am reporting this now because I think late is better than never.  I am sorry I did not report this earlier.
I can fix this if you think it is necessary.

PR #3707, opened 2024-10-10
Statistics Customization 1

Navid200:
This allows you to choose which items you would like to see on the statistics distribution page.
After this PR, you will only be able to remove some of the existing items.
If this is approved, we already have an open PR for adding capture rate.  I will work on that to add that to the menu and give credit to the person opening that PR.
Later, I would also like to add very high and very low.

Changing the pie chart to a bar chart as you requested will come later.

I chose using a simple menu so that after closing the menu, you will still be on the statistics page instead of going to the main screen.

I have avoided adding any strings so that we can make sure we are happy with English before any translation work starts.

So, why do we need this?
This is not a bug fix.  I think it can be useful.  The distribution page is already very cluttered and we have an open PR to add yet one more item to it.  This will allow one to remove what they don't need to clean up the page as they wish.

Please let me know what you think.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/stats/FirstPageFragment.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/stats/FirstPageFragment.java
@@ -10,6 +10,7 @@ import androidx.databinding.DataBindingUtil;
 import androidx.fragment.app.Fragment;
 
 import com.eveningoutpost.dexdrip.databinding.StatsGeneralBinding;
+import com.eveningoutpost.dexdrip.models.UserError;
 import com.eveningoutpost.dexdrip.models.UserError.Log;
 import android.view.LayoutInflater;
 import android.view.View;
@@ -28,6 +29,7 @@ import java.util.List;
  * Created by adrian on 30/06/15.
  */
 public class FirstPageFragment extends Fragment {
+    private final static String TAG = FirstPageFragment.class.getSimpleName();
 
     private View myView;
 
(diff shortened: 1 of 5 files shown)


Navid200:
https://github.com/user-attachments/assets/ccaac038-7a9b-48ac-a8ff-6b06b1e6094e

Navid200:
| Before | After |  
| ------- | ------ |  
| Screenshot_20241009-231358 | Screenshot_20241009-224313 |

Navid200:
The Include menu items:  
Screenshot_20241009-224411

Navid200:
I have a lot of thoughts and can use suggestions:
The AGP page can use the option to adjust the range.  Currently, you may get a lot of white space and any deviations from a straight line would be effectively invisible.  By adjusting the range, you can zoom in and see the variations better if you choose to.

The added menu has a title of "Include".  
We may need a different title for a different set of entries.  Or, we can add another item on the main menu.

When we change the pie chart and make it look more like TidePool or Clarity, we may need options there too.  

This is more of a multi-stage PR.  If you prefer, I can complete everything before submitting it.
The problem with that will be a very large PR.  
It may be easier for all to go one step at a time meaning reviewing this and possibly merging it.  And we make more changes as we move forward to complete the following steps.

Navid200:
I don't know if Include is the right word.  May be it should be `Content`.

Or I wonder if there should be `Settings` and then under `Settings`, we would have `Content` and other settings like `AGP Range` etc.

Let's think about this.

jamorham:
What are the default set of statistics which will be visible?

Is this currently inside a Scrollview? If not would it be better if it were?

Navid200:
I have not changed the settings that re visible by default in this PR.  I can.  I would make PGS invisible  for sure if I had a choice.  I would also like to make predicted A1C invisible as there is more and more recommendation to use time in range instead of focusing on lowering A1C.

If your question about the scroll view is about the menu that currently is called include, which I would like to rename, there is no option for scroll view or anything like that.  The only options are either menu or group.
https://developer.android.com/develop/ui/views/components/menus
I am sorry if I have misunderstood your question.

The persistent high alert problem is taking my time now.  I hope to have a solution for that as soon as I can.

Navid200:
The extra level of hierarchy I have added to the menu has caused the toggle action to be ignored the very first time it is requested.  
Now that I know what action I took has caused the problem, I can better work towards finding a solution.

Navid200:
Ah, I remember now what you told me about the defaults.  I had forgotten and what I was experiencing was exactly that.
I am fixing it following one of the options you suggested.

Navid200:
Thanks for reminding me about the defaults.

I am triggering the defaults update in IdempotentMigrations now.
I changed the heading from `Include` to `Visibility`.  

We are three statistics pages.  The new menu is only related to one of the 3 pages.  But, we may need  settings for the other 2 pages as well.

This is why I still am not certain about how to organize the menus and their hierarchy.

Navid200:
https://github.com/user-attachments/assets/4c2e4db5-522f-471f-9787-be896a9b75cd

Navid200:
This is ready for a review.
There is no change to defaults.  So, if this is merged and someone updates, they will see no difference unless they change something.
We can change the defaults later in another PR.  

If this is merged, the next PR will add capture rate as an option.

Navid200:
> What are the default set of statistics which will be visible?
> 
> Is this currently inside a Scrollview? If not would it be better if it were?

Navid200:
I have made too many comments here.  I also have made many commits.  
I will open another PR.

Navid200:
> What are the default set of statistics which will be visible?
They will all be visible as they are now.  
We can decide in 2 PRs after this if we want to change the defaults.  But, for now, I wanted to make the changes minimal to make the review easier.

> Is this currently inside a Scrollview? If not would it be better if it were?
Yes, the main window showing the parameters is already in a scrollview.  
Screenshot 2024-10-31 080125

jamorham (approved):
Thanks for a comprehensive PR. Sorry for the review delay

Navid200:
Thanks

PR #3719, opened 2024-10-16
Introduce proper wear-only mode for wearos4 and dexcom g6

t0st:
Implements #3676:

- fix direct-connect from wearos4 to dexcom g6
- implement nightscout upload from wearos4 lte
- prevent data transfer to mobile if __only_ever_use_wear__ is active


thepartisan:
@t0st  I am using this PR as basis for an attempt to add G7 support. Could you maybe quickly describe the pairing process for G6 for me here? After porting some more shortId-related things from app to wear I still struggle a bit here.

- As you are using the phone slot, will you pair G6 autonomously without prior phone pairing?
- Which name does G6 advertise?

thepartisan:
@Navid200 , reaching out to you because I saw you commit quite a bit in the wear section...
Do you have any resource that describes how the pairing of app vs wear is supposed to work (design-wise)?
Do I pair the sensors completely new with wear app? I've started an attempt to integrate G7, but I am stuck at the initialization of pairing now and saw that for app there is "the plugin" feature. Do you know if this is this something I would need for G7 or should it theoretically also work based on the existing code?

Is there any interest/place for technical discussions?

Navid200:
@thepartisan Please limit posts in a PR to what is specifically related to that particular PR.  
For everything else, please use [Discussions](https://github.com/NightscoutFoundation/xDrip/discussions).

PR #3716, opened 2024-10-13
Some new translation keys

Navid200:
I looked at the screenshots here:  https://github.com/NightscoutFoundation/xDrip/discussions/3715

And have taken the ones I thought would make sense to translate.  I have left a few that I was not sure about.

I have not made any change to the original English text except one.  There was Android 6 + in one.  I deleted that portion since we have now retired Android 6.

Please let me know what you think.  Thanks  
  
---  
  
The following images mark all the text that this PR moves to strings.  
Screenshot_20241107-165154  Screenshot_20241107-165247 Screenshot_20241107-165406 Screenshot_20241107-170041 Screenshot_20241107-170139

PR #3734, opened 2024-10-26
Persistent High threshold fix

Navid200:
This PR fixes two issues I caused adding Persistent High threshold (https://github.com/NightscoutFoundation/xDrip/pull/3704).

1- You can enter any value as persistent high threshold.
This PR rejects all values that are outside the 40-400 mg/dL range.

2- If xDrip is already installed and mmol/L is selected and xDrip is updated to our latest Nightly (2024.10.11), the threshold will be set to 170 mmol/L.
The user will need to change units to mg/dL and back to mmol/L to correct that to 9.4 mmol/L.
This PR does the correction automatically.


Navid200:
This is ready for a review.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
```diff
@@ -385,6 +386,9 @@ protected void onCreate(Bundle savedInstanceState) {
         setVolumeControlStream(AudioManager.STREAM_MUSIC);
 
         checkedeula = checkEula();
+        if (!Pref.getString("units", "mgdl").equals("mgdl")) { // Only if the selected unit is mmol/L
+            handleUnitsChange(null, "mmol", null); // Trigger the correction of values (defaults) if needed based on the selected unit
```
jamorham, line 390: I don't think this should be here, we should only need to call handleUnitsChange() when the units are changed.

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -755,6 +758,11 @@ public boolean onPreferenceChange(Preference preference, Object value) {
             String stringValue = value.toString();
             if (isNumeric(stringValue)) {
                 final boolean domgdl = Pref.getString("units", "mgdl").equals("mgdl"); // Identify which unit is chosen
+                double submissionMgdl = domgdl ? Double.parseDouble(stringValue) : Double.parseDouble(stringValue) * Constants.MMOLL_TO_MGDL;
+                if (submissionMgdl > MAX_GLUCOSE_INPUT || submissionMgdl < MIN_GLUCOSE_INPUT) {
+                    JoH.static_toast_long("The value must be between " + unitsConvert2Disp(domgdl, MIN_GLUCOSE_INPUT) + " and " + unitsConvert2Disp(domgdl, MAX_GLUCOSE_INPUT));
```
jamorham, line 763: If we want to continue improving localization this should be a string resource format string that has these values formatted in to it.

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -755,6 +758,11 @@ public boolean onPreferenceChange(Preference preference, Object value) {
             String stringValue = value.toString();
             if (isNumeric(stringValue)) {
                 final boolean domgdl = Pref.getString("units", "mgdl").equals("mgdl"); // Identify which unit is chosen
+                double submissionMgdl = domgdl ? Double.parseDouble(stringValue) : Double.parseDouble(stringValue) * Constants.MMOLL_TO_MGDL;
```
jamorham, line 761: tolerantParseDouble() is typically much safer to use and supports localized number input where , can be the decimal point.

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
Navid200: I left everything in English intentionally to make sure it all looks good before adding any of it to strings.

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
Navid200: I will create a video clip and add here showing why we need this.

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
Navid200: Thanks
Will change it

app/src/main/java/com/eveningoutpost/dexdrip/Home.java
Navid200: Install a previous version of xDrip on a test phone.  Change the unit to mmol/L.
Then, update xDrip to our current Nightly.  Go to the Persistent High Alert page.  You will see this as the threshold:

170 mmol/L

Change the unit to mg/dL and back to mmol/L.  You will see it change to 9.4 mmol/L.

Triggering the handle unit change right after xDrip start guarantees that will never happen.

jamorham: ok, but the onCreate method is not once on application start it is whenever the Home activity is created. It should be in `IdempotentMigrations.performAll()` but you need to check that there are no side effects as well because that `handleUnitsChange()` method does a few different things.

Navid200: Thanks a lot.  I will fix all of this in a few hours.

Thanks

Navid200:
Thanks for the catch.

I have changed to tolerantParseDouble().
I have moved the call to handleUnitChaange to IdempotentMigrations.

Thanks for the guidance.

With respect to translations, the same text for the glucose level alerts is still not translated.
I am happy to gradually bring in more and more strings.  This was my attempt along those lines:
https://github.com/NightscoutFoundation/xDrip/pull/3716

Navid200:
I don't know why/how the file "output-metadata.json" was added to the pull request.
Should I remove it somehow?  Or, can you exclude it from merge assuming you want to approve?

app/prod/release/output-metadata.json
```diff
@@ -0,0 +1,20 @@
+{
```
jamorham, line 1: This file should not be in the PR

Navid200: Removing it right now

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -755,6 +759,11 @@ public boolean onPreferenceChange(Preference preference, Object value) {
             String stringValue = value.toString();
             if (isNumeric(stringValue)) {
                 final boolean domgdl = Pref.getString("units", "mgdl").equals("mgdl"); // Identify which unit is chosen
+                double submissionMgdl = domgdl ? tolerantParseDouble(stringValue) : tolerantParseDouble(stringValue) * Constants.MMOLL_TO_MGDL;
+                if (submissionMgdl > MAX_GLUCOSE_INPUT || submissionMgdl < MIN_GLUCOSE_INPUT) {
+                    JoH.static_toast_long(xdrip.gs(R.string.the_value_must_be_between_space) + unitsConvert2Disp(domgdl, MIN_GLUCOSE_INPUT) + xdrip.gs(R.string.space_and_space) + unitsConvert2Disp(domgdl, MAX_GLUCOSE_INPUT));
```
jamorham, line 764: For good localization you shouldn't be building the string yourself with concatenation. Instead use a format string. See how `xdrip_software_changed_format` is handled for an example.

Navid200:
Thanks for the multiple reviews.  I don't know how a compilation file got added to the PR.

I hope you don't mind I will have to close this PR yet again and open a new clean PR without that file and with the string set up as you have advised.

PR #3736, opened 2024-10-27
Lower Voltage B warning level

Navid200:


Navid200:
This is an example of what people see these days:
xDrip-Status-Screenshot-2024-10-27-08-09-21

Navid200:
This change will lower the trigger so that we see red only if the voltage is lower by 25.

jamorham:
This is probably okay but can you provide your rationale for this change?

Navid200:
I used this as a reference: https://github.com/NightscoutFoundation/xDrip/discussions/3408#discussioncomment-9002235

jamorham (approved):

Issue #3749, opened 2024-11-02
System Status - No "Sensor Start/Age" display after xdrip update


Teute0815:
### Discussed in https://github.com/NightscoutFoundation/xDrip/discussions/3744

<div type='discussions-op-text'>

<sup>Originally posted by **Teute0815** November  1, 2024</sup>
I've been using
Libre3 Sensor -> juggluco (410 5.1.5) broadcasting through "Libre2 patched app" -> xdrip (dfcbe80-2024.09.17)
on a
Samsung S9+ Android 10
for quite some time.

The bug occured after automatic xdrip update through stable channel (dfcbe80-2024.09.17)
No other settings, versions or hardware have been changed.

After update, the Classic Status Page didn't display the sensor start date and the sensor age anymore. While being on a running sensor, I waited for that sensor to be replaced a few days later, to see if that would fix the "problem". It didn't. 

With my setting, xdrip doesn't display the sensor start date - and thus the sensor age (which is the quick information I'm missing most) anymore. I reinstalled the previous version again, to make sure it's an xdrip issue. It is - previous version instantly showed sensor start date and age again. It's not only a "**Libre2**" problem but also a "**Nightscout Follower**" source problem, which I tried in the meantime. (see images below) @Navid200

Could that be fixed?

**source "Libre (patched App)"**
![2024-11-02 12 19 42](https://github.com/user-attachments/assets/b9ea3c24-4d3c-4cd1-ab43-55108e886a7a)

In the meantime I tried something new and got my juggluco readings via local webserver with xdrip as nightscout follower. That way I can access missed historical values, which is not possible via juggluco - libre patched.

When using xdrip as nightscout follower, xdrip got the exact same problem:

**source "Nightscout Follower"**
![2024-11-02 12 17 15](https://github.com/user-attachments/assets/6b275904-c16c-40ad-8026-7a28e5662ec7)
</div>

Teute0815:
Nightlys involved are 06.11 (working) -> 06.25 (not working)

Navid200:
@PhilTu This post does not directly address you.  But, since you were a part of the corresponding discussion, I would like you to see this as well.  

I went ahead and opened a PR to fix this I thought.  And tested and could see a start time shown for the Nightscout follower.  

But, then I thought more about it and I don't think this is really an issue!

[...]

PR #3750, opened 2024-11-02
Classic status page session start fix

Navid200:
I removed session start time from the classic status page for G7 here: https://github.com/NightscoutFoundation/xDrip/pull/3531 

But, it is affecting other devices: https://github.com/NightscoutFoundation/xDrip/issues/3749  

This PR changes the logic used to determine if the start time should be shown or not.  It focuses on G7.  It shows it always except for G7.  This should address the problem we are facing now.  

Fixes: https://github.com/NightscoutFoundation/xDrip/issues/3749


Navid200:
I have tested this with G7, G6 and Nightscout follower.

jamorham (approved):

PR #3757, opened 2024-11-04
Enable Use xDrip Cloud by default

Navid200:
This PR enables `Use xDrip Cloud` by default.  
<br/>  
<br/>  
  
Screenshot 2024-11-03 204243


jamorham:
This will come, but not yet.

Navid200:
It was my understanding that this is the setting everyone should be using.
But, obviously I don't have all the facts.

Issue #1280, opened 2020-03-22
App Settings, BG readings Erased, Sensor Disappeared; cause: database deleted


adanvdo:
Today the app lost all BG readings somehow and was displaying "not getting any readings". The sensor is only 2.5 days old. logs showed readings being ignored because it said there was no sensor. Now system status shows sensor as 3 hours old on main screen but 2.5 on the detailed screen. 

Also, a few days ago my alerts were erased. I don't know what's going on but the app is losing data and settings somehow

![Screenshot_20200322-171552](https://user-images.githubusercontent.com/26498008/77260974-793eae80-6c61-11ea-985a-2357c2ddab45.png)
![Screenshot_20200322-171601](https://user-images.githubusercontent.com/26498008/77260975-79d74500-6c61-11ea-9f83-c99cf6868e10.png)
![Screenshot_20200322-171825](https://user-images.githubusercontent.com/26498008/77260976-79d74500-6c61-11ea-9ff5-7567a7e0e524.png)

adanvdo:
This happened to me again tonight on an 8.5 day old sensor @jamorham 
The entire app database is getting erased. the app was stuck in a loop of getting glucose data, logging that no sensor was detected, the bg reading was null, then getting a bunch of backfills.
this was looping for 30 minutes before I restarted the phone and then the app got readings again.
But ALL of my previous readings and settings are gone. Alarms, sync, everything

adanvdo:
@jamorham this happened to me again today. I sent you the logs. everything except the g6 collector info was wiped

adanvdo:
this happened again today @jamorham. I uploaded the logs. 

when I opened xdrip it said "autostarting sensor" but its a new sensor I started yesterday. everything was wiped. main screen had a select source button like i had never used the app

adanvdo:
This happened to me again in the middle of a LOW and I would really appreciate some kind of response @jamorham

tolot27:
Please can you report some more details about your app environment (exact phone type, Android version, patch level, etc.).
For me it looks more like a phone problem rather than xDrip related.

BTW: Did you enable daily database backup in the extended settings?

adanvdo:
Thank you so much for responding @tolot27 

I didnt realize there was a daily database backup. i have enabled that option.
![Attach11868_20200505_065756](https://user-images.githubusercontent.com/26498008/81059315-3bf35080-8e9e-11ea-9b63-b5d77e5fa7d8.png)
[...]

PR #3759, opened 2024-11-06
Database size on status page

Navid200:
This PR adds a new item to classic status page, the database size.  
It removes the phrase "downgradable to" from the same status page.  

| Before | After |  
| ------- | ------|  
| xDrip-Status-Screenshot-2024-11-05-21-53-00 | xDrip-Status-Screenshot-2024-11-05-23-07-26 |

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/SystemStatusFragment.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/SystemStatusFragment.java
@@ -43,7 +43,6 @@ import com.eveningoutpost.dexdrip.models.UserError;
 import com.eveningoutpost.dexdrip.models.UserError.Log;
 import com.eveningoutpost.dexdrip.services.DexCollectionService;
 import com.eveningoutpost.dexdrip.services.G5CollectionService;
-import com.eveningoutpost.dexdrip.ui.activities.DatabaseAdmin;
 import com.eveningoutpost.dexdrip.utilitymodels.CollectionServiceStarter;
 import com.eveningoutpost.dexdrip.utilitymodels.SensorStatus;
 import com.eveningoutpost.dexdrip.databinding.ActivitySystemStatusBinding;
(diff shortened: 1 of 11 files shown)


jamorham:
I think this is very messy what you're doing here. You're trying to use asynchronous features synchronously and creating a lot of weight for something that really doesn't need to be that complicated. See the new utility method I have provided for this. I suggest completely reworking this PR to just use that: https://github.com/NightscoutFoundation/xDrip/commit/53507c76c13de04a02c93359583492173e288e7e

Navid200:
Thanks you so much.
Will be done.

Navid200:
Thanks for the guidance.
It should be good now.
Thanks

olorinmaia:
@Navid200 What is the use-case that benefit of knowing database size?

Navid200:
@olorinmaia If you have been using xDrip for years, the database may contain data you will never need inside xDrip any longer.  Keeping data in xDrip corresponding to 20 years ago is not necessary.  
  
I welcome suggestions as to what could be the cause of this:  https://github.com/NightscoutFoundation/xDrip/issues/1280 
In the meantime, I would suggest to everyone to avoid allowing their database to unnecessarily grow.   
  
Being able to see the size of the database can be useful for deciding when you may want to do some cleanup.

olorinmaia:
Thanks for some info regarding this. If database cleaning is necessary, is this the way or is there other methods? 
https://navid200.github.io/xDrip/docs/Backup/SaveDaily.html
I mean setting Glucose retention to e.g. 365.

Navid200:
> Thanks for some info regarding this. If database cleaning is necessary, is this the way or is there other methods? https://navid200.github.io/xDrip/docs/Backup/SaveDaily.html I mean setting Glucose retention to e.g. 365.

Thanks for that link.  I will need to update that.

Yes, setting glucose retention to a value like 365 is a way to limit the size.

PR #3765, opened 2024-11-10
Persistent High Alert Threshold range

Navid200:
This PR only imposes a range on the persistent high alert threshold.  

I will address the other requirements in another PR to make this one independent and easier to review.  

The following images show the toast message shown when the input is out of range, for the two units, and is rejected.

 
Screenshot_20241109-214259 Screenshot_20241109-214346


Navid200:
> I don't really know what you're doing here. Why don't you just return false from the onPreferenceChange listener if you want to reject the new value?  
  
Done  
<br/>  
  
> If we want to continue improving localization this should be a string resource format string that has these values formatted in to it.  
  
Done  
<br/>  
  
> For good localization you shouldn't be building the string yourself with concatenation. Instead use a format string. See how xdrip_software_changed_format is handled for an example.  
  
Done

jamorham (approved):

PR #3767, opened 2024-11-10
Persistent High Threshold range verification

Navid200:
This PR corrects the value of the preference setting if it is out of range.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Constants.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Constants.java
@@ -61,6 +61,7 @@ public class Constants {
     public static final int ZXING_CAM_REQ_CODE = 49374;
     public static final int ZXING_FILE_REQ_CODE = 49375; // This is created by just incrementing the existing camera scan code from the zxing package
     public static final int SENSORY_EXPIRY_NOTIFICATION_ID = 2003;
+    public static final int OUT_OF_RANGE_GLUCOSE_ENTRY_ID = 2004; // Preference setting out of range
 
 
     // increments from this start number
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/IdempotentMigrations.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/IdempotentMigrations.java
@@ -1,5 +1,8 @@
 package com.eveningoutpost.dexdrip.utilitymodels;
 
+import static com.eveningoutpost.dexdrip.utils.Preferences.MAX_GLUCOSE_INPUT;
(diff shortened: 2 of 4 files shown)


Navid200:
> I don't think this should be here, we should only need to call handleUnitsChange() when the units are changed.  
  
Done.  
We don't use handleUnitsChange() any longer.

jamorham:
Hi, can you explain when you would expect this to activate? What are the circumstances that result in an out of range value? thanks.

Navid200:
Hi,

Thanks for the review.  There are two scenarios that will trigger this:  
  
1- A user already has xDrip installed and mmol/L selected as their unit of choice.
The version of xDrip installed is prior to when we introduced persistent high threshold.
When the user updates xDrip to a version containing the new persistent high threshold, it will be set to 170 mmol/L.
This will never be corrected to 9.4mmol/L unless the user changes their units to mg/dL and changes it back to mmol/L.  
  
This PR detects this situation and corrects it with no need for the user to do anything.  No toast or log will be shown.
<br/>  

2-  The user has xDrip installed prior to this PR but after we introduced persistent high threshold and by mistake or intentionally set the persistent high threshold to a value outside the 40-400mg/dL range.  

After this PR, this situation will be detected and the value will be set to min if the chosen value is less than min or max if the chosen value is greater than max.   A silent notification will be shown to notify the user.  A corresponding log will also be shown.

Navid200:
I forgot to show the silent notification.  here it is:  
<br/>  
Screenshot_20241126-082748

Navid200:
The silent notification only notifies the user that something needs their attention.  There is no much detail in it.  
But, there is more detail in the corresponding log.

Navid200:
I'm sorry about that redundancy.
And thanks for removing it.  I couldn't think of a way to remove it before one was merged.

Thanks so much for all the support.

Navid200:
Would it help if I rebase this now that the other one has been merged?
Or, is there anything else I can do to help?

jamorham:
I think something has got broken with the imports or similar as it is failing unit test:      

` JoH.static_toast_long(xdrip.gs(R.string.the_value_must_be_between_min_and_max, unitsConvert2Disp(domgdl, MIN_GLUCOSE_INPUT), unitsConvert2Disp(domgdl, MAX_GLUCOSE_INPUT)));`

jamorham:
I think it is just missing these imports:
```
import static com.eveningoutpost.dexdrip.EditAlertActivity.unitsConvert2Disp;
import static com.eveningoutpost.dexdrip.models.JoH.tolerantParseDouble;
```

jamorham:
Likely got broken when I tried to resolve the merge conflicts not realizing they were needed.

jamorham:
If you can just pull and add those and push that commit then that should fix it. thanks

Navid200:
Compiling

jamorham:
thanks for the fast response, sorry about the import issue

PR #3803, opened 2024-12-08
Sync battery level on status page

Navid200:
Show master battery level on the follower status page with no need to enable engineering mode:

| Before | After |  
| ------- | ------ |  
| Screenshot_20241208-184034 | Screenshot_20241208-185109 |  
<br/>  
  
---  
  
**Please note**  
This has an unintended ramification:
The battery levels of all followers will also be shown on the status page of the master.  
If this is unacceptable, we will need to complicate the code to only show this on the followers but not on the master.
I hope this is not a problem.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/RollCall.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/RollCall.java
@@ -296,7 +296,7 @@ public class RollCall {
         for (Map.Entry entry : indexed.entrySet()) {
             final RollCall rc = (RollCall) entry.getValue();
             // TODO refactor with stringbuilder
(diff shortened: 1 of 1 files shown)


jamorham (approved):
This will likely show all battery status for all master/followers to each other but it any additional clutter is unlikely to be a problem so its okay to move this from behind engineering mode.

Navid200:
@jamorham Thanks a lor for the review.  

Would you please also have a look at these when you can?  
https://github.com/NightscoutFoundation/xDrip/pull/3663
https://github.com/NightscoutFoundation/xDrip/pull/3707
https://github.com/NightscoutFoundation/xDrip/pull/3716

PR #3817, opened 2024-12-16
Time unit symbols for the Dex status page last connected

Navid200:
Please let's use time unit symbols on the status page for last connected at least.  
Then, I won't have to keep using a translator to find out if the translated unit is seconds or minutes.
  
The following table shows what the status page looks like before and after this PR when last connected is 2 minutes ago.  

| Before | After |  
| ------- | ------ |  
| Screenshot_20241216-114044 | Screenshot_20241216-114517 |  
  
The following image shows what the seconds symbol will look like after this PR:
Screenshot_20241216-114328

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/JoH.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/JoH.java
@@ -754,26 +754,29 @@ public class JoH {
     }
 
     // temporary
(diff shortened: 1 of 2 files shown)


Navid200:
We can close this PR if we can merge this: https://github.com/NightscoutFoundation/xDrip/pull/4015

PR #3820, opened 2024-12-18
Retire sensor location setting

Navid200:
I could be wrong.  If I'm not, we are not going to improve calibration schemes.  We can fix bugs.  But, I don't think we have a use for where users place their sensors.
But, there could be users who keep submitting that data.
If I am right and we are not going to need the data, we might as well remove the setting.

This PR removes the sensor location setting and disables it to be sure going back in time will not break anything.

| Before | After |  
| ------- | ------ |  
| Screenshot_20241217-221630 | Screenshot_20241217-221815 |


jamorham (approved):
Yes this was never really used as far as I know

PR #3823, opened 2024-12-19
Daily intent service no charging requirement

Navid200:
Please see: https://github.com/NightscoutFoundation/xDrip/discussions/3815  

I don't believe we have an issue if the daily intent service does not run daily as long as it runs at least once a week.

But, I have a concern about the requirements.  We currently have a requirement of the phone being in the process of being charged.  
What if the battery level is extremely low?  Wouldn't it be a bad idea to run then even if it was being charged?

This PR replaces the requirement of being charged with a requirement of the battery level not being low.  
<br/>  
  
---  
  
**Tests**  
In progress


Navid200:
The daily intent service runs everyday even when the phone is not being charged.
This is tested on Android 9.

jamorham (approved):
Seems like a good improvement

PR #3829, opened 2024-12-25
G7 Rapid Reconnect - Faster consistent user friendly behavior

Navid200:
This PR makes the user experience with G7 and xDrip consistent regardless of the random nature caused by the G7 Rapid Reconnect function:
  
| Problem | After this PR |  
| -------- | --------------- |  
| There is an 80% chance xDrip will miss the subsequent transmission to a pairing | xDrip is guaranteed to capture the subsequent transmission to pairing  |  
| There is 80% chance that the next capture after pairing occurs in 6-9 minutes | There is 80% chance that the next capture after pairing occurs in 1-4 minutes |  
| An error message may be shown after pairing | No error message will be shown after pairing |  
| xDrip may miss the second transmission after pairing | xDrip is guaranteed to capture the second transmission after pairing |  
| There is 20% chance that xDrip will pair on the once every 5 minute time grid.  Even then, xDrip may miss the second handshake after the pairing handshake. | There is 20% chance xDrip will pair on the once every 5 minute time grid.  xDrip is guaranteed to capture the second handshake after the pairing handshake. |  
  
---  
  
**What activates Rapid Reconnect Transition mode?**
It is automatically activated when xDrip pairs with a G7.  
I refer to this in the logs as the first Rapid Reconnect Transition mode handshake.
  
---  
  
**What happens in Rapid Reconnect Transition mode?**  
xDrip wakes once a minute.  
The Hunting Transmitter note is shown on the Dex status page after pair  approval until a subsequent handshake regardless of any event.
Subsequent to the pair request approval, after xDrip completes another handshake, scan always mode is activated.  
No brain status error is shown on the Dex status page.
  
---  
    
**What disables Rapid Reconnect Transition mode?**  
When either one of the following two conditions is satisfied, the Rapid Reconnect Transition mode is disabled.  
  
1- Two handshakes subsequent to the pair request approval handshake
2- 10 minutes after pair request approval  
  
---  
  
**Could it fail?**  
If the user walks away from the phone after pairing before two subsequent handshakes, the intended process will not complete.  
But, the user can just wait for xDrip to do its job.  Alternatively, the user can forget the device from the Bluetooth list and temporarily change the transmitter ID and wait for a pair request for the Rapid Reconnect Transition mode to activate again.  
  
---  
  
**Test**  
  
The following tests were performed on an Android 11 Motorola Stylus phone.
Before every test, xDrip was uninstalled and all devices starting with DXCM were deleted from the Bluetooth list. 
Then, the version under test was installed.  Then, xDrip battery optimization was disabled.  After that, source wizard was used to choose G7, background location access was granted, and the pairing code was entered.  

For both tests, the transmitter was disconnected more than 15 minutes prior and it was confirmed that it was transmitting once a minute.
For both tests, it was arranged for pairing to occur one 1-minute cycle before the 5-minute cycle transmission.  
For both tests, collection was enabled about 30 seconds before the transmission cycle start.  

Screens recordings in the following table are sped up by a factor of 16.  
| Before | After |  
| ------- | ------ |  
| https://github.com/user-attachments/assets/0e8c0306-fb9f-44ff-bb81-918fc7b0651d | https://github.com/user-attachments/assets/02a373ef-8c51-4f3c-a786-ecc06fd4c608 |  
| Before | After |  
  
---  
  
If the user waits patiently for several read cycles for xDrip to enable always scan mode, it will eventually discover the correct time grid.
The problem is that many users will intervene thinking that something is wrong considering the possibility of the error message and the longer periods of time with no connectivity.  So, they restart the phone or restart collector etc.  Sometimes, these will succeed.  

The guide asks the user to read this: https://navid200.github.io/xDrip/docs/Dexcom/G7OnceAMinuteScan.html
If user follows those instructions, the outcome will be identical to what this PR does.
The problem is that you cannot expect everyone to read all the pages of the guide. 
This PR does the right thing with no need for any human intervention other than keeping the old sensors away and staying with the phone for up to 10 minutes after pairing. 

This PR ensures connectivity as fast as possible taking into account everything that could possibly go wrong.


Navid200:
Let's see what happens if we use our latest Nightly and pair on the wrong grid without human intervention:
  
Screenshot_20241225-225334  Screenshot_20241225-225309
  
After 1 hour since pairing, there is still no proper connectivity.  As G7 goes in and out of Rapid Reconnect, xDrip connects.  But, it still has not found the correct time grid.
  
You can see what I mean by not having connected properly by looking at the time values on the left in the logs.  The time gap is never 5 minutes.  
  
Of course, this can be remedied if these instructions are followed:
https://github.com/NightscoutFoundation/xDrip/discussions/3813  
But, that requires human intervention.  

Now, lets see what happens if we do exactly the same thing after this PR.

Navid200:
@jamorham I'm sorry about the long description.  
You don't have to read the whole thing.  Please just tell me what you think or if you have any question.

Navid200:
On cannot recreate the problem with my Pixel.

jamorham:
Thanks for the very detailed PR. I will have to study it in some detail because making changes to the timing aspect is a very significant change which has to be handled really well. I appreciate there are some inconsistencies already and it is definitely worth a closer look to see if things can be improved but equally I don't find consistent behavior from the transmitter side which makes testing harder.

Navid200:
Before and after for a fully functional G7:
| Before | After |  
| ------- | ------ |  
| https://github.com/user-attachments/assets/d10b761d-509c-41a5-9a45-f60761fef97a | https://github.com/user-attachments/assets/f1c94e54-5918-4162-91ab-3bd364ffeba0 |  
| Screenshot_20250131-104557 | Screenshot_20250131-112400 |

Navid200:
> Thanks for the very detailed PR. I will have to study it in some detail because making changes to the timing aspect is a very significant change which has to be handled really well. I appreciate there are some inconsistencies already and it is definitely worth a closer look to see if things can be improved but equally I don't find consistent behavior from the transmitter side which makes testing harder.

Thanks
This is a critical change and I appreciate a critical review.

I hope I have not misunderstood the comment about testing challenge.
I identify the timing of the transmitter and write it down.  It could be something like: 00:02:50
This implies 00:07:50 as well as the time grid is cycles at 5 minutes.

Now that I have that, I know the rapid reconnect timings will be 00:3:50, 00:04:50, 00:05:50 and 00:06:50.  Of course, there are another 4 at 00:08:50 ...
Having the rapid reconnect timings, I connect to the transmitter.  Then, I disconnect from it and wait 15 minutes.  This guarantees that it is in rapid reconnect mode.  I can also verify using a scanner.

I install xDrip on the test phone and keep collection disabled during all scans.  Each scan lasts about 17 seconds.
When I know there is no scan, I enable G7, and scan the QR code etc.  
When xDrip is ready, I time enabling collection such that xDrip is scanning 15 seconds before the start of the particular rapid reconnect slot I want to test.
I use the exact same slot for my before and after tests for a valid comparison.

jamorham:
Do you think this is still needed? I thought that other changes had improved this overall. Please let me know what you think.

Navid200:
This can remove error messages that should not be shown because they are not real errors and are simply caused by the rapid reconnect behavior of G7.But, relatively speaking, I would look at a few other PRs before this.For example, there is a PR that upload treatments to multiple Nightscout sites.I followed your advice and believe that addressed your concerns.There is another PR that allows a delay to be added to the dex share follower.I hope you can consider those as they bought have been waiting for s while for a review.On Apr 25, 2025 5:45 a.m., JamOrHam ***@***.***> wrote:jamorham left a comment (NightscoutFoundation/xDrip#3829)
Do you think this is still needed? I thought that other changes had improved this overall. Please let me know what you think.

Navid200:
I cannot recreate a problem since the February 18, 2025 release.
But, I cannot find any change that could have caused it.
There is also a release of 2025.02.18b suggesting that something needed to be changed after the initial release.  But, there is no difference between the two!

I am closing this PR because I cannot recreate the problem even though I don't know why!

Issue #677, opened 2018-11-07
Eversense CGM


xLevelster:
I don't get any data when connceting the Eversense to xDrip, can anybody solve this problem?

xLevelster:
Hey thanks for your fast response. I am using the newest Eversense XL for 7 days now. I did not know that I have to patch the Eversense App, since I did not find anything about how to set it up, so I just used the newest xDrip Apk and the latest version of Eversense CGM from the PlayStore. How can I get that patched Eversense App?

AdrianLxM:
You do need the companion app "ESEL": https://github.com/AdrianLxM/Esel/

At the moment you do have to build it for yourself. As up until recently you had to be quite tech-savvy anyway (flash another operating system onto the phone) this was the way to go.

We are working on a more user-friendly option now that you don't have to root the phone anymore.

BernhardRo:
There is an easier solution availabe as well:
https://github.com/BernhardRo/Esel/tree/master/apk
This does not require a rooted phone and no compiling...

tremor:
@xLevelster: Can you provide some informations regarding the system itself? How long does the battery of the transmitter last? Is it easy to remove it (for snorkeling/diving) and attach again? How accurate is it? What  happens when you don't calibrate twice a day?

xLevelster:
Battery lasts up to 36 hours. It is very easy to remove because it is just sticked to your skin with an adhesive. For me it is super accurate. You won't get any readings after 2 hours when you don't calibrate.

BernhardRo:
Yes, the battery lasts for about 36h, but it takes only roughly 10 min to reload it. Therefore it is no problem to easily take it off every day (e.g. before taking a shower) and reload it. For me, this is even an advantage as my skin had problems with the 14 days of wearing a freestyle libre. With eversense my skin gets a break and a new plaster every day and with this I have no problems at all.
In terms of accuracy: the values are somehow noisy, but accurate.

tremor:
Can you fix it without a plaster with armband or 3d printed holder?

xLevelster:
> Can you fix it without a plaster with armband or 3d printed holder?

Yes not really necessary but that should work, depending on how good it stays over the sensor.

tremor:
Ich ask because I have to remove it on some days more than once when I go diving...

BernhardRo:
You can remove the transmitter as often as you want - thats no problem!
[...]

PR #3835, opened 2024-12-30
Eversense 365 US version companion app

Navid200:
https://play.google.com/store/apps/details?id=com.senseonics.eversense365.us&hl=en-US  

https://www.ascensia.com/siteassets/press/eversense-365/eversense-365_media-faq-vfinal_2.pdf


Navid200:
It doesn't look like the app produces a notification for every reading, only for alerts.

jamorham:
So its not worth merging this do you think?

Navid200:
No, I don't think so.  
Unless we consider it a test and remove it later.

Navid200:
It seems it works: https://github.com/NightscoutFoundation/xDrip/issues/677#issuecomment-2614614336 

But, it may need improvements.  So, please merge.

jamorham (approved):
Looks good, thanks for following this up

PR #3853, opened 2025-01-10
AOD bottom center setting fix

Navid200:
I have tested this for each one of the 5 sections and they all seems to be working fine.
Testing our latest Nightly as is with bottom center selected only, I can see that the readings moves to all regions.
After this PR it does not and stays where it should.

The following screenshot is meant to help with the review.
Screenshot 2025-01-10 161428


Navid200:
This PR fixes the issue reported here: https://github.com/NightscoutFoundation/xDrip/discussions/3852

jamorham (approved):
Good catch

PR #3854, opened 2025-01-12
Stop an expired sensor

Navid200:
After a G7 expires, we don't stop the internal session.
When xDrip connects to the next sensor, it thinks that there is an active internal session.  Therefore, it requests a full backfill causing a backfill overlap.

When we stop a G6, we also stop the internal session.  After that, when we connect to a G7, xDrip starts a new session and takes into account the last readings and avoid a backfill overlap.

This PR will avoid a backfill overlap for G7 to G7 as well.  
<br/>  
   
---  
  
**Testing**  
Unfortunately, I can test this in 10 days I hope.  I will update then.

However, I will be able to test G6 to G7 before then to make sure nothing is negatively affected there.  I will add this test's results in a day or two.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/g5model/CalibrationState.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/g5model/CalibrationState.java
@@ -109,6 +109,10 @@ public enum CalibrationState {
         return !stopped.contains(this);
     }
 
+    public boolean sensorStopped() {
+        return stopped.contains(this);
+    }
+
     public boolean sensorFailed() {
         return failed.contains(this);
     }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/g5model/Ob1G5StateMachine.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/g5model/Ob1G5StateMachine.java
@@ -1018,7 +1018,6 @@ public class Ob1G5StateMachine {
         }
 
         // TODO check firmware version
-        stopExpiredSession(); // Stop the internal session if the sensor has expired.
         if (glucose.calibrationState().readyForBackfill() && !parent.getBatteryStatusNow) {
             backFillIfNeeded(parent, connection);
         }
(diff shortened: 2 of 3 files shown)


Navid200:
@Der-Schubi This should fix the problem.  I will test in 10 days and update.

Der-Schubi:
Any chance we can re-enable the sensor start time in the first system status page?
Should show correct values now. We could both test this in a few days.
The wrong sensor start values only appeared after you've hidden the stop sensor menus for the G7.

Der-Schubi:
And one more question: Would the session be stopped when I enter a new sensor code? E.g. when the sensor fails and I have to set a new one. Maybe we can stop the session directly when pairing a new sensor. Automatic session start is shortly after that with the first values, right?

Navid200:
The most important piece of information is when the session will stop.
We provide that information on the Dex status page.  If the user chooses, they can also see the same on the main screen.

If the user needs to know when they started the sensor, they can easily calculate it by subtracting 10.5 days from the expected stop time.

If we bring back the start time to the status page, we will only add confusion for the inexperienced.  
And as a consequence, the number of questions in the forums will go up.  
And this will be the question:
xDrip shows that my session will end in 2 days.  When I look at the status page, it shows that the sensor started 8 days ago.  That's only 10 days!  Shouldn't the sensor work for 10.5 days?  Why is xDrip wrong?

And believe me, the question will actually be asked like that?  They won't say what am I missing?  They will ask why is xDrip wrong?
And who is going to have to answer those questions?  It will be me.
Please, let's not bring back the start time.  There is really no need for it.

If this PR is merged, the only time the internal session will be stopped is if the user waits for their sensor to expire before switching to their new sensor.
If you keep switching your sessions before the sensor expires, you could have an internal session that is 100 days old.  But, why does it matter?  
The only reason we needed an internal session was that we did everything for G5.  We calibrated G5 and we  ran it more than 7 days.  In that case, we needed to know when our session started.  There is no need for that any longer.
We can do everything differently now.  But, it would be a lot of grief to write everything from scratch.  So, we are still using the internal session being active or not for some logics.
So, the short answer to your question is, no, the internal session will not be stopped if you change from one to the other.

Der-Schubi:
> We provide that information on the Dex status page.

Where? I can't see that information anywhere in xDrip. I only see the sensor status, that says e.g. OK, 10.4 days. But that does not tell me when the session will stop. And it's resolution is only 0.1 day. Therefore I use a notebook to write down the sensor start time manually. Not very user friendly...

Navid200:
We should really move this conversation to a discussion to avoid reducing the chance of this PR being approved.  

Screenshot_20250112-090059

The item I have marked is reported by the device.  It is not something xDrip calculates.  It is as accurate as it can be.  So, is the expiry note at the bottom of the main screen if you enable it.

PR #3859, opened 2025-01-18
CareLinkFollower - v11 data endpoint outside US

benceszasz:
**Issue**
The old cloud data endpoint no longer works outside the US, so the new v11 data endpoint must be used in the EU  for 7xxG pump users . The data structure of the new data endpoint is different from the previous one, so the data structure must be extended and modified while maintaining compatibility with the old cloud data structure and also with the one used by the standalone sensors.

**Solution**
Implementation of the new v11 cloud data endpoint:
- replace endpoint in case of EU with v11 endpoint
- new and modified structures
- create new date correction logic for the new data structure (dates have no timezone)
- update data processing for new structures
- add JsonAdapter and structures to proguard

**Testing**
Additionally, other EU pump users have also checked the new version and found it to be working properly:
https://github.com/NightscoutFoundation/xDrip/discussions/3855#discussioncomment-11872395

### DIFF
--- a/app/proguard-debug.pro
+++ b/app/proguard-debug.pro
@@ -59,3 +59,12 @@
 -keep class com.newrelic.** { *; }
 -dontwarn com.newrelic.**
 -keepattributes Exceptions, Signature, InnerClasses, LineNumberTable, SourceFile, EnclosingMethod
+
+-keep class com.eveningoutpost.dexdrip.cgm.carelinkfollow.message.*
+-keep class com.eveningoutpost.dexdrip.cgm.carelinkfollow.message.util.*
+-keep class com.eveningoutpost.dexdrip.cgm.carelinkfollow.message.** { *; }
+-keep class com.eveningoutpost.dexdrip.cgm.carelinkfollow.message.util.** { *; }
+-keep class * extends com.google.gson.TypeAdapter
+-keep class * implements com.google.gson.TypeAdapterFactory
+-keep class * implements com.google.gson.JsonSerializer
+-keep class * implements com.google.gson.JsonDeserializer
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -150,3 +150,10 @@
 -keepattributes Exceptions, Signature, InnerClasses, LineNumberTable
 
 -keep class com.eveningoutpost.dexdrip.cgm.carelinkfollow.message.*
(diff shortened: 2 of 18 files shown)


benceszasz:
There may still be room for improvement and fine-tuning, but without v11 endpoint support, 7xxG pump users will not be able to use CareLink Follower in the EU.

Navid200:
Thanks a lot

jamorham (approved):
Looks good, many thanks for updating this.

kukuchta:
This new endpoint could be possibly derived from the discovery API endpoint. In the JSON loaded from https://clcloud.minimed.eu/connect/carepartner/v11/discover/Androida/3.2 there is a field:
"baseUrlCumulus":"https://clcloud.minimed.com/connect/carepartner/v11"

It is used this way in alternative Carelink python client from Ondrej Wisniewski.

benceszasz:
This would require significant changes, as the application configuration is currently loaded and used by the `CareLinkAuthenticator` class. This would need to be outsourced to make it generally available.

@kukuchta
If you have time, I'd be happy to leave this up to you :)
I've gone a little overboard with this now.

kukuchta:
Sure, I will try to follow up with this when I have some free time.

I have to finish my current idea for 780G first:


Sadly I think it does not fit the official version anymore, I had to adjust some of the tables in xDrip database a bit.

Sorry for off-topic😉

benceszasz:
Nice 🙂
I have added basal using `APStatus` in the `CareLinkDataProcessor`, but never had time to finish it. I had two issues and never had time to finish it:
1. manual basal data: detect start and end of manual basal, amount and time for the basal entries (small basals are not divided into 12 equal amounts)
2. upload basal to Nightscout from APStatus

jcsupelano:
> Sure, I will try to follow up with this when I have some free time.
> 
> I have to finish my current idea for 780G first: 
> 
> Sadly I think it does not fit the official version anymore, I had to adjust some of the tables in xDrip database a bit.
> 
> Sorry for off-topic😉

Hi @kukuchta, do you think would it be possible for me to try the version shown in your screenshot. I just think it's amazing. Cheers

kukuchta:
Sure, its here: https://github.com/kukuchta/xDrip/releases/tag/2025.01.24

PR #3872, opened 2025-01-26
Backfill overlap avoidance

Navid200:
I made two major mistakes in this PR: https://github.com/NightscoutFoundation/xDrip/pull/3557 
And it is not working.  
I believe this PR is a solution.  
  
My first mistake was that I overlooked that there are two different lines in the code that auto start a sensor for a session in progress.  That PR attempted to address only one of them (Ob1G5StateMachine) and not the other (Home).

My second mistake was that I incorrectly used a method for finding last readings overlooking the need to ignore the sensor.  By not ignoring the sensor, the method would detect that there is no local session and return null causing a full backfill causing overlap.  

This PR addresses those two oversights.  It also addresses a cosmetic issue.  it creates a log showing that we have started an internal session because there is a session in progress on G7.  
<br/>  
  
**Tests**  
I have tested G6 to G7.  The following shows an Android 11 Motorola.
Screenshot_20250126-115910
  
Screenshot_20250126-115940
  
I will test G7 to G7 in 5 days and report the results here.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -2624,9 +2624,11 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
         boolean isSensorActive = Sensor.isActive();
 
         // automagically start an xDrip sensor session if G5 transmitter already has active sensor
-        if (!isSensorActive && Ob1G5CollectionService.isG5SensorStarted() && !Sensor.stoppedRecently()) {
+        if (!isSensorActive && Ob1G5CollectionService.isG5SensorStarted() && (!Sensor.stoppedRecently() || shortTxId())) {
             JoH.static_toast_long(getString(R.string.auto_starting_sensor));
-            Sensor.create(tsl() - HOUR_IN_MS * 3);
+            Sensor.create(tsl() - Ob1G5StateMachine.getLatestReadingsTime()); // Create an internal session after last existing readings
+            UserError.Log.ueh(TAG, "Started internal session (native session in progress H)");
+            Treatments.sensorStartIfNeeded();
             isSensorActive = Sensor.isActive();
         }
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/g5model/Ob1G5StateMachine.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/g5model/Ob1G5StateMachine.java
@@ -134,7 +134,8 @@ public class Ob1G5StateMachine {
     private static volatile AuthRequestTxMessage lastAuthPacket;
     private static volatile boolean backup_loaded = false;
     private static final int OLDEST_RAW = 300 * 24 * 60 * 60; // 300 days
-    private static long relAutoSessionStartTime = HOUR_IN_MS * 3;
+    private static long noOverlapBackfillRelTime = HOUR_IN_MS * 3;
+    private static boolean IGNORE_SENSOR = true;
 
     public static long maxBackfillPeriod_MS = 0;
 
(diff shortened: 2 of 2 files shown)


Navid200:
I should be able to test G7 to G7 in 12 hours.

Navid200:
It seem to be working for G7 to G7 as well.  
The following screenshots show what I have now after an active sensor expired and I switched to my new sensor that had been running for about 12 hours.

Screenshot_20250131-205359 Screenshot_20250131-205535  
  
Screenshot_20250131-204127 Screenshot_20250131-204256

Navid200:
Test results with another sensor.  There is no overlap.  
  
Screenshot_20250210-203151 Screenshot_20250210-203725  

Screenshot_20250210-203736 Screenshot_20250210-203824

Navid200:
I am running this as a part of my xDrip and everything seems to be OK with it.

Navid200:
People are experiencing the overlap and have no idea what's going on:
481663686_9313854388672549_6891812943580043433_n
  
This is ready for the merge.  Or, please tell me what you don't like.

Navid200:
Please merge this.  Backfill overlap is broken without it.

Navid200:
Would you please merge this?  I have been using this for several weeks now.  It is working perfectly fine.

Navid200:
@Der-Schubi Have you tried this?
I see no issues with it.  Do you?

Der-Schubi:
> @Der-Schubi Have you tried this? I see no issues with it. Do you?

Yes, running it since 28th of January. Had no issues since then.

Navid200:
We don't auto start a G6 if it has just been stopped.  This is not going to work for G7.   G7 may stop (expire) and we may switch to the new sensor right away.  It should auto-start right away in that case.  That's why this PR adds an exception (for G7) to the requirement of it not having been stopped recently.

Instead of always auto-starting at 3 hours ago, we make it a variable.  If it is a G7, it may need to auto-start 24 hours ago.  But, if we have had readings more recent than that, we only auto start from the last reading.  So, we now have a new method, getLatestReadingsTime(), that looks at everything and come up with exactly how long ago we should auto start at.

The new method returns: 
**24 hours ago** if we are using G7 and have had no readings since
**3 hours ago** if we are using G6 and have had no readings since
**How long ago** the last reading was if we are using G7 and last reading was less than 24 hours ago
**How long ago** the last reading was if we are using G6 and last reading was less than 3 hours ago

@jamorham Please tell me what you don't like about this.  I cannot fix it if I don't know.

jamorham:
Hi, can you edit the PR description to describe the problem that this is intended to solve? 
There are likely some fixes needed but I'm concerned that this can break existing logic in certain situations. It needs to be very carefully handled.

Navid200:
> Hi, can you edit the PR description to describe the problem that this is intended to solve? There are likely some fixes needed but I'm concerned that this can break existing logic in certain situations. It needs to be very carefully handled.

I have edited the first post adding exactly what problem this fixes.
Please let me add that the consequence of the problem can also be observed when a user switches from G6 to G7.

Navid200:
What happens if the user was following Nightscout or xDrip sync?  What would happen then when switching?  Does xDrip have a sensor for a follower?
This PR will avoid an overlap regardless of what the previous readings are for, another G7, or xDrip Sync follower or Nightscout follower or even fake data source.

I don't know why you mention code duplication.  I have created a method to avoid code duplication.  There is no code duplication in this PR.

We are using xDrip code that has dependence on a local sensor for a device that does not care about a local session.  But, we cannot rewrite xDrip now.  Considering we approved raw capability for modified G6, we will be stuck with this arrangement, dependence on local session, for a very long time.  So, we need patch work.
This PR and what you suggest both are patch works.  I am not saying that as a criticism.  But, it is only fair to state that both alternatives you are looking at are patch works.

One (this PR) reduces the dependence on the local session and is already tested.
The other (what you propose) further increases the dependence on the local session.  And it still needs to be tested.

Navid200:
My intent was to fix the problem with as simple as a change as possible.  If I look at this a year from now, I want to understand it right away.
It is possible I misunderstand what you are suggesting.  But, if we have to stop a sensor and then start another, that is already more complicated than what this does.  So, I worry that it will be something harder to understand after I forget them both in a year from now.
Regardless of how we address this, we are dealing with a local sensor that does not really exist.  The only session that really matters is on G7.  So, this is what we have to do to fix the problem without rewriting xDrip from scratch.  Let's use the simplest shortest solution.

jamorham:
I will check out the other code paths for other collectors to see what issues there could be.
You have code duplication between `Home` and `Ob1G5StateMachine` from a single responsibility design point of view what I'm saying is that if we can put logic in to `Sensor` to make it more consistent then that could help in other areas like backfilling.

We already shouldn't be asking for backfill beyond the sensor start time. One of the issues at the moment is that sensor sessions can overlap but we don't want to allow that because xDrip isn't designed to track multiple sensors simultaneously.

Navid200:
Users have been overlapping sessions since G6 transmitters became available at low cost.
So, it is not valid to think this is only related to G7.

So, overlapping remote sessions is a reality.  We have no choice over that.
As you have stated, we don't backfill prior to the start of a session.  To avoid backfill overlap, all we need to do is to make sure we never auto start a sensor prior to any existing readings.
That's exactly what this PR does.

Navid200:
I did some tests to remind myself of the details.
This problem of backfill overlap only occurs if the internal sensor sis stopped.  
If there is an internal session and we just switch to a new G7 or G6, there will be no backfill overlap.  
And I believe it will be the same with Nightscout.

But, if the user stops the internal session or if G7 expires, which stops the internal session, switching to a new G7 will do an auto start because xDrip detects a remote session with no active internal sensor.  But, the auto start does not have the required intelligence to start the session such there there would not be an overlap with the previous readings.

In fact, if I switch to my new G7 just before my current G7 expires, there will never be an overlap.
But, we cannot expect users to always switch before their current G7 expires.
I wish I could read your mind to be able to exactly understand what concerns you about this.
To me, it is the easiest way to address an annoying problem with no impact on anything else.
The impact is limited to G7 and G6.
If you tell me, I will happily make it even limited to just G7.  But, it should be only a plus if we offer the same protection to ?G6 users.

Navid200:
The new method returns the time of the last reading.  To limit this PR to G7, I will add a check inside the method to detect of we are using short ID.  If not, it means we are using G6 and in that case, I will have the method return 3 hours ago.  This effectively makes this PR exactly as xDrip does now for G6.
I cam do that if that will make this less risky in your opinion.

Navid200:
> So I think a different approach might be better to tighten up the overall handling of things. I've done some work on this already, but please tell me what you think.
> 
> 1. Make it so that only a single sensor object can be active at any time. So that creating a new sensor will close the old session either with the current timestamp or the last received reading timestamp that is not before that sensor's start time.
> 2. Make it so that when we create a sensor object, the start time of this cannot be before the stop time of the previous sensor object.
> 3. Avoid creating backfill entries before the current sensor start time so that sensor X's backfill data cannot be put in to sensor Y's time period.
> 
> I believe this would all the same flexibility while ensuring that backfill cannot overlap without having to apply fancy rules to decide exactly how to create the sensor and avoid code duplication where it is used in different places. Let me know if you can think of cases where this could fail to work correctly.

Your 1, 2 and 3 are perfect.  I see nothing wrong with any of them.  Except that I prefer you don't spend time on this.
I prefer your time to be reserved for when we have a serious problem like xDrip sync stopping to work or Dex collector breaking after Android 11.  That's when we need you when no one else can do anything to help us.

This PR does what's needed and starts a session intelligently not ignoring existing readings.
If you would like to spend time on the side and work on this, you still can.  If you merge this PR, it will not stop you from working on what you have suggested.
But, if I dare say I have a lot more serious issues I prefer you spend time on.  I give you a few examples: Audio focus for speak readings - Volume profile for persistent high, forecast low and missed reading alert - A unique notification channel for the number icon in the notification area ...

Please let me address easy things.  This PR is good enough for what we need and will not break anything.  Please look at it from that angle.  Thanks for your consideration.

PR #3873, opened 2025-01-28
Broadcast rounded sensor start time

Navid200:


Navid200:
I opened this PR and then realized it was a wrong approach.

As a G7 user, I don't care when exactly my sensor started.
What matters to me is how much longer it is going to work.  
You may say those two are the same.  Well, those two are related but are not the same.

The transmitter continuously tells us how long ago it started.  That is the time parameter the transmitter sends.  The transmitter has no idea what time it is.  It sends an integer value which is the number of seconds since it started.  The transmitter only talks with respect to relative time.   It has no idea where on the planet it is and what time zone it is in and it has no idea what season it is and what day of the week it is.  All it knows is how long ago it started.  

We have already changed xDrip.
We used to show on the xDrip classic status page when the sensor started.  We don't do that any longer for G7.

xDrip only uploads to Nightscout a sensor start when we start a new sensor or when it discovers a session in progress and automatically starts a sensor.  

I suspect that AAPS looks at the sensor start time and when it sees a change, it concludes that there must be a new sensor.
Can that be changed?  Can we add a threshold to it?  For example, can we say if the start time changes by more than an hour, we conclude that we must have a new sensor?

PR #3875, opened 2025-01-30
Notify and log failure to upload to all intended Nightscout sites

Navid200:
When uploading to more than one site, a silent notification is presented and a log is created once a day if there is a failure to upload to one site but not the other.  

If some day we add dedicated (per site) queues, all of this will be removed because there will be no need for any of it then.

I am sorry I have added a new setting.  Unfortunately, someone may use Nightscout and not care about inconsistent uploads.  In that case, having to acknowledge a notification everyday may not be convenient.  
  
| Before | After |  
| ------- | ------ |  
| Screenshot_20250130-092153   | Screenshot_20250130-080824 |  

**Notification**  
Screenshot_20250130-080725
  
**Log**  
Screenshot_20250130-080638  

**Translations**
The notification and its title will be translated.
The log will not be translated.  
<br/>  

**Tests**
This has been tested on Motorola Android 11 and Motorola Android 8.
There is no notification if all uploads fail (no WiFi).
There is notification and log when there is a failure to upload to one site but not the other.  By default, the notification is enabled.
Disabling the notification will still generate a log but no notification.
To cause failure to upload to only one site, the Google Cloud Nightscout site was temporarily stopped.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/DailyIntentService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/DailyIntentService.java
@@ -20,6 +20,7 @@ import com.eveningoutpost.dexdrip.models.UserError.Log;
 import com.eveningoutpost.dexdrip.utilitymodels.BgSendQueue;
 import com.eveningoutpost.dexdrip.utilitymodels.CalibrationSendQueue;
 import com.eveningoutpost.dexdrip.utilitymodels.IncompatibleApps;
+import com.eveningoutpost.dexdrip.utilitymodels.NightscoutUploader;
 import com.eveningoutpost.dexdrip.utilitymodels.Pref;
 import com.eveningoutpost.dexdrip.utilitymodels.UploaderQueue;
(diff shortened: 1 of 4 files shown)


Navid200:
This is now being used as my active xDrip on Pixel 6a Android 15 and is working fine.

Navid200:
I get this notification almost everyday.  I have disabled upload on mobile.  I suspect that when I leave a Wifi zone, it is possible that one site upload completed but the other has not and that may be why I have this everyday.

Please merge this so that users are notified when this happens.

Navid200:
I have used statis parameters for track when there has been a failure.  But, it could take close to 24 hours for the notification to be issues for the parameter to be reset.  In the meantime, if the user restarts the phone, I don't believe the static parameter will survive.

I am going to change this to use the PersistentStore option.

jamorham (approved):
Looks like it should work

Navid200:
Thanks

Navid200:
Yes, I have been running it for a month at least

PR #3877, opened 2025-01-31
Retire G4/G5

Navid200:
Can you tell me what you think of this?

I can change anything you don't like.

| Before | After |  
| ------- | ------ |  
| Screenshot_20250130-205023 | Screenshot_20250313-224626 |  
| Screenshot_20250130-205044 | Screenshot_20250312-115136 |  
| Screenshot_20250313-230053 | Screenshot_20250313-224716 |  
| Screenshot_20250130-205110 | Screenshot_20250130-204701 |  

I have tested this and it is ready for review.


Navid200:
There are people who are still using G5.  This I suspect is partly because they see no reason to use them and keep them just in case.  

If we retire G5, it doesn't mean no one can use G5 any longer.  The older releases can still be used.  It is not a good idea to force people to use our older versions.  But, our current stable release supports G5.  I am not suggesting that they use a 2 year old release.  As long as we continue to support G5, people will have no reason to use their sensors.  They may never use them.  
Their thinking is if some days something happens that causes them not to get G7, they have something to get them going.  if something like that happens, they can use an older version of xDrip as it is an emergency case.  it is not something that is going to continue for weeks.  it is something until they resolve the hold up on their new supplies.

Navid200:
This comes up often when someone cannot find the specific device name in source wizard.
<br/>  
Screenshot 2025-03-01 080933
<br/>  
  
This PR addresses that by having everything under Dexcom.

jamorham:
Yes we can do this but I don't want to use trademarked names if we can avoid it, so I would prefer Dex and picture of G6 and G7 as options.

Navid200:
Is it OK now?

Navid200:
It is not clear since I am showing two devices in the image.
I can add two different options one for G6 and one for G7 both choosing the same setting.  But, that is going to cause the misunderstanding that if someone chooses one, they will have to change a setting to switch to the other, which will be incorrect.

Can the image just show a frame with Dex shown in the middle?  So, the title will be Dex and the corresponding image will also show Dex in a frame.

Navid200:
I added a border around each one of the two images hoping it would make it clear that the option Dex stands for either G6 or G7.

Under Hardware Data Source, I have changed G6/G7/1/1+ Transmitter to Dex.

jamorham:
Okay this looks good but I need confirmation that you own the copyright for the image such that you can submit it in to the project where it will be under the GPL license. Typically this means that the original source of the image is your own work/photograph and hasn't been taken from another source?

Navid200:
I took the two photographs.  I then used Paint in Windows to merge them together into a single image.
It is my work and I am happy for it to be used in xDrip.  That's why I created it.
I have not used any material from the web or anywhere else.

jamorham:
Ok thanks that's what I needed.

PR #3886, opened 2025-02-08
Offering Forecast Low Threshold as an independent setting

Navid200:
This PR adds the option to adjust your forecast low threshold without an impact on statistics or the colors in your graphs.
After this PR, by default, xDrip acts as it does now using the Low Value as the forecast Low Threshold.
But, the user will have the choice of disabling the setting "Threshold: Low Value", in which case, the specified threshold will be used instead.  

This is almost identical to the PR below except that was for the Persistent High threshold: https://github.com/NightscoutFoundation/xDrip/pull/3704  

| Before | After |  
| ------- | ------ |  
| Screenshot_20250207-190148 | Screenshot_20250207-185952 |

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/BgGraphBuilder.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/BgGraphBuilder.java
@@ -129,6 +129,7 @@ public class BgGraphBuilder {
     private SharedPreferences prefs;
     public double highMark;
     public double lowMark;
+    public double forecastLowMark; // Marker used for forecast low analysis
     public double defaultMinY;
     public double defaultMaxY;
     public boolean doMgdl;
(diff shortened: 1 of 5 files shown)


Navid200:
| Threshold | Threshold is Low Value |        |  
| ----------- | ------------------------- | ---- |  
| 3.5 | Unselected | Screenshot_20250207-192152 |  
| 4 | Unselected | Screenshot_20250207-192225 |  
| 4.5 | Unselected | Screenshot_20250207-192246 |   
| 6 | Unselected | Screenshot_20250207-192310 |  
| 6 | Selected | Screenshot_20250207-192329 |

jamorham:
This looks like it should work. I'm not sure what will happen if a user puts some stupidly high value in

Navid200:
Let me answer your question differently.

1- What happens if the user enters a threshold greater than 400mg/dL (22.2mmol/L)?  xDrip will reject it with a toast message informing the user what the acceptable range is (40-400).

2- What happens if the user enters a threshold that is less than 400mg/dL (22.2mmol/L), but is still an unreasonably high value?
xDrip will set the threshold to that value.  As a result, the forecast low alert will never trigger. 

This is the same as if the user sets a glucose level low alert threshold to a very high value.  It will never trigger.

If you want us to reject the value, we can.  But, we will have to do the same for the persistent high alert and all the glucose level alerts.

For example, we can change the acceptable range from 40-400 to something like 100-400 for all high alerts and 40-100 for all the low alerts.
I can do that.  But, that can all be handled in a separate PR.

jamorham (approved):

Navid200:
Thanks

PR #3888, opened 2025-02-09
CareLinkFollower - v11 endpoint for US also

benceszasz:
**Issue**
The old cloud data endpoint no longer works inside the US, so the new v11 data endpoint must be used in case of US also for 7xxG pump users .

**Solution**
Use the new v11 cloud data endpoint for every region in case of cloud datasource.

**Testing**
Additionally, other US pump users have also checked the new version and found it to be working properly:
https://github.com/NightscoutFoundation/xDrip/discussions/3855#discussioncomment-11953103


Navid200:
Thanks

jamorham (approved):
Thanks

marioyeloy:
Muchas gracias, donde me lo puedo descargar?

benceszasz:
@marioyeloy 
In the latest nightly release:
https://github.com/NightscoutFoundation/xDrip/releases/tag/2025.02.14

PR #3901, opened 2025-02-16
Add strings to make xDrip more translated and fix logging of BG alert to be in English only

olorinmaia:
Some obvious places missing translation. This will help getting xDrip app seem more complete in terms of localisation.

Below are screenshots of where these strings are seen in app.

| List of Glucose alerts: | Warning when Data Source is disabled in main screen:|
|--------|--------|
| image | image|
| BG alert:| Touch and hold on xDrip logo on main screen to display below settings: |
|Alert in notification:imageAlert in log now English only:image| image|

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/AlertList.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/AlertList.java
@@ -52,7 +52,7 @@ public class AlertList extends ActivityWithMenu {
 
     String stringTimeFromAlert(AlertType alert) {
         if (alert.all_day) {
-            return "all day";
+            return getString(R.string.all_day);
         }
         String start = timeFormatString(AlertType.time2Hours(alert.start_time_minutes), AlertType.time2Minutes(alert.start_time_minutes));
         String end = timeFormatString(AlertType.time2Hours(alert.end_time_minutes), AlertType.time2Minutes(alert.end_time_minutes));
@@ -61,9 +61,9 @@ public class AlertList extends ActivityWithMenu {
 
     HashMap<String, String> createAlertMap(AlertType alert) {
         HashMap<String, String> map = new HashMap<String, String>();
-        String overrideSilentMode = "Override Silent Mode";
-        if (alert.override_silent_mode == false) {
-            overrideSilentMode = "No Alert in Silent Mode";
+        String overrideSilentMode = getString(R.string.override_silent_mode);
+        if (!alert.override_silent_mode) {
+            overrideSilentMode = getString(R.string.no_alert_in_silent_mode);
         }
         // We use a - sign to tell that this text should be stiked through
         String extra = "-";
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -2465,7 +2465,7 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
             updateCurrentBgInfoCommon(collector, notificationText);
         }
         if (collector.equals(DexCollectionType.Disabled)) {
-            notificationText.append("\n DATA SOURCE DISABLED");
+            notificationText.append(getString(R.string.__data_source_disabled));
             if (!Experience.gotData()) {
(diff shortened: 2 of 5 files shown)


Navid200:
Thanks for the submission.
We should never translate any logs. 

I just checked and can see that the following for sure affects a log:
Screenshot 2025-02-15 200240  
  
I have no idea about any other one of the items you are translating in this PR.
Would you please verify?

If you feel it is necessary to translate something that affects logs, would you please create a new variable for it and leave the one used for logs alone?

If the logs are not modified, I don't see a problem myself.

Thanks

Navid200:
Fundamentally speaking, if the app did not heavily rely on external guides/documentation, you could translate a lot of its content.
But, just look at the home shelf you are translating.  How is someone who needs help with an item that happens to be there is going to find it if all the documentation is in English?  https://navid200.github.io/xDrip/docs/HomeShelf.html

It may seem that translating as much as you can is going to help those who don't speak English.
But, I ask you to consider the documentation.  Who is going to translate the documentation to those languages?  I cannot do that.

Are you sure you are helping those who don't speak English?

olorinmaia:
> Fundamentally speaking, if the app did not heavily rely on external guides/documentation, you could translate a lot of its content. But, just look at the home shelf you are translating. How is someone who needs help with an item that happens to be there is going to find it if all the documentation is in English? https://navid200.github.io/xDrip/docs/HomeShelf.html
> 
> It may seem that translating as much as you can is going to help those who don't speak English. But, I ask you to consider the documentation. Who is going to translate the documentation to those languages? I cannot do that.
> 
> Are you sure you are helping those who don't speak English?

Thanks for feedback!

I don't agree with you on this Navid. I use xDrip in Norwegian, but I read your documentation in English, if I can't find the setting I change the language in xDrip to English. If any other languages want to make a version of wiki in their native language there is noone stopping them, but for Norwegian at least, it's not necessary. For AAPS we got local language translations of wiki, thus it helps that the app is fully translated.

It looks very strange to have 7/8 items in the HomeShelf translated, why not the last one? 

Regarding issue with translating log entry I agree that it shouldn't be translated. I will try to address that in a commit in a few moments. I didn't see that when testing, thanks :)

olorinmaia:
@Navid200 Actually, the Alerting of BG as it today got translation in it. I fixed so that it is purely in english now with latest commit. So it will improve upon that also.

See screenshots:
Alert in notification:
image

Alert in log:
image

Navid200:
Many of those who ask for help using xDrip are not as tech savvy as you are.  Not even close.  

Yes, I am aware that items have got translated in xDrip that were not supposed to.  I had seen translated alert logs on facebook.  That is why I would like us to make it a procedure to ensure no log gets translated from now on.  

Thanks a lot for seeing and correcting that.

olorinmaia:
> I have no idea about any other one of the items you are translating in this PR.
> Would you please verify?

I've checked the other strings, and I can't see them being used in logs.

Navid200:
@olorinmaia 
Thanks,

Another challenge we sometimes have with translations is when a translated value becomes much larger the original English resulting in an overlap with another filed that results in a problem that the original designer of the layout never considered.  

Is there anything we need to do to ensure none of these could result in that?
If we say let's merge these and when someone shows a problem, we will deal with it, it may not be the best plan.  
Do you have any suggestions?

olorinmaia:
> @olorinmaia Thanks,
> 
> Another challenge we sometimes have with translations is when a translated value becomes much larger the original English resulting in an overlap with another filed that results in a problem that the original designer of the layout never considered.
> 
> Is there anything we need to do to ensure none of these could result in that? If we say let's merge these and when someone shows a problem, we will deal with it, it may not be the best plan. Do you have any suggestions?

Yes it can happen, but rarely. Crowdin got nice tools for translaters to give feedback about strings. If a translation is added I always check afterwards it looks good in app, if not I adjust it in crowdin so it will get updated in next release.

When I add strings like this. I test by adding norwegian translations manually and with longer text just to see that something don't break if text is to long. In this case no issues was seen.

Example of discussion in Crowdin.

image

jamorham (approved):
Looks good thanks.

PR #3908, opened 2025-02-21
Add strings to make Reminders and Calibration graph fully translated

olorinmaia:
* Tested OK on my test-rig with all features in Reminders
* Tested with adding Norwegian translations in longer text length to ensure nothing breaks
* As far as I can see no logging are impacted by new strings.

Some screenshots:

| Before | After |
|--------|--------|
| image | image|
| image| image |
| image| image |
| image| image|
| image| image |
| image| image|

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
@@ -88,7 +88,7 @@ public class CalibrationGraph extends ActivityWithMenu {
             DecimalFormat df = new DecimalFormat("#");
             df.setMaximumFractionDigits(2);
             df.setMinimumFractionDigits(2);
-            String Header = "slope = " + df.format(calibration.slope) + " intercept = " + df.format(calibration.intercept);
+            String Header = getString(R.string.calibration_slope_intercept, df.format(calibration.slope), df.format(calibration.intercept));
             GraphHeader.setText(Header);
 
             //red line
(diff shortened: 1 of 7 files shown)


jamorham:
Thanks for a detailed PR. Can you confirm that this is simply a string extraction without any other changes to functionality or formatting?

olorinmaia:
> Thanks for a detailed PR. Can you confirm that this is simply a string extraction without any other changes to functionality or formatting?

Hi @jamorham, your welcome. I've ran xDrip with this PR since publishing it and I haven't discovered anything not working as before. As far as I can see when I take a final look at the code changes it is only string extraction.

Navid200:
In my opinion, items that are only visible in engineering mode should never be translated.  
We will have a very hard time working with a developer who cannot speak in English.

olorinmaia:
@Navid200 Not sure if I see an issue in translating items behind Engineering mode. Users will regardless choose the language of their preference. Is anything here hidden behind engineering mode? At least I wasn't aware of that. When I disable engineering mode on my test rig Calibration graph is still there (I guess that is what you mean)

I see else in the app some engineering mode related items are translated, some not. If @jamorham also think items behind engineering mode shouldn't be translated, I can take it out of this PR.

Navid200:
There is a benefit in translating items: Those who don't speak English may benefit.
There is a price to pay.  And I know that we disagree about this part.  

Because of the latter, in my opinion, we should not translate anything that is behind engineering mode.
If you disable engineering mode, you will not be able to modify the slope and intercept of your calibration graph.

If it was my call, we would limit translations to what was absolutely necessary.  Definitely not every single word.

Navid200:
I think you want to say that the items you are translating are visible whether engineering mode is enabled or not.  it is only the 3-dot menu that goes way if you disable engineering mode.  The slope and intercept will be shown regardless.

If that's true, then I am sorry I made a mistake.

Navid200:
I don't see a problem.

But, we are at the time we usually identify the next stable release.  It is possible that we are waiting for the translators now to translate everything that is already there so that we identify the next stable release before adding any more strings.
That may be why this is still waiting and will be added soon.

jamorham (approved):

PR #3921, opened 2025-03-02
Minor cleanup of code to simplify reading and

ahaarrestad:
maintainability. No changes in any method signatures of code behavior. This change is
part of coding and testing in xDrip


Navid200:
Wow!
It has been a long time.  It is nice to see you again.
Thanks for this.

I' wonder if I should add this to my active setup so that I test it.

ahaarrestad:
Long time indeed :-) And I see that the gradle.properties file should not have been added to the commit. That is just for building on my machine.

ahaarrestad:
I have also cleaned up (for next commit) the mix of accounts. sorry about that

jamorham (approved):
Looks good, thanks

PR #3922, opened 2025-03-02
Extracting general Retrofit factory

ahaarrestad:
Extracting general Retrofit factory to be utilized in other services. This was done as pr comment in file
and enables common code for new versions of Nightscout follow


Navid200:
A major issue we have with Nightscout is with respect to the uploader.  It does not maintain a unique queue for each URL.  This is a problem when the user chooses to upload to more than one URL and one of them fails.  The queue is cleared as soon as one of the sites succeeds.  The one that failed will never get the content that failed once.

This was discussed here: https://github.com/NightscoutFoundation/xDrip/pull/3630#issuecomment-2585750173

ahaarrestad:
If I understand your comment, the problems is that we might need to reset the connection if we xDrip is missing readings?

Navid200:
@ahaarrestad I mentioned something that I thought you may be able to work on to help us fix.  It has nothing to do with this PR.

A user can choose to enter two URLs to upload readings from xDrip to two different Nightscout sites.
xDrip does not have a dedicated queue for each site.  If one of the sites is down or if for any other reason, xDrip cannot upload a reading to one of the two sites but can upload the same reading to the other site, the single queue will be cleared.  The result is that when xDrip resumes uploading to both sites, it will only upload the rest of the readings.  The single reading that failed to upload to one site will never be uploaded to it.

The only workaround for this is for the user to backfill in a caser like this.
The problem is that the user may never know that this problem has ocured.
That's why I am adding a notification: https://github.com/NightscoutFoundation/xDrip/pull/3875  
That will not fix the problem.  But, will at least inform the user that he/she should backfill.

We need to create a dedicated queue for each site.  And we may limit the number of sites to only two to make the solution possible.

ahaarrestad:
Got it. Will keep that in mind :-)

Currently playing around with the V3 of the nightscout REST API to see if we can manage to follow in more "real time" from nightscout, so in some way it ties in with downstream work of that.

jamorham:
Just to mention we already have the nightscout sdk as a dependency if that simplifies anything in the future..

jamorham:
Please see the build/test failures

```
NightscoutFollowTest.java:10: error: cannot find symbol
import static com.eveningoutpost.dexdrip.cgm.nsfollow.NightscoutFollow.UNRELIABLE_INTEGER_FACTORY;
^
  symbol:   static UNRELIABLE_INTEGER_FACTORY
  location: class NightscoutFollow
```

jamorham:
@ahaarrestad would you be able to resolve the test failures? thanks

ahaarrestad:
I will look i to it shortly


tir. 15. apr. 2025, 13:05 skrev JamOrHam ***@***.***>:

jamorham:
` error: cannot find symbol
public class ShareFollowCallback<T> extends BaseCallback<T> {
                                            ^`

BaseCallback appears missing.

ahaarrestad:
I will fix this as soon as possible. sorry for the delay

jamorham (approved):
Looks good, thanks for the quick update

PR #3925, opened 2025-03-05
Missed Readings Alert Sound

Navid200:


Navid200:
The behavior when you tap on Choose file is identical to what you would see when you do the same for the Persistent High alert or Forecast Low alert.

Screenshot_20250305-150953

jamorham:
Can you just confirm that this is safe for backwards compatibility when users upgrade etc? thanks

Navid200:
When user has already changed their other alert sound, and they update to this after, the update will set the missed reading alert sound to the other alert sound.

This PR is much more deserving of being reviewed now.  Not just because of it being much older.  But, also because it affects G7 users who are already in a disadvantage because of many xDrip limitations:
https://github.com/NightscoutFoundation/xDrip/pull/3872

Navid200:
@jamorham Years ago, you told me that we didn't change the behavior of xDrip in a way to break a working system.
This for me translated to something more conservative:
We will not allow an xDrip update to surprise a user who has been using xDrip before the update.  

Now, let's look at this PR and see what can possibly surprise the user.

1- The user has been using xDrip and have changed their Other Alerts sound.  If they just update xDrip to this PR, will their existing Missed Reading alert sound change just because of the update?

No, it will not.  xDrip will always use the existing Other Alerts sound as the default sound setting for the Missed reading alert after this update.  

<br/>  
  
2- The user has got used to the fact that changing their Other Alerts sound also changes their Missed Reading alert.
After they update to this PR, they don't realize this change and go ahead and change their Other Alerts sound.  Will their Missed Reading alert sound change as it did before?

No!
This could surprise the user because it is a change of behavior just caused by this update.

I was going to add a notification that would come up the very first time only when the user changes the Other Alerts sound informing them that their Missed Reading alert sound is not going to change and if that's what they want, they can change it on the Missed Reading alert page.

Instead of adding this notification, I decided to open this PR to seek your advice since you neve responded to this discussion: https://github.com/NightscoutFoundation/xDrip/discussions/3916 

If you want, I can add such a notification.  It will require added complexity to the code.  It will cost two additional preferences.
If you are OK with this as is, I will make an announcement when this is merged to inform users of this change.

jamorham (approved):
Looks good

PR #3926, opened 2025-03-08
YD instead of YTD

Navid200:
The abbreviation YTD is often mistaken for year to date when in fact it is meant to stand for yesterday.
This PR changes it to YD hoping that less will be confused.  
Shown next to TD for today, I hope that most will accept and understand YD standing for yesterday.  
  
---  
  
Edit:  **Outdated - See next post**  

| Before | After |  
| ------- | ------ |  
| Screenshot_20250307-194634 | Screenshot_20250307-195145 |

### DIFF
--- a/app/src/main/res/layout/activity_statistics.xml
+++ b/app/src/main/res/layout/activity_statistics.xml
@@ -24,26 +24,30 @@
             <Button
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
-                android:text="TD"
+                android:text="Td"
                 android:id="@+id/button_stats_today"
                 android:layout_marginTop="1dp"
                 android:layout_marginBottom="1dp"
                 android:paddingTop="1dp"
                 android:paddingBottom="1dp"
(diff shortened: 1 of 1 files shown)


Navid200:
| Before | After |  
| ------- | ------ |  
| Screenshot_20250307-194634 | Screenshot_20250307-220058 |

jamorham:
Should this string also be extracted for localization? I know its going to be a challenge for translators to get it right given the space restrictions..

Navid200:
In my opinion, for the same reason we do not translate units, neither should we translate abbreviations.

Navid200:
There was a suggestion to place Today, Yesterday etc. in front of the title at the top as the user chooses the different options.
That sounds good and I can work on it.  
But, the priority goes to items that are broken and already have open PRs that have not been reviewed or have been reviewed and the review has been addressed.

Navid200:
I should listen to myself:
Let's not change xDrip unless absolutely necessary.
I will open a PR when I get to setting up the title to represent the selected period.

PR #3934, opened 2025-03-12
CareLinkFollower - Use Cloud servers for standalone sensors

benceszasz:
**Issue**
The old servers are no longer providing new data as standalone sensors have been migrated from old servers to the cloud servers.
This migration affects every region (US and outside US).
Following standalone sensors is currently not possible in any region.

**Solution**
Using cloud servers for data retrieval in the case of standalone sensors.

**Testing**


jamorham (approved):
thanks for the update

PR #3941, opened 2025-03-19
Sensor: fix inconsistencies

jamorham:
This pull request is designed to resolve some inconsistencies with how the `Sensor` class works which are likely to contribute to other problems.

In simple terms it should not be possible to have more than one "active" sensor within the database and sensor sessions should not be able to overlap.

Prior to this change, if you have two sensor records which do not have stop times and you then stop the sensor, the last one is stopped but then the previous one is considered to be active which is not intended.

The current code allows for creating a new sensor session without stopping the previous one and doesn't correctly handle the situation where the latest sensor is stopped at which point we should not have any active sensor sessions.

The code changes also allow for new sensor starts which point the start time to the past to then correctly stop the previous sensor and prevent the start time of the new sensor overlapping the stop time of the previous sensor. 

In order to allow the maximum number of readings where physical sensors may actually overlap we can automatically make the stop time of the previous sensor clamp to the last recorded reading time so that we have opportunities for backfilling before the current time but after the physical end time of the previous sensor.

Much of this relates to legacy support where individual sensor sessions are important for calibration reasons but there are clear inconsistencies in how this class handles data vs the modelling of how we would expect it to work and this PR is designed to resolve those which should then potentially resolve other related issues.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/Calibration.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/Calibration.java
@@ -11,6 +11,7 @@ import com.activeandroid.ActiveAndroid;
 import com.activeandroid.Model;
 import com.activeandroid.annotation.Column;
 import com.activeandroid.annotation.Table;
+import com.activeandroid.query.Delete;
 import com.activeandroid.query.Select;
 import com.eveningoutpost.dexdrip.GcmActivity;
 import com.eveningoutpost.dexdrip.Home;
@@ -1359,6 +1360,12 @@ public class Calibration extends Model {
         String msg = "Deleted all calibrations for sensor";
         Log.ueh(TAG, msg);
         JoH.static_toast_long(msg);
+    }
+
+    public static void deleteAll() {
+        new Delete()
+                .from(Calibration.class)
+                .execute();
     }
 
 }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/Sensor.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/Sensor.java
@@ -1,10 +1,13 @@
 package com.eveningoutpost.dexdrip.models;
 
+import static com.eveningoutpost.dexdrip.models.JoH.tsl;
+
 import android.provider.BaseColumns;
 
 import com.activeandroid.Model;
 import com.activeandroid.annotation.Column;
 import com.activeandroid.annotation.Table;
+import com.activeandroid.query.Delete;
 import com.activeandroid.query.Select;
 import com.eveningoutpost.dexdrip.GcmActivity;
 import com.eveningoutpost.dexdrip.Home;
@@ -21,7 +24,11 @@ import org.json.JSONObject;
 
 import java.util.Date;
 import java.util.List;
+import java.util.Objects;
 import java.util.UUID;
+
+import lombok.Setter;
+import lombok.val;
 
 /**
  * Created by Emma Black on 10/29/14.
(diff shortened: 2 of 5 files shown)


Navid200:
Thanks a lot.
Please let me do some testing.  I will report all here.

jamorham:
I've fixed up the unit tests in the latest commit - where data was not properly being cleaned up and where I had used overlapping sensor sessions in the unit tests this is no longer allowed so those were failing.

I haven't tested this change against the original backfill problem you reported and where you have made a PR to fix that but I think that the issues could be very strongly related and so it is possible that this would fix that as well so I'm keen to hear the results of your testing but overall this is designed to prevent problems that overlapping or inconsistent sensor class results could occur. 

I had seen this manifest in the past where when you stop a sensor from the menu it apparently doesn't stop.  I believe the issues resolved in this PR are the root cause of that but there are other problems it may resolve too.

I'm very keen to hear if you think there are any operational use cases that my changes here could break. I can't think of any but please let me know if you do.

Navid200 (approved):
I didn't know it was possible to have more than one internal session in xDrip.
I cannot think of a case when we would need more than one xDrip internal session.

I see references to starting a sensor 3 hours in the past in the SensorTest file.  I am not sure why 3 hours.

After a G7 sensor expires, we have a mechanism (after this PR: https://github.com/NightscoutFoundation/xDrip/pull/3526) that equates expired sensor to stopped sensor.  So, the internal session is stopped.
xDrip has an auto start that kicks in when xDrip detects a native session in progress on the transmitter when we switch TX ID to the next G7.  That mechanism started the sensor 3 hours in the past.  After G7 came out, I added a conditional to make that 24 hours in the past.
This blind auto start collides with the previous session that just stopped and a backfill overlap results.

I don't know what will happen after this is merged in that scenario.
I can test that on Saturday when  my current sensor will expire.

Navid200:
I did test by setting up Nightscout follower and then switched to an existing session in progress.  There was no backfill overlap neither with this nor with our latest Nightly.  
It's not easy if not impossible to test what G7 users experience when they switch from an existing sensor to one that they started 12 hours ago.
I can test on Saturday.

Navid200:
I'm sorry I just had an idea for another test.  The results will be here by your tomorrow.

Navid200:
I repeated the test but this time switched to a G7 session in progress.  
The latest Nightly backfilled the past 24 hours.
This PR only backfilled starting from the sensor stop.  So, it is working.
Does this mean the auto start mechanism is now redundant.

Anyway, this seems to solve the problem.

Navid200:
A G7 user could decide to switch to their next sensor before their current sensor expires.
If they do that right now, the internal session keeps going.  I believe after this PR, when they switch, the internal session will stop and a new one will start.
Effectively, I see no problem with this.

Navid200:
After a G7 expires, the sensor is stopped.  
When the transmitter ID is changed right away to that of the new G7, there will be no active session (because it was just stopped) and a new sensor should be started.  But, this conditional will never trigger for a G7 because the sensor was recently stopped.  Don't we need to change this?
Screenshot 2025-03-21 084312 

My test did not catch this because I think the similar code in ObG5StateMachine does trigger for G7 after this PR: https://github.com/NightscoutFoundation/xDrip/pull/3557 
But, does it always?  

I never understood why the same code was needed in two different spots and I cannot say that it is needed.  But, I just know that it is in two places and right now, only one triggers for G7.  

I will test this tomorrow.

Navid200:
I let my G7 expire using this PR.  Then, switched the transmitter ID to that of my new G7 that had been running for 12 hours.  
It connected and only backfilled starting from after the previous sensor expired.  So, there was no backfill overlap.
So, this is is good.  Please go ahead and merge it.

Thanks

jamorham:
Thanks for all the testing, glad to hear that it seems to be doing what it is supposed to.

Navid200:
I have a G7 expiring tomorrow.  It is my plan to run a different test with it.  Shortly before it expires, I am going to stop collection in xDrip, peel it off and move it away, and switch to the new sensor.  
I hope to see proper operation with no backfill overlap and no other malfunction.  I will report the results here.

I have already tested letting it expire and switched after and confirmed proper operation.  This other condition should also be tested and verified, which would I will tomorrow.

Navid200:
No backfill overlap.  So, it looks like we are OK.

Thanks

PR #3961, opened 2025-04-06
Making buttons in system status match parent to be more symmetrical

WojSkwie:
Hello everyone :)

I have noticed a tiny UI misalignment that occurs at least when using polish strings.
The system status activity has buttons on different heights because one of the texts is longer that usual.
image

That could be potentially fixed with replacing wrap_content with match_parent on those two buttons. From my perspective this is rather safe change that should not break the view for other languages. Please correct me if I am wrong.

Screen after change
image

### DIFF
--- a/app/src/main/res/layout/activity_system_status.xml
+++ b/app/src/main/res/layout/activity_system_status.xml
@@ -231,7 +231,7 @@
                         android:id="@+id/restart_collection_service"
                         style="?android:attr/buttonStyleSmall"
                         android:layout_width="wrap_content"
-                        android:layout_height="wrap_content"
+                        android:layout_height="match_parent"
                         android:text="@string/restart_collector" />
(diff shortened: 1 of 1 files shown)


Navid200:
Thanks a lot.
I will test this in a day.

Navid200:
I just tested this.  For English, it looks identical before and after.
For Nederlands:  
  
|  Before | After |  
| ------- | ------ |  
| Screenshot_20250408-083453 | Screenshot_20250408-083601 |  
  
It looks good.
Thanks for the submission.

WojSkwie:
Are You approving it then?

Navid200:
@WojSkwie I wonder how long this problem had existed and not one did anything about it.  But, you took the initiative and fixed it.
Thank you so much for that.

PR #3960, opened 2025-04-06
ican companion additional package

Cyxapuk-git:
Added additional package ID for ican health app which founded in russian GooglePlay

PR #3962, opened 2025-04-08
Add sensor iCan Health Asia version

thiagomsoares:
Add support for iCan Health Asia app (com.sinocare.ican.health.ce)

This commit adds the package name for iCan Health Asia app to both coOptedPackages  and coOptedPackagesAll lists. This enables xDrip to receive and process glucose  values from iCan Health app notifications, allowing users to use this app as a  Companion App data source.

The standard Sinocare CGM CE app was already supported, but the iCan Health app  uses a different package name that needed to be explicitly added for compatibility.


Navid200:
Is there a reason for this PR considering this other existing one?
https://github.com/NightscoutFoundation/xDrip/pull/3960

Navid200:
Thanks for your contribution.
This feature has now been integrated in xDrip.

Thanks

PR #3978, opened 2025-04-20
Retrieve LibreLinkUp data every two minutes

tk154:
This commit adjusts the `WebFollowerService`'s sampling period so that the app retrieves glucose data every two minutes from `LibreLinkUp`. Currently, the sampling period is set to five minutes.
For that change to work, I also needed to decrease the `WebFollower`'s rate limit, where I chose a value of 90 seconds.
First, I wanted to retrieve the data every minute but my `Libre 3` app seems to retrieve data from my `Libre 3 Plus` test sensor every two minutes at most although it's advertised otherwise.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/webfollow/WebFollowService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/webfollow/WebFollowService.java
@@ -119,7 +119,7 @@ public class WebFollowService extends ForegroundService {
                     }
                 }
                 if (context != null) {
-                    if (JoH.pratelimit("last-web-follow-poll", 180)) {
+                    if (JoH.pratelimit("last-web-follow-poll", 90)) {
                         Inevitable.task("WebFollow-Work", 200, () -> {
                             try {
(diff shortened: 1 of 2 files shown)


Navid200:
Thanks

Have you tested this?

The problem is that a lot of what we have has been written with the sample period as a constant of 5 minutes rather than a variable.
So, there is no easy way to verify if everything will work after such a change you are suggesting.

Have you verified if the forecast low alert works as expected?
How about the persistent high alert?
How about other alerts?
Does the trend arrow look correct?

SeIex:
>... First, I wanted to retrieve the data every minute but my `Libre 3` app seems to retrieve data from my `Libre 3 Plus` test sensor every two minutes at most although it's advertised otherwise.

Hi,

If this would be done, it should work with one minute readings.
The Libre 3/+ is usually delivering one reading every minute.
Usable in xDrip with GlucoDataHandler, getting data also with LibreLinkUp connection.

Maybe your sensor is not OK.
I think Ive had one Libre 3 with way slower readings some time ago.

You could check a LibreView .csv export for the readings, or GlucoDataHandler with Patched Libre source.

PR #3983, opened 2025-04-28
SheverdyaevAV_test_task

underbreathing:
Добавил круговой прогресс бар.
Время обновляется при получении нового значения глюкозы. 

*Увидеть работу прогресс бара без подключения к устройству можно на "dev" сборке.


Navid200:
Can you provide more details?
Can you include screenshots?  
What happens if the cycle is 1 or 3 minutes and not 5?
What happens if you walk away from the phone?

Navid200:
xDrip as is shows how long ago the glucose was measured.  For example, mine shows 3 minutes ago.  I know that I get a new reading once every 5 minutes.  So, I know that I will get a new reading in 2 more minutes.

Can you describe the problem with xDrip as is that this PR is addressing?

PR #3984, opened 2025-04-30
Added circular Progressbar

Pikosha2101:
Added a more informative and visual view of the time display since the last data reception. The principle is as follows: 0-5 - green (ok), 5-10 - yellow (almost ok), 10+ - red (alarm).
if approved, it will be possible to remove "x minutes ago" which is located to the left of the bar. 
The bar is shown only when the application receives data. Otherwise, the bar is hidden.
On my test device everything works fine. If you have any questions - I'm always happy to help

photo_2025-04-30_04-39-00 (2)
photo_2025-04-30_04-39-00 (3)
photo_2025-04-30_04-39-00
photo_2025-04-30_05-42-18


Navid200:
Have you seen this?
https://github.com/NightscoutFoundation/xDrip/issues/2680  

For a device that receives one reading a minute, having readings once every 5 minutes is not normal.
There are other devices coming to the market that provide one reading every 2 or 3 minutes.

We already have a missed reading alert.  You can choose how long you are OK having no readings and after how long of no readings you want to receive a missed reading alert.

Can you explain what exactly the problem is with xDrip as is that this PR is addressing?

PR #3990, opened 2025-05-07
Merge 悬浮窗

allisprelude:


jamorham:
Please describe the purpose of this PR?

iceyu07:
Sorry, I made a mistake — this shouldn't have been merged into the official main branch.

But let me explain:
My child uses an Android smartwatch that lights up the screen when the wrist is raised.
I added a floating window for xDrip, so every time they raise their wrist, they can see their blood glucose level immediately without needing to open xDrip.

allisprelude:
> @allisprelude Would you please explain what this PR does and why we need it?

Sorry, I made a mistake — this shouldn't have been merged into the official main branch.

But let me explain:
My child uses an Android smartwatch that lights up the screen when the wrist is raised.
I added a floating window for xDrip, so every time they raise their wrist, they can see their blood glucose level immediately without needing to open xDrip.

PR #4010, opened 2025-05-23
Use new blootooth API for SDK version > 32 

NiklasMehner:
This in turn enables filtering bluetooth devices that are not glucose meters.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/BluetoothGlucoseMeter.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/BluetoothGlucoseMeter.java
@@ -1002,12 +1002,14 @@ public class BluetoothGlucoseMeter extends Service {
                     scanLeDevice(false); // stop scanning
                     connect(btDevice.getAddress());
 
-                    if ((lastScannedDeviceAddress.equals(btDevice.getAddress())) && (!JoH.ratelimit("bt-scan-repeated-address", 2))) {
-                        if (d)
-                            Log.d(TAG, "Ignoring repeated address: " + btDevice.getAddress());
-                    } else {
-                        lastScannedDeviceAddress = btDevice.getAddress();
-                        sendDeviceUpdate(btDevice);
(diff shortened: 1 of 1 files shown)


jamorham:
So the force old was likely put in for the reason that the old api was much more reliable, however that may be something that was circa android 6. Since android 12 the bluetooth stack is a lot better. So the question here is how many different android versions and handsets has this change been tested on?

NiklasMehner:
This is only tested on my phone (android 13). I don't really have the option to test many devices.

What is you typical approach for something like this? Release a beta? Add a configuration toggle?

Navid200:
I have my old phones and use them for testing.
I have an Android 8 phone, 2 Android 9 phones, one Android 11 phone and one Android 16.

Can you get your hands on an Android 7 phone?

NiklasMehner:
@Navid200 No, sorry, I normally don't do android development, so I am not keeping old phones.

I have updated the PR to only use the new API for SDK > 32. That value can be decreased if tests with older versions succeed and it is now only one location that has to be changed.

NiklasMehner:
But the PR doesn't change the conditionals for the old versions.

The old condition was:

`(Build.VERSION.SDK_INT < 21) || (force_old)`

with "force_old == true" this was always true. So this condition did not actually matter. The new approach is to use the new API for version > 32. So force_old is "Build.VERSION.SDK_INT > 32".

But then the above condition becomes:

`(Build.VERSION.SDK_INT < 21) || Build.VERSION.SDK_INT< 33`

and since <33 is implied by <21 the first prat can be dropped without changing anthing.

Navid200:
Some time in the future, this new conditioal will become redundant.  But, you will not come back then to remove it from the code.  It will become my responsibility.  This is not fair.
That was my point.

parapenT1sta:
@ahaarrestad Is this needed after Android 7 retirement in xDrip?

https://github.com/NightscoutFoundation/xDrip/commit/10f57b82badf907430d87e5be4b2ad6f40629d67

PR #4014, opened 2025-05-25
Override Silent Mode to reduce dependence on notification channels

Navid200:
#### **Why do we need this?**  
Currently, if you need to have different override silent mode settings for Persistent High, Forecasted Low, Missed Reading and Other alerts, your only option is notification channels.  
This PR offers the user the option with no need to the notification channels.  
<br/>  
  
---  
  
#### **New Settings**  
**Functional**  
The only new functional setting is "Override Silent Mode" on the following three alerts at the bottom of each page.  
  
Screenshot_20250525-153333 Screenshot_20250525-153404 Screenshot_20250525-153246  
<br/>  

**Hidden**  
There is a new (boolean) setting that the user will never see, `has_been_explicitly_set_persistent_high_alert_override_silent`.
We check the value of this setting the first time xDrip runs.  If the value is false, it means the new settings have not been initialized.  In that case, we copy the value of Override Silent Mode from Other Alerts and set the Persistent High and Forecasted Low and Missed Reading alerts to have the same value.  
This way if someone has been using xDrip and got used to the fact that the Override Silent mode setting of Other Alerts dictates the behavior for Persistent High and Forecasted Low and Missed Reading, the behavior will remain unchanged after the update.

However, after the update, if the user changes the Override Silent Mode setting for Other Alerts, the value will not change any longer for the other three alerts.
We will make an announcement about this as soon as this is released to inform the users about this change. 
<br/>  
  
---  
  
#### **New Log**  
With our latest Nightly, if any of the three alerts, Persistent High, Forecasted Low or Missed Reading, triggers with Override Silent Mode disabled, the log only shows that the alert triggered.  The user waking up in the morning and looking at the log will have no idea if the alert made a sound or not wondering why they never woke up.

A new log is now added only if Override Silent Mode is not enabled.  This will inform the user that the alert never made a sound.  
The three new logs corresponding to the three alerts are shown below.  
  
  
  
<br/>  
  
---  
  
#### **Tests**  
  
1- Installed our latest Nightly on an Android 9 phone
Enabled Override Silent Mode for Other alerts.
Then, installed this version over it.  The new Override Silent Mode setting was automatically enabled for the Persistent High and Forecasted Low and Missed Readings alerts.  
  
2- Installed our latest Nightly on an Android 11 phone.
Left the Override Silent Mode for Other alerts at the default setting (disabled).
Then, installed this version over it.  The new Override Silent Mode setting was disabled for the Persistent High and Forecasted Low and Missed Reading alerts.  
  
3- Installed this version on an Android 9 phone.
Set up fake data source.  Enabled Override Silent Mode for Other Alerts.  Disabled Override Silent Mode for Persistent High and Forecasted Low alerts.

The Persistent High and Forecasted Low alerts did not make a noise.   But, they created a log showing that the alert makes no noise in silent mode.  
  
4- Set up this xDrip as a Nightscout follower.  Enabled Override Silent mode for Other Alerts.  Disabled Override Silent Mode for Missed Reading Alert.  Disabled WiFi and Cellular.  Missed reading alert made no sound.  But, created a log showing that the alert makes no noise in silent mode.
  
5- I am now running this as my active everyday xDrip on Android 16.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/IdempotentMigrations.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/IdempotentMigrations.java
@@ -180,9 +180,9 @@ public class IdempotentMigrations {
     // Set new settings such that a version update does not cause a surprise
     private static void inheritPrefSettingsAfterUpdate() {
         if (!Pref.getBooleanDefaultFalse("has_been_explicitly_set_persistent_high_alert_override_silent")) { // If override silent mode has never been explicitly set for the Persistent High alert
-            Pref.setBoolean("persistent_high_alert_override_silent", Pref.getBoolean("other_alerts_override_silent", false)); // Inherit Persistent High override silent mode from other alerts
-            Pref.setBoolean("bg_predict_alert_override_silent", Pref.getBoolean("other_alerts_override_silent", false)); // Inherit Forecasted Low override silent mode from other alerts
-            Pref.setBoolean("bg_missed_alerts_override_silent", Pref.getBoolean("other_alerts_override_silent", false)); // Inherit Missed Reading override silent mode from other alerts
+            Pref.setBoolean("persistent_high_alert_override_silent", Pref.getBooleanDefaultFalse("other_alerts_override_silent")); // Inherit Persistent High override silent mode from other alerts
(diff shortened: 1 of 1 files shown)


Navid200:
Assuming this is merged, I wish to add one more control, vibration.

The user can control override silent mode after this PR and vibration after the next PR.  Then, they will have no need to use notification channels.  Notification channels have pros and cons.  I wish to offer the option not to rely on notification channels.

jamorham (approved):

PR #4015, opened 2025-05-26
Dex collector Last Connected - Color

Navid200:
This will let me see if the "Last Connected" value is less than 5 minutes or not without having to understand the translated words for minutes or seconds when helping xDrip users.  
Screenshot_20250526-091650  
    
If this PR is merged, there will be no need for the following any longer:
https://github.com/NightscoutFoundation/xDrip/pull/3817

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
@@ -2179,7 +2179,7 @@ public class Ob1G5CollectionService extends G5BaseService {
 
         if (static_last_connected > 0) {
             long since = msSince(static_last_connected);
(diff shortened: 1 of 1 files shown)


Navid200:
The following is an example of what a user may add to a post requesting for help in facebook.
500265123_10229130337431934_1159663331901557176_n  

Looking at that image, I have no idea if last connected is 1 minute ago or one hour ago.  I will have to ask the user to clarify.  This takes time that I prefer not to have to spend.
This PR removes the question.  The color, after this PR, will tell me what I need to know regardless of what language the text is in.  Then, there will be no need to ask the user to clarify.

jamorham:
Looks like a good improvement but can you use the constants like `MINUTE_IN_MS` with a multiplier to make it easier to read.

Navid200:
I am now using a constant for the time value as suggested.

jamorham (approved):

Navid200:
Thanks so much.

PR #4021, opened 2025-05-29
Home screen layout fix for very small phones

Navid200:
Please see the issue here: https://github.com/NightscoutFoundation/xDrip/discussions/4018 

A few pixels are overlapped and not sown at the bottom of the glucose value.
  
The problem only happens for devices with a very small number of pixels in their display.  Since we use absolute number of pixels for gaps, these gaps become a large percentage of the screen when the total number of pixels on the screen is very small.


Navid200:
Tests performed on Android Studio Virtual devices:  
  
| Device | Res |Before | After |  
| ------- | ---- | ------- | ------ |  
| 2.7" QVGA | 240 x 320 | 240_320_bad | 240_320_good |  
| 3.2" QVGA | 320 x 480 | 320_480_bad | 320_480_good |  
| 3.3" WQVGA | 240 x 400 | 240_400_bad | 240_400_good |  
| 3.4" WQVGA | 240 x 432 | 240_432_bad | 240_432_good |  
| 3.7" WVGA | 480 x 800 | 480_800_bad | 480_800_good |  
| 4.7" WXGA | 720 x 1280 | 720_1280_bad |  |

Navid200:
Thanks for the review.
I had thought that it was the Forecast Low frame that was overlapping the BG reading frame.  But, it wasn't.  So, I have set the Forecast Low frame top padding back to -2.

I believe it is the Glucose treatment frame before confirmation that overlaps.  
I repeated the tests with the new fix (only increasing the BG reading bottom padding from 1 to 4) and they still look good.  I have updated the screenshots shown above to reflect the update.

The following shows the fix after entering a glucose treatment.  You can see that it is close but not overlapping.
1b

Navid200:
Screenshot 2025-05-31 072754

Navid200:
Screenshot 2025-05-31 075246

Navid200:
Screenshot 2025-05-31 084752  
  
Screenshot 2025-05-31 084712

Navid200:
Screenshot 2025-05-31 112751

Navid200:
Thanks

PR #4025, opened 2025-05-30
Removing redundant build version conditionals

Navid200:
I wanted to see how many conditionals exit based on the build release that is unconditionally true or false.
I was surprised to find so many.
I have removed such conditionals.  Set the statements conditioned on true values to always run.  Deleted the statements conditioned on false values.

How could a change this extensive ever be adequately tested?

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/ui/graphic/TrendArrowFactory.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/ui/graphic/TrendArrowFactory.java
@@ -26,11 +26,9 @@ public class TrendArrowFactory {
         final String arrow_type = Pref.getString(PREF_TREND_ARROW_TYPE, "");
 
         // yucky workaround for broken LG lack of gradient fill.
-        if (Build.VERSION.SDK_INT == 23 && Build.MANUFACTURER.startsWith("LG")) {
-            if (arrow_type.equals("JamTrendArrowImpl")) {
-                JoH.static_toast_long("LG Phones have problems with some arrows, choose another");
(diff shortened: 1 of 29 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/ui/graphic/TrendArrowFactory.java
```diff
@@ -26,11 +26,9 @@ public static ITrendArrow create(ImageView imageView) {
         final String arrow_type = Pref.getString(PREF_TREND_ARROW_TYPE, "");
 
         // yucky workaround for broken LG lack of gradient fill.
-        if (Build.VERSION.SDK_INT == 23 && Build.MANUFACTURER.startsWith("LG")) {
```
asavageiv, line 29: `Build.MANUFACTURER.startsWith("LG")` is always `true`?

jamorham:
Typically I would just consider removing these whenever the associated bit of code is being worked on. It does mean though that the code base then cannot be compiled for a lower android sdk version and function correctly. 

In general I think that makes sense because if someone wants to be using, for example, android 4 then really they just shouldn't bother. I think the range of handsets we support with the current minsdk level is already very good so I'm okay with removing these.

But a PR like this just involves a lot of meticulous reading to ensure nothing gets broken for quite a small gain and at the risk of introducing potentially serious bugs if anything is got wrong which is why it makes more sense to adjust these as we are working on particular sections of code for other reasons because then the testing should be happening already.

For example one that is in this PR I have already eliminated in my working tree where I am adjusting some of how the G7 sensor start is handled.

PR #4029, opened 2025-05-31
Y axis range menu orientation fix

Navid200:
This is not anything that is broken.  But, I think it looks better if the ranges are shown with the maxes at the top.  
This is actually something I added a few years ago.  I'm not sure why it didn't look wrong then.  It sure looks wrong now.  

| Before | After |  
| ------- | ------ |  
| Screenshot_20250531-154811 | Screenshot_20250531-155111 |  
Screenshot_20250531-154835 | Screenshot_20250531-155130 |

### DIFF
--- a/app/src/main/res/values/arrays.xml
+++ b/app/src/main/res/values/arrays.xml
@@ -501,30 +501,30 @@
     </string-array>
 
     <string-array name="ymin_entries">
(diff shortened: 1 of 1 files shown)


jamorham (approved):

Navid200:
Thanks

PR #4030, opened 2025-06-01
Notification records (Logs)

Navid200:
There are many notifications that create a corresponding log.  
However, we also have notifications that do not.
The ones that are part of an interactive sequence, do not need a log.  As an example, if you are going through hard resetting a device and get a notification, it is unlikely that you would later need to see it again.  Even so, you can see it again by going through the same process again.

There are notifications that are not part of an interactive sequence and do not create a corresponding log.
After a user swipes away such a notification, they will have no way of accessing any information about it.  
It happens when a user swipes away an incorrect notification by mistake.
There are also times a user swipes away a notification after seeing what it is.  But, later needs info about it.   For example, they may need that in order to show it to us in a forum.

This PR adds corresponding logs to notifications that are not part of an interactive sequence and do not create a corresponding log already. 

I believe it is a very good practice to have a log for every notification as notifications are swiped away; but, logs remain for a few days.


Navid200:
@jamorham I need to redo this to fix the tags and also to avoid translating the logs.
So, this is not ready for a review.
But, can you tell me what you think of the idea?

jamorham (approved):

Navid200:
Thanks

PR #4032, opened 2025-06-01
G7 Battery level and Silent Notification

Navid200:
We should not wake up the user to notify them that their transmitter battery voltage is below the recommended level.
https://github.com/NightscoutFoundation/xDrip/discussions/3893  

That's why I have changed the notification to silent.  There will still be a notification.  It just will not make a sound or vibrate.
<br/>  
<br/>  
  

There are G7 devices that show lower voltages.  Therefore, I have slightly lowered the warning level for G7. 
Screenshot_20250531-203801


Navid200:
More of these are coming up in forums:  
506168037_10236559365993620_6964291016974673848_n

Navid200:
More of these keep coming up:
524293744_10235468697911784_3873076195464118801_n

Navid200:
@jamorham Please correct me if I am wrong.
Once every 12 hours, xDrip reads the battery level.
If the battery level goes below the warning level, a notification will be created to inform the user that the transmitter battery has dropped below the acceptable limit and may stop working.
When this notification is created, if the phone is not being charged, the notification will also make a sound.  

The problem I see with this arrangement is that not everyone charged their battery every night.  I have had phones that I have charged every other night.  That means, some nights, I have slept with my phone not being charged.
Why would we make a sound to inform the user that the battery may be low?

What makes this worse with G7 is that we don't even allow a G7 user to adjust the warning level, which we do for G6.  

Would you please review this PR and tell me what you don't like about it?

PR #4036, opened 2025-06-03
Tidepool new authorization only

Navid200:
I have been uploading to Tidepool for over a year now with this setting enabled.   I have had no problems.
This PR enables the setting permanently, resolves any conditionals related to it, removes the setting from the UI, and ensures any regression will enable the setting.

I am currently running this PR.


jamorham:
Yes this makes sense

Navid200:
Thanks

PR #4041, opened 2025-06-07
Temporary Sensor Issue

Navid200:
This PR changes the text and adds the word "Temporary".
This is to convey that the issue may not be final.  

Unfortunately, this is not something I can test.  I am currently running this.  But, over the past year and half I have been using G7, I have experienced this error only 3 times.   It is highly unlikely, I will see the impact soon.  
And I cannot think of a way to intentionally cause this error.

If you can think of a way to test this, please let me know.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
@@ -2162,7 +2162,11 @@ public class Ob1G5CollectionService extends G5BaseService {
             l.add(new StatusItem("Scan Error", lastScanError, BAD));
         }
         if ((lastSensorStatus != null)) {
-            l.add(new StatusItem("Sensor Status", lastSensorStatus, lastSensorState != Ok ? NOTICE : NORMAL));
+            String sensorStatus = lastSensorStatus;
+            if (shortTxId() && lastSensorStatus == "Sensor Failed 7") {
+                sensorStatus = "Temporary Sensor Issue";
+            }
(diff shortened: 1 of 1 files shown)


jamorham:
Error 7 doesn't necessarily ever resolve. Its quite easy to test, often if you just pull off a G7 a few hours before the session would naturally end then it will go in to this state and stay that way instead of going to Stopped state when the session should expire. I think it likely means that the glucose measurement is completely out of range (likely too low) and it registers as a fault. I don't really know how this tallies with the reports we had that usable glucose was still available during this state.

Navid200:
> Error 7 doesn't necessarily ever resolve. 

I agree.  
This PR does not say that either.  It says that it MAY recover.  It has for me.  
I have got sensor errors from xDrip while I was hiking.  I just ignored it for a while and the readings came back with a gap left.  The gap represented a period during which G7 did not trust the validity of the sensor.

I don't want an xDrip user to see the current terminal error and peel it off when in reality it may recover.  
The fact is that Dexcom app also states that it may recover.  So, why shouldn't we?

Navid200:
@jamorham 
> Error 7 doesn't necessarily ever resolve. 

I agree.  
This PR does not say that either.  It says that it MAY recover.  It has for me.  
I have got sensor errors from xDrip while I was hiking.  I just ignored it for a while and the readings came back with a gap left.  The gap represented a period during which G7 did not trust the validity of the sensor.

I don't want an xDrip user to see the current terminal error and peel it off when in reality it may recover.  
The fact is that Dexcom app also states that it may recover.  So, why shouldn't we?

Issue #4044, opened 2025-06-10
G7 Sensor Age uploaded to Nightscout is wrong


Navid200:
![Image](https://github.com/user-attachments/assets/e25f528b-d872-45a4-b578-071616f5bac1) 

If you wait for your old sensor to expire, when you switch to the new sensor, the sensor age uploaded to Nightscout is very close to reality.
But, if you don't wait for your old sensor to expire before switching the transmitter ID to that of the new one, xDrip continues to upload the age of the old sensor.

xDrip reports the correct expected sensor expiry for G7 regardless.  So, this should be easy to fix by changing what xDrip updates to Nightscout for sensor age for a G7.

PR #4047, opened 2025-06-11
Session start time update for alt (corrects Nightscout SAGE) 

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/4044

If you don't wait for your current sensor to expire before switching the transmitter ID to the new one, your local session will never end and new one will never start.
The result is that xDrip will continue to upload the previous start time to Nightscout.
Nightscout uses that start time to calculate the age of the sensor:
Screenshot 2025-06-11 183503
  
In the above screenshot, you can see the sensor age reported by Nightscout (SAGE) to be close to 11 days.  But, that's really the age of my previous sensor that I peeled off yesterday before its expiry because I wanted to recreate the problem reported on facebook.  

I have also reported this in the following issue after getting the report on facebook:
https://github.com/NightscoutFoundation/xDrip/issues/4044

After this PR, the following is what Nightscout shows:
Screenshot 2025-06-11 184439
It now shows 1 day and 11 hours, which is the correct age of my current sensor.  

And the following shows the added treatment when the session really started (yesterday morning): 
share_5177174540425584126

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/Treatments.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/Treatments.java
@@ -18,6 +18,7 @@ import com.activeandroid.query.Select;
 import com.activeandroid.util.SQLiteUtils;
 import com.eveningoutpost.dexdrip.GcmActivity;
 import com.eveningoutpost.dexdrip.Home;
+import com.eveningoutpost.dexdrip.g5model.DexSessionKeeper;
 import com.eveningoutpost.dexdrip.models.UserError.Log;
 import com.eveningoutpost.dexdrip.R;
 import com.eveningoutpost.dexdrip.services.SyncService;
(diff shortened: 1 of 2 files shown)


Navid200:
I need to test this in 9 days when the next sensor start happens to ensure it will not cause a backfill overlap.

jamorham:
Can you explain exactly the logic you're using to determine if the start time should be updated?

We need to be careful of potential clock drift both forwards and backwards that we don't end up updating it too frequently and that there are not any other issues caused by any adjustments.

Navid200:
> Can you explain exactly the logic you're using to determine if the start time should be updated?  
  
I remember the fuzzer issue.  The magnitude of that drift was never even close to 5 minutes per 15 days.  The first time the start time is updated, the two will line up perfectly at the millisecond accuracy.  I keep comparing the two and only if the delta is not less than 5 minutes, I update again.    Even after the 15-day devices are released, we should never trigger this more than once in the lifetime of the device.

Navid200:
> We need to be careful of potential clock drift both forwards and backwards ...  

In the grand scheme, we would update even if someone keeps switching back and forth between the same two transmitters.  This is not something I am attempting to address.  I don't think anyone should be doing that.  
I am assuming that someone is running a device and inserts a new device and before waiting for the old device to expire, they switch.  In that case, after the switch, the local sensor will be older than the one on the new transmitter.  I am banking on that.  

After we first take the start time from the transmitter and update the local sensor with it, you are absolutely correct that there could be a drift in either direction.  I don't believe that drift could ever be greater than 5 minutes in either direction in 15 days.  
Therefore, we will never update the start time just due to the drift in either direction.

If you want to provide more margin, we can change the conditional from 5 minutes to 10 or 15 minutes.  But, I don't think that is necessary.

jamorham (approved):
Okay lets try this

PR #4042, opened 2025-06-09
Add strings to Edit Alerts + Statistics to make xDrip+ more translated

olorinmaia:
This mainly adds strings for translation for various error messages when editing BG level alerts.

* Tested OK on my test-rig with all features in Edit Glucose Level Alerts
* Tested with adding Norwegian translations in longer text length to ensure nothing breaks
* As far as I can see no logging are impacted by new strings.

@Navid200 @jamorham English strings are untouched. If there is any adjustments/improvements to be made to the English strings before this possibly is merged let me know.

Some screenshots to illustrate where change is made, and that it is working.
| Before | After |
|--------|--------|
|image| image |
| image | image| 
| image | image |
| image | image |
| image | image |
| image | image |
| image | image|
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

PR #4051, opened 2025-06-16
Ascending as the volume profile title

Navid200:
| Before | After |  
| ------- | ------ |  
| Screenshot_20250616-151630 | Screenshot_20250616-152324 |

### DIFF
--- a/app/src/main/res/values/arrays.xml
+++ b/app/src/main/res/values/arrays.xml
@@ -168,7 +168,7 @@
     <string-array name="BgAlertProfileEntries">
         <item>High</item>
         <item>Medium</item>
-        <item>Ascending volume</item>
+        <item>Ascending</item>
         <item>Vibrate only</item>
         <item>Silent</item>
     </string-array>


jamorham:
In some ways this is more consistent but also the ascending profile is the easiest to be misunderstood so the clearer we can be about its function the better?

jamorham:
There are related changes here so we have to be careful: https://github.com/NightscoutFoundation/xDrip/pull/4042

Navid200:
> There are related changes here so we have to be careful: #4042

I am aware.  That's why I wanted this sorted out before anyone translates anything because getting approval after translation becomes more difficult.

Navid200:
This guide complements the menus:
https://navid200.github.io/xDrip/docs/Ascending-volume-profile.html 
It helps the user understand what the option is and does.

I will have to review it and make edits if needed if this is merged.

Navid200:
This other page in the guide talks about volume profiles in general: 
https://navid200.github.io/xDrip/docs/Volume-profiles.html 

And it already calls in "Ascending".

tzachi-dar (approved):

PR #4056, opened 2025-06-18
Add more sensor type: 7f 0e 31

tzachi-dar:


Navid200:
https://github.com/NightscoutFoundation/xDrip/discussions/4028

Navid200:
So far, I know two different individuals have tried the test release and confirmed that it works.

PR #4063, opened 2025-06-21
Retire Parakeet Alerts

Navid200:
My limited understanding of Parakeet is that it was used with G4.
We have retired G5.  So, we should retire Parakeet as well.
This PR removes the alerts related to Parakeet.  

I may need help with the data sources though.  That's why this PR is only focused on these particular alerts.  
I'm not sure if most users understand what they are.  I believe it is best to just remove them.  

| Before | After |  
| ------- | ------ |  
| Screenshot_20250620-203151 | Screenshot_20250620-203936 |


Navid200:
Assuming this is approved, I believe the remaining alert under the other alerts heading is only active when xDrip is a Sync Follower.
I will open another PR to move that alert as a conditional UI that would only appear if xDrip data source is follower and it would appear right under hardware data source under settings.  The same way when you choose Nightscout follower, the follow URL appears there.

Navid200:
My hope is to make the alerts menus more intuitive and easier to navigate.

Navid200:
Regression tests failing.

jamorham (approved):
Yes parakeet is not used any more

Navid200:
Thanks

Issue #3705, opened 2024-10-08
ZXing does not work on Android 14


Navid200:
xDrip has been using ZXing for years.

It seems that it's not being updated any longer:
https://github.com/zxing/zxing  
![Screenshot 2024-10-08 161104](https://github.com/user-attachments/assets/47f79d7e-3fbd-4bab-b57c-b49a5f942402)

We have already had several reports of the scanner screen not fully opening.  
  
I tested the option of scanning from a saved image on Android 15 as well.  So, that option works.  But, the option to use the camera does not work on some phones sometimes.  

I hope we can find a solution.

Navid200:
I wonder if this may be a solution: https://github.com/mchehab/zbar

Navid200:
It turns out all you have to do is to rotate the phone to go into the landscape mode before attempting to scan.  Then, the scan window opens in the center of the screen:
https://github.com/NightscoutFoundation/xDrip/pull/4064

Navid200:
As of the 2025.10.04 release, the problem has been resolved.

PR #4064, opened 2025-06-21
QR scanner camera title/summary clarification

Navid200:
Our QR scanner runs assuming you are in landscape mode.
On some phones, this looks OK and the window is landscape centered on the phone screen as you rotate it.

However, on some phones, if you are holding it in portrait mode, the scanner preview is panned to the side.  Rotating the phone after starting the scanner will not center the preview screen.

There are suggested fixes for this.  But, it is much easier to just hold the phone in landscape mode before starting to scan.
The following video shows the problem on one phone (Pixel 6a) that suffers from it:

https://github.com/user-attachments/assets/da9f57e7-86ed-4e54-8800-81e82c737769

<br/>  

The following video shows the same phone in landscape mode before starting to scan.  

https://github.com/user-attachments/assets/f42d603c-54c9-48a4-9fc7-bc997302f941

<br/>  

I believe we may be better off just helping users know that all they need to do is to rotate the phone before scanning instead of trying to change the code to fix it, which would require lots of testing.  

| Before | After |  
| ------- | ------ |  
| Screenshot_20250621-164330 | Screenshot_20250621-165804 |  
  
  
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/3705

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -40,7 +40,7 @@
     <string name="auto_configure_title">Auto configure</string>
     <string name="prefs_auto_config_summary">Auto configure using a barcode.</string>
     <string name="auto_config_cam_title">Landscape Camera</string>
-    <string name="auto_config_cam_summary">Rotate the phone to landscape mode before tapping \'Landscape Camera\'.  Scan a QR code with your mobile (recommended)</string>
+    <string name="auto_config_cam_summary">Rotate the phone to landscape mode before tapping \'Landscape Camera\' to scan a QR code with your mobile (recommended)</string>
     <string name="auto_config_image_title">Image file</string>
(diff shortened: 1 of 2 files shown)


jamorham:
I think we can probably fix this with some updates to the library. I did look at this before but didn't get that far with it.

Navid200:
It works perfectly fine as is as long as the user knows to rotate the phone before tapping on camera.  I didn't know for years.
This is the purpose of this PR to help the users.
If we fix it, it will not contradict this PR.
When that day comes, I will undo this PR if you like.  But, until then, this will help users.

Navid200:
@jamorham I have added dedicated strings to represent the suggested title and summary while the existing title and summary are left alone.

When we eventually fix this, we can just change the UI to use the original unchanged strings.  The result will be that the translators will neither have to change the strings now nor after we fix the problem.

Navid200:
Thanks for fixing this.
I have one phone, Pixel 6a, that experienced this problem.  It does not anymore after this update.

Thank

PR #4066, opened 2025-06-23
Add insulin stock tracking feature

schmidtienet:
- Adds preferences to enable/disable insulin stock tracking and set initial stock.
- Introduces InsulinStockManager to handle stock logic using SharedPreferences.
- Modifies Treatment creation to decrease stock when insulin is administered.
- Updates settings UI to display current stock and manage initial stock value.
- Initializes stock with a default of 300 units, user-configurable.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/InsulinStockManager.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/InsulinStockManager.java
@@ -1,0 +1,89 @@
+package com.eveningoutpost.dexdrip.models;
+
+import android.content.Context;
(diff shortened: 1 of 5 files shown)


Navid200:
Thanks for your submission.

It would help if you start by explaining what is missing in xDrip as is that this PR addresses.

Adding new settings should be a last resort as xDrip already has too many settings.

If you have to add a setting, it is a good idea to show before and after screenshots of the user interface.

We should coordinate the menus such that the user would be able to find the settings intuitively.  This will be easier if settings are placed in the right menu/submenu.   But, the preferences should always be given to a solution that des not require a new setting.

PR #4074, opened 2025-07-06
No Follower calibration except xDrip Sync

Navid200:
Problem:
The main menu offers an "Initial Calibration" option for the Dex share follower:
https://github.com/NightscoutFoundation/xDrip/discussions/4071 

In fact, it also shows an option to Stop Sensor.  
But, neither of those functions can actually be performed.  You cannot calibrate a sensor through Dex share servers.  You also cannot stop a sensor running on the master transmitter through Dex share servers.

Suggested solution:
This PR removes the options to start sensor, or stop sensor or calibrate for either Nightscout follower or Dex share follower or CareLink follower or Companion app.  

Before - After  

Tests


Navid200:
@jamorham Thanks so much for taking the time guiding me.  I am grateful.  
I will be working on this and post another message here when it is ready.

PR #4078, opened 2025-07-08
Follower Chime menu update

Navid200:
The follower chime alert is only functional when the hardware data source is xDrip Sync follower.
For that reason, I believe it would be best to locate the setting under the hardware data source and only visible when the hardware data source is set to xDrip Sync follower.  That's all this PR does.  
There is no change to functionality.  There is only a change to the user interface.  

| Before | After |  
| ------- | ------ |  
| Screenshot_20250708-150633 | Screenshot_20250708-150447 |  
| Screenshot_20250708-150654 | Screenshot_20250708-150514 |


Navid200:
Thanks

PR #4082, opened 2025-07-14
Retire Go on without calibrations

Navid200:
This PR removes the setting "Go on without calibrations" from the Dex Debug Settings page.
I don't believe anyone would have a reason to disable the setting since we retired G5.
Please correct me if I'm wrong.  If not, there is no reason for the setting to remain.

| Before | After |  
| ------- | ------ |  
| Screenshot 2025-07-13 193244 | Screenshot 2025-07-13 192123 |


jamorham:
I think this is theoretically still possible to get from G6 and possibly G7 but its very advanced feature which probably is a bit risky as it allows use of glucose values that are marked as not usable. Its research stuff only really. I'm wondering whether we should, for example, instead lock its use to be dependent on engineering mode in the preferences so that user's can easily enable it with a view to eventually retiring it.

Navid200:
@jamorham This setting is enabled by default.  Our guide states that it should be left enabled.
Are you saying if I disable it, I may get readings when otherwise, I would have no readings at all?!  

Or are you saying now that it is enabled, we are getting readings when we wouldn't get any if we had disabled it?  

Should the guide state that this should be disabled by default?

I'm kind of lost with respect to this setting after reading your comment above.

jamorham:
Thank you for pointing out that it is enabled by default. In that case I must have seen it operationally during testing where we would get data with this flag set. It is hard to know what to do in this situation without more data as to how often the situation occurs and whether the glucose is in fact usable. I think I wasn't sure about this at the time in terms of whether `InsufficientCalibration` is just a notice or means the data isn't usable which is why I put in that conditional in the first place so we could turn it off if we needed to. Obviously we never have but I still don't know that if/when we receive data marked like this whether we should be allowing it or not.

PR #4101, opened 2025-07-23
Overwrite calibrations for fake data source only

Navid200:
It is dangerous to be able to modify the calibration constants.  Because if you make a mistake, it could have dangerous results.
I suppose that is why it can only be done in engineering mode.

This PR changes this so that it can only be done for the fake data source.  So, it can still be done for testing (fake data source),  But, it will not be an option when using xDrip with a real data source.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/CalibrationGraph.java
@@ -20,6 +20,7 @@ import com.eveningoutpost.dexdrip.utilitymodels.Constants;
 import com.eveningoutpost.dexdrip.utilitymodels.Pref;
 import com.eveningoutpost.dexdrip.calibrations.CalibrationAbstract;
 import com.eveningoutpost.dexdrip.utils.ActivityWithMenu;
+import com.eveningoutpost.dexdrip.utils.DexCollectionType;
 
 import java.text.DateFormat;
 import java.text.DecimalFormat;
@@ -220,8 +221,8 @@ public class CalibrationGraph extends ActivityWithMenu {
     @Override
     public boolean onCreateOptionsMenu(Menu menu) {
 
-        //Just generate the menu in engineering mode
-        if (!Pref.getBooleanDefaultFalse("engineering_mode")) {
+        if (DexCollectionType.getDexCollectionType() != DexCollectionType.Mock) {
+            // Generate the menu only for fake data source
             return false;
         }
 


Navid200:
**Tests:**  
I installed this PR and confirmed that with fake data source, calibration settings could be overwritten.  

Then, I installed the current Nightly over it and changed the data source to Dex.  It would still allow changing the calibration constants.  
I then installed this PR over it and the option to overwrite the constants was gone with the data source set to Dex.

Navid200:
@jamorham There is nothing "hidden" about the engineering mode.
There are many who leave engineering mode enabled and use what's behind it.  I wish it would end there.  But, it doesn't.  Those individuals also create guides.
As a result, there are users who have no idea what engineering mode means and follow such guides and think engineering mode is just another menu that contains some additional settings.

There is no safety in leaving something possibly dangerous behind engineering mode.  

I am worried about people modifying calibrations and ending up with high readings when they are actually low and take more insulin.
Please reconsider.

SeIex:
Hi @Navid200, @jamorham,

I can only assume I am the reason for this? See: #4098 
Sorry Navid, if you read my text as "guide", assuming you really relate to me.
I just wrote half a sentence describing my terrible, but sadly needed workaround, without that I am personally not able to use most Libre sensors. As of calibration plugins are a little inconsistent in integration and the not working "Non-fixed Libre slopes" setting for Libre (patched app) source is the basic problem of this, maybe not only for me and @jukefoxer, what is the basic topic of  #4098.

So I don't think I recommended doing it that way, I literally said, that this shouldn’t be needed.
I sure see the risk and would also prefer a better way!

Thanks anyway for all your work on xDrip!

SeIex

SeIex:
Hi again,

>...
>there were no sensors in production that offered factory calibration. Today, there are no sensors in production without it. Considering that, I suggest we update the behavior of xDrip in that respect.
> ...

The new Accu-chek Smartguide CGM needs two calibrations after 12 hours to change from "Trend Mode" to "Therapy Mode" and make therapy decisions based on sensor reading: [https://www.accu-chek.com/sites/g/files/papvje226/files/document/dms/English/User%20Manual%20Accu-Chek%20SmartGuide%20App/SmartGuide-app-man-html-en_2024-06/index/calibratingy/calibratingy.html](url)



Here (german): [https://insulinclub.de/index.php?thread/35565-accu-chek-smartguide-cgm/&pageNo=1](url)
Are some first experiences and it looks like it is less off for those before calibration, then almost every Libre 3 in my (maybe special) case.
I also got a SmartGuide test sensor, but didn't test it yet, will take some time till i do.
What I don't like, is that new calibrations after the initially needed ones, is not possible.


OK, then I was probably not the point of writing risky guides, just of removing the calibration override. ;)

As I wrote here #4102, I personally need, with almost every Libre 3, a slope between 0.7 and 0.9, maybe max ever around 0.6 and 1.1, there may be other cases.
In my "direction" of slope, what plugins deliver and present valid (see #4093), was on a safe-ish border for me personally. Without searching, no clue where borders are set in both directions of slope and intercept for different sources and/or plugins.
More than I40 is invalid I think, override max I39 (I40 override isn't working, bug?), Libre patched app source with fixed slope 1 is only max I20, what would not be near enough in my case with slope <1.

SeIex:
> Where does this block intercept to go above 39? This blocks all manual calibrations.

@tzachi-dar 
It is not directly related to this PR, I just mentioned it because Navid started a discussion on changing limited calibrations and there might be a bug using overrides right now.


> @tzachi-dar This has nothing to do with blocking intercept of 39.
> 
> If you enable engineering mode, you can set intercept and slope with no restrictions. This PR removes that from engineering mode expect for fake data source.

@Navid200 
There seems to be some restriction, maybe a bug, setting an intercept override >39

Navid200:
A safety warning will be added to the UI for this.

Issue #3616, opened 2024-08-06
Time of "sensor expired" is only for English speakers ok


Lutz9357:
### Discussed in https://github.com/NightscoutFoundation/xDrip/discussions/3615

<div type='discussions-op-text'>

<sup>Originally posted by **Lutz9357** August  6, 2024</sup>
Please add a new function to Xdrip+. 
I haven't the possibility to change the time format In "sensor expired at" from PM and AM into 24 hour format.
I read now (in german) "Sensor endet um 8.30 PM" - In future please "Sensor endet um 20.30 Uhr"
The english time format is confusing for me. </div>

PhilTU:
This isn't an "English" issue. My phone is set to 24h clock and I would prefer xDrip to honour this setting.

wirbel-at-vdr-portal:
I would really like to see this change in future xdrip.

No longer messing up a sensor change by 12 hours, because stumbling once again over the 12hr/24hr AM/PM diff between english and non-english users would be an strong improvement.

Navid200:
@wirbel-at-vdr-portal 
It is on my todo list.  Although, I am not sure how easy it will be.  
I have open submissions that have not been reviewed yet.  I don't want to flood the repository with submissions.
It will be done if possible when we get to it.

Navid200:
@wirbel-at-vdr-portal I am sorry about your loss.  You can unsubscribe from this issue so that you won't receive a notification every time someone posts in it.
You have now posted in the associated PR.  As a result, you will also receive a notification every time there is a post in that PR.  Again, you can unsubscribe.

PR #4131, opened 2025-08-17
24-hour format support

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/3616 

Currently, we show sensor expiry using am/pm regardless of what time format is set in Android system settings.
This PR changes xDrip so that when 24-hour format is selected in Android settings, sensor expiry time uses the 24-hour format.

The following 2 screenshots show sensor expiry with the 24-hour time format enabled and disabled.

Screenshot_20250816-203244  
  
Screenshot_20250816-204607


wirbel-at-vdr-portal:
Sorry, i can't test anymore. The person that used the sensor passed by 10 months ago.

PR #4133, opened 2025-08-19
Session start time update for alt (corrects Nightscout SAGE) (phase 2)

Navid200:
After you insert a new sensor with your old sensor still active, you have 2 options:

1- You can disable collection in xDrip, peel off the old sensor and move away, re-enable collection and change transmitter ID to the pairing code of the new device. 

2- You can wait for your old sensor to expire.  Then, peel it off and move away. Then, change the transmitter ID to that of the new device.  

In either case, there will be a discrepancy between the local session start time and the real session start time as reported by the new device transmitter.  

This discrepancy was addressed for case 1 here: https://github.com/NightscoutFoundation/xDrip/pull/4047 
It has been tested for a month now.  
This PR addresses the discrepancy for the remaining case 2 as well.  

I will be able to test this in 10 days making sure there will be no backfill overlap.  Hopefully, it will pass and I will then bring this out of draft mode.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/Treatments.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/Treatments.java
@@ -453,8 +453,8 @@ public class Treatments extends Model {
         val lastSensorStart = Treatments.lastEventTypeFromXdrip(Treatments.SENSOR_START_EVENT_TYPE);
         long localStartedAt = lastSensorStart.timestamp; // When the xDrip local session started
         long dexStartedAt = DexSessionKeeper.getStart(); // When the current session on the transmitter started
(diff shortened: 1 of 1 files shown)


Navid200:
This will be tested, for backfill overlap, in 36 hours from now.

Navid200:
I allowed my old sensor to expire.  I then peeled it off and moved it away.
I then switched the transmitter ID to that of my new sensor that I had inserted 12 hours ago.
It took an hour for the session start time to be updated.
There was no backfill overlap.

I had disabled Nightscout upload because in case there was an overlap, I did not want the overlap to be uploaded to Nightsocut.
After enabling Nightscout, it has not backfilled.  
The result is that the session start time in xDrip is correct now.   But, the session start time is still incorrect in Nightscout.  
I have always noticed that everything is not always backfilled to Nightscout.  This can be another issue we may want to look at some time.

I did a manual backfill in Nightscout.  All missing readings backfilled.   However, the treatment that represent the updated start time never backfilled.  So, Nightscout still shows an incorrect start time.  
I'm not sure why the treatment did not backfill.

I hope that this corrects the session start time broadcast to AAPS.

PR #4137, opened 2025-08-22
Statistics time period label

Navid200:
The following PR was an attempt to clarify what YTD stood for:  https://github.com/NightscoutFoundation/xDrip/pull/3926 
But, it could not be translated.  

I have taken a different approach this time.
I have added a single label that only represents the selected option and can be translated.
This way, YTD is clarified and they all can be effectively translated.  

Screenshot_20250822-082054  Screenshot_20250822-082108

Screenshot_20250822-084120 Screenshot_20250822-084157


Navid200:
The buttons still do exactly what they did before.  The only difference is that now there is a label that is updated when you tap any of the buttons and shows the description of the tapped button.

jamorham (approved):
Good solution

Navid200:
Thanks

PR #4141, opened 2025-08-25
Override silent mode log correction

Navid200:
I added a new log to xDrip for when an extra alert triggers when the phone is silenced and override silent mode is not enabled:
https://github.com/NightscoutFoundation/xDrip/pull/4014  
  
Due to my oversight, this log is also added when the phone is not silenced.  
This PR changes that and adds the log only when the phone is silenced and override silent mode is disabled.  
  
I have tested this on Android 11 set up as a follower with no Sim card.  
To cause missed reading alert to trigger, I have disabled WiFi on the phone.  
The tests were:
| Phone DND active | xDrip missed reading override silent mode On | Log created |  
| -------------------- | ------------------------------------------------------ | ------------- |  
| No | No | No |  
| No | Yes | No |  
| Yes | No | Yes |  
| Yes | Yes | No |

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Notifications.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Notifications.java
@@ -747,6 +747,32 @@ public class Notifications extends IntentService {
         player.start();
     }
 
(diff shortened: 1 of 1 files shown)


jamorham (approved):

Navid200:
Thanks

PR #4148, opened 2025-08-30
CareLink Follower - Simplera CGM

benceszasz:
**Issue**
CareLink Follower is not working for standalone Simplera CGM users.

**Solution**
In case the device type is standalone Simplera CGM the data is downloaded through the same API as for 7xxG pumps with some minor modifications.
Data available in case of standalone Simplera CGM:
- sensor glucose measurements
- blood glucose measurements
- carbs
- insulin

**Testing**
I have tested it with all kinds of accounts (patient, carepartner) with different devices (standalone Guardian 4 CGM, standalone Simplera CGM  and 7xxG pump) in EU on different phones and it is was working fine without any errors and was stable. Since US region uses the same API endpoint versions, it should be working there also.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/carelinkfollow/CareLinkDataProcessor.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/carelinkfollow/CareLinkDataProcessor.java
@@ -18,6 +18,7 @@ import com.eveningoutpost.dexdrip.cgm.carelinkfollow.message.RecentData;
 import com.eveningoutpost.dexdrip.cgm.carelinkfollow.message.SensorGlucose;
 import com.eveningoutpost.dexdrip.cgm.carelinkfollow.message.TextMap;
 
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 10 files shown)


Navid200:
Thanks for this.
Please let me know if you want me to ask users to test a test version.

benceszasz:
If you think so, you can try to ask users to test it.

Navid200:
Do they have to set the hardware data source to CareLink Follower?

benceszasz:
Yes, hardware datasource is CareLink Follower. Patients can continue to use their offical CGM app, which nam is also Simplera and they can use xDrip in parallel with the same patient account. Followers either have to create a new caregiver account or suspend using the official app, which name is CareLink Connect.
I found a discusson of xDrip users about using xDrip Companion mode with Simplera app (offical CGM app): https://github.com/NightscoutFoundation/xDrip/discussions/3267

Navid200:
Yes, xDrip can get readings from  Simplera using Companion app.
But, Companion App mode is a workaround.  It is not perfect.  So, if the users can use CareLink follower, it will be a good additional option.

benceszasz:
Yes and also it only works if the CGM app is running on Android. If for example a child with T1D uses iPhone, then even the parent has an Android phone it will not be possible to use xDrip as a follower.

jamorham (approved):
Looks good thanks

PR #4180, opened 2025-09-20
New libre sensor

tzachi-dar:


jamorham (approved):

ponzao:
Hopefully not a problem asking here about any idea when this is going to get released? Just headed abroad with a new sensor and ran into this issue. The patched version seemed to interact correctly with the sensor. Thanks for the great app and the work that has been put into it!

ponzao:
> Hopefully not a problem asking here about any idea when this is going to get released? Just headed abroad with a new sensor and ran into this issue. The patched version seemed to interact correctly with the sensor. Thanks for the great app and the work that has been put into it!

Actually nevermind, just realised there are official nightly builds, trying the latest out (it did successfully scan at least). In any case thanks very much!

Navid200:
@ponzao Please use discussions for asking questions.  Thanks

PR #4202, opened 2025-10-07
Sensor expiry note overlap fix

Navid200:
I cannot guarantee this is a perfect fix for all phones and all languages.
But, it is a very simple change that addresses the issue for the flagged cases here: https://github.com/NightscoutFoundation/xDrip/discussions/4187 
  
A more elegant solution may be to change the layout and stack them vertically.  But, that's going to make things look different for those who have got used to things.  And it will be much more work.
I will do it if you want.

To make testing possible, I created a custom version of xDrip to trigger the note early.

The following shows the problem for Russian before this PR:  
share_3198234421738001446  
  
The following shows Russian after this PR:  
Screenshot_20251007-145732  
  
The following shows German after this PR:  
Screenshot_20251007-163112


jamorham (approved):
Lets see how this goes on different devices

PR #4214, opened 2025-10-15
Tidepool uploader queue

Navid200:
Problem:
If you only upload to Tidepool, you will not be able to see the queue because the Uploaders status page will not exist. 
But, if you are uploading to Tidepool as well as Nightscout, you can see both queues on the Uploaders status page.

This PR changes that such that even if you only upload to Tidepool, you will still be able to see the uploader queue.
The following image shows the queue on a test phone set to upload to Tidepool only after this PR:
Screenshot_20251015-140947


jamorham (approved):

Navid200:
Thanks

jamorham (approved):

PR #4216, opened 2025-10-16
Vibrate on alert choice for extra alerts

Navid200:
You can disable vibration only for glucose level alerts.
If you want to disable vibration for any of the extra alerts or for any member of 'other alerts' group, you will need to rely on notification channels.

This PR adds settings so that you can disable or enable vibration for any alert through xDrip settings.  
<br/>  

**How  to enable/disable vibration for each alert:**    
| Alert | Before | After | Setting |  
| ----- | ------- | ------ | ------------- |  
| Glucose level | Existing setting | Existing setting<br/> (no change) | Screenshot_20251016-152631 |  
| Missed reading | **Notification channel** | New setting | Screenshot_20251016-100320 |  
| Other alerts | **Notification channel** | New setting | Screenshot_20251016-152154 |  
| Persistent high | **Notification channel** | New setting | Screenshot_20251016-151823 |  
| Forecasted low | **Notification channel** | New setting | Screenshot_20251016-154040 |  
<br/>  
  
This does not mean that you cannot use notification channels any longer.  It means you won't have to any longer.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/MissedReadingActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/MissedReadingActivity.java
@@ -36,6 +36,7 @@ public class MissedReadingActivity extends ActivityWithMenu {
     private CheckBox checkboxAllDay;
     private CheckBox checkboxEnableReraise;
     private CheckBox checkboxOverrideSilent;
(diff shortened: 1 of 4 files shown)


jamorham:
Once android 17 is out I think we're going to need to increase the targetsdk - notification channels is just one of those upgrade headaches.

PR #4253, opened 2025-11-23
Valid warmup time for remotely started Libre session

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/4215 
  
@jamorham Does this help?
If it gives you ideas for fixing the problem, please use it.
If it is good enough or if it needs minor tweaks, please let me know.

PR #4257, opened 2025-11-25
Skip warmup if warmed up already

### LINKED ISSUE #4215: Libre sensor "wait 60 minutes" on already running sensor
A number of users have identifed a problem using xDrip+ with Libre 2/2+ sensors. If a sensor has been operating for some time, after the initial one hour warmup, it was previously routine to start the sensor in xDrip+ and say "not started today" in order to bypass a warmup phase in xDrip. Some months ago this appeared to change.

The problem behaviour is that users are now told to wait one hour before being asked for calibration input, even though the sensor may be days old. There is also a one hour delay before data is sent to AAPS causing a break in the loop. 

Previous discussion https://github.com/NightscoutFoundation/xDrip/discussions/4004 

@AdLouwerse identified when the change of behaviour occurred - ~" Nightly build xDrip-plus-20250312-14494e7.apk works fine, the problem starts at nightly build xDrip-plus-20250323-0addc89.apk."

https://github.com/NightscoutFoundation/xDrip/compare/2025.03.12...2025.03.23

_Originally posted by @PhilTU in https://github.com/NightscoutFoundation/xDrip/discussions/4201_

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/4215

This I believe is the same as my previous attempt (https://github.com/NightscoutFoundation/xDrip/pull/4253) without the redundancies.


Navid200:
I have tested it with fake data source and it skips the warmup time if the sensor is started previously.

Navid200:
If you merge this please, it will right away address the issue Libre users have reported.
It will also fix the issue I have experienced with fake data source.
It will not block the improvement you have suggested and I will be happy to pursue it following your guidance.

If you insist that we have to do it that way, xDrip will remain problematic for a while and we will be driving users away.

Navid200:
@jamorham On a test phone, set up fake data source and start sensor.  Then, stop it so that we have time for stopped sensor recorded.

Now, start sensor and when asked when you started it, enter a time 2 days ago.  You will see that it will go into a 2-hour warm-up.

This problem cannot be solved by addressing the Libre problem only.

This PR fixes the issue for any and all data sources.

I admit that it is useful to access the device and find out exactly when it started in case the user has made a mistake in entering a wrong time.  And I can work on this to improve it.
But, please merge this so that users can use xDrip now until we improve it even more.
Improving xDrip will never stop.  But, waiting for a perfect solution leads to delays.

jamorham:
@Navid200 As far as I know this is only an issue for Libre and the code is already there to prevent it so it doesn't make sense to me to change database schema. It should just automatically handle the situation where the sensor is already started and out of warm up. That's the whole point of the computed time. That is the bug that needs fixing imho.

With this PR can you also test this: Roll back these changes. Compile xDrip - Start sensor, Stop sensor etc. Restore code changes in sync with this PR. Compile xDrip - Upgrade xDrip with the new version. Does it work?

Navid200:
> @Navid200 As far as I know this is only an issue for Libre 

I respectfully disagree.
The fake data source was also destroyed by the commit that caused the same problem for Libre.  I explained how you can recreate the problem for fake data source.


> With this PR can you also test this: Roll back these changes.  Compile xDrip

That will take this back to the Nightly 2025.11.22.

> Start sensor, Stop sensor etc. 

That effectively will be like installing our current Nightly, starting fake data source and then stopping it.   I can do that. 

>Restore code changes in sync with this PR. Compile xDrip - Upgrade xDrip with the new version. 

That means I will now install this PR over 

>Does it work?

I am going to do that right now.  I hope I am reading your instructions correctly.

jamorham:
@Navid200 I don't think you're managing the database schema correctly. Upgrade is more complex. This is another reason I don't really want to change it unless we have to.  I think there is another solution.

My current plan is to try to reproduce the problem here and see what needs fixing. `SensorDays` should handle this already so things like sensor start time shouldn't restrict display of data for exactly the reason to avoid this type of problem. Hopefully I will have time tomorrow to do that.

Navid200:
@jamorham It's so great to have you.
I made a mistake and I am glad you caught it.
I will try to fix the regression problem with my PR.  My only reason for this is because I wish the fake data source to be fixed as well.  But, I will not insist.  I will just remove draft if I can fix the regression problem and you can consider or ignore this PR then.

jamorham:
I don't think fake data source is a significant issue as it is developer only but it can likely be fixed the same way in `SensorDays` by providing a custom (short) warm-up period or using `fake data source automatic calibration` command I would expect.

jamorham:
I've removed the 3 hour window restriction on use of `computedStartedAt` in today's nightly. This works in my tests but I am only using a single sensor and stopping/starting it so it isn't quite the same as the situation users are reporting. I believe this has a good chance of fixing the problem though so I am interested on feedback as to whether it does.

Navid200:
It seems the issue is resolved for the Libre sensors.  I close this.
I still need to fix this for the fake data source.

PR #4276, opened 2025-12-08
Silencing the alert-in-the-past log

Navid200:
This log appears occasionally.

I wish I could say I have figured out why and have stopped it from happening.
All I know is that the user cannot do anything about it.  Yet, the log is seen by many and some ask why:
Screenshot 2025-12-07 182937
  
This PR changes the log to silent so that it is not visible by default.  .


jamorham (approved):
Weird I thought this had been done some time back already.

PR #4282, opened 2025-12-12
Add support for Diaexpert sensor

Der-Schubi:
The Diaexpert sensor is a version of the Microtech Lynx.
https://diaexpert.com/de


Der-Schubi:
The sensor is called Diaepert, but the package id misspelles it as diaexport.

PR #4291, opened 2025-12-18
More accurate missed reading alert

Navid200:
Problem (reported: https://github.com/NightscoutFoundation/xDrip/discussions/4283):
Set up a test phone running Android 9 or newer as a follower with our latest Nightly.
Set Missed reading alert to trigger after 7 minutes of missed readings.
Let it get a few readings.   Then, disable access to the internet.  Wait and see exactly when the missed reading alert comes up.
Run this test multiple times.  You will see that sometimes, it will trigger at values greater than 7 minutes even at values greater than 10 minutes.

Now, run the same test with this PR.  You will see that it will trigger consistently at 7 or 8 minutes.


Navid200:
@tzachi-dar Would you please have a look at this PR and let me know what you think?
Thanks

Navid200:
As you can see on line 195, the value of the parameter force, which this PR changes, only makes a difference if the parameter alarmIn is less than 5 minutes.

I have added a conditional log that appears only if alarmIn is less than 5 minutes.
I have performed tests under different conditions.  My conclusion is that this PR does not change the timer frequency.
It only changes the delay when a reading is missed.  The following goes into detail analyzing my test results.
<br/>  
  
---  
  
Tests  
I have made the MissedReadingService logs verbose using "Extra logging settings" for the following tests.  
I am running this PR only for all the following tests.
xDrip is set as Nightscout follower.  

**Missed reading alert disabled**  
Screenshot_20260408-101605
There is no sign of the conditional log.  This means this PR makes no difference when the missed reading alert is disabled.  
  
  
**Missed reading alert enabled and set to 30 minutes with no missed readings**  
Screenshot_20260408-102912
There is no sign of the conditional log.  This means this PR makes no difference when there are no missed readings with the missed reading threshold set to 30 minutes.  
  
  
**Missed reading alert enabled and set to 8 minutes with no missed readings**  
Screenshot_20260408-104408
There is no sign of the conditional log.  This means this PR makes no difference when there are no missed readings with the missed reading threshold set to less than 10 minutes.  
  
**Missed reading alert enabled and set to 30 minutes with missed readings (no WiFi)**  
Screenshot_20260408-112857  
We finally see the conditional log.  Indeed the timer rate goes up.  But, only during the very last 5-minute cycle leading to the alert.  It could be up to 4 additional wakes.  
  
But, this is not something that happens frequently.  It only happens when there is a missed reading.  
The only time this could lead to a battery drain is if a user has a disconnect; they fix the problem so the readings return;  then, the problem returns;  they fix the problem; and the problem returns.  

If the connectivity keeps going away and keeps coming back over and over again, there will be more wake cycles.  This is a very unusual case.  I don't think we should be concerned about this scenario.

jamorham:
What is the behavior when missed readings alert is disabled before/after this change in terms of frequency of wake up? 

As far as I know a lot of things are hanging off the missed readings service which is also used as a general background timer which is why I'm wary about changing it unless we're sure.

Navid200:
According to the enabled detailed logs in my tests provided above, there is no change in wake frequency if the missed reading alert is disabled.

Navid200:
I am using my private version of xDrip that has this in it for testing.
My missed reading alert is set to 20 minutes.
I see no negative impact on the battery run time.  I see no degradation in xDrip performance as a G7 collector or broadcast to AAPS or upload to Nightscout or Tidepool.  

Please merge this.  If anything breaks, undoing it will take one second.   Nothing will break.  I prommise.  
By releasing this, we will be supporting a parent of a diabetic.   That's what we do.   You told me that.

jamorham (approved):

Navid200:
Thanks you.
I will be monitoring the forums for anything that could suggest a degradation.

Navid200:
Please merge this too, which you did approve.

PR #4294, opened 2025-12-18
Changed the BG reading timer to show seconds in addition to minutes

644755:
WhatsApp Image 2025-12-18 at 22 42 42


644755:
Yes, You're right, I should omit the "seconds ago" and just leave it as 4:22 Minutes ago. For the sake of translations and not to clutter the screen. 
For my opinion it'll be better this way than to only display seconds in a certain range, but if others thinks different let's do what suites the most, or as gitan offered we can leave the option for the user to decide, but this shall take some more work. 
I'll also take it off the widget.

644755:
Fixed. Let me know if you want me to do anything else with that @Navid200

Issue #2173, opened 2022-06-10
Y-axis scaling can lead to confusing views


Navid200:
### Discussed in https://github.com/NightscoutFoundation/xDrip/discussions/2144

<div type='discussions-op-text'>

<sup>Originally posted by **christian-rummey** May 25, 2022</sup>
Hi - I just posted this in the xDrip FB group, and someone asked me to replicate it here: 

The y-axis (BG value) currently is auto-scaled to the highest level within the last 24h - see an example on the screenshot (there was a 20+high the afternoon before). This makes sense during the high, and maybe few hours thereafter, for high insulin loads, or similar issues after the high. 

But one day later, the curve can give confusing impressions. As we all know, the slope is actually more important than the actual value, and if the y-axis is scaled up to very high values, really steep slopes (eg, +/- 1.5 or more)  apper much flatter than they are. 

My suggestion is to let the y-axis scale to a normal range by default (e.g, 0-12 or so?), and only scale it up if higher values are visible on screen. Or at least let the user decide how he wants to scaling to work.  

hope that makes sense? 
thanks
- Christian

![Screenshot_20220525-092910_xDrip+](https://user-images.githubusercontent.com/41680029/170321866-55f28f8b-abf7-455c-a716-e305f77339fe.jpg)

</div>

thedewi:
The x-axis scaling is also unpredictable, which I find biases my reading of the chart. Can we lock the aspect ratio of the chart so gradients are consistent?

thedewi:
In cases where scaling is warranted, it might be better to scale both axes consistently, so at least the gradient would be preserved.

Personally I would prefer if it stayed constant until manually changed.

Navid200:
@thedewi You can force the x axis range using time buttons.  
https://navid200.github.io/xDrip/docs/TimeButtons.html

thedewi:
Thanks! Never knew you could long-press that.

thedewi:
I probably never would have figured out you need to long-press the time, as well, to lock it in

christian-rummey:
So I liked the discussion in #874, and the time buttons. 

But - I think the simple solution for this  issue would be instead of the last 24h, for the y-axis scaling we should use 4 hours (or so). 

Maybe a time when it's safe to say short-term insulin is out of the body after the peak?

It would keep us focused on what is relevant at the moment (and not what happened hours earlier). 

Thanks for your help & discussion

thedewi:
[...]

PR #4310, opened 2025-12-30
Auto Y-axis pan

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/2173  
  
My previous attempt to address this problem failed: https://github.com/NightscoutFoundation/xDrip/pull/4286 
I had overlooked the impact on predictive simulations.
But, the objective and the main approach was correct.  I am not going to repeat all the reported test results in the other PR because they are still the same.  So, please have a look at that other PR if you have forgotten.

I will go over the additional measures this PR takes.  

There is now a new setting that enables auto Y panning.  By default, it is disabled.  When disabled, there is no change in xDrip's behavior.  

|  Before | After | Notes |  
| -------- | ----- | ------- |  
| Screenshot_20251230-113454 | Screenshot_20251230-113310 | See the last item on the shown page |  
| Screenshot_20251230-113515 | Screenshot_20251230-113330 |  
<br/>  
<br/>  
  
When Predictive simulation is enabled, insulin on board or carbs on board cause the Y axis to extend down to 0.  
After this PR if you enable "Auto-pan Y-axis", insulin on board and carb on board will be drawn on screen with reference to the minimum value shown on screen instead of with respect to 0.  Therefore, the Y axis will not be extended down to 0.  If Auto-pan Y-axis is disabled (default), xDrip behaves exactly as it did before this PR. 
  
| Before | After <br/> Auto-pan disabled | After <br/> Auto-pan enabled |  
| ------- | ------------------------------------ | ----------------------------------- |  
| Before | AfterOff | AfterOn |  
<br/>  
<br/>  
  
When you have a readings that is out of your chosen range, if auto pan is enabled, xDrip will pan.  When doing that, all the non-glucose additions will also pan.  This is done by adding an offset to the values.
As an example, step count is placed at 6 mmol/L.  Let's say you have high readings causing the window to pan up by 1mmol/L.  This will cause the steps to move up to 7mmol/L.  So, the relative location of the step count with respect to the window remains unchanged.  
  
The following table shows the offset amount for different items when the `Auto-pan Y-axis` setting is enabled.  
  
| Item | Offset value <br/> withut out-of-range readings | Offset value <br/> with out-of-range readings |  
| -------------------- | ---------------------------- | ------------------------- |  
| BG Readings | 0 | 0 |  
| Insulin on board |  Ymin | Ymin + auto-pan amount |  
| Carbs on board | Ymin | Ymin + auto-pan amount |  
| Insulin activity | Ymin | Ymin + auto-pan amount |  
| Predicted glucose | 0 | 0 |  
| SMB | 0 | Auto-pan amount |  
| Step count | 0 | Auto-pan amount |  
| Heart rate |  0 | Auto-pan amount |  
  
I have been using the original PR and then this modified version all along.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/BgReading.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/BgReading.java
@@ -1003,9 +1003,10 @@ public class BgReading extends Model implements ShareUploadableBg {
                 .limit(number)
                 .execute();
 
-        // Restore ascending order for graph logic
+        // Restore ascending order for graph logic and remove invalid readings.
         if (list != null) {
             Collections.reverse(list);
+            list = filterInvalidReadings(list);
         }
 
         return list;


Navid200:
**What about Dex prediction?**  
  
Dex prediction is an option under Settings > xDrip display settings > Graph settings  
If you are using xDrip as a collector from a Dexcom transmitter and you enable this setting, you will get an additional point for every reading shown in dark purple small points.  

These prediction points often present wider variations compared to the readings.  
Only readings are used for calculating the required panning in order to keep the latest readings visible.  
Therefore, if you enable Auto-pan Y-axis and have Dex prediction enabled as well, the Dex prediction points may be cropped out at times.  But, you will be able to manually pan to see them when needed.

PR #4324, opened 2026-01-03
Companion: Read IoB from german MiniMed Mobile EU companion app

cmeissl:
The notification of the MiniMed Mobile EU companion app will be translated using the system locale. For german the IoB value will be shown as "Actives Insulin 1.234 IE".
Extend the companion app IoB regex to also read the IoB value in this case.

see https://github.com/NightscoutFoundation/xDrip/discussions/4316

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/UiBasedCollector.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/UiBasedCollector.java
@@ -149,6 +149,8 @@ public class UiBasedCollector extends NotificationListenerService {
         companionAppIoBRegexes.add(Pattern.compile("IOB: ([\\d\\.,]+) U"));
         // MiniMed Mobile (EU): "Active Insulin" label and "1.234 U" value in separate TextViews
         companionAppIoBRegexes.add(Pattern.compile("^([\\d\\.]+) U$"));
+        // MiniMed Mobile (EU): "Aktives Insulin" label and "1,234 IE" value in separate TextViews
+        companionAppIoBRegexes.add(Pattern.compile("^([\\d\\,]+) IE$"));
     }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eveningoutpost/dexdrip/services/UiBasedCollector.java
```diff
@@ -149,6 +149,8 @@ public class UiBasedCollector extends NotificationListenerService {
         companionAppIoBRegexes.add(Pattern.compile("IOB: ([\\d\\.,]+) U"));
         // MiniMed Mobile (EU): "Active Insulin" label and "1.234 U" value in separate TextViews
         companionAppIoBRegexes.add(Pattern.compile("^([\\d\\.]+) U$"));
+        // MiniMed Mobile (EU): "Aktives Insulin" label and "1,234 IE" value in separate TextViews
+        companionAppIoBRegexes.add(Pattern.compile("^([\\d\\,]+) IE$"));
```
mortenfyhn: How about just merging the now two Minimed regexes? So it'd be something like
```regex
^([\\d\\.,]+) (U|IE)$
```
(Plus the escape slashes.)

mortenfyhn (commented):
I think it makes sense to merge this to get German support until we have a general solution.

app/src/main/java/com/eveningoutpost/dexdrip/services/UiBasedCollector.java
mortenfyhn: Either way is probably good :)

mortenfyhn:
> I finally was able to test this change, the test was done on top of #4329

Beautiful!

jamorham (approved):

PR #4287, opened 2025-12-14
Fetch IoB from MiniMed Mobile EU companion app

mortenfyhn:
I noticed the MiniMed app I have shows insulin on board in the notification, so I wanted to fetch it, mostly so I can add it to my Pebble watchface ([here](https://github.com/mortenfyhn/xDrip-Pebble-E)).

With the existing Omnipod companion app IoB fetching, this turned out to be quite easy. Thanks @bart2 !

Here's the notification and the successfully fetched data displayed in xDrip:

(Test code to display  IoB value on xDrip home screen not in this PR.)

PR #4329, opened 2026-01-05
Store companion app IoB in PumpStatus

mortenfyhn:
I previously added IoB fetching from the Minimed companion app in https://github.com/NightscoutFoundation/xDrip/pull/4287. IoB from the Omnipod app was already supported. The IoB was stored inside the UiBasedCollector class.

I suppose it's likely that all companion app IoB sources are pumps, so I think this change is reasonable. Maybe if someone makes an insulin pen with an app, I'll have to reconsider.

This change saves companion app IoB data to PumpStatus instead. That makes it possible to display it on the home screen with the existing "extra status line pump status" setting:

image


cmeissl:
I rebased the changes from #4324 on top of this and it it seems to work fine.

jamorham (approved):

PR #4330, opened 2026-01-06
feat(carelink): upgrade to v13 api

m0rt4l1n:
- upgrade v11 to v13 api url
- Support for Instinct sensors

Todo:
- [x] Let Instinct users test 
- [x] Check differences for resulting patient data for non-Instinct and Instinct users
- [x] check if any checks need to be changed for instinct support

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/carelinkfollow/CareLinkDataProcessor.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/carelinkfollow/CareLinkDataProcessor.java
@@ -223,7 +223,7 @@ public class CareLinkDataProcessor {
         }
 
         //PUMP INFO (Pump Status)
-        if (recentData.isNGP()) {
+        if (recentData.isNGP() || recentData.isCC()) {
             PumpStatus.setReservoir(recentData.reservoirRemainingUnits);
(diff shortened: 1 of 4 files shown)


Navid200:
@m0rt4l1n Thanks for your work on this.
Have you tested it?  

If you know that this will not break the existing users' systems but are not sure if it will fix Instinct or not, we can ask a user to test it.  Do you need help with that?

Or, are there other concerns you have with this?

Navid200:
This is what we would have if this PR was merged now (2026.01.13) with the current Nightly:
https://drive.google.com/file/d/1cEVWsTr8rvFRk2WeMjrh1OwMgZC9RNMm/view?usp=sharing 

It is a test release that can be installed in parallel with an existing xDrip with no need to uninstall first.  
It will have to be set up from scratch.  It will have no reference to your existing settings in your existing xDrip installation.  Therefore, you will need to set it up from scratch.  You can save your settings from your current xDrip and then load settings in this test version.

This explains how you can create a backup, which generates a backup of your database as well as your settings.  There should be no need to restore the database for testing.  But, you can if you want to.  
https://navid200.github.io/xDrip/docs/Backup-Database.html

jamorham (approved):
looks good thanks

PR #4360, opened 2026-01-26
Hard reset UI improvements

Navid200:


jamorham:
I think I know what this is for but could you describe the intention of this this change?

PR #4361, opened 2026-01-27
Auto-pan default

Navid200:
I had set the default to disabled because at the time, I was worried about the crash you had experienced.
There has been no report of any crash.  I have been using this set to enabled ever since I was testing it and since it was merged.
So, I believe it is safe.

I would like to change the default to enabled so that newcomers to xDrip benefit from it and get used to it this way.

The procedure the users follow will not change.  If anyone does not like the behavior, they can switch it off.  And hopefully, they inform us by posting in forums to let us know what they don't like about it.


jamorham:
Yes I'm 99% certain the crash was being caused by the infinite glucose value coming out of SmartGuide and not the panning.

PR #4365, opened 2026-01-28
Glucose colors on lock screen

Navid200:
There was a request on facebook to allow the number wall on lock screen to use the text color representing the glucose range the same as the main screen.  This PR adds that ability.  

| Before | After |  
| ------- | ------ |  
| Screenshot_20260127-225511 | Screenshot_20260128-141310 |  
  
<br/>  
  
By default, the new setting is enabled forcing xDrip to behave exactly as it did before.
Disabling the new setting will dynamically change the text color on lock screen to represent the glucose range.


Navid200:
Screenshot_20260127-114330  Screenshot_20260127-114551 Screenshot_20260127-121601

jamorham (approved):
looks good, thanks

PR #4368, opened 2026-01-31
Use Abbott’s NFC manufacturer code `7A` when getting the newer Libre 2/3 “patch infos”

gui-dos:
[deleted]


psonnera:
I agree that correctly identifying a non supported sensor would help troubleshooting.

Navid200:
Thanks a lot

Have these been tested with compiled xDrip?

PR #4369, opened 2026-02-01
Novopen translations

Navid200:
I have never used this menu.  
I have added the English strings for translation because there has been a request for a while.
It is quite possible they were left out for a good reason.
Please ignore if that is the case.

I have also replaced the switch with a checkbox.  I believe we should replace all switches with checkboxes for consistency.
A checkbox provides more room in the layout for the summary more suitable for smaller devices specially.


jamorham (approved):
thanks

jamorham:
Out of interest why do you want to convert feature switches to checkbox?

Navid200:
The switch occupies too much space.  We need space for description (summary).  Specially, on small devices, this can be a problem.

There is a way to customize the switch so that it would not take as much space.  But, that would complicate the code.

We already use checkboxes instead of switches in xDrip.  I believe so does Google on Pixels.

Considering all of that,  I thought it was the best approach.

PR #4381, opened 2026-02-13
Check for update dialog

Navid200:
Everyday, there is a post in facebook asking for help by someone who needs to update.
When they are told to update, they say they are using the latest version.
After a lot of time is spent asking them to show the status page, it turns out they are using the stable release.

This PR adds a dialog that comes up only if the user is not using the Nightly channel and there is no update.  It stays on the screen until the user taps on close.
This should inform the user that even though there is no update in their current channel, they may still be able to get an update using other channels. 

If the user is using the Nightly channel and there is no update, this is the toast message they will see:
Screenshot_20260213-000950

If the user is on a channel other than Nightly and there is no update, this is the dialog they will see:
Screenshot_20260213-001044
  
In other words, there is no change if there is an update or if the user is in the Nightly channel.  In the latter case, the toast message is slightly changed.  
Only if the user is not in the Nightly channel and only if there is no update, there will be no toast message.  Instead there will be a dialog the user will have to tap on to close.


Navid200:
Thanks

This is also ready for your review:  https://github.com/NightscoutFoundation/xDrip/pull/4376

PR #4384, opened 2026-02-14
Disable wear auto update

Navid200:
This is my attempt to address this: https://github.com/NightscoutFoundation/xDrip/discussions/4383 

I cannot test anything.  I can create a test version if needed.
Can we use a Nightly as a test vehicle?


Finky314:
Using nightly is totally fine.

Edit: I realize now that you probably weren't asking me.

Navid200:
Thanks for the review.

Does it look OK?

Screenshot_20260225-164715

PR #4397, opened 2026-02-18
Status page cleanup

Navid200:
Adds a button that can be used to reread transmitter.
Avoids showing expected parameters on the status page the same as other devices.  

The following shows the classic and Dex status pages after.

Screenshot_20260218-182113 Screenshot_20260218-183252


Navid200:
Thanks you so much

PR #4405, opened 2026-02-23
Update proximity margin to respect sample period

Navid200:
This is meant to let 1-minute readings from another app broadcast to be accepted.
Tested here: https://github.com/NightscoutFoundation/xDrip/discussions/4265#discussioncomment-15902427


trizt:
Test setup: [juggluco (primary device)] ---{mirror}--->[juggluco (secondary device)] <---{EverSense Broadcast}--->[xdrip (640/Eversense)]<---{Nightscout sync}--->[remote nightscout server]

Tested:
xDrip receives eversense value every minute: OK
xDrip reports values to nightscout every minute: OK

PR #4423, opened 2026-03-08
Nightscout follower 1-minute interval option

Navid200:
This was requested on facebook and is presented here:
https://github.com/NightscoutFoundation/xDrip/discussions/4404 

xDrip as is shows all readings present in Nightscout at one reading per minute.  However, xDrip updates the database once every 5 minutes.  Therefore, every 5 minutes, 5 new points are added, one with a 4-minute delay, which is what could delay alerts with respect to those on the master.  
This PR when set to the 1min option for the sample period adds one reading per minute to the database. 

While working on this, I realized that I had messed up Anticipate 2.5 years ago here: https://github.com/NightscoutFoundation/xDrip/pull/3179
This PR fixes that (changes to Anticipate undone) as well so that other (than Nightscout follower) collectors that also use Anticipate are not affected in any possible way by the optional lag.    The optional lag is added at a higher level of hierarchy (NightscoutFollowService).

The existing delay option presented to the user is currently in minutes.   This PR allows the user to change the sample period.  Therefore, it will be inappropriate to define the delay in seconds.  It should instead be defined in relation to the sample period.  This PR also changes that presenting the option as a percentage of the sample period.

| Before | After |  
| ------- | ------ |  
| Screenshot_20260308-144859 | Screenshot_20260308-141750 |  
|  |  Screenshot_20260308-142057 |  
| Screenshot_20260308-145228 | Screenshot_20260308-142125 |

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -357,6 +357,7 @@ dependencies {
     testImplementation 'org.hamcrest:hamcrest-library:1.3'
     testImplementation "org.robolectric:robolectric:4.11.1"
     testImplementation "com.google.truth:truth:1.1.3"
+    testImplementation 'com.squareup.okhttp3:mockwebserver:3.12.13'
 
     testImplementation 'org.mockito:mockito-inline:2.13.0'
     testImplementation 'org.mockito:mockito-core:4.11.0'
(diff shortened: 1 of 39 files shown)


Navid200:
This has been tested and is ready for merge in my opinion.

jamorham:
Hi, I'm getting a merge conflict, I think maybe due to another of your PRs I just merged. I'm not sure exactly how to resolve them, can you take a look?

PR #4430, opened 2026-03-10
Wrong TX ID notifications

Navid200:
Post like the following come up more and more on facebook:  
  
Screenshot 2026-03-10 094504  
  
And you can see an example in our discussions as well here: https://github.com/NightscoutFoundation/xDrip/discussions/4422  
    
This is explained on the Dexcom page: https://www.dexcom.com/faq/why-does-my-transmitter-serial-number-have-1-at-the-end-of-it
And we have explained it here: https://navid200.github.io/xDrip/docs/Starting-G6.html

But, documentation is really the second line of defense.  If we can address something in the UI, we should.  That's what this PR does.
This PR informs the user by a dialog as well as by showing a note on the status page.

| Before | After |  
| ------ | ------- |  
| Screenshot_20260310-091731 | Screenshot_20260310-091930 |  
| Screenshot_20260310-091817 | Screenshot_20260310-091959 |

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1783,7 +1783,7 @@
     <string name="start_sensor_confirmation">Data Source is set to: %s\n\nDo you want to change settings or start sensor?</string>
     <string name="state_not_currently_known">%s  State isn\'t currently known. Next connection will update this</string>
     <string name="state_tx_id_ends_with_dash_1">Use only the first 6 characters of the transmitter serial number as the transmitter ID.</string>
-    <string name="message_transmitter_id_ending_with_dash1">Enter the first 6 characters of the transmitter ID without “-1”.</string>
+    <string name="message_transmitter_id_ending_with_dash1">Enter the first 6 characters of the transmitter serial number without “-1”.</string>
     <string name="title_transmitter_id_ending_with_dash1">Invalid Transmitter ID</string>
     <string name="message_removed_dash1_from_transmitter_id">Using the first 6 characters; “-1” has been removed.</string>
     <string name="title_removed_dash1_from_transmitter_id">Corrected transmitter ID</string>


Navid200:
If we just remove -1 from the end without informing the user, the problem is that it may still not work.  Users make mistakes.  They may have it already paired with Dexcom app.  They may have a typo in the 6-character transmitter ID.  If we just remove -1 and it doesn't work, what will the user think and how have we helped them?

If we do inform them and remove -1, the dialog will become longer.  This PR will not become simpler.

If we only accept 6 characters, and the user is confused, hopefully they will ask for help and they will be helped.   But, that's what the dialog does already.  Helping the user understand what they must do.

I'm not sure which part is not simple.  I just used your existing new firmware mechanism for the changes I have made to the status page.
And the dialog only states facts.

Navid200:
Please ignore the previous posts that are still there as a record.

If the user enters -1 at the end of the transmitter ID, we will remove -1 and correct the transmitter ID if there will be exactly 6 characters remaining, and we will show a dialog to inform the user as shown below.
Screenshot_20260330-174949
  
If the user enters -1 at the end of the transmitter ID and the number of characters including -1 is not 8, we leave the transmitter ID as is.  But, we show a dialog to inform the user as shown below.
Screenshot_20260330-175029

In this case only, we also change the red text shown on the main screen to state the issue as shown below.
Screenshot_20260330-175055

jamorham:
The issue I have here is that this code is very specific and narrow to solve a single problem and that can increase the maintenance overhead later. So if, for example, we do see -2 ending txid's then the whole thing needs refactoring. 

I think you should be using regular expressions to check and also for the replacement of the string, that way its easier to extend later.

For example, we probably should also be looking in the future at validating all transmitter id's for allowed characters and sequences. 

If we wrap that logic in functions in a separate class then its easier to use elsewhere. For example this PR doesn't address the transmitter id entry in `DexCollectionHelper.assistance()` - We want reusable code that's also easy to plug in in multiple places. We don't want to be duplicating logic and we want to make life easy for ourselves down the road.

So I would suggest moving the functions to a separate class, creating static methods there which can do the logical determination and provide corrected transmitter id's as well as be passed in an activity and then do informative pop-up dialogs as needed. That way we can extend that for anything else we need to do in the future.

I'm happy to take what you've done here and restructure it like I'm describing if that would be helpful.

Navid200:
Thanks a lot.  
I will do that.

jamorham:
I've been working on a utility class which I think would be really helpful here. You may want to wait until that is done. It also handles allowed characters and sequences in transmitter ids.

jamorham:
This is in today's nightly. Hopefully you like it. Thanks for raising this as a priority.

PR #4441, opened 2026-03-13
Removing "xDrip" from some setting titles

Navid200:
This PR removes the word xDrip from some xDrip settings titles.
This is done only in cases where I believe no confusion will be caused.
The benefit is shorter titles on smaller devices and/or languages that result in longer titles.

As an example, please consider "xDrip display settings".  Since this setting is only visible inside xDrip and it is about settings affecting xDrip, I don't believe excluding the word xDrip removes any crucial content.
 
As an example of a case where this is not done, please consider xDrip Sync.  Replacing "xDrip sync" with "Sync" as the title of an xDrip submenu, would cause confusion.

Most of the changes are shown in the table below.  
  
| Before | After |  
| ------- | ------ |  
| Screenshot_20260312-232020  | Screenshot_20260314-225434 |

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -222,9 +222,9 @@
     <string name="deel_settings_for_algs">Deep settings for algorithms</string>
     <string name="low_level_prediction_values">Low level prediction values</string>
     <string name="glucose_drop_for_one_unit">Glucose drop for 1 Unit</string>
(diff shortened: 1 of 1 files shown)


ahaarrestad:
looks like a good change for the users. agree that the settings suffix might be reduntant.

Navid200:
@ahaarrestad Which setting suffix?

Navid200:
Oh, I see.
OK.  Let me do that.  Thanks @asavageiv and @ahaarrestad

PR #4476, opened 2026-03-28
NSFollow V1: network guard, wakeup reorder, uploader battery & reservoir

ahaarrestad:
## Summary

Reliability and performance improvements to the V1 Nightscout follower (\`NightscoutFollow\` / \`NightscoutFollowService\`).

- **Network guard** — skip the poll when there is no active data connection; saves wake lock, DNS resolution, and TCP handshake. The next alarm is always re-armed first so the service never goes silent.
- **Wakeup reorder** — \`scheduleWakeUp()\` moved before the poll condition so the alarm fires even on early returns (previously it was only called at the end of the try block).
- **Server-side date filter** — replaces \`MissedReadingsEstimator\` with \`find[date][$gt]=<lastTimestamp>\` on the v1 API. Only entries newer than the last local reading are returned; the server does the filtering. The cutoff is capped at 24 hours back so devices offline for a long time do not over-fetch. The safety count is fixed at 2880 (2× headroom for a full 24-hour gap at 1 reading/min), making it independent of the configured sample period. Falls back to \`count=10\` on first run (no prior readings). Reduces bandwidth and server load on every steady-state poll.
- **Uploader battery & charging** — fetches \`/api/v1/devicestatus.json?count=1\` (rate-limited to 5 min) and applies uploader battery and charging state to the existing status line. Handles both the legacy flat format (\`uploaderBattery\`) and the modern nested format (\`uploader.battery\`) used by current xDrip REST uploads. Charging state displayed alongside battery percentage in one combined line.
- **Pump battery & reservoir** — applies \`pump.battery.percent\` and \`pump.reservoir\` from devicestatus to \`PumpStatus\`, displayed in the existing pump status line. Pump battery takes priority over uploader battery in PumpStatus.
- **MegaStatus display** — uploader battery and charging shown in MegaStatus alongside existing pump fields.
- **Clear on remove** — \`resetInstance()\` clears battery, reservoir, and charging from \`PumpStatus\` so stale data is not shown after the user switches away from NSFollow.

## Changes

| File | What changed |
|---|---|
| \`NightscoutCallback.java\` | Added \`Consumer<T> populator\` overload (4-arg constructor); 3-arg delegate unchanged |
| \`NightscoutFollowService.java\` | \`isNetworkAvailable()\` helper + guard + \`scheduleWakeUp()\` reorder; routes uploader status to service |
| \`NightscoutFollow.java\` | \`getEntriesSince\` on Retrofit interface (\`find[date][$gt]\`); \`getDeviceStatus\`; \`applyDeviceStatus()\`; \`resetInstance()\` clears PumpStatus; date-filtered entry fetch with 24h cap and fixed 2880 safety limit |
| \`messages/DeviceStatus.java\` | New POJO — \`uploaderBattery\`, \`uploader.battery\`, \`isCharging\`, \`pump.battery.percent\`, \`pump.reservoir\` |

## Tests

Unit tests across 7 test classes covering every new production line:

| Test class | Coverage |
|---|---|
| \`NightscoutCallbackTest\` | Populator routing, legacy path, exception safety |
| \`NightscoutFollowServiceNetworkTest\` | \`isNetworkAvailable\`: null CM, no network, network present |
| \`NightscoutFollowDeviceStatusApplyTest\` | \`applyDeviceStatus\`: both battery formats, charging, pump battery, null guards, reservoir, reset clears PumpStatus |
| \`NightscoutFollowDeviceStatusTest\` | Retrofit interface: path, auth header, count param, response parsing, edge cases |
| \`NightscoutFollowDeviceStatusWorkTest\` | \`work()\`: devicestatus dispatched, rate-limited, battery/reservoir applied |
| \`NightscoutFollowCallerTest\` | Devicestatus through full RetrofitService stack |
| \`NightscoutFollowEntriesWorkTest\` | \`work()\`: date filter used when prior reading exists; count-only on first run; safety limit is 2880; 24h cap applied when last reading is older than 24 hours |
| \`NightscoutFollowIntegrationTest\` | \`getEntriesSince\`: path contains \`find[date][$gt]\`, count, rr, auth header; response parsing |

All tests pass. No changes to \`Session.java\`, \`EntryProcessor.java\`, \`NightscoutTreatments.java\`, \`NightscoutUrl.java\`.

## Screenshots

System Status before changes:
before_v1_fixes

System Status after changes - not charging:
after_v1_fixes_not_charging

System Status after changes - charging:
after_v1_fixes_charging

Home Screen with pump status enabled:
home_screen_after_v1_fixes


ahaarrestad:
@Navid200 Added screenshots of changes

Navid200:
It is nice to have charging in parenthesis.  I see no problem with that.
It is fantastic that you have added the ability to see the master battery on the follower status page the same way an xDrip Sync follower can see it.  That is going to be very welcomes by users.  Thanks you.

It is nice that you have found a way to remove the redundant horizontal dividing lines.  Fantastic!

I have no problem with this PR as an incremental improvement.  

I hate it when a PR is blocked because it can be better.
After this is done, not necessarily now, I would like us to remove two items from this page.
1- Last BG time as it is redundant with respect to the existing "Latest BG".
2- Next poll time as it is redundant with respect to the existing "Next poll in".

The page is clearly cluttered.  The user has to scan it to find what they are looking for.  We should not have redundant items on it.

Thanks

ahaarrestad:
@jamorham are there any changes needed to this pr before it can be a candidate for merge?

Navid200:
I would really like the feature to show the master battery on the status page.

I cannot speak for jamorham.  I wonder if this needs to be broken down into smaller PRs.
The more changes you include in a single PR, the more likely it is that the entire PR will be held off.

jamorham:
Overall this looks good.

- Do you have any details on integration testing which has been done?
- What happens if the data retrieval for a cycle goes beyond the scheduled next wake-up time?

ahaarrestad:
@jamorham 

* I have tested this PR both on emulators and on real phone towards a real nightscout server for an extended periode of time. No issues found. 
* I have done some analysis and comparison between head and this PR with regards to efficiency, and it is a bit more efficient (don't have the numbers at hand, but it saves a bit of battery when offline (no radio) and it also downloads fewer elements from nightscout for each request due to the "fetch from latest read" instead of "fetch x elements and filter from latest read" which offloads filtering to nightscout and saves download volume
* If you miss a download due to offline or turned off phone, it works as before with retries and backfill, but backfill works on time filter, now number of elements, downloading max 24 hours of data regardless of sampling frequency (handles sampling frequency up to every 30 seconds as a failsafe)
* Only tested with AAPS as uplader to nightscout (latest official version), and G7 as source for AAPS.

jamorham:
@Navid200 are you happy with this PR - should it be merged?

Navid200:
@jamorham There is a lot in this PR that I cannot comment on.
I can and I will if you want me to review it in detail.

But, I am very happy that it allows a follower to see the master phone battery on the status page.  In that respect, I am very happy with this.

I have to add that later, I will open a PR to remove two redundant settings from the NS follower status page.  
Those being "Latest BG" and "Next poll in".  
And I hope you will be OK with it when that time comes.
I prefer not to hold back a PR for not being perfect and I believe in incremental updates as opposed to big perfect updates.

Navid200:
So, yes, I am OK if this is merged.

ahaarrestad:
Bumped to latest main - no changes

ahaarrestad:
@jamorham merged latest changes from main and still builds and works ok :-)

jamorham (approved):

Navid200:
Thanks

So that I can document and communicate with the users properly, what exactly is the purpose of the pump battery?

Is this so that a parent can see the pump battery of a Nightscout master?
Or, is this for someone to see their own pump battery in xDrip?  

If there is pump battery, what will be shown on the Nightscout follow status page?  The master phone battery level or the pump battery level?

ahaarrestad:
@Navid200 thanks for documenting! If you show me how to, i can participate. 

The pump battery will either be the battery level for the pump - If supplied - or the battery of the uploader If pump battery is not supplied...

ahaarrestad:
@Navid200 Yes, I'm ok with removing the redundancy. But your two comments name opposite items, so we should agree on which ones before the PR.

[April 5](https://github.com/NightscoutFoundation/xDrip/pull/4476#issuecomment-4188804176): remove "Last BG time" and "Next poll time" (the absolute timestamps).
[May 29](https://github.com/NightscoutFoundation/xDrip/pull/4476#issuecomment-4575492863): remove "Latest BG" and "Next poll in" (the relative ones).

My vote is the April version: drop the absolute timestamps, keep the relative ones. On a status page you want to know how old the reading is, and "5 min ago" answers that directly. With a wall-clock time you have to do the subtraction yourself.

So keep "Latest BG" and "Next poll in", remove "Last BG time" and "Next poll time" (NightscoutFollowService.java:274 and :276).

Open the PR whenever you like, or say the word and I'll do it.

ahaarrestad:
Opened it: #4620. Removes "Last BG time" and "Next poll time", keeps the relative rows.

Navid200:
I'm sorry about my unclear statements.

As a user looking at the status page, I find the two I have circled in the following screenshot redundant:
Screenshot_20260711-143507

The reason I think that is that first, when I see "Next poll in", that is what I need to know if I am checking the page to see how long I need to wait for the next poll.  If I need to know what time exactly that is going to be, I can figure that out.

Similarly, when I see "Latest BG", that tells me how long ago the current reading was.  Again, if I need to know exactly what time that was, I can figure it out.  
So, I suggest we remove the two I have circled in the screenshot from the page.

With respect to Buggy handset, I didn't know that it was relevant for a follower.  But, I could be wrong.
Regardless, this setting on the Dex status page is only visible if it is enabled.  If the phone follows Android specifications, which is for the majority of the cases, it is not shown.  

That is my opinion.

Thanks a lot

ahaarrestad:
Thanks — the screenshot settles it. Those are the two #4620 removes, so no change needed there.

On "Buggy handset": it is relevant for a follower. `NightscoutFollowService` runs `buggySamsung.evaluate(wakeup_time)`, so the follower uses the same wakeup workaround as the collectors. It's about alarm/wake jitter, not BLE.

But you're right that it shouldn't sit there saying "No". The collector pages (OB1, DexCollection, Wifi) and Web Follow only add the row when `buggy_samsung` is true — NS Follow, Share Follow and CareLink show it unconditionally. I've made NS Follow follow the collector convention in #4620: the row now only appears on a handset that actually needs the workaround.

Share Follow and CareLink have the same pattern, but I'll leave those for a separate PR.

PR #4479, opened 2026-04-04
[pattern] Replace android.preference.PreferenceManager with androidx in Agreement

ahaarrestad:
## Summary

- Replaces deprecated `android.preference.PreferenceManager` import with `androidx.preference.PreferenceManager` in `Agreement.java`
- Pure import swap — no call-site changes, no behavioral changes
- Adds characterization tests (`AgreementTest`) verifying existing behavior before and after the change
- Adds migration note `docs/cleanup/shared-preferences.md` for use in subsequent cleanup PRs

This is the **pattern PR** for Category 2 of the deprecated API cleanup initiative. It establishes the migration pattern for the remaining 64 files that use `android.preference.PreferenceManager`.

## Test plan

- [x] `./gradlew :app:testFastDebugUnitTest --tests "com.eveningoutpost.dexdrip.AgreementTest"` passes (2 tests)
- [x] Agreement screen opens normally and checkbox/save button behave as before
- [x] No other behavioral changes

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Agreement.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Agreement.java
@@ -3,7 +3,7 @@ package com.eveningoutpost.dexdrip;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.os.Bundle;
-import android.preference.PreferenceManager;
+import androidx.preference.PreferenceManager;
 import android.view.View;
 import android.widget.Button;
 import android.widget.CheckBox;
--- a/app/src/test/java/com/eveningoutpost/dexdrip/AgreementTest.java
+++ b/app/src/test/java/com/eveningoutpost/dexdrip/AgreementTest.java
@@ -1,0 +1,46 @@
+package com.eveningoutpost.dexdrip;
+
+import android.content.SharedPreferences;
+
+import androidx.preference.PreferenceManager;
+
+import org.junit.Before;
+import org.junit.Test;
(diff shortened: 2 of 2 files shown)


Navid200:
I have wanted to use code that I couldn't because xDrip still had not fully adopted Androids.  Thanks for working on this.  I cannot review.  But, thanks for your work.

jamorham:
Has this been tested that it works correctly including searching in preferences? I had to do some workarounds when that was first implemented and it might be tightly coupled. Before merging I want to confirm properly has been end-to-end tested?

ahaarrestad:
@jamorham Good point — let me be more precise.

Agreement.java only calls PreferenceManager.getDefaultSharedPreferences() to read a single boolean. It has no connection to the preference screen hierarchy, fragments, or the search infrastructure. Both android.preference.PreferenceManager and androidx.preference.PreferenceManager return the same underlying SharedPreferences file, so there is no behavioral change.

Preferences.java — where the search lives and where you had to add workarounds — uses android.preference.* throughout its UI structure (PreferenceFragment, PreferenceScreen etc.) and is explicitly excluded from this migration pattern. That is a separate and more complex migration that will need its own PR with careful testing.

The 64 files in this migration series all use PreferenceManager in the same simple way as Agreement.java — only getDefaultSharedPreferences(). None of them touch the preference screen hierarchy.

That said, I did test preference search end-to-end on this build and it works fine (screenshots above). I'll include the same confirmation on each subsequent PR in this series.

image

image

jamorham (approved):

PR #4485, opened 2026-04-10
Add Event Log for tracking glucose-influencing activities

Kulitorum:
## Summary

Adds a structured event logging system for tracking activities that affect blood glucose beyond carbs/insulin.

## Changes

**New: `UserEvent` model** (`models/UserEvent.java`)
- `EventType` multi-field enum — name, icon, color all defined in one place per type
- 10 built-in types: Food, Exercise, Stress, Sleep, Illness, Alcohol, Medication, Caffeine, Hydration, Other
- **Custom user-defined types** via app preferences (comma-separated names stored in `user_event_custom_types` pref)
- Sub-options for Food (size/type), Exercise (type/intensity), Sleep (quality/event) stored as JSON in `details` column
- ActiveAndroid table with idempotent migration

**New: Home screen button + dialogs**
- Calendar icon button below the note button
- Event type picker → sub-options + description dialog
- Edit/delete via graph marker tap → snackbar → edit dialog

**New: Long-press on graph**
- Long-press anywhere on the glucose graph to add an event at that historical time
- Intercepts HelloCharts gesture handler, converts screen X to timestamp via chart computator

**New: Graph markers**
- Color-coded icon markers per event type on the glucose graph
- Overlapping events stack vertically (alternating above/below center)
- Sleep events show distinct icons: crescent moon (bedtime) vs sun (wakeup)

**Modified files:**
- `Home.java` — button init, gesture handler, dialog methods, intent handling
- `BgGraphBuilder.java` — event marker rendering, click handler
- `activity_home.xml` — new button in sidebar
- `PointValueExtended.java` — new type constant
- `IdempotentMigrations.java` — migration registration
- `AndroidManifest.xml` — AA_MODELS registration
- `strings.xml` — string resources
- 13 new vector drawable icons

## Screenshots

_Will attach before/after screenshots_

## Test plan
- [ ] Event log button visible on home screen below note button
- [ ] Tap → event type picker with all built-in types
- [ ] Food/Exercise/Sleep show sub-option spinners
- [ ] Other types show description field only
- [ ] Save → colored marker appears on graph
- [ ] Tap marker → snackbar with event summary
- [ ] Tap snackbar action → edit dialog with pre-filled data
- [ ] Delete button removes event
- [ ] Long-press on graph → event picker at that timestamp
- [ ] Multiple events close in time stack vertically
- [ ] Sleep bedtime → moon icon, wakeup → sun icon
- [ ] Custom types defined in preferences appear in picker


jamorham:
I think this will need some significant discussion as to the use cases and what would be useful to people. Initial technical consideration of the PR:

- Please provide before/after screenshots
- This looks like it contains out of scope stuff relating to the PDF rendering which should not be part of this PR?
- The docs with agentic workflow should not be included in the PR
- The way that UserEvent is structured is not easily maintainable. Instead I suggest using a multi-field enum where all the associated constants are defined in the enum definition and getters provided in the same object rather than separate functions from each. Let me know if there is some reason not to do that.

From a user choice perspective I would suggest that there are quite a lot of other entry types that people will want and also there should be the ability for users to add their own text in to the app preferences to define customized log types to cover specialized cased specific to individual users.

Kulitorum:
This adds one button on the main screen, so no large GUI change — if you don't use it, it's just a small calendar icon below the note button.

I want it because I'm newly diagnosed and I want to understand how this whole thing works. So I want to log anything that has an impact on my glucose so I can analyze patterns, and understand what influences my glucose. Just looking at a curve adds no real information, this does.

There is already an "Other" entry type, so the flexibility is there. The event types are based on several AI's thoughts on what should be tracked — what would influence your glucose.

I have implemented your suggestions:

- **Multi-field enum**: `UserEvent.EventType` now holds name, icon, and color all in one enum definition with getters — no more separate switch statements for each property
- **Custom user-defined types**: Users can define their own event types via app preferences (comma-separated names stored in `user_event_custom_types`). These appear in the picker alongside the built-in types
- **PDF rendering removed**: Moved to a separate PR (#4487) — this PR is now event log only
- **Docs removed**: No agentic workflow files in this PR

Screenshot of the feature in use:

xDrip Event Log

Navid200:
You can add notes.

What is nice about that is that you can later perform a search in notes and find the ones you are looking for.
As an example, you can enter notes for exercise.  You can enter notes for strenuous exercise.   And you can for example enter notes for easy exercise.
Later if you search for just exercise, it will fins all of those.  But, you can search for easy exercise and that will only bring up the easy ones.  

This makes what we have extremely useful and powerful.

If we go ahead with what you are suggesting here, later someone is going to open a discussion and say why is there no symbol for easy exercise?  How can I distinguish those because easy exercise and regular exercise affect my glucose differently.
So, we will be asked to add more and more symbols.

Can you explain why you cannot use xDrip as is please?

jamorham:
Another significant issue with this proposed feature is I don't see that it fits in to Nightscout ecosystem very well. As these data elements are not synced anywhere else. I also think the structure is too prescriptive and doesn't provide enough customization for the user as well as missing key event types which are likely to be common. For example menstruation, cannula problem etc. 

I think the concept deserves more discussion within the community for what would actually be useful to people and how it integrates in to data sync definitely needs to be considered more as well as what kind of user interface would balance quick ease of use and customization. Also conceptually I think there needs to be a facility for an event to have a duration and the ability to enter them historically.

PR #4486, opened 2026-04-10
Raise max from 400 to 500mg/dL

Navid200:
xDrip currently shows HIGH for any value that is equal to or greater than 400 mg/dL.

There are sensors that show values greater than 400.  

This PR changes that and sets the maximum to 500 mg/dL.
The max remains at 400 if the user uses plugins.  

I have tested this with fake data source and see that it clips at 500.
I am currently using this as my xDrip.


jamorham:
Which type of sensors have range >400 out of interest?

Navid200:
I'm sorry I don't have a list.
But, I know Dexcom limits to 400.  But, Libre limits to 500: https://www.freestyle.abbott/sa-en/support/faq/question-answer.html?q=the-freestyle-libre-system-tab-question-97 

I don't have any info about the others.  But, I can do research if you want.

Navid200:
This is Libre 3: https://www.support.freestyle.abbott/hc/en-us/articles/14772750216343-What-do-HI-and-LO-results-mean

It shows 400 as the max.  

Thanks for the review.  Let's not do this for now.  I need to do some more research.

PR #4487, opened 2026-04-10
Add PDF export for glucose reports

Kulitorum:
## Summary

- Adds **PDF export** that generates a printable glucose report for health professionals
- Landscape A4 pages with glucose curve, event markers, treatment list, and statistics
- User configures date range, page count (time-slicing), and content toggles via dialog
- Glucose curve rendered as vector graphics directly to PDF Canvas (crisp at any zoom)
- Smart peak/valley detection with 1-hour smoothed curve + prominence filter
- Peak labels avoid overlapping the curve via bbox collision detection
- Start/end value readouts in reserved margins at exact Y position
- Dynamic Y-axis (starts at 0, auto-scales to data, includes high/low thresholds)
- No new dependencies — uses Android native `PdfDocument` API

## Dependencies

- Depends on #4485 (Event Log feature) for `UserEvent` model used in event markers

## New files
- `PdfExportConfig.java` — date range, page count, content toggle data class
- `PdfReportRenderer.java` — full PDF rendering with Canvas drawing

## Modified files
- `Home.java` — export dialog with date pickers, spinner, checkboxes, AsyncTask
- `menu_home.xml` — "Export PDF Report" menu item
- `strings.xml` — PDF export string resources

## Test plan
- [ ] Hamburger menu > Import/Export > "Export PDF Report" opens config dialog
- [ ] Date pickers default to last 7 days
- [ ] Page count spinner (1-30) works
- [ ] Checkboxes toggle events, treatments, statistics
- [ ] 1-page export shows full date range
- [ ] Multi-page export time-slices correctly
- [ ] Glucose curve renders with correct units (mg/dl or mmol/l)
- [ ] High/low threshold lines at correct positions
- [ ] Peak/valley labels don't overlap the curve
- [ ] Start/end readouts sit in margins at correct Y
- [ ] Events listed in side panel with color dots
- [ ] Treatments listed in side panel
- [ ] Statistics bar shows per-page stats
- [ ] Share button opens Android share intent


Kulitorum:
Here's an example report. I'm a new user, so I have only few data points:
[xDrip-Report-20260410-192453.pdf](https://github.com/user-attachments/files/26636645/xDrip-Report-20260410-192453.pdf)

Navid200:
Have you looked at Statistics under the main menu?

Can you explain why you want to change xDrip?  I'm not suggesting that xDrip is perfect.  But, it helps very much to know what your needs and intents are.  If xDrip can already do what you need, why would you want to change it?

jamorham:
I think for PDF reports the best route is to upload the data to tidepool because the print report there which produces a PDF is really nice and follows common standards that clinicians are used to.

In terms of improving the data we send to tidepool, I believe that currently missing is information about the profile, in terms of insulin and carb ratios, baseline basal rates etc. 

I think effort in to improving that aspect would be more worthwhile than adding a separate PDF report feature which will take a very considerable amount of work to bring it up to the same quality as is available from tidepool and of course the maintenance overhead.

PR #4493, opened 2026-04-18
Bluetooth page cleanup

Navid200:
The Bluetooth settings page is too cluttered.
This PR removes 4 of the 18 settings.

Before and after are shown below.

| Before | After |  
| ------- | ------ |  
| Screenshot_20260417-230833 | Screenshot_20260417-230600 |  

The Dex Bluetooth watchdog is not used anywhere in the code.
The Bluetooth frequent reset is not used anywhere in the phone code.  It is transferred to Wear.  But, the setting does not affect anything on the phone.  If this causes a problem on a watch, I will fix it.  I mean xDrip does not support G5 any longer.  So, what sensor is the watch directly communicating with that needs to be reset continuously?
The other 2 settings are related to experimental devices, which I doubt any user is using.  

The page will still need cleanup work.  
I am hoping this will be easy and go through easily.  Then, I will open other PRs to hopefully address some other settings on this same page.  

**Testing**  
I am currently running this as my main active xDrip, a collector from G7.


ahaarrestad:
I really like cleanups of both UI and code.

jamorham:
Yes this is a reasonable change

jamorham (approved):

PR #4497, opened 2026-04-18
More prominent extra alerts

Navid200:
I wanted the Persistent high and forecasted low and sensor expiry alerts to have dedicated settings pages so that we could add more settings to each.  This was the PR that did it: https://github.com/NightscoutFoundation/xDrip/pull/3704 
It accomplished my objective to separate those alerts by putting each in its own page.

But, the consequence is that each one of those alerts moved down into a lower level of hierarchy.  That was not really my intent and I did not think about that consequence at the time.  I am sorry about that.

This PR moves those alerts up in the hierarchy.  They stay separated from each other.  But, they will all move up to the same level of hierarchy as the missed reading alert or the calibration alerts.  

Please let me know what you think.  

This has been tested and I am currently running it.

| Before | After |  
| ------- | ----- |  
| Screenshot_20260418-154556 | Screenshot_20260418-154505 |

### DIFF
--- a/app/src/main/res/xml/pref_notifications.xml
+++ b/app/src/main/res/xml/pref_notifications.xml
@@ -154,6 +154,117 @@
                 </PreferenceScreen>
 
             </PreferenceScreen>
+            <PreferenceScreen
+                android:key="persistent_high_alert_page"
+                android:title="@string/persistent_high_alert">
+                <SwitchPreference
+                    android:defaultValue="false"
+                    android:key="persistent_high_alert_enabled"
+                    android:summary="@string/alarm_if_above_high_value"
+                    android:switchTextOff="@string/short_off_text_for_switches"
+                    android:switchTextOn="@string/short_on_text_for_switches"
+                    android:title="@string/persistent_high_alert_enable" />
+                <SwitchPreference
+                    android:defaultValue="true"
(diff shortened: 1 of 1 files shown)


jamorham:
I'm not sure what the resolution of the screenshot is, my only concern generally is that on smaller screens might require scrolling which might make it less obvious there are other alarm settings. It probably does make sense to have laid out like this though.

Navid200:
Thanks,


I will use the virtual device to see how it will look on a small device.

Navid200:
You are right that on a small device, it will require scrolling.
But, the is the same with the main settings page as well.
On a very small screen, even this same page as is (before this PR) will also require scrolling.

I understand your point.
But, this is not something that is specific to this PR.  

When you go to the page, a scroll symbol is visible on the right side of the screen.  But, it disappears shortly.  The same is true with the main settings page.  

The following table shows this page before and after this PR on a small device 240 X 320.

|      | Before | After |  
| -- | ------ | ------- |  
| Top | Screenshot 2026-04-26 175307 | Screenshot 2026-04-26 180251 |  
| Scrolled down | Screenshot 2026-04-26 175344 | Screenshot 2026-04-26 180317 |  
  
Do you have any suggestion?

jamorham (approved):

PR #4498, opened 2026-04-19
Get only last measurement from Contour Next meter

psonnera:
Do not pull all measurement but only the last one performed.
https://github.com/NightscoutFoundation/xDrip/discussions/4495

Tested functional but need review to make sure it doesn't interfere with other meters


psonnera:
That's a botched PR. Let me redo that

PR #4499, opened 2026-04-19
Get only latest measurements from Contour Next meter

psonnera:
Do not pull all measurement but only the last one performed.
https://github.com/NightscoutFoundation/xDrip/discussions/4495

Tested functional but need review to make sure it doesn't interfere with other meters

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/glucosemeter/GlucoseReadingRx.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/glucosemeter/GlucoseReadingRx.java
@@ -50,7 +50,7 @@ public class GlucoseReadingRx extends BluetoothCHelper {
             final boolean sensorStatusAnnunciationPresent = (flags & 0x08) > 0;
             contextInfoFollows = (flags & 0x10) > 0;
 
-            sequence = data.getShort(1);
+            sequence = data.getShort(1) & 0xFFFF;
             year = data.getShort(3);
             month = data.get(5);
             day = data.get(6);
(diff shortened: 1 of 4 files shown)


Navid200:
Thanks

Does this respect multiple meters?
Let's say a user has one at home and one at work.  They use both with the same xDrip back and forth.  
Will this code recognize the two different meters and still operate properly getting only new readings from both meters and not skip readings on either?

psonnera:
From what I see in the code, the sequence is specific to each meter as it's indexed to the device mac address.
[```private static int getHighestSequence() {
return (int) PersistentStore.getLong(PREF_HIGHEST_SEQUENCE + mBluetoothDeviceAddress);```](https://github.com/NightscoutFoundation/xDrip/blob/82b0ff5952c9d0ec26f60e8d0596550f6beefbdb/app/src/main/java/com/eveningoutpost/dexdrip/services/BluetoothGlucoseMeter.java#L1163)
So yes, I believe only new readings will be uploaded without interference between devices.
I also wouldn't be able to test that with different meters.

jamorham:
Could you explain the PR in more detail? Is this requesting the last reading or reading since time of last sync so that its incremental but will get all the readings initially or does this just always get the single last reading?

psonnera:
@jamorham this PR is to avoid uploading all measurements but only those not already uploaded into xDrip+: all reading since last sync. I know it works for the last one. The logic using the index should work for more but I will test the second option and confirm.

jamorham:
Please let me know if you have any tests results regarding your last message. thanks

psonnera:
@jamorham sorry for the delay. Yes: it's pulling all new (non already sync'ed) measurement. Not only the last one. I believe it's working.

jamorham (approved):

PR #4509, opened 2026-04-26
fix: UiBasedCollector uses source package unit instead of xDrip preference

ryceg:
## Summary

I switched from a Dexcom sensor to a Freestyle Libre 3 Plus, both in companion mode with CamAPS FX. After the switch, xDrip stopped receiving any glucose readings. CamAPS FX (mmol variant `com.camdiab.fx_alert.mmoll`) was sending notifications with values like "7.1", but `UiBasedCollector.tryExtractString()` was trying to parse them as mg/dL integers because my xDrip is configured for mg/dL — resulting in `NumberFormatException` on every reading.

The root cause is that `tryExtractString()` used `Unitized.usingMgDl()` (xDrip's own display preference) to decide how to parse notification values, but the companion app's unit is determined by its package name, not xDrip's setting.

## Fix

- Added `sourcePackageUsesMgDl()` which detects the unit from the companion app's package name (`.mmol`/`.mmoll` suffix → mmol/L, `.mgdl` suffix → mg/dL), falling back to xDrip's preference for packages without a unit suffix
- This correctly handles all existing co-opted packages: CamAPS regional variants, Dexcom regional variants, Microtech variants, etc.

## Test plan

- [x] Tested live on device — CamAPS FX mmol notifications now parse correctly (7.0 mmol/L → 126 mg/dL)
- [x] Added unit tests for `sourcePackageUsesMgDl()` covering all package name patterns
- [x] Added unit test for end-to-end mmol parsing via `tryExtractString()`

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/UiBasedCollector.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/UiBasedCollector.java
@@ -368,7 +368,7 @@ public class UiBasedCollector extends NotificationListenerService {
         int mgdl = -1;
         try {
             val ftext = filterString(text);
-            if (Unitized.usingMgDl()) {
+            if (sourcePackageUsesMgDl()) {
                 mgdl = Integer.parseInt(ftext);
             } else {
                 if (isValidMmol(ftext)) {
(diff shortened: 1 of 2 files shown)


Navid200:
Why not just use a version of the CamAPS that uses mg/dL?

ryceg:
Would if I could, it's unavailable in my country.

PR #4511, opened 2026-04-30
Daily job even on mobile

Navid200:
xDrip will never run the daily job for someone who never uses WiFi.  
This results in the Google Drive automated daily backup never running.  
This was reported on facebook.  I asked them to post in our discussions for reference:
https://github.com/NightscoutFoundation/xDrip/discussions/4502 

I don't think this is right.  If someone does not want to upload using mobile, they can control that on the Nightscout settings page or on the Google Drive backup settings page.
I don't think daily job should impose such a limitation.  

This PR changes xDrip's behavior such that the daily job is run as long as there is connectivity to the internet even if it is through cell service and not WiFi.

I am currently using this.  I disabled WiFi on my phone last night after having installed this PR.  The following shows the logs and the log associated with the daily job that highlights that it ran while using mobile data.

Screenshot_20260430-093121  
  
Please ignore the previous logs showing mobile that are related to a previous design I decided to change.  
I wanted to give priority to WiFi.  But, that became too complicated.  I don't see a reason to give priority to WiFi.  The user has full control over limiting uploads to WiFi already.  
Only the log dated April 30 is related to this PR.  

What do you think?

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/jobs/DailyJob.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/jobs/DailyJob.java
@@ -27,7 +27,9 @@ public class DailyJob extends Job {
     protected Result onRunJob(@NonNull Job.Params params) {
         final long startTime = JoH.tsl();
         DailyIntentService.work();
-        UserError.Log.uel(TAG, JoH.dateTimeText(JoH.tsl()) + " Job Ran - finished, duration: " + JoH.niceTimeScalar(JoH.msSince(startTime)));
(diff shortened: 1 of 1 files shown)


ahaarrestad:
I agree this looks like a good way to ensure daily jobs run periodically. We could, in a later PR, try to make all this configurable, but at the same time, most people don't care for all the technical parts of this and just want it "to work" :-) 

Ideally I would love some tests for this. I really like TDD and it helps to both document and prove behaviour.

jamorham (approved):
Seems like it makes sense given that a fair few things hang off that service.

Navid200:
Thanks

PR #4516, opened 2026-05-04
Nightscout follower backfill for a 1-minute sample period

Navid200:
This PR affects the backfill request behavior of the Nightscout follower.
If the sample period is set to 1 minute, the backfill request requests for more samples respectively.
The problem this PR addresses is explained here: https://github.com/NightscoutFoundation/xDrip/discussions/4515


Navid200:
Test:
After this PR, setting the sample period to 1 minute and turning off the follower phone for 15 minutes and turning it back on will leave no gap for the period the phone was off.

However, if you have the follower sample period set to 5 minutes when you turn the phone back on, there will be a gap.  And correcting the sample period on the follower after will not fill the gap in this case.  

After this PR, as long as the user sets the sample period correctly, xDrip will behave correctly leaving no gaps and requesting no exsesive backfill.

ahaarrestad:
Alternative solution baked into #4476 filtering on 24h instead of number of entries

Navid200:
It seems the problem is resolved.  So, I close this PR.

PR #4517, opened 2026-05-04
AAPS basal rate fix

psonnera:
Trim space before U/h from AAPS external status line to allow correct basal display.
Space seems to always have been there, so I don't know why this happened.
This PR should allow both versions to be parsed: with and without space.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/wearintegration/ExternalStatusService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/wearintegration/ExternalStatusService.java
@@ -139,7 +139,7 @@ public class ExternalStatusService extends IntentService {
 
     public static String getAbsoluteBR(final String statusLine) {
         if (JoH.emptyString(statusLine)) return "";
-        val pattern = Pattern.compile(".*(^|[^0-9.,])([0-9.,]+U/h)", Pattern.DOTALL); // match last of any number followed by units per hour
+        val pattern = Pattern.compile(".*(^|[^0-9.,])([0-9.,]+ ?U/h)", Pattern.DOTALL); // match last of any number followed by units per hour
         val matcher = pattern.matcher(statusLine);
         val matches = matcher.find();       // was at least one found?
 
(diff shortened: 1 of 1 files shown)


Navid200:
I set up xDrip and AAPS on an Android 11 test phone.
Screenshot_20260505-223412

I started with our latest Nightly version of xDrip, 2026.04.29.
Screenshot_20260505-193658
And it doesn't seem to work.  

I then compiled this PR added to our latest Nightly.  
Screenshot_20260505-212141
It seems to work now.  Although, I don't understand why it looks to be upside-down.  But, I am not a user.  So, let's forget that for now.  

I then installed our latest Nightly without this PR again.
Screenshot_20260505-223853
And sure enough, it has started working again.  

**Edit**
The above line contains a typo.  It was meant to say "sure enough, it has stopped working ...".  Please see my post below.

But, as you stated on facebook, there has been no change to xDrip where this PR intervenes.  And you state that there has also been no change to AAPS in this respect.  So, the question remains, why does this fix the problem?

psonnera:
If you return to the release you'll see the graph displays but doesn't change anymore. The amplitude of basal graph depends on your xDrip+ profile, and unless it is synchronized with AAPS, it can just go up and down as the calculation of the basal rendering is based on a % of the local profile. The fact it's downwards is because the icicle style is used, and it's easier to handle since higher basal rate goes lower in the graph hence avoid risking to go too high in the display. This should be changed to absolute basal rate but raises the issue of scaling. I would like the PR to be merged anyway as it's not disrupting external status recognition: with or without space.

jamorham:
Which version of AAPS does this affect? I'm using xDrip with AAPS (probably older version) and have not had any problems with the basal handling. That said the change looks to be safe an would improve the tolerance.

psonnera:
@jamorham I believe it started with AAPS 3.2 but there is no clear indication anything changed in AAPS status information.

jamorham (approved):

Navid200:
Thanks for this Patrick.

I find that my posts on GitHub are changed often and I assume it is done by a spellchecker or something.
My post was about to state that after I installed the latest Nightly before this PR was merged, it stopped working again.  But, my post stated that it started working.  

My point was that this PR seemed to fix the problem and the problem returned after undoing this PR and going back to the Nightly that did not have your fix.

I need to proof read every one of my posts after I am done with it and consider it done.

Anyway, thanks for fixing this.

PR #4524, opened 2026-05-10
Language submenu

Navid200:
This PR adds a new item "Language" to the main settings page between "Glucose units" and "Alarms and alerts".

Currently, language settings are located under the "Display" menu.
But, language is a more general factor with respect to display.
I have found myself looking through the menus looking for language at times.
I believe the top level of the settings is more suited for the setting.

Since there are currently two different settings related to language, this PR moves them both under the new "Language" setting.  It removes those two settings from the display menu.  

I am currently running this and see no issues.  

| Before | After |  
| ------- | ------ |  
| Screenshot_20260510-112653 | Screenshot_20260510-112037 |  
| | Screenshot_20260510-112120 |  
| Screenshot_20260510-112925  | Screenshot_20260510-112100 |

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
@@ -1138,6 +1138,8 @@ public class Preferences extends BasePreferenceActivity implements SearchPrefere
             units_pref = findPreference("units");
             bindPreferenceSummaryToValue(units_pref);
 
+            addPreferencesFromResource(R.xml.pref_language);
+
             addPreferencesFromResource(R.xml.pref_notifications);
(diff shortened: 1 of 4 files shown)


jamorham:
No I don't think this should be at a top level, the language selection is really only an edge case for users using a different language on their phone than the one they want xDrip in. Its also a setting they're only likely to use very rarely.

Navid200:
I will place it where you like if you don't want it at the top level.    But, language is not a display setting.
Even if you look at your phone (Android), you will never find language under display settings.

Navid200:
I agree that the very top level may not be the very best place for language.
I still believe that "Display" is not the right place for Language.

That's why I will close this instead of changing it.  
I will open another PR to put the two language-related settings on a submenu on the Display settings page since you insist.
Hopefully, some day, we can find a more appropriate place for it.

PR #4525, opened 2026-05-14
Retiring Android 7

Navid200:
Edit:
See this instead: https://github.com/NightscoutFoundation/xDrip/pull/4601 
  
---  
  
I hope that Android 8 phones can be found at reasonable cost in third world countries.

Only tested:
G7 collector
G6 collecor
Upload to Nightscout
Upload to Tidepool
xDrip Sync master
xDrip sync follower
Nightscout follower
Companion app mode
Broadcast to AAPS
Receiving broadcast as 640G/Eversense

I cannot test:
Libre
Wear OS
Tasker


Navid200:
| Before | After |  
| ------- | ----- |  
| Screenshot_20260514-073905 | Screenshot_20260626-220615 |  
| Screenshot_20260626-221220 | Screenshot_20260626-221032 |

Navid200:
This is what I am running (Pixels 6a, Android 16) and I also have used a test phone (Moto Android 11).

Since I have opened this, I have had Low alert, forecasted low, high alert and persistent high alerts all working fine.

Navid200:
Please tell me what I have overlooked here.

When we are ready for merge, I will announce on facebook reminding people that after reporting any issues, they can easily undo an update by installing the previous version over it.

Navid200:
I'm seeing an unexpected (by me) behavior.
Even though I have "Group notifications together" disabled, when I get an alert, the alert notification and the ongoing notification are grouped!

I just checked and number icon in the notification area works fine.

jamorham:
There are two sets of changes which I think we need to handle carefully. What I'd like to do is push the current nightly to alpha and then start with potentially breaking changes like this. For android 8 isn't there an issue with background services being started from the background? That was the main thing that worried me about the move to android 8 target?

Also I have a significant set of changes which I think I should put in to a new branch because it might be too risky for nightly just yet. I should get time for that this week.

jamorham:
So I have now merged my gradle 8 changes so we can look again at this modernizing.

Back when Android 8 came out I looked at this and thought we would have significant pain relating to services being started from the background.

I think this maybe we get a pass on when we're using battery whitelisting - which we must ensure is set, it is no longer optional on later android versions.

The issue is not with cost/availability of handsets but to do with the target sdk restrictions and what problems they might cause us. But we will need to update because of the problems with side-loading if we are too far behind and also if we can modernize and the app still work successfully then we should do it.

Its worth checking things like after a reboot if the app isn't opened, does it manage to restart the collector correctly (even with screen off etc)

Navid200:
In a few hours, I will address the conflict this PR has and I will run the tests you have suggested.
Thanks

Navid200:
Just to be clear about my statement about availability of the devices, I am concerned about retiring Android just because Google has decided we have to retire it.  I don't want us to do it unless we are certain people living in every country and at every economical capability can survive.   I hope that there will be Android 8 devices given away considering many bank apps do not accept Android 8 any longer.  This is my hope that it is not too soon retiring Android 7.  

That statement did not mean to suggest that I did not care about what may break if/when we retire Android 7.  I am very well aware of those concerns and have never suggested they did not matter.  

 Those are two different concerns.  They both matter and are important.

Navid200:
I have updated this so that it is based on the latest Nightly now including Gradle 8.
I am running it.  

I just restarted the phone.  After unlocking the screen after reboot, I can see xDrip's ongoing notification.
  
I didn't open xDrip.  But, am looking at Nightscout on my computer in a browser.  This is Nightscout that my xDrip directly uploads to.  It's showing last reading to be 3 minutes ago.
Still not opening xDrip on the phone.
Nightscout is now showing last reading to have been 7 minutes ago.

I think you have a point and we do have a problem that we need to resolve before we can merge this.

Thanks for suggesting this test.  Now that I understand this test option, I can work on fixing this.

Navid200:
The collector starts now automatically after a reboot.

But, we need to figure out why all the notifications are grouped together.  Swiping an alert away also removes the ongoing notification.  It will return in up to 5 minutes.  But, the notifications should not be grouped.

Navid200:
I feel I have addressed challenges.
I am now testing updating from our Nightly and reverting back (regression).
When this is removed from draft, I wish someone else tests it as well before we merge it to Nightly.

Navid200:
I have tested an upgrade from Nightly to this on real phones Android 16, Android 11 and Android 8.
The alerts continue to make sound respecting the chosen custom sounds.  The upgrade does not make any alert silent and does not change any sound files.

Going back to the latest Nightly does not cause any crash and the alerts remain functional.   
  
---  
  


1- Why are there zombie notification channel after I change a sound file?
This is what xDrip does now.  If we decide to change this, we can.  But, I'd rather we do it separately in a different PR.

2- Why are different notifications grouped?
The ongoing notification will never be grouped?  But, if you have a low alert and forested low alert triggered at the same time, they will be grouped.  Why would you want to snooze one and not the other?  Even if you want to do that, you can  by tapping on the arrow to extend the group.  
Likewise, if you have a persistent high alert and a high alert trigger at the same time, they will be grouped.

3- Why is the option to group notifications gone?
Sorting out notifications was a nightmare.  It took so long.  If it is absolutely necessary to maintain that option, this PR will be delayed even more.  I hope we can be happy without that option.

Navid200:
No way!  You've got to be kidding me!  The number icon in notification area is gone!

Navid200:
This PR already has 16 commits.  
It is time to replace this PR with a clean one.  
I am closing this PR now and I will open a new PR later today.

PR #4526, opened 2026-05-16
Dynamic api version check

Navid200:
This is what we currently have in Home.java:

        if (Build.VERSION.SDK_INT < 17) {
            JoH.static_toast_long(gs(R.string.xdrip_will_not_work_below_android_version_42));
            finish();
        }
        
And the corresponding string is this:
xDrip+ will not work below Android version 4.2

We have not updated either the comparator in Home or the corresponding string since we have raised minSDK.

This PR makes several changes:
1-The comparison is now dynamic.  As we raise minSDK, It will automatically point to the updated minSDK.  Therefore, there will be nothing to forget in that respect.
2- The corresponding string is more general not identifying any particular Android version.  Again, there will be no need to update it when/if we raise minSDK.
3- Instead of a toast message, we will show a dialog, which will stay on the screen until the user acknowledges it.
4- If the dialog fails to show, we will only then show a toast message.
5- After showing the dialog or the toast, the code will terminate instead of continuing to run.  There is no reason to continue to run considering we cannot test xDrip on older versions of Android.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -92,6 +92,7 @@ android {
         minSdkVersion 24
         // increasing target SDK version can cause compatibility issues with Android 7+
         //noinspection ExpiredTargetSdkVersion
+        buildConfigField "int", "minSDK", minSdkVersion.mApiLevel.toString()
         targetSdkVersion 24
         // change versionCode only when downgrade should be prevented
         // eg, when data structures are incompatible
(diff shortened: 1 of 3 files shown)


ahaarrestad:
Looks good with a dynamic check instead of fixed

jamorham:
Guys, I don't think you can install an app below the min-sdk level so this condition will never fire. It dates back to a very long time ago when the min-sdk was actually lower than the version xDrip would work with. I think probably a better thing to do is remove this check and message.

Navid200:
I can remove this.

When I looked at it and tried to figure out why it may have been added, I decided someone must have thought it may be possible  to restore some form of backed up image from one phone to another when the destination phone runs an older version of Android.  Then, if someone did that with xDrip, it could come up in an Android version lower than minSDK.  I thought it was a safety valve and did not want to just remove it.

I will if you confirm please that you are sure you want it removed.

PR #4529, opened 2026-05-18
Feature/nocturne uploader

ryceg:
Adds Nocturne as a remote upload target, authenticated via OAuth 2.0 Dynamic Client Registration (RFC 7591) and the Device Authorization Grant (RFC 8628).

Now built on the official Java SDK (`org.nightscoutfoundation:nocturne-java` 0.2.4 from Maven Central) rather than bare HTTP. The SDK's generated code is verified compatible with xDrip's pinned OkHttp 3.12.13 / gson by a floor-compile gate in the SDK's publish pipeline (nightscout/nocturne#530), so its OkHttp 4/gson/jackson transitives are excluded here in favor of xDrip's own pinned versions — no dependency changes to the rest of the app. R8 keep rules cover the SDK's gson-serialized models.

Uploads glucose (bulk), calibrations, meter readings, treatments (boluses/carbs/meals/notes/device events, with typed by-sync-identifier deletion), heart rate, step count, motion tracking, and uploader device status. Treatments routing prevents re-upload loops for records that arrived via Nightscout.

The feature is gated to Android 8.0+ (API 26) following the HealthConnect precedent, since the SDK models use java.time and xDrip does not enable desugaring; older devices see an explanatory message and the uploader circuit never activates. The upload-failure retry (6 min) is scoped to the Nocturne queue only.

The previously bundled UiBasedCollector unit fix has been split out to #4648.

33 unit tests cover treatment routing, DTO field mapping, and time/trend conversion.


Navid200:
Why is there a need for this?
What can this offer a diabetic that Nightscout cannot?

Many volunteers have spent years creating, supporting and improving Nightscout.  I see no reason it cannot continue to be improved.

If this conversation has already been had and a decision has been made, I hope that it can be shared with the rest of the community by Nightscout representatives.
  
There is significant value in the title Nightscout.  Even if the decision is that it is not possible to improve Nightscout and it needs to be rebuilt from scratch, there is no reason for the title to change.  
When the new Nightscout is ready for release, we can add a switch to xDrip under the existing Nightscout menu to allow the user to use the new and improved Nightscout.

ryceg:
@Navid200 This might be a discussion better had over a call- happy to organise a one on one, or we can add it to the agenda at the next unified AID meeting (which is iirc the 14th or 15th next month). 

A long story short, the decision to shift past Nightscout has been a long time in the making, and as we're inching towards a full public release of Nocturne, we want to get the supporting applications up to standard. Similar to how there's a Tidepool uploader, there's no reason that this has to supplant Nightscout entirely within XDrip- it would be a bad move to force everyone using XDrip to shift over to Nocturne, and that's explicitly not on our agenda, this is for people that want to move to Nocturne, at their own pace.

Navid200:
jamorham is the one who decides on PRs in this repository.  I ask questions that I anticipate he would have to help move the PR review process forward.  

What exactly is Nocturne?  Is it a replacement for cgm-remote-monitor supported by the Nightscout Foundation?  If so, why should the title Nightscout change?  There is already a Nightscout menu as an uploader option in xDrip.  Why can Nocturne not be placed on that existing menu as an option?

What are the pros and cons of Nocturne?    Also, how many people are currently using it?

From a support perspective, we have to be very careful to avoid confusion.  Every new option increases the volume of questions in xDrip support groups.  If I don't understand the "why" behind a feature, I will not be able to support it.  
Once a change is made to the UI, it will become harder to modify later.

Nightscout can be used for following a loved one in real time.  Tidepool cannot.  
Many clinics accept Tidepool for reporting.  Not many accept Nightscout.
Tidepool and Nightscout are both supported in xDrip because they offer complementary values to the users.  
If Nocturne offers exactly what Nightscout already does, I'm trying to understand the necessity of another integration.

ryceg:
Nocturne is a replacement for cgm-remote-monitor, and is endorsed by the Nightscout Foundation. The title change is because it's a totally different stack, and it's precisely because of community-led support that we are _not_ using the Nightscout name- how confusing would it be needing to specify whether you were on Nightscout 1 v15.0.6 or Nightscout 2 v1.0.6? What happens if we have 15 major versions and end up with Nightscout 2 v15.0.6? A clean break of naming helps avoid a simple category of user error. It also leaves the door open for the original Nightscout to potentially one day have work done on it again.

Nocturne is written with a modern stack and best practice security from the ground up. It is currently in pre-release. Its API is a superset of Nightscout- all Nightscout endpoints are valid Nocturne endpoints, but not all Nocturne endpoints are valid Nightscout endpoints (in practice: we accept v1 - v3 endpoints, but reserve v4 for Nocturne specific endpoints). A separate upload mechanism is necessary because Nocturne accepts data types that Nightscout does not (see: heart rate and steps, included in this PR). 

Nocturne ships with 20+ reports that can be used in a clinical setting, as well as a full node and leaf alert engine, realtime endpoints, statistics, and first party connectors to almost all CGM / pump platforms. It's been built with clinicians and the common patterns of use (visiting an endocrinologist, temporary guest access, allowing limited permissions for school nurse, etc.) in mind.

jamorham:
What are the dependencies of the java sdk (that you're having problems with) ? 

Can you provide some more end user facing information about nocturne, I didn't see anything obvious on the github repository.

app/src/main/AndroidManifest.xml
```diff
@@ -216,7 +216,24 @@
                 <category android:name="android.intent.category.DEFAULT"/>
                 <category android:name="android.intent.category.BROWSABLE"/>
                 <data android:scheme="xdrip"
-                    android:path="/callback/tidepool"/>
+                    android:host="callback"
```
ahaarrestad: This looks like a separate change and should possibly be in a separate PR?

app/src/main/java/com/eveningoutpost/dexdrip/utils/Preferences.java
```diff
@@ -1160,6 +1161,15 @@ public void onCreate(Bundle savedInstanceState) {
             bindPreferenceSummaryToValue(findPreference("cloud_storage_mongodb_collection"));
             bindPreferenceSummaryToValue(findPreference("cloud_storage_mongodb_device_status_collection"));
 
+            try {
+                findPreference("nocturne_connect").setOnPreferenceClickListener(preference -> {
+                    NocturneConnectHelper.startConnectFlow(getActivity());
+                    return true;
+                });
+            } catch (Exception e) {
```
ahaarrestad: Small issue here: this catch block is completely empty, so any failure silently disappears with no log entry. The intent — handling the case where the preference isn't found — is better served by a null check, which avoids accidentally swallowing unrelated exceptions:

```java
Preference connectPref = findPreference("nocturne_connect");
if (connectPref != null) {
    connectPref.setOnPreferenceClickListener(preference -> {
        NocturneConnectHelper.startConnectFlow(getActivity());
        return true;
    });
}
```

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneOAuthService.java
```diff
@@ -0,0 +1,492 @@
...
+                    case "access_denied":
+                        return TokenPollResult.DENIED;
+                    default:
+                        UserError.Log.e(TAG, "pollForToken: unexpected error=" + error);
+                        return TokenPollResult.DENIED;
+                }
+            }
+        } catch (Exception e) {
+            UserError.Log.e(TAG, "pollForToken failed: " + e.getMessage());
+            return TokenPollResult.DENIED;
```
ahaarrestad, line 282: One tricky edge case here: this catch returns `DENIED` for all exceptions, including transient network errors (`SocketTimeoutException`, `UnknownHostException`, etc.). From the user's perspective, they've just approved the device flow in their browser — but xDrip tells them "Authorization denied" because of a momentary connectivity hiccup.

Would it be possible to distinguish network failures from actual OAuth rejections?

```java
} catch (IOException e) {
    UserError.Log.e(TAG, "pollForToken network error, will retry: " + e.getMessage());
    return TokenPollResult.PENDING; // keep polling
} catch (Exception e) {
    UserError.Log.e(TAG, "pollForToken unexpected error: " + e.getMessage());
    return TokenPollResult.DENIED;
}
```

That way a momentary network blip keeps the poll alive instead of killing the session.

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneUploader.java
```diff
@@ -0,0 +1,710 @@
...
+        final Request request = new Request.Builder()
+                .url(baseUrl + path)
+                .header("Authorization", "Bearer " + token)
+                .header("Origin", baseUrl.endsWith("/") ? baseUrl.substring(0, baseUrl.length() - 1) : baseUrl)
+                .post(RequestBody.create(JSON, jsonBody))
+                .build();
+
+        try (Response response = httpClient.newCall(request).execute()) {
+            return response.code();
+        }
```
ahaarrestad: The response body is closed without being read, so when a 4xx comes back the only thing logged is the status code. If the server returns something like `{"error": "field 'mgdl' must be a positive integer"}`, that information is lost — which makes diagnosing payload or auth scope issues very difficult without attaching a proxy.

Logging the body on non-2xx would help a lot:

```java
try (Response response = httpClient.newCall(request).execute()) {
    if (!response.isSuccessful() && response.body() != null) {
        UserError.Log.e(TAG, "POST " + path + " HTTP " + response.code()
            + " body=" + response.body().string());
    }
    return response.code();
}
```

(Same applies to the `delete()` method below.)

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneOAuthService.java
```diff
@@ -0,0 +1,492 @@
...
+            // Token missing or near expiry — attempt refresh if we have a refresh token
+            final String refreshToken = PersistentStore.getString(KEY_REFRESH_TOKEN);
+            if (refreshToken.isEmpty()) {
+                return null;
+            }
+
+            if (!refreshAccessToken()) {
+                // Refresh failed but tokens may still be valid for transient failures
+                final String current = PersistentStore.getString(KEY_ACCESS_TOKEN);
+                return current.isEmpty() ? null : current;
```
ahaarrestad, line 380: Returning the existing token on refresh failure makes sense for transient network blips, but if the token is already expired the upload will get a 401, retry in 6 minutes, get another 401, and so on — indefinitely, with no notification to the user. CGM data quietly stops flowing to Nocturne with no indication anything is wrong.

It might be worth having the upload callers check for a 401 response specifically and surface something to the user — even a log line at warning level saying "Nocturne: re-authorization needed" would help them know to reconnect rather than discover the gap in their data days later.

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneUploader.java
```diff
@@ -0,0 +1,710 @@
...
+        if (Pref.getBooleanDefaultFalse("nocturne_upload_calibrations")) {
+            uploadCalibrations(calibrations);
+        }
+
+        if (Pref.getBooleanDefaultFalse("nocturne_upload_bloodtests")) {
+            uploadBloodTests(bloodTests);
+        }
+
+        if (Pref.getBooleanDefaultFalse("nocturne_upload_treatments")) {
+            uploadTreatments(treatmentsAdd);
```
ahaarrestad: `uploadTreatments()`, `uploadCalibrations()`, and `uploadBloodTests()` all return a `boolean`, but the return values aren't captured here. If any of these fail, `upload()` still returns `true` (based only on `sgvSuccess`), so `UploaderTask` marks those queue entries as completed — meaning the data is permanently dropped without a retry.

Something like this would close the gap:

```java
boolean calibrationSuccess = true;
boolean bloodTestSuccess = true;
boolean treatmentSuccess = true;

if (Pref.getBooleanDefaultFalse("nocturne_upload_calibrations")) {
    calibrationSuccess = uploadCalibrations(calibrations);
}
if (Pref.getBooleanDefaultFalse("nocturne_upload_bloodtests")) {
    bloodTestSuccess = uploadBloodTests(bloodTests);
}
if (Pref.getBooleanDefaultFalse("nocturne_upload_treatments")) {
    treatmentSuccess = uploadTreatments(treatmentsAdd);
    deleteTreatments(treatmentsDel);
}

return sgvSuccess && calibrationSuccess && bloodTestSuccess && treatmentSuccess;
```

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneUploader.java
```diff
@@ -0,0 +1,710 @@
...
+        } catch (Exception e) {
+            UserError.Log.e(TAG, "Error uploading treatments: " + e.getMessage());
+            return false;
+        }
+    }
+
+    private void deleteTreatments(final List<String> uuids) {
+        if (uuids == null || uuids.isEmpty()) return;
+
+        final String[] deletePaths = {
```
ahaarrestad: Looks like `nutrition/meals/by-sync-id` is missing from this array. Meals (treatments where both insulin > 0 and carbs > 0) are uploaded to `api/v4/nutrition/meals` in `uploadTreatments()`, but there's no corresponding delete path — so if a meal is deleted in xDrip, all four paths are tried, none return 204, and the entry is logged as "not found" and marked done. The record remains on the Nocturne server permanently.

Adding `"api/v4/nutrition/meals/by-sync-id"` to the array should cover it.

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneUploader.java
```diff
@@ -0,0 +1,710 @@
...
+
+        if (Pref.getBooleanDefaultFalse("nocturne_upload_heartrate")) {
+            uploadHeartRates();
+        }
+
+        if (Pref.getBooleanDefaultFalse("nocturne_upload_stepcount")) {
+            uploadStepCounts();
+        }
+
+        if (Pref.getBoolean("nocturne_upload_devicestatus", true)) {
```
ahaarrestad: All other Nocturne upload options (`nocturne_upload_sgv`, `nocturne_upload_heartrate`, `nocturne_upload_stepcount`, etc.) use `getBooleanDefaultFalse()`, making them opt-in. `nocturne_upload_devicestatus` is the odd one out — it defaults to `true`, so device status (including battery level) uploads automatically without the user ever explicitly enabling it.

Would it make sense to align this with the rest and change it to `getBooleanDefaultFalse("nocturne_upload_devicestatus")`?

app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/UploaderTask.java
```diff
@@ -188,6 +192,9 @@ public Void doInBackground(String... urls) {
                             SyncService.startSyncService(2000);
                         }
 
+                    } else {
+                        Log.d(TAG, UploaderQueue.getCircuitName(THIS_QUEUE) + " upload failed, scheduling retry");
+                        SyncService.startSyncService(Constants.MINUTE_IN_MS * 6);
```
ahaarrestad, line 199: Worth flagging that this `else` branch runs for *every* failing uploader circuit — Nightscout REST, MongoDB, InfluxDB, and Wear — not just Nocturne. Before this PR, the `retry_timer` flag was always `false`, so no per-circuit retry was ever scheduled on failure. This change quietly enables that retry behaviour for all existing uploaders.

It may well be intentional and a good thing! But since it's a behaviour change for existing users who aren't using Nocturne at all, it might be worth calling out explicitly in the PR description so it gets deliberate attention during review.

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneConnectHelper.java
```diff
@@ -0,0 +1,217 @@
...
+            });
+            builder.setNegativeButton(android.R.string.cancel, (d, w) -> runOnUiThreadSafely(activity, onFinish));
+        } else {
+            builder.setPositiveButton(android.R.string.cancel, (d, w) -> runOnUiThreadSafely(activity, onFinish));
+        }
+
+        final AlertDialog dialog = builder.create();
+        dialog.show();
+
+        new Thread(() -> {
```
ahaarrestad, line 149: The polling thread captures `activity` and `dialog` and holds onto them until `deadline` — which can be up to `expiresIn` seconds (potentially many minutes). If the user rotates the screen or navigates away during this window, the activity is destroyed but the thread keeps running, holding a reference to the dead context.

`runOnUiThreadSafely` guards the posts nicely, but doesn't stop the thread itself. A simple cancel flag that gets set when the dialog is dismissed would let the thread exit cleanly:

```java
final AtomicBoolean cancelled = new AtomicBoolean(false);
// in dismiss/cancel listener: cancelled.set(true);
// in polling loop: if (cancelled.get()) return;
```

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneConnectActivity.java
```diff
@@ -0,0 +1,127 @@
...
+                    .show();
+            return;
+        }
+
+        if (!existingUrl.isEmpty() && !existingUrl.equals(trimmedUrl)) {
+            new AlertDialog.Builder(this)
+                    .setTitle(R.string.nocturne_connect_switch_title)
+                    .setMessage(getString(R.string.nocturne_connect_switch_message, existingUrl, trimmedUrl))
+                    .setPositiveButton(R.string.nocturne_connect_switch_positive, (dialog, which) -> {
+                        new Thread(() -> {
```
ahaarrestad: If `revokeToken()` throws an unchecked exception here (e.g. a `NullPointerException` from `PersistentStore`), the thread dies silently and `connectToInstance()` is never called. The user tapped "Switch" and the activity just finishes with no feedback — they have no idea whether the switch happened or not.

Wrapping the thread body in a try/catch ensures `connectToInstance` is always reached regardless:

```java
new Thread(() -> {
    try {
        new NocturneOAuthService().revokeToken();
    } catch (Exception e) {
        UserError.Log.e(TAG, "revokeToken failed during switch: " + e.getMessage());
    }
    runOnUiThread(() -> {
        if (isFinishing() || isDestroyed()) return;
        connectToInstance(trimmedUrl);
    });
}).start();
```

The same pattern applies to the equivalent thread in `NocturneConnectHelper.java`.

app/src/main/res/xml/pref_nocturne.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ahaarrestad: Just noticed this file seems to duplicate the Nocturne preference block already in `pref_data_sync.xml`, and I couldn't find a `addPreferencesFromResource(R.xml.pref_nocturne)` call anywhere. Is this intended for a future alternative settings screen, or is it safe to remove?

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneUploader.java
```diff
@@ -0,0 +1,710 @@
...
+                          final List<String> treatmentsDel) {
+        if (!ready) return false;
+
+        boolean sgvSuccess = true;
+
+        if (Pref.getBooleanDefaultFalse("nocturne_upload_sgv")) {
+            sgvSuccess = uploadSgv(bgReadings);
+        }
+
+        if (Pref.getBooleanDefaultFalse("nocturne_upload_calibrations")) {
```
ahaarrestad, line 149: Curious about the four data streams gated on `nocturne_upload_calibrations`, `nocturne_upload_bloodtests`, `nocturne_upload_treatments`, and `nocturne_upload_motion` — they don't seem to have corresponding toggles in the settings UI yet, so users can't enable them. Are these placeholders for a future release, or did the toggles end up somewhere I'm not seeing?

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneOAuthService.java
```diff
@@ -0,0 +1,492 @@
...
+import okhttp3.Response;
+import okhttp3.MediaType;
+
+/**
+ * OAuth service implementing Dynamic Client Registration (RFC 7591)
+ * and Device Authorization Grant (RFC 8628) for Nocturne.
+ */
+public class NocturneOAuthService {
+
+    private static final String TAG = "NocturneOAuth";
```
ahaarrestad: Small one — the rest of the project tends to use `ClassName.class.getSimpleName()` for TAG definitions to keep them refactor-safe. Would you be open to aligning the Nocturne classes with that pattern?

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneOAuthService.java
```diff
@@ -0,0 +1,492 @@
...
+                    return null;
+                }
+
+                final JSONObject result = new JSONObject(responseBody);
+                final String clientId = result.getString("client_id");
+                PersistentStore.setString(KEY_CLIENT_ID, clientId);
+                UserError.Log.d(TAG, "registerClient: registered client_id=" + clientId);
+                return clientId;
+            }
+        } catch (Exception e) {
```
ahaarrestad, line 182: One small thing across the catch blocks here — `e.getMessage()` can return `null` for exceptions without a message, which makes the log line read `"registerClient failed: null"`. Something like `e.getClass().getSimpleName() + ": " + e.getMessage()` would make those logs a lot more useful when debugging in the field. There are about six of these in `NocturneOAuthService.java`.

ahaarrestad:
Really appreciate that `NocturneUploaderTest` is included — the routing and mapping tests that are there are well-structured and cover some real regression risks. A few areas caught my eye that might be worth adding coverage for, especially given the medical context:

**`sourcePackageUsesMgDl()`** — the existing tests cover the main paths nicely, but the `.contains(".mmol.")` and `.contains(".mgdl.")` branches don't seem to be exercised. A bug here could produce glucose readings off by a factor of 18, so it feels worth pinning down.

**`mapBgReading()`** — the slope conversions (`* 60000` for trendRate, `* 5 * 60000` for delta) and the `* 1000` scaling on filtered/unfiltered look untested. Since these are values patients use for dosing decisions, would it be possible to make the method package-private like the other `map*` methods so it can be covered directly?

**`isOAuthErrorResponse()`** — this is what decides whether a failed refresh should wipe stored credentials, so getting it wrong (e.g. treating a Cloudflare 403 as an OAuth error) would silently stop uploads until the user manually reconnects. It's a pure function that should be straightforward to unit-test without Android dependencies.

**`getBaseUrl()`** — the http→https upgrade and trailing-slash logic look untested too. Would be good to have a few cases pinned down since a regression here would break uploads for all internet-hosted Nocturne instances.

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneUploader.java
```diff
@@ -0,0 +1,710 @@
...
+        m.put("Pump Battery Change", "PumpBatteryChange");
+        m.put("Pod Change", "PodChange");
+        m.put("Reservoir Change", "ReservoirChange");
+        m.put("Cannula Change", "CannulaChange");
+        m.put("Transmitter Sensor Insert", "TransmitterSensorInsert");
+        DEVICE_EVENT_TYPE_MAP = Collections.unmodifiableMap(m);
+    }
+
+    static TreatmentRoute routeTreatment(final Treatments t) {
+        // Loop prevention
```
ahaarrestad: The `// Loop prevention` comment is clear about *what* this does, but the *why* is easy to miss for someone new to the codebase. Would it be worth expanding it slightly to capture the full picture?

```java
// Loop prevention: skip treatments that originated from Nightscout to avoid
// round-trip cycles (xDrip → Nocturne → Nightscout → xDrip)
```

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneUploader.java
```diff
@@ -0,0 +1,710 @@
...
+
+        if (Pref.getBoolean("nocturne_upload_devicestatus", true)) {
+            uploadDeviceStatus();
+        }
+
+        if (Pref.getBooleanDefaultFalse("nocturne_upload_motion")) {
+            uploadMotionTracking();
+        }
+
+        return sgvSuccess;
```
ahaarrestad: `upload()` returns only `sgvSuccess`, so if a user has SGV disabled but heart rate or step count enabled, the method always returns `true` regardless of whether those uploads actually succeeded. `UploaderTask` then marks all queue entries as completed even on failure.

Would it make sense to have the return value reflect the overall outcome? Something like returning `true` only if all enabled data streams succeeded?

jamorham: @ahaarrestad thanks for the technical review. I'm still waiting for general information about Nocturne from the author to understand whether this is worthwhile being integrated.

ryceg:
> What are the dependencies of the java sdk (that you're having problems with) ? 
> 
> Can you provide some more end user facing information about nocturne, I didn't see anything obvious on the github repository.

XDrip uses okhttp3, while the open API generator uses okhttp4. XDrip also forces findings to 1.3.9 while SDK requires 3.0.2. there's a couple others but those are the main issues. 

https://getnocturne.dev is the end user facing info site, still under construction. If you have a particular query in mind I'm happy to explain.

Navid200:
@ryceg If you are a volunteer developer who enjoys helping the community and provide support for years for an app, and someone brings a new app and requests to replace your app, the community owes you a little bit of consideration.  The community needs to give you a chance to continue providing support for your app.

Many volunteers have put blood and tears into cgm-remote-monitor for years.  Why has the time come to replace it?  
Being the boss is not always as glorious as one may think.  There will be times the boss has to make a choice.   And there are times regardless of which choice the boss makes, there will be pros and cons.  That is really when we need a boss.  There will be people who will focus on the negative outcomes of the decision of the boss.  It is not a great position to be in.

Is there anything I can do to convince you to work with the existing cgm-remote-monitor team?
It is out of the question that you are extremely clever and resourceful.  You have already proven that.  It is obvious you are a very capable individual and can accomplish what you put your mind to.  
The individuals who have worked on cgm-remote-monitor have done nothing unacceptable.  They may have made mistakes.  Who amongst us has never made any mistakes?  
Give them another chance.

xDrip should not be used to push any new tool onto the user community.  Usually, the community hears of a new sensor or tool and they approach us and ask xDrip to support the new sensor or the new tool.
It's never the other way around.  An app developer never comes to us asking us to add support for their app.  It would be very inappropriate on us (xDrip) to add support for a new app that no user has ever asked for.  

If this is an impropriate channel and you would like to have a chat in a different channel, please let me know.
I can even provide an email address if it helps.  

Thanks

TheConen:
I have been using both Nocturne and a self-compiled version of this PR for the past few weeks without any issues - in parallel with Nightscout, as there is no need to immediately replace Nightscout with Nocturne. Both can be used at the same time without any problems - xDrip+ just uploads data to both, same as it would upload data to TidePool or Dex Share in addition to Nightscout.

Nocturne is a HUGE (!!!) improvement over Nightscout. I encourage you to try it out, you can just set it up next to your Nightscout and use both at the same time. I don't think there needs to be a discussion along the lines of "Do we want this" or "Why do we need this" - this has already been decided by the Nightscout Foundation. Nocturne is a project under the umbrella of and endorsed by the Nightscout Foundation.  The Nocturne repository is https://github.com/nightscout/nocturne - right next to the cgm-remote-monitor repository. The Discord channel for Nocturne is on the official Nightscout Foundation Discord. The question for xDrip+ is: Do we want to support what the Nightscout Foundation does or not?

Without this PR, users would not have the option to use Nightscout and Nocturne both at the same time. It would be possible to use the Nightscout uploader to upload data to Nocturne since the Nocturne API is fully Nightscout compatible, but then you could not upload to Nightscout anymore.... not adding a Nocturne uploader forces users to make a choice between Nocturne and Nightscout.

Yes, this uploader is still missing features like uploading treatments and backfilling data - before Nocturne releases v1.0, those should probably be added. But Nocturne is still in development, and the purpose of this PR is to add the foundation needed to support Nocturne once it is released.

Adding Nocturne support is not "xDrip pushing any new tool onto the user community". It is "xDrip giving users the choice of which platform(s) they want to use". There is no need for anyone to replace Nightscout with Nocturne if they don't want to - unless there is no Nocturne Uploader and users are forced to use the Nightscout Uploader to upload data to Nocturne.

Navid200:
@TheConen Is there a public platform where you can link that I can see anyone like Adrian Lxm or Sulka Haro or Ben West or John Costik or Lane Desborough say the time has come to replace Nightscout?

dnzxy:
> Without this PR, users would not have the option to use Nightscout and Nocturne both at the same time. 

Technically, xDrip+ users could just put in their Nocturne instance as the Nightscout instance, since Nocturne accepts all v1 and v3 type requests. You would lose out on providing additional data points that Nocturne accepts and uses, but Nightscout does not.

> @TheConen Is there a public platform where you can link that I can see anyone like Adrian Lxm or Sulka Haro or Ben West or John Costik or Lane Desborough say the time has come to replace Nightscout?

Of all those names, only a single one is still somewhat present and doing work. However, none of those people reside on the board of the Nightscout Foundation, and as an outside person, I'm somewhat curious as to why adding in this uploader is harming the existing project in any way. It just adds an alternative to xDrip+ uploading to Nightscout, just as if there was a Glooko uploader, or you decided to add an uploader to an app like Diabetes Cockpit or Enhance-D. It does not hurt any of the existing projects behind existing uploaders; it just gives users more choices.

TheConen:
Again: there is no need to replace Nightscout, unless xDrip is not adding a Nocturne uploader and forces users to use the Nightscout Uploader to upload data to Nocturne (and therefore losing the ability to upload data to Nightscout). Both projects can coexist unless xDrip forces users to make a choice between the two.

Regarding your question, I'd refer to @ryceg 's initial offer ( https://github.com/NightscoutFoundation/xDrip/pull/4529#issuecomment-4473851363 ) or feel free to talk to the Nightscout Foundation Board on https://discord.gg/HqCWAa7YR

Navid200:
@dnzxy There is no reason to add redundancies in xDrip.
Tidepool cannot be used for following someone in real time.  But, if the only reason for you uploading your data is to have everything in a repository, Tidepool will do.  Then, you have no reason to set up a Nightscout server.  So, Tidepool and Nightscout are not redundant.  xDrip supports Tidepool because there are individuals who are happy by just using Tidepool with no need to pay someone for a Nightscout server or no need to spend time setting up their own Nightscout server.  

But, let's say someone comes along and says I am offering something that does exatly what Tidepool does and I want xDrip to upload to it.  I will then have no choice but to ask, what does it do that Tidepool cannot?  I need to ask that question.  We cannot just keep adding item after item to xDrip with no reason.  

Nightscout is a there because people have supported it and people have needed it.  There has to be a very good reason for adding something to xDrip that offers nothing new.

But, more importantly, everything you can think of that cgm-remote-monitor is missing can be added to it.  There is no reason to reinvent the wheel.  

Who has a Nocturne server that does not have a Nightscout server?  And why?  
What can Nocturne do that Nightscout cannot?

dnzxy:
> […] So, Tidepool and Nightscout are not redundant. 

Neither are Nightscout (cgm-remote-monitor) and Nocturne. 

> Nightscout is a there because people have supported it and people have needed it. There has to be a very good reason for adding something to xDrip that offers nothing new.
> 
> But, more importantly, everything you can think of that cgm-remote-monitor is missing can be added to it. There is no reason to reinvent the wheel.
> 
> Who has a Nocturne server that does not have a Nightscout server? And why? What can Nocturne do that Nightscout cannot?

@Navid200 , I think you should probably get in touch with @ryceg and do that 1on1 and let him show you Nocturne, or read up on Nocturne. I can emphasize with a little skepticism, but it very much does not do the same thing or is the same thing if you look at the feature set. Plus, just because I really loved my 1998 Ford Focus, I'm still driving a new 2024 build car. Both cars took me from A to B, and if you're asking why did I buy a new car if the old one was fine taking me places, then I have a bad time denying you're right. Still, I upgraded, got new features, and a more modern, safer car. Figuratively speaking, I think that's Nightscout vs. Nocturne. (In all transparency: I am writing this currently still only using Nightscout to store diabetes data, as Nocturne is not yet GA and I currently don't have the headspace to help release-test it.)

Navid200:
@dnzxy I am delighted you used that analogy.   

The manufacturers of the cars you pointed out do what they do because they make money.
There is no money here in xDrip.  I am not here to earn a salary.  

Why are you buying a car when you have one?  I really don't care.  It's really not relevant.  There are people who cannot buy even one car and have to take the bus to work everyday.  
That analogy is so wrong on so many levels.  

Do you know how much time supporting an open source app takes?
Forgive me I don't recognize your avatar. May be you are already a contributor.  But, if you are not, you are missing what this is all about.
It takes really hard work to support an open source app.

The least the community can do in return for the time the volunteer developers have  spent is to give them a little bit of respect.   

It takes a long time to fix a bug.  Sometimes, a lot longer than it takes to add a new feature.  

There is no money in this.  It is so misguided to create an unnecessary competition in a domain that is starved with developers.  It is such an unnecessary mistake.  
It makes perfect sense for businesses to compete for your money.  But, when open source apps compete, what are they competing for?  

I keep asking why can you not improve cgm-remote-monitor.  Have you noticed none of you has even attempted to answer that question?

Every one of you is telling me that I have to go somewhere and ask someone to explain this to me?  Why is this hidden?  What is there to be ashamed of?  If the developers of cgm-remote-monitor have committed a crime, let's hear it.   But, if not, why are we treating them like dirt?
How would you like it if we go ahead and support Nocturne and tomorrow someone else comes along and says I have something better?  Do you want us to flush Nocturne down the toilet then just because someone behind closed doors has decided that it is time to replace it?

Do you have the time and knowledge to help us with xDrip?
I can give every single one of you projects to work on and for each one of those projects, I will present a case why the diabetic community needs it and there is absolutely nothing else that is doing what is needed.  
Do you?

dnzxy:
I'm leading developer of Trio, and I've probably worked the equivalent of a full-time job the last ~4 years in the iOS OS-AID space contributing to code, community, traveling to conferences, and furthering the envelope of what our ecosystem here does and provides. Trust me when I say I do understand the work and hardship it takes :) 

The analogy was not to bring _money_ into it, but to compare a system that was once the top of the line, and solved many issues, and made many people happy with a more modern replacement. I intentionally used the word _upgraded_, this is not a money thing. Apologies if you feel like it is. I am as unpaid in my contributions at this point in time as everyone else here is. I could have also used a different analogy – I was just attempting to draw a very simple picture that tries to point out the differences without discarding any progress or achievements made for anyone's project :) 

The basic answer to your question, I think, is that people decided it not worth the resource invest (read: resource is also developer hours) to modernize the cgm-remote-monitor stack and solve incurred tech debt, but rather modernize by approaching a green field. I suspect that as an outside contributor, I have not pushed code to neither NS nor Nocturne. This is just how I am perceiving this development, both from the NSF board and from discussions I have been a part of as listening participant. Having said that, both projects can co-exist; I don't see either just being removed or deleted. 

> Do you have the time and knowledge to help us with xDrip?

Conceptually I can probably provide knowledge, yes. Same for access to a network around industry and research. I commented the other day on your post about making xDrip+ available for MDI users and providing forecasting. I do see the work you do, we work in the same space :) However, I am not an Android developer, nor do I use Android. And ultimately, this is not about _me_, but about this PR, and I am just trying to provide outside arguments as to why co-existing projects and co-existing uploaders for projects make sense to me :) 

I'm not sure, this may be due to translation and text-based communication, but I am a bit surprised by how personal you seem to be taking this. There's no hardship here Navid. We all sit in the same boat, we all got this condition involuntarily, and we are spending our freedom and acquired skills to make diabetes suck a little less for thousands of people around the world.

Navid200:
@dnzxy I hear you.

It becomes personal for me because I cannot help drawing a parallel in my mind with xDrip.  I have spent a lot of time helping to improve xDrip.
It is very easy for someone to look at everything we have done today and come up with something else and present it to the community as an improvement.  
When I see this PR, I cannot help wondering what if someone goes to the AAPS project and opens a PR to replace xDrip without giving us a chance to improve whatever it is that needs  to be improved.

That is basically why this has become personal to me.   

There is so much coming with xDrip.  But, it takes time.  I have been working for months on this:  https://github.com/NightscoutFoundation/xDrip/pull/4525  
It takes time.  But, we are not even close to where xDrip is going.  We are going to fix so many items.  But, we have to do all of this running tests after tests to avoid disrupting systems that are serving the community.  Running all of those tests take time.  It is not easy as you must know.  We need help not competition.  I can only imagine the same is true with cgm-remote-monitor.   They need help not competition.

ryceg:
@Navid200 I think we're talking past each other, so let me try to answer your questions directly and reset the frame.

**This isn't a replacement PR.** Nobody is asking xDrip to drop the Nightscout uploader, and the cgm-remote-monitor project is not being shuttered. This adds an *optional* upload target, exactly like the Tidepool uploader already in xDrip. A user who never touches Nocturne sees no change. That's the whole ask.

**On user count:** I'd gently push back that it's the wrong question for an opt-in uploader. The Tidepool uploader's value isn't measured by how many people use it today either. The right question is "does this harm anyone who doesn't use it?" and the answer is no. It's a toggle, off by default.

**What it does that Nightscout can't, concretely and in this PR:** xDrip already tracks heart rate and step count, but Nightscout does not store them, so that data is dropped. Nocturne ingests both. So this isn't redundant with the Nightscout uploader, it lets xDrip upload data it's already collecting and currently throwing away.

**On endorsement:** Nocturne was presented at the DData conference in New Orleans, alongside the announcement that Omnipod 5 had been cracked. I offered a 1:1 or for you or @jamorham to join the monthly unified AID meeting earlier in this thread, both of which still stand- we often have NSF board members join the call. 

I get that this is a repo you support and field questions on, and that's a real burden. The intent here is to give your users a choice, not to create a support headache or push anything onto anyone.

Navid200:
I appreciate the direct answers and the context regarding Nocturne’s capabilities.  
  
To be clear, my concern isn't about whether Nocturne is a "good" project—I'm sure it is. My concern is about the fragmentation of the community.  
  
You mentioned that Nocturne handles heart rate and step count while Nightscout does not. If you have the skill to build a new uploader and backend, you certainly have the skill to contribute those schema changes to cgm-remote-monitor.  Doing so would benefit the entire existing user base immediately, rather than requiring users to migrate to a new platform to get those features.  
  
In the open-source world, "competition" often leads to a drain on limited resources. When we split the developer pool and the user base across multiple backends, we increase the support burden on those of us managing the front-end apps like xDrip.   Even an "opt-in" toggle becomes a maintenance and support obligation for this repo once it is merged.  

I believe the community is strongest when we consolidate our efforts.  I would much rather see your talents applied to improving the universal standard (Nightscout) than adding another target for us to maintain in xDrip.  

As I do not have approval privileges in this repository, this is my advice to the maintainers: I strongly advise against approving this PR, not because of the quality of your work, but because I believe it is in the best interest of the diabetes community to avoid further fragmentation.

ryceg:
@Navid200 appreciate the thoughtful reply, and I want to engage the fragmentation point seriously because it's the right thing to be thinking about.

On "contribute the schema changes to cgm-remote-monitor instead": that's a reasonable instinct, but the decision to build on a new stack rather than extend cgm-remote-monitor in place wasn't a unilateral call by me. It was the Nightscout Foundation's direction and agreement, on the assessment that modernizing the existing stack's tech debt in place was a larger and riskier job than a clean foundation. @dnzxy described the same thing earlier in this thread. So "just add it upstream" is really a disagreement with the Foundation that hosts both projects, and that conversation is best had with the board directly rather than through this PR. The only commonality that the two tech stacks have is the API. Nocturne had to be a different project because it uses relational data, while Nightscout uses Mongo. 

On the maintenance burden, which I think is the part that actually matters for this repo: you're right that every uploader is a support and maintenance surface, and that's a fair thing for a maintainer to weigh. xDrip already carries this exact category of burden for Tidepool, InfluxDB, Mongo and Wear, though; this is the same shape of obligation, not a new one. We're not creating new data types throughout the application, just POSTing in a Nocturne-native format.

I'd frame the actual decision in front of the maintainers not as "consolidation vs fragmentation" but as "do xDrip users get the choice." Without this, a user who wants Nocturne has to point their Nightscout uploader at it and give up uploading to Nightscout. Given that Nocturne is still not GA, I am uneasy with this idea. With it, they can run both in parallel and decide for themselves. That's strictly more choice for the user and zero change for anyone who doesn't opt in.

Happy to keep answering anything the maintainers need to make the call.

TwistaTim:
Hi @Navid200 ,

Tim Gunn from the Nightscout Foundation here.

Nocturne is a project the foundation is fully supporting going forward.
We started support for the development for it at the HackDiabetes event in Vancouver, Canada in 2025.
There are reasons for us taking this step and very happy to discuss them with everyone.

I warmly invite you to attend https://hackdiabetes.io/ where we will be convening OSS Diabetes Developers from around the world to attend, where we will be discussing this and other developer subjects in future detail and how we all can best support the community.

Anyone reading that has developed or contributed to xDrip/CGM-in-the-cloud/Nocturne, please save the dates Nov 14–16, 2026 and add few days for travel.

We will be opening interest registrations shortly.

Feel free to reach out to me via email:
Screenshot 2026-06-28 at 9 32 01 AM

Kind Regards

Tim Gunn
Nightscout Foundation

jamorham:
@ryceg thanks for engaging. I've managed to view the website now. xDrip is focused on benefits for users and acting as a data hub - so assuming Nocturne works and is something users want to use then it would likely meet the criteria for being implemented.

@ahaarrestad has provided some really good technical feedback on the PR and I would be interested in your responses to those points? There is some work ongoing at the moment to update to okhttp4 so that dependency issue may be resolved.

Due to the age of the PR it has now picked up some other merge conflicts which would also need resolving ahead of time.

With the Oauth authentication is there a situation where the user can get logged out due to inactivity? - I prefer if we can avoid situations like that.

Likewise is there a demo account viewable? I didn't see any full-screen shots on the website to illustrate the advantages of the system.

Changes to UI based collector should be in a separate PR

ahaarrestad:
@ryceg PR to get xDrip to okhttp4: https://github.com/NightscoutFoundation/xDrip/pull/4589

ryceg:
Thanks @ahaarrestad I'll take a looksie at getting this implemented this weekend, building off your PR under the assumption that it'll get merged in :) If I encounter any more dependency-based incompatibilities, I'll submit them as PRs so that we can use the SDK generated by Nocturne rather than the hand-rolled http requests, as it is much less verbose.

Navid200:
Please also provide instructions for setting up a free server so that we can test this.  

This is what we have for Nightscout for reference:
https://nightscout.github.io/nightscout/new_user/#free-diy 

We also have paid options for Nightscout: 
https://nightscout.github.io/nightscout/new_user/#vendors 

There are users who do not have the time or patience or know-how to do it themselves.  There has to be a way for them to still get support.  The paid options are for them.  

It would be nice if you can provide the list of available options for Nocturne.

ryceg:
> Please also provide instructions for setting up a free server so that we can test this.  
> 
> This is what we have for Nightscout for reference:
> https://nightscout.github.io/nightscout/new_user/#free-diy 
> 
> We also have paid options for Nightscout: 
> https://nightscout.github.io/nightscout/new_user/#vendors 
> 
> There are users who do not have the time or patience or know-how to do it themselves.  There has to be a way for them to still get support.  The paid options are for them.  
> 
> It would be nice if you can provide the list of available options for Nocturne.

This is already available at https://getnocturne.dev/docs/installation :)

SeIex:
Hi,

@ryceg 
Just asking, because I'm not sure right now:
The insulinJSON decomposition wouldn't be used to add the insulin type as bolus and basal injection to nocturne, would it?
If so, ignore this and sorry for my question.

At the moment there is no way to upload the JSON stored insulin type treatments anywhere where they will really show up as basal injections.
Old open issues here and at nightscout.

It would be really nice if this would work. ;)
And it looks like you added basal injection support to nocturne allready. :)

ahaarrestad:
Ran the SDK port through some verification. Four things.

**1. The dependency substitution checks out.** I decompiled `nocturne-java:0.2.4` and cross-checked its method references against xDrip's pinned versions rather than relying on the floor-compile gate:

- the okhttp3 calls (71 of them) all resolve against 3.12.13
- the okio usage (ByteString/GzipSink/ForwardingSink etc.) resolves against okio 1.15.0, which ships with okhttp 3.12.13
- gson + gson-fire 1.9.0 resolve against gson 2.8.6
- `com.fasterxml.jackson.core`: 0 references, so excluding it is safe. `JsonNullable` is used heavily though — good that `jackson-databind-nullable` is kept

So the SDK calls no okhttp4-only API. The substitution is sound.

**2. Only an SGV failure triggers a retry — the other streams can be lost.** `upload()` returns `sgvSuccess`, and the upload methods for calibrations/treatments/bloodtests/HR/steps/motion swallow their exceptions and return void. In `UploaderTask`, all queue items are marked `completed()` when `uploadStatus` is true. So if SGV succeeds but a bolus or calibration fails transiently (500/timeout/429), the queue bit is cleared and the record is dropped permanently — no retry.

**3. The Android 8 gating is now dead code.** `isSupported()`, the "requires Android 8" toast and the `&& isSupported()` conjuncts were written against minSdk 24. Master set minSdk 26 in #4601, so `SDK_INT >= O` is always true. It can be dropped to slim the PR down (this branch is 179 commits behind master, so it's probably not visible on your end yet).

**4. Test coverage is uneven.** The routing and DTO-mapping tests are solid. But `mapBgReading` (the SGV mapping, including the slope `*60000` / `*5*60000` and the `*1000` scaling) has no test, and neither does `isOAuthErrorResponse` — the function that decides whether a failed refresh wipes stored credentials. Those two are the highest-risk pure functions in the change; `getBaseUrl` and `NocturneOAuthService` are untested too.

app/src/main/AndroidManifest.xml
jamorham: yes, what is the reason for this change?

jamorham:
There still are two merge conflicts remaining. I'm expecting the okhttp4 PR to be merged soon also, I'm not sure how that affects things here.

jamorham:
The okhttp4 PR is now merged.

Navid200:
Are there options for creating and hosting Nocturne for free?

Please post a screenshot of any user interface page that is new or that you are changing.

Please don't use switches.  Please only use checkboxes.
The tiles do not wrap.  On devices with small screens, long titles may cut out for a switch while still readable for a checkbox.

Why is there a checkbox to enable uploading readings to Noctrune?
What would be the point of xDrip uploading anything to Nocturne if it is not uploading CGM readings?  

Why is mobile data enabled by default?

What is nocturne_upload_motion?

It's best not to enter any strings in a PR.
Every now and then, we add strings to some menus.  
Otherwise, some additional time will have to be spent on reviewing the strings.  

It is my suggestion not to use title case in text.  This is what is done in iOS but not in Android.  
xDrip currently has both.  I would like to gradually sort it out and make it consistent considering xDrip runs on Android.  
For new text, we should not add title case.   

So, I would remove all strings and for the text (even titles), I would stick to sentence case.

ryceg:
Hi Navid,

> Are there options for creating and hosting Nocturne for free?

Yes. You can run it on Oracle for free. 

> Please don't use switches. Please only use checkboxes.

The only switch is the one which turns it on/off. Every other cloud uploader uses that pattern. I don't see a compelling reason to break from that pattern. The Enable Nightscout MongoDB sync already cuts out. 

> Why is there a checkbox to enable uploading readings to Noctrune? What would be the point of xDrip uploading anything to Nocturne if it is not uploading CGM readings?

Uploading other data. A user might prefer XDrip's interface for logging carbs or boluses, I'm not going to prescribe that they must upload glucose data.

> Why is mobile data enabled by default?

 Because the Nightscout upload has it enabled by default. Keeping with the patterns that are already established.

> What is nocturne_upload_motion?

Recognised activity classifications from Google Health.

> It's best not to enter any strings in a PR.

I don't think that hardcoding UI text is a good practice. 

> It is my suggestion not to use title case in text.

Good call, will change.

Screenshots forthcoming :)

Navid200:
There are existing practices on xDrip that will change as we move forward.  Since there are not many developers, it takes time.  That is the nature of open source.
We are improving xDrip continuously.  Just because something has been done in the past, does not mean we should continue.

>  The Enable Nightscout MongoDB sync already cuts out

And that is the reason we are changing the practice.  
Do you have a better suggestion?  Can you suggest a way to make title text wrap?
If not, the best solution I can think of is to avoid switches altogether.  

We will have to review every single text if you leave them as strings.  It is up to you.  It will delay the review process.  It does not mean that we  will never convert them to strings.  But, since we do not have too many developers, which I expect will be the same with Nocturne, it takes time to change the outcome of a practice that we decide to change.

That brings up something else.  I see that one of the options for server is your business.  Isn't that a conflict of interest?  I'm sure Nightscout foundation is aware, are they?
But, don't you think you should have clarified that when you opened this PR?  

What I have suggested is what is best for xDrip.  You are in the xDrip repository.  

Change xDrip+ to xDrip.  

By the way, is Nocturne an option in AAPS?   If not, is there a PR to add it?  

The URL to the server should not be visible on the page showing all the checkboxes.  Users will post screenshots in forums revealing it without realizing.  The  URL should only be visible after tapping on the item.  

Why is there a string for Nocturne?
Wouldn't that just require every translator to add the same word?  Or, do you expect Nocturne to be translated to different words in different languages?

jamorham:
@ahaarrestad Do you think the major review items you mentioned have now been resolved? My main concern is no negative impact on existing features.

ahaarrestad:
@jamorham Re-reviewed at c119de0, built it and ran the unit suite. Three of my four items are resolved, one isn't.

**1. Dependency substitution — resolved.** Master is on okhttp 4.12.0, which is what `nocturne-java:0.2.4` targets. I resolved `fastDebugRuntimeClasspath` on this branch and on master and diffed them: four artifacts added (`nocturne-java`, `gson-fire`, `jackson-databind-nullable`, `jakarta.annotation-api`), none removed, no existing version moved. `com.fasterxml.jackson.core` is genuinely absent from the graph, and `JsonNullable.class` carries no jackson references, so the exclusion holds at runtime and not just at build time.

**2. Silent data loss — fixed.** `upload()` now aggregates across all four queue-driven streams and keeps the watermark-driven ones out of the return value. That is the right split.

**3. Dead Android 8 gating — not fixed.** minSdk is 26, so `isSupported()` is always true. Harmless in itself, but the new comment in `SyncService` justifies it with settings imported from another phone, which can't happen — the app won't install below 26.

**4. Test coverage — partly.** Still my main reservation; below.

**On existing features.** Seven existing files are touched. The manifest change fixes a latent bug — `android:path` is ignored without a host, so the old Tidepool filter matched every `xdrip://` URI; I've replied in that thread. `Treatments.getInsulinInjections()` is only widened to public. The new `UploaderQueue` bit doesn't collide and is set only when the pref is on. In `UploaderTask` the new retry is explicitly scoped to NOCTURNE, so the established uploaders keep their behaviour. `SyncService` gains one OR term and goes via AlarmManager, so it adds no background-start crash path on 12+. `Preferences` gets two null-guarded lookups. proguard, strings and the prefs XML are additions only.

With `nocturne_upload_enable` off, which is the default, none of the new code runs. `:app:testFastDebugUnitTest` and `:wear:testFastDebugUnitTest` are green on c119de0.

**My inline comments.** I re-checked all seventeen. Fifteen are done and I have nothing further on them — @ryceg feel free to resolve those. Two still have something open and I've replied in the threads: the manifest filter, and basal injections having no delete endpoint.

**Tests.** This is my remaining objection. The tests cover routing and DTO mapping; the functions that decide whether patient data is kept or credentials are destroyed have none. In priority order:

1. **`NocturneUploader.upload()` aggregation.** The fix for the data-loss finding has no test of its own. A failing calibration or treatment stream must return `false`, a failing watermark stream must not. That is what stops a transient 500 from permanently dropping a bolus, and it is the exact thing that was wrong before.
2. **`mapBgReading`.** Made package-private as I asked in May, still untested. `trendRate` × 60000, `delta` × 5 × 60000, `filtered`/`unfiltered` × 1000.
3. **`NocturneOAuthService` — 454 lines, no test file at all.** `isOAuthErrorResponse` decides whether stored credentials are wiped, and a Cloudflare 403 with an HTML body must not clear tokens. `getBaseUrl`/`isLocalHost` decide whether OAuth traffic is upgraded to https; getting that wrong either sends credentials in cleartext to an internet host or breaks every LAN instance. `oauthErrorCode` drives the device-flow state machine.

`DeleteOutcome.merge` is tested, and it is exactly the pattern I mean — small, pure, pinned.

app/src/main/AndroidManifest.xml
ahaarrestad: Answering my own question: this isn't a separate change, it's required by this PR, and it fixes a latent bug at the same time.

`android:path` is ignored unless the filter also carries an `android:host`. The old filter had `scheme="xdrip"` and `path="/callback/tidepool"` with no host, so it matched *every* `xdrip://` URI — including the new `xdrip://connect/nocturne` deep link, which would otherwise have landed in the Tidepool callback activity.

Adding `host="callback"` with `path="/tidepool"` narrows it to what it was always meant to match. `AuthFlowOut` uses `xdrip://callback/tidepool`, so the Tidepool flow still matches. No objection from me.

app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneUploader.java
ahaarrestad: The meals half is answered: the code now explains that the server stores a meal as a correlated bolus and carb intake pair sharing the meal's sync identifier, so the two deletes cover it.

What's left is basal injections. The comment says there's no `basalInjectionDeleteBySyncIdentifier` yet, so deleting a long-acting entry in xDrip reports "not found", the queue entry completes, and the record stays on the server permanently. And basal injections go up with suffixed sync identifiers from `suffixedSyncId` (`uuid`, `uuid-1`, …) while deletion only ever tries the bare `uuid`, so the suffixed ones won't match even once the endpoint exists.

Not a blocker, but worth capturing in the code comment so it doesn't get lost when the endpoint lands.

ahaarrestad:
Re-reviewed at `09957bcb1`. Both items I had open are resolved — objection withdrawn.

Gate: `isSupported()` and all five call sites are gone. `minSdkVersion` is 26 in both build blocks, so every one of those conditions was always true — no runtime change. String removed, no orphan translations, nothing in proguard referenced it.

Tests: the decision logic I named is covered — OAuth error classification, `getBaseUrl`/`isLocalHost` boundaries, refresh-token rotation, `mapBgReading` scaling, and `upload()`'s per-stream aggregation including the watermark streams staying out of the return value.

Suite green on the merged head: `./gradlew :app:testFastDebugUnitTest :wear:testFastDebugUnitTest`.

Still open, not a blocker: basal injections can't be deleted, no `basalInjectionDeleteBySyncIdentifier` endpoint upstream. The new comment at the delete loop records the suffixed-sync-id trap for when it lands.

jamorham (approved):
Ok lets give it a try

PR #4533, opened 2026-05-20
mg/dL to mmol/L conversion factor

Navid200:
This PR changes the conversion factor (18.0182) to a more accurate value (18.01559).
The rounding error imposed for showing a value in mmol/L could be significantly greater than the error caused by this minor imperfection.     The absolute error due to this is not the reason for this PR.  The reason is to address a discrepancy between xDrip and Nightscout.

Nightscout (cgm-remote-monitor) was modified in 2021 to use a more accurate conversion factor (18.01559).  xDrip still uses a less accurate value (18.0182).
For two particular values (100 mg/dL and 109 mg/dL), the conversion done by xDrip and Nightscout result in different values.
xDrip shows 5.5 mmol/L while Nightscout shows 5.6 mmol/L for 100 mg/dL.
xDrip shows 6 mmol/L while Nightscout shows 6.1 mmol/L for 109 mg/dL.

Considering the extremely small difference between the two conversion constants, it may seem surprising for this discrepancy.  The discrepancy is caused because the rounded values (after conversion by the two different constants) end up in two different adjacent rounding bins as demonstrated below.

100 / 18.0182 = 5.5499 ~ 5.5
100 / 18.01559 = 5.5507 ~ 5.6

109 / 18.0182 = 6.0494 ~ 6.0
109 / 18.01559 = 6.0503 ~ 6.1  

The following image shows a screenshot of a Nightscout server.
Screenshot 2026-05-20 000354

The following photo shows two xDrip following this same server.  The one on the left is our latest Nightly (2026.05.06).  The one on the right is this PR added to our latest Nightly.
PXL_20260520_040415438

This same difference will appear if a user looks at their Nightscout and xDrip that uploads to it.  The discrepancy only appears if the value in mg/dL is 100 or 109.  
This PR removes the discrepancy.  So, xDrip and Nighscout will show the same values regardless of what the value may be.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Constants.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Constants.java
@@ -4,7 +4,10 @@ package com.eveningoutpost.dexdrip.utilitymodels;
  * Various constants
  */
 public class Constants {
-    public static final double MMOLL_TO_MGDL = 18.0182;
+    // Molar mass of glucose (C6H12O6) = 180.1559 g/mol
+    // Source: NIST Chemistry WebBook, SRD 69 (https://webbook.nist.gov/cgi/cbook.cgi?ID=C50997)
+    // Conversion factor: 1 mmol/L = 18.01559 mg/dL
+    public static final double MMOLL_TO_MGDL = 18.01559; //
     public static final double MGDL_TO_MMOLL = 1 / MMOLL_TO_MGDL;
 
 
--- a/app/src/test/java/com/eveningoutpost/dexdrip/utilitymodels/UnitizedTest.java
+++ b/app/src/test/java/com/eveningoutpost/dexdrip/utilitymodels/UnitizedTest.java
@@ -8,11 +8,11 @@ public class UnitizedTest {
 
     @Test
     public void mmolConvertTest() {
-        assertWithMessage("to mmol").that(Unitized.mmolConvert(99.1)).isWithin(0.001).of(5.5);
+        assertWithMessage("to mmol").that(Unitized.mmolConvert(99.086)).isWithin(0.001).of(5.5);
     }
 
     @Test
     public void mgdlConvertTest() {
-        assertWithMessage("to mgdl").that(Unitized.mgdlConvert(5.5)).isWithin(0.001).of(99.1);
+        assertWithMessage("to mgdl").that(Unitized.mgdlConvert(5.5)).isWithin(0.001).of(99.086);
     }
 }


ahaarrestad:
Looks like a great improvement, and I would like it to be merged.

If you find time, you could add a couple of more conversion tests, in particular for the cases highlighted in the description to verify the expected behavior :-)

jamorham:
Are we absolutely completely sure 1000% that this is the correct constant to use? 

I recall that it was previously changed before after suggestions that the constant we were using was wrong. As far as I know 100mgdl should be 5.5mmol but perhaps the medical profession has been getting this wrong consistently for a long time. Think no matter what we do there will be inconsistencies especially with things like web followers.

Navid200:
I am not very happy and cgm-remote-monitor updated theirs in 2021 without telling us.
AAPS actually is very messed up and is already using something that is also close to what cgm-remote-monitor does.
I agree that one of them may decide to change again.  But, it is my hope that the three projects can work together.

If we merge this, xDrip will be 100% in agreement with cgm-remote-monitor and closer to AAPS.

I can go into detail why this happened years ago and why there are still references pointing to 18.0182.  But, I don't want to add too much detail unless you prefer.

I believe this should be merged now to avoid confusion by users who use xDrip and Nightscout or by users who use xDrip and AAPS.  

All that said, this is not a major error.  But, the discrepancy can be confusing to someone who is not aware of the cause of it.

jamorham:
So I want to get this right and the molecular mass of glucose does make a compelling argument for the constant being slightly off. Would be be correct in thinking that the current conversion constant is incorrect due to rounding errors using the reverse constant of 1/0.555 or similar? The constant we're using is also quite widely cited so I'm not sure if those inaccuracies exist elsewhere already and due to the small difference (not clinically significant) that these just never have been corrected.

Navid200:
The 0.0555 constant itself is not perfectly accurate.  
The accurate value is like Pi.  You have to decide how many decimal points you want to have.
Someone must have decided that 4 decimal points were enough for the conversion constant and came up with 0.0555.

If you take the molar mass of Glucose and invert it, you will get 1 / 180.1559 = 0.0055507479910455.  And that's where the calculator in Windows 11 stopped.  But, it will keep going.  

The accuracy the conversion rate of 18 provides is more than adequate compared to the error we get from the rounding that has to be made to show mmol/L with only one decimal point.  So, xDrip would have been perfectly accurate if we even just used 18.  
The problem is not any inaccuracy in xDrip.  The problem is the inconsistency between xDrip and Nightscout.

So, you may ask how do we know if Nightscout or AAPS are not going to change it again tomorrow?
The answer is we don't.  A perfect solution is to have an overseeing body that would advise all of us (xDrip, AAPS, cgm-remote-monitor, ....) to keep us consistent.  Such a body does not exist.  If you merge this, it is my plan to bring this up with AAPS.  I have already brought it up with cgm-remote-monitor.  
It is my hope, that we all cooperate to avoid this kind of problem.  The problem being xDrip and Nightscout acting an a vernier magnifying a very small error.

jamorham (approved):
Okay lets try it

PR #4534, opened 2026-05-20
Update to support gradle 8

jamorham:
To support newer Android Studio versions and modernize the code base, we need to use gradle 8 because gradle and android studio and related toolchains are tightly coupled.

This PR should do that, but when I first implemented this it caused crashes in certain activities and so before merge I would like others to test that the app is stable because there are some fundamental changes to how it builds the byte code with gradle 8.

I've been running this for some time and I believe all the issues should be resolved and that I've committed all the needed changes. I haven't tested the wear component though.


Navid200:
I will attempt to compile this in an hour.
Thanks

Navid200:
OMG, Panda?  I'm not complaining.  I am pleasantly surprised.

I am running it now.  Collecting from G7 on Android 16 and uploading to Nightscout, broadcasting to AAPS with no problem.
I will be using this now and will notify if I see any issue.

Navid200 (approved):
Looks good to me.
Thanks

Navid200:
I'm still running this.  
No issues.

jamorham:
Ok good, I will merge it soon I think. One thing which crashed for me originally was the emergency message screen. Can you check that is working okay for you?

Navid200:
I just enabled emergency message.
It asked me to allow permission.  I did.  It disabled it.    
I then noticed a note saying add a contact.  So, I did.  I then enabled it.  It stayed enabled.
I tapped on test emergency message.  it sent it.
I received the text message (I had added my own number for the test).

So, I see no problem.

jamorham:
Ok l will merge

PR #4535, opened 2026-05-20
Add tighter y-axis range options for graph display

alanshurafa:
This adds tighter preset values to the existing graph Y Range dropdowns so users can focus the graph on lower glucose bands without changing the current defaults or adding new UI.

The change is intentionally limited to additive resource-array entries while a longer-term input/spinner approach is considered.

Validation:
- Parsed `app/src/main/res/values/arrays.xml` successfully and confirmed the y-axis entry/value arrays still have matching counts.
- `git diff --check upstream/master...HEAD` passed.
- `./gradlew.bat :app:processFastDevResources` was attempted, but the build stopped before resource processing on the existing `:app:mapFastDevSourceSetPaths` / `processFastDevGoogleServices` provider error.

### DIFF
--- a/app/src/main/res/values/arrays.xml
+++ b/app/src/main/res/values/arrays.xml
@@ -503,12 +503,16 @@
     </string-array>
 
     <string-array name="ymin_entries">
+        <item>90 mg/dl (5.0 mmol/l)</item>
+        <item>80 mg/dl (4.4 mmol/l)</item>
         <item>70 mg/dl (3.9 mmol/l)</item>
         <item>60 mg/dl (3.3 mmol/l)</item>
         <item>50 mg/dl (2.8 mmol/l)</item>
         <item>40 mg/dl (2.2 mmol/l) ></item>
     </string-array>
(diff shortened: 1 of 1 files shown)


Navid200:
Thanks for this PR.

We are in the process of making a significant change to the code right now:  https://github.com/NightscoutFoundation/xDrip/pull/4534

After that is settled, which I hope will be done in less than a week or two, I hope to be able to open a PR to improve the Y axis range settings.

Navid200:
After thinking about this more, I see that this is not as simple as I had thought.
We don't force auto-pan.  The user may choose to disable auto-pan.  
If auto-pan is disabled and the user chooses a very narrow range, a very high reading will lead to the range expanding dramatically with respect to the now very narrow regular range.

If the range changes all of a sudden, the user may get a visual illusion that the readings are not changing as fast as they really are.  This could lead to the user not treating themselves by misunderstanding the real slope.

The only way I can think of to avoid this is to offer narrower ranges only if auto-pan is enabled.

Issue #1632, opened 2021-02-01
Dexcom sensor state inconsistently shown


tolot27:
The Dexcom sensor state is not consistently shown at the main screen, menu and status which is likely quite confusing for new users and annoying, especially at sensor session beginnings or ends.

### Your environment

- xDrip version 2101292241
- Samsung Galaxy S4 mini
- LineageOS 15.1-20191216 (Android 8.1.0)
- Hardware data source: Dexcom G6 native

### Expected behavior

Sensor state should be consistent at the main screen, in the menu and at the system status page.

### Actual behavior

If the sensor session is ended, the main screen correctly shows "Stopped":

<img src="https://user-images.githubusercontent.com/1370732/106516111-e3a5b980-64d6-11eb-8250-c8fb735180ee.png" width="300">

But in the menu still "Stop sensor" is shown. 

<img src="https://user-images.githubusercontent.com/1370732/106516797-dfc66700-64d7-11eb-8db9-3289241a55a1.png" width="300">

Before starting a new sensor, one has to stop it in xDrip. If the sensor was already started with the Dexcom receiver and "Stop sensor" was issued (accidentally), it is necessary to clear the transmitter queue (which is currently only possible by entering a voice message or typing it in the treatment dialog).

After a sensor was started using the menu and also sent to the transmitter (queue is empty), the system status page still shows: "Sensor status: stopped" and the main screen "Stopped" next to "Ends in 10 days" and "Please wait while sensor warms up!" on top of the screen:

<img src="https://user-images.githubusercontent.com/1370732/106517906-662f7880-64d9-11eb-903c-aa83a551dee1.png" width="900">

This is really confusing.

emp-00:
I have also noticed this behavior (Samsung A8, Dexcom G6). AFAIK after approx. 3 connections those messages are again consistent on all relevant pages. Nevertheless, I agree that all those messages should show the correct status "everywhere instantly" (if possible).

tolot27:
The sensor start and stop commands should be sent to the transmitter immediately, as well. But this is another issue.

jamorham:
This behavior is in part a hang-over from when we were coexisting both native and non-native and that the concept within xDrip of a sensor session isn't strictly tied to the native session. I don't really think that mode is used anymore and so it is possible to change this but it does have to be very carefully handled to ensure it doesn't break other things.

[...]

PR #4552, opened 2026-06-03
Fix/preserve sensor row on stale stop ack

savek-cc:
This patch fixes the following issue (discussed in https://github.com/NightscoutFoundation/xDrip/discussions/4542):

1. User stops the current sensor session in xDrip. xDrip queues a SessionStopTxMessage. No BLE traffic has happened yet; the command sits in the queue waiting for the  next connection.
2. User starts a new sensor session in xDrip. Sensor.create() closes the previous local Sensors row and inserts a brand-new active row pointing at the new session. xDrip queues a SessionStartTxMessage. At this moment xDrip's local state already considers the new session active.
3. The transmitter eventually receives the queued Stop and ACKs it. Its on-board calibration state moves to Stopped.
4. On a subsequent BLE connection, xDrip reads glucose and gets a calibration state of Stopped. processCalibrationState(Stopped) enters the !is_started && was_started branch.
5. That branch calls Sensor.stopSensor() against Sensor.currentSensor() - which is the row the user just created in step 2. The new local row gets stopped_at set. The transmitter session is unaffected and continues to warm up.

Result: xDrip's local database no longer has an active Sensors row. The main screen reverts to "Please start sensor," nothing is broadcast, no BG values are processed -- even though the transmitter is healthy. The detailed G6 view stays correct because Ob1G5StateMachine tracks the physical session separately from the local table.

The patch fixes step 5: when pendingStart() is true at the moment of step 4, we know the row about to be killed is the new session, not the prior one the transmitter is confirming stopped. So we skip the Sensor.stopSensor() call (and the duplicate notification/treatment) and let the new row stand.

This issue was also described in https://github.com/NightscoutFoundation/xDrip/issues/1632 previously.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
@@ -1988,6 +1988,15 @@ public class Ob1G5CollectionService extends G5BaseService {
 
 
         if (!is_started && was_started) {
+            // A SessionStartTxMessage queued at the moment a stopped-state calibration
+            // is processed means xDrip has already moved on to a new sensor session
+            // locally - whether the Start was queued behind a still-pending Stop, or in
+            // the gap between the connection that delivered the Stop and the one now
+            // reading the transmitter's resulting Stopped state. Either way, the stop
+            // ack we are now processing applies to the prior session, so do not clear
+            // the newly-created Sensor row or emit duplicate stop notifications or
+            // treatments.
+            final boolean staleStopAck = pendingStart();
             if (Sensor.isActive()) {
                 if (Pref.getBooleanDefaultFalse("ob1_g5_restart_sensor")) {
                     if (state.ended()) {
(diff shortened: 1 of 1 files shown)


Navid200:
This is my opinion. I am not writing this on anyone else's behalf.  I don't have merge authority.  
  
---  
  
Thanks for the discussion.  Thanks for spending the time creating a PR instead of just fixing it for yourself.
It is so wonderful to have developers contribute here.  

If you remove a transmitter from a sensor and snap it into another sensor between two cycles, the transmitter has no way of knowing it has moved from a sensor to another sensor with a different calibration code.  The only time the transmitter may react and stop working is if the readings change significantly.  Even then, I am not sure what the trigger is.  

I have a concern about this PR related to that.
xDrip as is stops the local session after a stop is submitted.  It then transmits the stop to the transmitter.  If the user right away submits a start, xDrip creates a queue and will not proceed to start a local session.  
When the current read cycle completes on the transmitter, the transmitter first submits a reading to xDrip.   The transmitter performs this reading while in the new sensor.  xDrip does not have a local session at this point because of the queue.  Therefore, xDrip never shows this last reading.  Neither does it broadcast it to AAPS.  And it does not upload it to Nightscout or Tidepool.  
The transmitter then proceeds to stop its session.  On the next transmission cycle, it acknowledges that it has stopped.  

Your PR addresses an existing problem.  I admit that we do have a problem.
Your PR allows the start command to create a local session.  
The consequence of that is that xDrip will see an active local session when the transmitter submits its last reading while sitting inside a completely new sensor which will require using a different calibration code and has not even warmed up yet.    This reading may be different with the correct calibration code and after the required warmup time.  Using this reading as a valid reading may have a negative impact on glucose control.  So, I prefer the current behavior of xDrip in this respect, rejecting the reading.

I would like to help improve xDrip by building on your work. I feel we should address this side-effect before moving forward. I am happy to collaborate on this, or even do the heavy lifting on the code if you aren't interested in addressing this specific edge case yourself.  

How would you like to proceed?
We can continue our existing email conversation.

savek-cc:
You are mistaken with "If the user right away submits a start, xDrip creates a queue and will not proceed to start a local session."
In fact xDrip DOES start a new local session in its database as soon as you queue a start command.

That behaviour is the reason for this patch because the new session is directly destroyed again when the stop ack is received from the transmitter. No new session is started on a start ack though.

I did not want to modify this general behavior, so the patch is minimal to NOT destroy the local session (created by queueing a start after the stop) if we receive a stop ack while already having queued a start command. No more, no less.

Navid200:
I just tested our current Nightly multiple times.  I am surprised.  I see that it accepts the readings after stop.  But, I also do not see it stop the session after.
It looks like it does exactly what you want.

Have you tested our latest Nightly?

savek-cc:
The stop only happens if you queue a stop, let the stop be sent to the sensor and then directly queue a start in the next window. Not if you do both in one ble window.

jamorham:
This makes sense and looking at the PR I'm thinking it should work. With native mode calibration we don't strictly need xDrip sensor sessions to track actual sensor sessions so I think this probably should be safe for most use cases. I can't think where a problem could result from this.

Main question is whether this has been tested and confirmed to work?

savek-cc:
Running live in my setup for 4 real sensor changes across two persons with both regular and modified G6.

jamorham (approved):

PR #4553, opened 2026-06-03
fix: trust user-installed CA certificates for self-hosted Nightscout

ahaarrestad:
## Summary

- Adds `network_security_config.xml` with `<certificates src="user"/>` so Android trusts user-installed CA certificates
- Wires the config into `AndroidManifest.xml` via `android:networkSecurityConfig`

## Root cause

In May 2024 ([commit 5ac6b761d](https://github.com/NightscoutFoundation/xDrip/commit/5ac6b761d68a51e6c6b09cc052ebe80b0cb936d5)) `targetSdkVersion` was bumped from 23 → 24 to support Android 15. Android's Network Security Configuration policy for `targetSdkVersion >= 24` trusts **system CAs only** by default — user-installed CAs (the kind home Nightscout users add via *Settings → Encryption & Credentials*) are silently excluded.

This is exactly what causes `SSLHandshakeException: Trust anchor for certification path not found` reported in [discussion #4139](https://github.com/NightscoutFoundation/xDrip/discussions/4139).

Evidence:
- `NightscoutUploader` and the Nightscout follow clients use `OkHttpWrapper.getClient()` with no custom `TrustManager` — they fully delegate to Android's NSC policy
- OkHttp 3.12.x respects Android NSC globally (no bypass)
- `ShareRest` (Dexcom Share) has a trust-all implementation but it is scoped to Dexcom Share only and unrelated to this bug
- The Wear module communicates with the phone via Wearable API only — no direct Nightscout HTTPS, so no change needed there

Why some users are unaffected: users running xDrip builds from before June 2024, or with CAs installed in the system store (rooted device / MDM), are not affected.

## Fix

```xml
<!-- app/src/main/res/xml/network_security_config.xml -->
<network-security-config>
    <base-config>
        <trust-anchors>
            <certificates src="system"/>
            <certificates src="user"/>
        </trust-anchors>
    </base-config>
</network-security-config>
```

`src="system"` is kept so existing system CAs are not regressed. `src="user"` re-enables the trust behaviour that existed when `targetSdkVersion` was 23.

**Security note:** Trusting user-installed CAs is opt-in by the device owner (the user must manually install a CA in Android settings). The existing `android:usesCleartextTraffic="true"` already signals an intentionally permissive network posture for a self-hosted-server use case. This change restores the behaviour users relied on before the SDK version bump.

## Verification

This fix is pure Android Network Security Configuration — platform behaviour that only takes effect on a real Android runtime. It **cannot** be meaningfully covered by a JVM unit test (such a test builds its own `SSLContext` and never exercises the app's NSC, so it stays green with or without the fix). Verified instead on a rooted API 34 emulator (`google_apis` image) with a home CA installed into the **user** trust store and a local HTTPS server (`https://10.0.2.2:4443`) signed by that CA. Same emulator, same user CA, same URL — only the APK's manifest differs.

**Without the fix** (no `networkSecurityConfig` ref in manifest) — reproduces #4139:
```
D/OkHttp  --> GET https://10.0.2.2:4443/api/v1/entries.json?count=288
D/OkHttp  <-- HTTP FAILED: javax.net.ssl.SSLHandshakeException:
              java.security.cert.CertPathValidatorException:
              Trust anchor for certification path not found.
```

**With the fix** (`android:networkSecurityConfig="@xml/network_security_config"`):
```
D/OkHttp  --> GET https://10.0.2.2:4443/api/v1/entries.json?count=288
D/OkHttp  <-- 404 File not found https://10.0.2.2:4443/... (1047ms)
```
The TLS handshake now succeeds (the user CA is trusted); the 404 is just the test server lacking that path — i.e. the request got *past* the trust layer, which it never did before.

Without fix
ssl-error

With fix
fixed-ssl

Closes #4139


ahaarrestad:
@jamorham PR is ready for review

jamorham:
This looks like it should work. Thanks for a comprehensive PR.

jamorham (approved):

PR #4566, opened 2026-06-14
Replace legacy IOB model with Oref biexponential insulin curve

p-louis:
## Summary
  - New ExponentialInsulin class implementing the Oref biexponential model, matching the OpenAPS/AAPS formula. Both the activity curve `((s/tau^2) * t * (1-t/td) * exp(-t/tau))` and IOB curve are computed analytically, producing a smooth bell-curve that reaches zero at exactly DIA.
  - Replaces the old piecewise quadratic approximation in Treatments.java for the non-MultipleInsulins path. The legacy calculation used a hard-coded 75-minute peak scaled to a 3-hour DIA with two quadratic segments, and derived jActivity from finite IOB differences between timeslices rather than from a proper activity formula. Both are removed.
  - Enforces a minimum DIA of 300 min (5 h): the AAPS/OpenAPS standard for rapid-acting insulin. The xDrip legacy default of 3 h placed the peak/DIA ratio near 50%, which collapsed the exponential tail and truncated the curve prematurely.
  - Caps the peak at ≤25% of DIA so the exponential decay has room to form a proper tail.
  - Three new exponential insulin profiles added to insulin_profiles.json: FIASP-exp (peak 55 min, DIA 300 min), Novorapid-exp, and Humalog-exp (both peak 75 min, DIA 360 min), usable via the MultipleInsulins path.
  - Routing fix: users who explicitly select a non-exponential insulin profile via MultipleInsulins continue to use the existing calculateIobActivityFromTreatmentAtTime path unchanged.
## Why
The old piecewise quadratic IOB model was a rough approximation ported from early Nightscout/oref0 code. It doesn't respect DIA correctly at values other than 3 h, produces a flat/truncated activity curve, and computes jActivity via backward differences rather than a real derivative — making BGI predictions inaccurate. Aligning with the Oref biexponential model (used by AAPS and OpenAPS) gives a smoother, physiologically more accurate IOB decay and activity curve that correctly integrates to 1 over the full DIA.


ahaarrestad:
Thank you for taking the initiative on this and for putting in the work! It’s a very interesting PR. I have a couple of questions and thoughts:

1. **User Choice / Opt-in**: Would it be possible to make this a user choice initially? (A question for both @p-louis  and @jamorham  / @Navid200). If we make it an "opt-in" feature, it will be less invasive for current users, and it allows us to gather valuable feedback on how it performs before potentially making it the default algorithm?
2. **Unit Tests**: Given that this involves a lot of math and formulas, I think it would be a great idea to add some tests for the new algorithm. Verifying the expected behavior through tests makes the implementation more robust and ensures everything works exactly as intended.

What do you think?

p-louis:
Hey @ahaarrestad 
1. As it stands it is sort of Opt-In - at least when using the multiple-insulin option, one would have to explicitly select the exponential-model variant of the fast-acting insulins in order to activate it. As for the single-insulin default: I'm sure an option could be added to toggle between the two models instead of replacing the linear one.

2. Fair point! Given that the formula behind it is not something I came up with, I'm not entirely sure what would be a good approach for unit testing this. Maybe I should look for 'inspiration' in the AAPS-tests.

Navid200:
Thanks for this work.  

To help put this PR into perspective, I have a question about the real-world impact on the user.

Most people who rely on xDrip's predictive simulations are on MDI (Manual Injections). They use pens or syringes with a resolution of 0.5 units at best. In contrast, a system like AAPS (which does its own math) has pump resolution down to 0.01 units.

If a user is looking at a bolus recommendation or IOB value:
What is the typical 'delta' (difference) this PR introduces?
For a standard bolus—say 5 units—does this change the simulation result by 0.001 units, 0.1 units, or more?

If the math results in a recommendation of 3.142 units, a manual user will still have to round to 3.0 or 3.5. If the PR changes that 3.142 to 3.143, the user experience remains identical.

We can always improve mathematical accuracy, but at some point, we hit diminishing returns. Knowing where that point is for the manual user would help in reviewing the necessity of this change.

p-louis:
Hey @Navid200

That's a fair question to ask, but the differences are considerably bigger than one might expect. 
Say we compare the linear-trapezoid model and exponential model of Novorapid using a dose of 5 IE as you suggested.

We find that the maximum difference in IOB is roughly 1.335 IE at 142 min after the dose was given. Given my profile, that would result in a -BGI in excess of 100 mg/dl

To me the biggest advantage lies beyond simulation results though:
Say I need to go grocery shopping, I'm currently at 120 mg/dl and had lunch about 3h ago with a 5 IE bolus. Given the current model, my IOB would be 0. 
Whereas with the exponential model I'd see that I still have about 1 IE on board. This would tell me that I should eat a snack before heading out or I'd go very low by the time I get to the supermarket.

I hope the above example gives a good idea of where I see the added value in a more accurate IOB simulation - not for the more accurate BG predictions, but the improved situational awareness when making decisions. I'd assume that, generally, having more accurate information is always preferable to a less accurate one.

Navid200:
A discussion so that we avoid turning this PR into a discussion.:
https://github.com/NightscoutFoundation/xDrip/discussions/4570

jamorham:
Can we get on/off graphs for the new behavior to visually show the difference in the curve?

I do favor making significant changes like this something controllable via a user preference so that existing users don't suddenly see a difference unless they opt in and if we are satisfied that new behavior is better then we make that default for new users (and otherwise promote it)

When I originally implemented this (which was a long time ago) I had to do some hacky stuff as I remember to get it to work "right" so we might need to ensure that those items are preserved if needed.

Also worth noting is that I've never seen an insulin curve (from manufacturers) which seems to match actual experience and that also seems to vary on a day by day basis - so simulation is difficult in general. That doesn't mean that this modelling isn't better than what we are currently using.

p-louis:
As requested: here, are the comparison images.
Multiple Insulin-types: Off, Bi-Exponential Insulin Model: On
image
Multiple Insulin-types: Off, Bi-Exponential Insulin Model: Off
image

Multiple Insulin-types: On, Regular Novorapid selected as Bolus-Insulin
image

Multiple Insulin-types: On, Exponential Novorapid selected as Bolus-Insulin
image

I've added the toggle-option in the predictive simulation category towards the end as shown here:
image

Navid200:
Thanks.  Please  wait for more from jamorham with respect to what really matters, the main concept.

However, with respect to the switch, it is not a very good idea to use switches in xDrip because they waste a lot of real estate and on a small screen device, the summary will end up occupying too many lines.  
It's best to always use checkboxes.  Even though switches and checkboxes are presented to address different purposes, for xDrip that we want to be usable on as many devices as possible, checkboxes are superior.  

You can see the existing use Trend Momentum option on the same page.  We could have used a switch for that.  But, we have used a checkbox.  I suggest you use a checkbox as well.  

I suggest avoiding to use capital letters for more than the very first word.  I presented this here: https://github.com/NightscoutFoundation/xDrip/discussions/4408 
We have not been consistent in xDrip and that is what I would like to improve gradually as we move forward as I explained in that linked discussion.
Please note that Apple devices use title case.  Android devices, most of them, use sentence case.  You can look at Android settings on a Pixel device as an example.  xDrip is made for Android.  
  
So I suggest the following as the English title:
Bi-exponential insulin model

For the summary, I suggest:
Use a bi-exponential model to calculate IOB and insulin activity curves

p-louis:
> Also worth noting is that I've never seen an insulin curve (from manufacturers) which seems to match actual experience and that also seems to vary on a day by day basis - so simulation is difficult in general. That doesn't mean that this modelling isn't better than what we are currently using.

That is certainly true, but that may also be related to the sheer number of factors that affect BG - which is also what makes accurate simulation near impossible without having an absurd amount of information available.
I guess my approach differs in that, the BG simulation plays less of a role than COB and IOB calculations. If have somewhat accurate information on these, I can then use that information to draw my own conclusions based on the other information I have available (sleep, stress, physical activity, other medication,...you name it).

Navid200:
I have 3 requests and if I am wrong, @jamorham can correct me.  

1- With respect to the switch you have added, would it be OK to change the string so that it would use sentence case as explained here?
https://github.com/NightscoutFoundation/xDrip/discussions/4408 

2- Can you confirms that after this PR, the xDrip behavior will not change at all unless the user enables the new checkbox?  In other words, every change is gated by a conditional that the new switch blocks unless enabled.

3- Would you mind removing any string translation from the PR?  The translations are done through Crowdin.  I could be wrong.   But, there are translators that can translate.  You may be one of the translators already.  But, doing it this way, you are bypassing the regular process of new strings coming in and all interested translators receiving an automated email informing them, and they having a chance to participate.  I am new to the Crowdin process and could be overlooking something here.  

I am sorry about the delay.

p-louis:
Hey, 

I'm sorry, I should have made it clear that I already changed the switch to a checkbox and the wording to sentence case in my last commit: 
https://github.com/NightscoutFoundation/xDrip/pull/4566/commits/7da398991a2afd5d7c04ffd9f6609854dfb85238
I didn't upload a new screenshot, but I can do that if you want.

As for number 2: Sadly I can't make any guarantees as to whether there are behavior changes in places I'm unaware of, for that I'm not familiar enough with the project. However, I can state that to the best of my knowledge the changed behavior has to be explicitly enabled via either the checkbox or by selecting one of the exponential insulin-types as Bolus insulin when Multiple insulin-mode is enabled.

I'll remove the German strings if keeping them messes with your regular process of providing translations. 

Cheers!

Navid200:
Thanks,

Considering jamorham's concern, we need a way to verify that nothing will change until that switch is enabled.
Let me look at this more carefully.   There may be a way to insert some optional logs to let us see the behavior.
The test can even be a very crude test, which then of course can take time.  For example, one can use the fake data source and enter a set of treatments with our latest Nightly and record IOB.
Then, repeat the test with this version with the switch disabled.
Then, repeat the test with this and with the switch enabled.  

That can be a lot of work.  But, that is something we have to worry about if changing something that could affect people who are already using the app.  There are individuals who just update without paying attention to what the update does.  

Let me se if I can run some of those tests.

Navid200:
I don't mean to delay this.
But, we had this go through: https://github.com/NightscoutFoundation/xDrip/pull/4601
Now, there are some broken things with override silent mode etc. in the current Nightly.  I have to fix those before anything else.
Your work is very valuable.  I will get to testing it soon I hope.  Unless you do the tests yourself.

p-louis:
I can run the test you suggested and provide documentation of the results if that makes your life easier

Navid200:
I'm not sure if what has been requested, and why it is needed, is clear.  I will go into detail.  I hope no one is offended.  That is not the intent.

There are xDrip users who just update when a new update appears in their selected xDrip update channel without reading all the release notes.  

Imagine a user who uses the predictive simulation curves to control their diabetes.
This PR will result in the IOB curve to stay high longer.
A user who experiences this change without realizing that there has been a change to the curves may take extra carbs as they have got used to the IOB curve to diminish fast.  This will result in a disturbance in their control.  

There has been a request by our lead developer to have a switch included for this PR guaranteeing that no change will be made to any curves unless the switch is flipped.  This means the new switch by default must keep xDrip's behavior exactly as it was before this PR:  

> I do favor making significant changes like this something controllable via a user preference so that existing users don't suddenly see a difference unless they opt in  
  

Then, a user who just updates without ever knowing that there is a new feature will never be impacted.  
A user who is aware may choose to keep the switch disabled until their next visit with their doctor.  Then, they may flip the switch and be very careful with their control while they adapt their behavior to the new curve(s).  

You have stated:
  
> As for number 2: Sadly I can't make any guarantees as to whether there are behavior changes in places I'm unaware of, for that I'm not familiar enough with the project.  
  
---  
  
Considering all of that, we have a problem.  
If you cannot guarantee that, we cannot make this change.  Effectively, you are saying you cannot do what our lead developer has asked.  

Even though the acknowledgement a user has to check before they can start using xDrip states xDrip should not be used for dosing decisions, you and I as developers still have a responsibility to keep the app stable and avoid throwing any curve balls to users.  

I suggested that you can add local logs in your working version of xDrip and run tests to convince yourself that there will be no change by this PR if the switch is not flipped.  
If the tests help you discover changes that are being made to the curves without the switch being flipped, you can then edit the PR to add more conditionals to ensure no change will be made until the switch is flipped.
You need to understand xDrip as is and your changes enough so that you feel confident.
  
If you cannot do that, I offered to run those tests myself.  However, that will take a very long time.  I have no idea when I will be able to run those tests myself.  

Again, I was not pedantic here to offend anyone.  I want xDrip to be continuously improved while we do everything we can to keep it safe for all users.

p-louis:
Hey @Navid200,

Now some time has passed and I've had enough opportunity to perform some tests, I can confidently assert that there are no unexpected side-effects introduced by this - at least as far as my tests go.

I do however have a proposal to make for further improvement, that I'd like to run by you:

When not using the multiple insulins option, I use the minimum of 75 and 0.25*the insulin duration setting as the time of the peak activity. This ensures a proper bell curve that doesn't cut off the long tail. However, some users may prefer to set the timing of the peak duration manually to improve the simulation based on their experience. 

The question now is: should I introduce such a setting alongside the one for the insulin duration? 
This would come at the cost of increased complexity and yet another setting.

Cheers

Navid200:
I may be misunderstanding you.  I see no reason for further changes until the concerns we brought up already are addressed.
I am going to repeat what we talked about before.
And before saying anything, I am not the owner of this repository.  

1- Translations are done through Crowdin.  
To follow protocol, a PR should never introduce any strings into any file other than English.
This PR changes strings-de.xml.  That is not good.  

2- I see only one checkbox: exponential_model_enabled 
If this PR is merged, will an existing xDrip user who updates see any difference in behavior if they do not enable that checkbox?
Can you guarantee that nothing will change in behavior as long as that single checkbox is not enabled?

p-louis:
> I may be misunderstanding you. I see no reason for further changes until the concerns we brought up already are addressed.
Fair enough I'll leave that as is for the time being then.

1) The translation I can remove - though I have to admit I find it rather pointless to undo work already done just so somebody else can re-do it...

2) There are 2 paths to enable this feature:
 * the checkbox you've mentioned when not using the multiple insulin option
 * selecting one of the bi-exponential insulin types as bolus-insulin when using the multiple insulin option

In either case a user has to explicitly enable the feature post-update to see any difference in behaviour.

PR #4567, opened 2026-06-14
add glucose offset calibration

luborjurena:
Add a user-configurable glucose offset that shifts the displayed glucose by a fixed amount. The value is entered in the user's units; a positive value raises every shown reading, negative lowers it. Default 0 (disabled).

Some CGMs (Syai, Instara, ...) have no built-in calibration, and from time to time we need to adjust their accuracy.

A ⚠ warning icon is shown before the main-screen value while the offset is non-zero, as a reminder that the value is shifted.

edited_9b55f9a5-f352-4f23-b503-82a5be1ce34a13849555906928206087

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/BestGlucose.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/BestGlucose.java
@@ -258,6 +258,9 @@ public class BestGlucose {
             warning_level = 2;
         }
 
+        // offset calibration: shift the displayed value here (after delta/slope are computed). No-op when unset.
+        estimate = BgReading.applyGlucoseOffset(estimate);
+
         dg.unitized_value = BgGraphBuilder.unitized(estimate, doMgdl);
         final String stringEstimate = BgGraphBuilder.unitized_string(estimate, doMgdl);
         if ((lastBgReading.hide_slope) || (bg_from_filtered)) {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -3131,6 +3131,8 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
                 String stringEstimate = bgGraphBuilder.unitized_string(estimate);
                 currentBgValueText.setText(stringEstimate + (itr == null ? " " + BgReading.activeSlopeArrow() : ""));
             }
+            // offset calibration active: warn that the shown value is shifted
+            if (extrastring.isEmpty() && BgReading.getGlucoseOffsetMgdl() != 0) extrastring = "⚠";
             if (extrastring.length() > 0)
                 currentBgValueText.setText(extrastring + currentBgValueText.getText());
         }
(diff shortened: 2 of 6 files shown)


Navid200:
Thanks for your suggestion and the work put into this PR.

In my opinion, the role of calibration is often misunderstood in the modern CGM landscape.

There was a time when CGMs were in their infancy and the impact of the immune system on the sensor had to be manually addressed via frequent calibration. At that time, sensor accuracy was volatile enough that a blood glucose meter (BGM) was a necessary point of reference.

Nowadays, manufacturers account for the impact of the immune system and physiological changes by embedding time-varying correction factors directly into the sensor's behavior. This is what is known as "Factory Calibration."

Many users are not aware that calibration is no longer as necessary as it once was. It is expected that a CGM and a BGM will not be in perfect agreement, but this is because both devices have a margin of error. When you calibrate, you are not necessarily "correcting" the CGM to the truth; you are simply forcing it to agree with a BGM reading that may itself be inaccurate compared to a lab result. Because the errors of the two devices are uncorrelated and they are measuring different fluids (interstitial vs. blood), you will always see the error fluctuating between positive and negative values.

There are specific scenarios where calibration is often incorrectly used as a "fix":

Compression Lows: If a sensor is being compressed (e.g., sleeping or sitting on it), readings will be consistently lower. The solution is physical (moving the sensor), not a software calibration.

Hardware Faults: If blood or moisture enters the transmitter/sensor contacts, it creates an electrical leakage path. This is a faulty sensor. Calibration cannot fix a hardware failure; the sensor should be cleaned or replaced.

If a manufacturer has designed a sensor to be factory-calibrated and does not provide a mechanism for user calibration, I believe xDrip should respect that design. Introducing a manual calibration mechanism in these cases is error-prone and could make xDrip the weakest link in the user's safety chain.

I respectfully ask you to find a different way to address the underlying concern, as I do not believe manual calibration is the appropriate solution for these sensors.

luborjurena:
I fully agree with you about the factory calibration. I've been talking with few local CGM distributors and patients. The pattern they describe is quite specific. Group of patients for whom the new CGM is otherwise good — trend and reaction time are perfectly fine — but the readings sit constantly shifted.

That case is different. It isn't a compression low (which is intermittent and positional), and it isn't a contact or hardware fault (which is jumpy and unstable). It's a stable, reproducible bias with a faithful trend — which is the one situation where a single fixed offset actually fits, rather than a calibration loop chasing a BGM reading.

Navid200:
https://github.com/NightscoutFoundation/xDrip/discussions/4568

jamorham:
So I don't believe this PR is the way to solve the problem. 

If the value is inaccurate and cannot be calibrated on the manufacturer side then a simple offset is not likely to be reliable either. Instead I think you want to be able to apply a calibration algorithm with slope and intercept to the pre-calibrated values.

There may already be some support for this but also it is quite a risky thing to do so it hasn't received much attention.

luborjurena:
Closing this PR. The discussion has been interesting, but it's clear this won't move forward here.

The feature will live in my fork - those who need it know where to find it.

PR #4569, opened 2026-06-14
okhttp 4 migration — Phase A: characterization-test baseline

ahaarrestad:
> **Phase A of the okhttp 4 migration — ready for review.** This PR adds tests only: no production code and no dependency versions change. It establishes the behavior baseline and the approach to review before the dependency bumps (Phase B) begin.

## Why we are doing this

xDrip's HTTP stack is pinned to **okhttp 3.12.13**. The 3.12.x line is okhttp's old *legacy LTS* branch (kept alive for Android < 5.0 / Java 7) and it is **end-of-life** — it has received no fixes since 2021. We want to move onto a maintained line.

The end goal is okhttp 5.x (the actively maintained line), but we get there in two careful steps. This effort takes us to **okhttp 4.12.0** first, as a safe halfway point across okhttp's big internal change (the 4.x Kotlin rewrite). A later, separate effort will take 4.12.0 → 5.x.

The hard rule for the whole migration is: **no change in observable behavior.** To guarantee that — rather than hope for it — we first write **characterization tests**: tests that pin the *current* behavior on okhttp 3.12.13. Once they are green, every later dependency bump must keep them green, which proves the bump preserved behavior.

## What is in this PR (Phase A)

Six new characterization tests covering the HTTP code paths that previously had **no test coverage**. Each pins the exact current behavior:

| Test | Pins |
|------|------|
| `tidepool/BaseMessageTest` | JSON request body bytes + `content-type` (incl. okhttp's auto `; charset=utf-8`) |
| `cgm/nsfollow/messages/BaseMessageTest` | same, for the Nightscout-follow message base |
| `cgm/webfollow/CmdMediaTypeTest` | the `Cmd.JSON` media-type constant (`application/json`) |
| `cgm/nsfollow/GzipRequestInterceptorTest` | gzip compression, `Content-Encoding: gzip` tagging, and pass-through cases |
| `tidepool/GzipRequestInterceptorTest` | same, for the tidepool copy (package-private class) |
| `tidepool/InfoInterceptorTest` | request and response pass through unchanged |

- **No production code changed** (`app/src/main` diff is empty).
- **No dependency versions changed** — still okhttp 3.12.13 / Retrofit 2.4.0.
- Full `./gradlew test` suite is green.

## What comes after this PR (not included here)

- **B1** — bump okhttp 3.12.13 → 3.14.9 **and** Retrofit 2.4.0 → 2.9.0 together (this is the exact pair upstream tests against).
- **B2** — bump okhttp 3.14.9 → 4.12.0 (the Kotlin-rewrite step), isolated so any regression is easy to bisect/revert.
- **C** — replace deprecated okhttp API calls, one area at a time.
- **D** — remove dead pre-Lollipop TLS-1.2 code (separate, independent).

## Note for reviewers

Incidentally found: `GzipRequestInterceptor` exists as **three identical copies** (nsfollow `public`, tidepool package-private, and an inner class in `NightscoutUploader`). Consolidating them is **out of scope** here, but is a good follow-up cleanup PR — and the tests in this PR would de-risk it.

## Test plan

- [x] `./gradlew test` green (full suite, all flavored variants)
- [x] No changes under `app/src/main`
- [ ] Reviewer: confirm the characterization-test approach is acceptable before Phase B begins

### DIFF
--- a/app/src/test/java/com/eveningoutpost/dexdrip/cgm/nsfollow/GzipRequestInterceptorTest.java
+++ b/app/src/test/java/com/eveningoutpost/dexdrip/cgm/nsfollow/GzipRequestInterceptorTest.java
@@ -1,0 +1,117 @@
+package com.eveningoutpost.dexdrip.cgm.nsfollow;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import com.eveningoutpost.dexdrip.RobolectricTestWithConfig;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.util.zip.GZIPInputStream;
+
+import okhttp3.MediaType;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.RecordedRequest;
+
+/**
+ * Characterization tests for {@link GzipRequestInterceptor} on okhttp 3.12.13.
+ *
+ * @author Asbjørn Aarrestad
+ */
+public class GzipRequestInterceptorTest extends RobolectricTestWithConfig {
+
+    private static final MediaType JSON = MediaType.parse("application/json");
+
+    private MockWebServer server;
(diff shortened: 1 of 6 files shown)


ahaarrestad:
@jamorham part of an effort to update dependencies, but make sure we don't break anything. Step by step approach is safest. This PR does not change anyting, just adding tests.

jamorham (approved):
Tests only, looks like a good place to start

PR #4587, opened 2026-06-29
Tslim sync support

danpowell88:
This is a proof of concept to allow xdrip to interact with Tslim, it needs polishing and feedback but its a start

I've tested it so far on a Pixel 7 Pro running android 17 and a Tslim x2 running Control-IQ v7.8.1.

The integration is one way and currently would not allow bolusing etc without modifications to the code

This was created with the help of AI, I am a developer but don't normally do mobile app development, happy to iterate on this with some feedback and direction from others as to how it should work.

---

Lets xDrip read a Tandem **t:slim X2 / Mobi** pump directly over Bluetooth (via the MIT-licensed [pumpX2] https://github.com/jwoglom/pumpX2) library) and show its **boluses, carbs, basal** and (optionally) **CGM** on the normal graph / treatments / IOB-COB screens

---

### Features
- **Direct, read-only pump link** - pairs with the pump's 6-digit code (JPAKE, firmware v7.7+).
- **Background sync that survives restarts** - runs as a foreground service and reconnects on its own.
- **History import** - boluses, carbs and basal-rate changes flow into xDrip's native stores.
- **Near-real-time updates** - new data appears within seconds while connected (see below).
- **Pick what to sync** - per-type opt-in (boluses / carbs / basal / basal profile / glucose); glucose is **off by default** so it doesn't clash with an existing CGM.
- **Basal mini-graph** - basal on its own graph in real U/hr (details below).
- **Basal profile import** - reads the pump's active basal schedule into xDrip's profile editor.
- **Live Pump tab** - battery, cartridge, IOB, current basal, sensor glucose + trend, Control-IQ  state, total daily insulin, and carb ratio / correction (ISF) / target / insulin duration, shown in  your configured units.

---

### How the sync works
1. **Connect & pair** over Bluetooth using the pump's 6-digit code.
2. **History first, then a small live-status read set** - every request is sent **one at a time**, because the pump drops the link if it's hit with many at once.
3. **Write into xDrip's native stores** (treatments, basal, optional CGM), **de-duplicated** so a re-sync never doubles anything up.
4. **Remember a position cursor** - the first sync pulls a recent window (not the pump's entire log); reconnects then fetch only what's new.
5. **Stay current two ways while connected:** it **reacts to the pump's own change notifications** (so a new bolus / basal change / CGM reading lands within seconds) **and** re-checks every **30 s** as a backstop. Both are rate-limited so the link is never flooded.
6. **Clock anchoring** - if the pump's internal clock is wrong, recent events are re-anchored to the phone's time so they land in the right place on the graph.

---

### Chart changes
These touch the normal xDrip graph for anyone on this build, not just Tandem users:

- **Basal moved onto its own mini-graph above the glucose chart.** Previously basal was drawn as a temp-basal *percentage* on the glucose axis, which put a basal value up near a glucose reading confusing and not in real units. It's now a dedicated graph in **U/hr** with a labelled scale, the rate marked **at each change**, **solid** = delivered basal and **dashed** = the upcoming scheduled profile.
- **A vertical "now" line and a grey-shaded future region** on both graphs, so what has already happened reads clearly apart from what's projected/scheduled.
- **The default view extends a couple of hours ahead** so the next scheduled basal change (and near-future IOB/COB) is visible, while keeping "now" on screen.

Home graph - basal mini-graph, now line and shaded future

---

### The one added screen
Three tabs - **Sync** (connect / pair / status), **Pump** (live read-only values), **Settings** (choose what to pull). Reached from **xDrip → Settings → Experimental → "Tandem pump (read-only)"**.

| Sync | Pump | Settings |
|---|---|---|
| Sync | Pump | Settings |

---

### ⚙️ Requirements & notes
- **Minimum Android raised to 8.0 (API 26)** - required by pumpX2's Bluetooth APIs (was 7.0 / 24).
  `targetSdk` is unchanged (24); the unit-test suite runs at minSdk so it stays green. *This minSdk
  bump is the only app-wide change - flagging it for maintainer review.*
- New dependencies: **pumpX2 v1.9.0**


Navid200:
This is great work.  Thanks.

I have a few observations.

You have raised the min SDK to 26.  But, you have left the target SDK at 24.  Have you done this intentionally?  Were there no warnings during compile?

I have been working on raising both to 26 for over a month now: https://github.com/NightscoutFoundation/xDrip/pull/4525 
That affects the behavior of the collectors as well as the alerts (notification channels).     
It is very challenging.  I will get it done.  But, it is not done yet.  We may have to wait for that to succeed before this PR can be considered.  

With respect to the UI, you have made many changes.  I can see new items on the manus.  But, you are not showing them in the description using Before/After tables.
Such a table will make it easier for the reviewer to see all the changes in one place.

We currently have a top-level menu item titled "Smart watch features".   Under it, there are all watches that xDrip supports. 
Shouldn't we add a new item to the settings menu titled "Insulin pumps" or something like that instead of adding "Tandem pump"?  This will avoid cluttering the top level if more pumps are added in the future.  

A useful part of a PR explains from the user perspective what the PR offers.  It is not the technical part of the PR.  It is the part that answers the question "why do we need this".  
Do I understand correctly that this will allow a Tandem user to not have to constantly look at their Tandem and get the satisfaction they need by just glancing at their xDrip?
They will still need to use their Tandem when they need to make adjustments.  
It is very easy to overlook this because as the developer who has felt the need for something and having worked for a while on it, it is obvious why it is needed.  But, it helps for a reviewer to get that in a description.  Otherwise, that will be the first question they will ask.  

Do I understand correctly that the insulin you added by Tandem will automatically be added to xDrip as bolus and basal treatments?  

When you add graphics to xDrip, you need to state clearly what the trademark status is.  If you have created the graphics yourself, you need to explain what material you have used to ensure there will never possibly be a trademark infringement possibility in the future.   For that reason, the addition of any graphics should be done only when absolutely necessary.  It is best to use text instead and avoid including graphics in menus.  
It is not just copyright that can be a problem.  A symbol can look inappropriate to a user for many different reasons.   We have translators that translate words.  But, the translators cannot translate graphics.  A symbol could have many meanings in different cultures.
Do the best you can to avoid adding graphics.  

Thanks a lot for this work.  
  
---  
  
Please don't take any actions just based on my comments.  I have no approval privileges in this repository.   
I love xDrip and do what I can to help move it forward.  Look at my comments as just that.  Comments by an interested party.  
Thanks

danpowell88:
> This is great work.  Thanks.
> 
> I have a few observations.
> 
> You have raised the min SDK to 26.  But, you have left the target SDK at 24.  Have you done this intentionally?  Were there no warnings during compile?
> 
> I have been working on raising both to 26 for over a month now: https://github.com/NightscoutFoundation/xDrip/pull/4525 
> That affects the behavior of the collectors as well as the alerts (notification channels).     
> It is very challenging.  I will get it done.  But, it is not done yet.  We may have to wait for that to succeed before this PR can be considered.  
> 
> With respect to the UI, you have made many changes.  I can see new items on the manus.  But, you are not showing them in the description using Before/After tables.
> Such a table will make it easier for the reviewer to see all the changes in one place.
> 
> We currently have a top-level menu item titled "Smart watch features".   Under it, there are all watches that xDrip supports. 
> Shouldn't we add a new item to the settings menu titled "Insulin pumps" or something like that instead of adding "Tandem pump"?  This will avoid cluttering the top level if more pumps are added in the future.  
> 
> A useful part of a PR explains from the user perspective what the PR offers.  It is not the technical part of the PR.  It is the part that answers the question "why do we need this".  
> Do I understand correctly that this will allow a Tandem user to not have to constantly look at their Tandem and get the satisfaction they need by just glancing at their xDrip?
> They will still need to use their Tandem when they need to make adjustments.  
> It is very easy to overlook this because as the developer who has felt the need for something and having worked for a while on it, it is obvious why it is needed.  But, it helps for a reviewer to get that in a description.  Otherwise, that will be the first question they will ask.  
> 
> Do I understand correctly that the insulin you added by Tandem will automatically be added to xDrip as bolus and basal treatments?  
> 
> When you add graphics to xDrip, you need to state clearly what the trademark status is.  If you have created the graphics yourself, you need to explain what material you have used to ensure there will never possibly be a trademark infringement possibility in the future.   For that reason, the addition of any graphics should be done only when absolutely necessary.  It is best to use text instead and avoid including graphics in menus.  
> It is not just copyright that can be a problem.  A symbol can look inappropriate to a user for many different reasons.   We have translators that translate words.  But, the translators cannot translate graphics.  A symbol could have many meanings in different cultures.
> Do the best you can to avoid adding graphics.  
> 
> Thanks a lot for this work.  
>   
> ---  
>   
> Please don't take any actions just based on my comments.  I have no approval privileges in this repository.   
> I love xDrip and do what I can to help move it forward.  Look at my comments as just that.  Comments by an interested party.  
> Thanks

Valid points thank is for taking the time!

SDK was just due to requirements in using the linked library for tslim integration, semes to compile fine, but I was a little questionable on that one

I'll update PR with before and after examples. I was thinking maybe insulin pumps should be a menu item above or below insulin pens perhaps since they are vaguely similar in what we interact with them, and tandem could be an option under that perhaps?

Worth considering what we would sync from pumps as well, do we want to sync basal profiles etc and does it get confusing if we allow BG data from pumps but not as a data source like normal cgms, I've added the setting screen as a starter to let the user opt in to what they want to sync for now.

The main point of this PR is to reproduce parts the functionally found in the official tslim app for showing bolus /basal and settings etc and would sync it into xdrip including basal profiles etc, and eventually I'd like to make it available on the http server for use in watch faces to see iob/cob etc.

The user will still need their tandem for dosing, while technically possible to imemwnt dosing I don't want to introduce that responsibility and so far Xdrip seems more around data retrieval than being able to deliver insulin etc.

Bolus and basal will get automatically retrieved from the pump shortly after they are administered and added to xdrip treatments as well as carbs.

Graphics were auto-generated using AI to somewhat be close to a tslim pump look but happy to add / change just wasn't sure what else made sense.

I'll update the main description with more of this detail soon!

jamorham:
Looks interesting. The SDK level changes would have to be coordinated with overall xDrip development because it isn't a simple change.

The changes to the UI graph should be controlled by a preference switch to preserve the existing behavior and the new mode as an option. The home shelf might be a good place to locate that option. Ideally the graph change should be a separate PR.

Navid200:
Is there anything I can do to support you?
I feel the community will benefit from this.

PR #4589, opened 2026-06-29
okhttp 3.12.13 → 4.12.0 (B1 + B2 version bumps)

ahaarrestad:
**WIP / draft** — this is the next step of the okhttp upgrade (after the characterization baseline in #4569). Opening it early so the dependency bump can be reviewed on its own before I start cleaning up the deprecated call sites (Phase C).

## What this does

Two commits, kept separate on purpose so they can be reviewed and bisected independently:

1. **okhttp 3.12.13 → 3.14.9 + Retrofit 2.4.0 → 2.9.0.** This gets us off the okhttp 3.12.x LTS branch, which has been EOL since 2021 and gets no more fixes. I bumped okhttp and Retrofit together because 3.14.9 is exactly the okhttp version Retrofit 2.9.0 is built and tested against — it's the matched pair, so we land on a combination upstream already tests, and I didn't need any `resolutionStrategy.force`.
2. **okhttp 3.14.9 → 4.12.0.** This is the Kotlin-rewrite step, so I kept it in its own commit — if anything ever regresses in the field, this is the one to suspect and revert in isolation. It pulls okio up to 3.6.0 transitively. Everything resolves cleanly to 4.12.0.

4.12.0 is deliberately a stepping stone, not the final destination — it's the last 4.x release and the line is frozen. A later, separate effort takes us to the maintained 5.x line. I'm doing it in two hops because 4.12.0 is the safe halfway point across the Kotlin rewrite.

## The one real behavior decision: TLS

From okhttp 3.13.0, TLSv1 and TLSv1.1 are no longer enabled by default. We don't set a custom `ConnectionSpec` anywhere, so after this PR xDrip offers only TLS 1.2/1.3 in the handshake.

I opened #4583 to check this with the community and we landed on **accepting the new default**. In practice this changes nothing for a normal user: a connection only ever used TLS 1.0/1.1 if the *server's* maximum was 1.1, and those servers have basically died out (and are insecure). Everyone on TLS 1.2+ already negotiates 1.2/1.3 today, so for them nothing changes. The only person who could be affected is someone on an ancient TLS-1.1-only Nightscout server, and the right fix for them is to update that server.

## One test-only change

Retrofit added `Call.timeout()` in 2.6.0, so the anonymous dummy `Call` in `AuthenticatingCallbackTest` had to implement it. I returned `okio.Timeout.NONE`. This is test scaffolding only — no production code changed.

## Tests I ran

The whole point of #4569 was to pin current behavior with characterization tests *before* touching any dependency, so the bar here is simple: the full suite has to stay green through every step.

- Full `./gradlew test` green on the baseline (3.12.13) **before** any change.
- Full `./gradlew test` green after the B1 bump (3.14.9 + Retrofit 2.9.0).
- Full `./gradlew test` green after the B2 bump (4.12.0).
- Verified resolved versions with `./gradlew :app:dependencies` — okhttp 4.12.0 and Retrofit 2.9.0 everywhere, no surprise transitive pulls, no forcing.

One run on 4.12.0 hit a JVM crash — a SIGSEGV in the C2 JIT compiler (`Node::uncast`) while it was compiling a Robolectric-shadowed SQLite method, not anything to do with okhttp. To make sure the bump hadn't somehow made it more likely, I ran the failing test task 10 times on 3.14.9 and 10 times on 4.12.0: 0 crashes on both. So it's a rare JDK JIT flake, with no difference in crash rate between the two okhttp versions.

## Manual smoke test on a real device (okhttp 4.12.0)

The unit tests all run on the JVM (Robolectric), so they don't prove the new Kotlin bytecode (okhttp 4.12.0 + okio 3.6.0) actually loads and runs on Android's ART runtime, and they don't exercise a real TLS handshake. So I ran a quick on-device smoke test of the 4.12.0 build to close that gap.

**Setup:** the `app-fast-debug` APK built from this branch, installed on an Android 14 emulator (API 34). A local Nightscout (cgm-remote-monitor) running over **HTTPS** on the host, reached from the emulator at `https://10.0.2.2:1337`, with a self-signed (mkcert) certificate installed in the device's user trust store. I seeded one SGV reading (120 mg/dL) into Nightscout and pointed xDrip at it as a Nightscout follower.

**Result:** xDrip connected over TLS, pulled the reading, and parsed it — no SSL or runtime/linkage errors anywhere in logcat. With verbose logging enabled on the uploader tag I also got the negotiated TLS version logged explicitly:

```
06-30 09:31:08.875 D/NightscoutUploader: doRESTtreatmentDownload() starting run
06-30 09:31:09.982 V/NightscoutUploader: TLS: TLS_1_3, CipherSuite: TLS_AES_256_GCM_SHA384
06-30 09:31:09.989 D/NightscoutCallback: NS entries download success
06-30 09:31:10.023 D/NightscoutFollowEP: Glucose value: 120
06-30 09:31:10.040 D/NightscoutUploader: doRESTtreatmentDownload() finishing run
```

So on a real Android runtime, okhttp 4.12.0 loads cleanly, negotiates **TLS 1.3** with a modern cipher suite, and the full Nightscout REST round-trip works. The TLS 1.2/1.3-only default from this upgrade is exactly what we want, and it's confirmed working end-to-end. A grep of the full logcat for `SSLHandshakeException`, trust-anchor errors, `NoSuchMethodError`, `NoClassDefFoundError`, `VerifyError` and `AbstractMethodError` came back with zero hits.

(This was the Nightscout follower/download path. I didn't exercise the BG upload POST, because follower-sourced readings aren't re-queued for upload — but that path uses the same okhttp client that's shown above negotiating TLS 1.3.)


jamorham:
Is this the next PR which should be merged on this upgrade path? Note that it has picked up a merge conflict.

ahaarrestad:
This is the next PR on the upgrade path. I will resolve the merge conflict.

ahaarrestad:
@jamorham The merge conflict is resolved and the branch is synced with master. Tests are green. This is ok for merge.

Phase E (#4593) is next after this one.

jamorham (approved):
Thanks for the work on this

PR #4593, opened 2026-07-02
okhttp4 Phase E: consolidate GzipRequestInterceptor into one shared class

ahaarrestad:
## Summary

Consolidates three separate `GzipRequestInterceptor` implementations into one shared class, with **no observable behavior change**. Part of the okhttp4 migration (Phase E — independent of the B1/B2 version bump in #4589; this branch is off `master` on okhttp 3.12.13).

## What was actually there (the premise "3 identical copies" was wrong)

| Copy | Visibility | Behavior | Consumers |
|------|-----------|----------|-----------|
| `cgm/nsfollow/GzipRequestInterceptor` | `public`, `// TODO make reusable` | gzip unless `body == null` or `Content-Encoding` present | 4 cross-package consumers: `SendFeedBack`, `NightLiteClient`, `RetrofitService`, `sharefollow/RetrofitBase` |
| `tidepool/GzipRequestInterceptor` | package-private | byte-for-byte identical to nsfollow | none — only usage was commented out → **dead code** |
| `NightscoutUploader.GzipRequestInterceptor` | inner `static` class | **different** — also gates on `!supportsGzip(host+port)` | `NightscoutUploader` |

So: 2 identical copies + 1 with a per-server capability gate — not 3 identical copies.

## Change

- New shared `com.eveningoutpost.dexdrip.utils.framework.GzipRequestInterceptor` + a small `GzipDecider` functional interface.
  - No-arg constructor → always gzip (byte-for-byte equal to the old nsfollow/tidepool behavior); used by the 4 nsfollow consumers.
  - `GzipRequestInterceptor(GzipDecider)` → per-request gate.
- `NightscoutUploader` now passes `NS_GZIP_DECIDER = req -> supportsGzip(host+port)`, preserving its per-server gzip gating **verbatim** (same id construction, same `PersistentStore` marker `ns-end-supports-gzip-`, same short-circuit order).
- Migrated the 4 production consumers (+ 4 pre-existing test files) to the shared class.
- Removed the dead `tidepool/GzipRequestInterceptor` copy and the dead commented-out usage line in `TidepoolUploader`.

Note: `okhttp3.*` / `okio.*` package names are correct and remain so under okhttp 4.x (4.x keeps the `okhttp3` package). The deprecated `RequestBody.create(MediaType, String)` arg order is intentionally left for Phase C (deprecation cleanup); Phase E actually shrinks Phase C's gzip scope from 3 copies to 1.

## Testing

- New TDD characterization test pins `NightscoutUploader`'s `supportsGzip` gate through the shared class (marker true → gzip; absent → passthrough; GET/no-body and already-encoded passthrough).
- New shared-class test covers no-arg (gzip / no-body / already-encoded) + both decider outcomes — real `MockWebServer` round-trips, gunzip assertions.
- Full `./gradlew test` → **BUILD SUCCESSFUL**, 0 failures.


jamorham:
should this be merged next? Is it safe to do so?

ahaarrestad:
Safe to merge in parallell with the others.

ahaarrestad:
Synced this to master (703395e73) and re-ran the tests, since the branch was written against okhttp 3.12.13 and #4589 has landed since.

Merge was clean, no conflicts. This PR only moves and consolidates the interceptor, #4589 only changed version numbers, so they never touch the same lines.

Full unit suite green. The two relevant test classes ran against okhttp 4.12.0 and mockwebserver 4.12.0:

- `GzipRequestInterceptorTest` — 5 tests
- `NightscoutUploaderGzipInterceptorTest` — 4 tests

Nothing skipped, nothing failing.

The interceptor only uses API that is unchanged in 4.x: `Interceptor`, a `RequestBody` subclass, `Okio.buffer` and `GzipSink`. It never calls `RequestBody.create` or `MediaType.parse`, which are the two that need handling in the deprecation cleanup (Phase C).

jamorham (approved):

PR #4594, opened 2026-07-02
nsfollow: fix Juggluco follower breakage (empty {} entries, devicestatus 400, airplane-mode loopback)

ahaarrestad:
## Problem

Reported in discussion #4592: after 2026.07.01, users following **Juggluco** via the Nightscout Follower (URL `http://127.0.0.1:...`) stopped importing data, with repeating errors:

```
NS entries download Failed: java.lang.IllegalStateException:
    Expected BEGIN_ARRAY but was BEGIN_OBJECT at line 1 column 2 path $
NS devicestatus download was not successful: 400 Bad Request
```

Reverting to the previous release fixed it. The regression came from the nsfollow v1 improvements in #4476.

## Root cause (verified against Juggluco source)

Juggluco ships a *partial* Nightscout emulator on `127.0.0.1:17580` (`Common/src/main/cpp/net/watchserver/watchserver.cpp`). Two behaviours break the newer follower code:

1. **Empty entries → `{}`, not `[]`.** `Sgvinterpret::getdata()` calls `givenothing()` (returns HTTP 200 body `{}`) when no entries match. With the new `find[date][$gt]` filter, the steady state between readings has *no new* entries, so almost every poll returns `{}` and Gson deserialization into `List<Entry>` throws `Expected BEGIN_ARRAY but was BEGIN_OBJECT` → nothing imported. The old count-based query always matched historical readings (non-empty array), which is why the previous release worked.
2. **`devicestatus.json` is not implemented** — the route is commented out, so requests fall through to `wrongpath()` → HTTP 400. This is the "master battery" devicestatus fetch added in #4476.

Additionally, Juggluco binds loopback only, which is reachable in airplane mode even though `ConnectivityManager` reports no active network — so the new network guard skipped polling entirely.

## Changes

- **Tolerate `{}` empty-object entries as an empty list** (`RetrofitService.EMPTY_OBJECT_AS_EMPTY_LIST`, gated to `List` targets; `RetrofitService.getRetrofitInstance`'s only caller is `NightscoutFollow`). No exception on the common empty-poll path; the efficient `find[date][$gt]` query is retained. Real Nightscout returns `[]`, so it is unaffected.
- **Self-healing, per-URL skip of `devicestatus` after HTTP 4xx** (`NsServerCapabilities`): record the rejection and stop polling, but allow one re-probe after a 6h recheck window, clear immediately on a later 200, and reset when the follow URL changes. Never triggers on transient network errors. Uses `JoH.tsl()` for time (project convention; `java.time.Clock` needs API 26+ and minSdk is 24 with no desugaring), with `@VisibleForTesting` `...At(url, nowMs)` seams for deterministic window tests.
- **Loopback-aware network guard** (`NightscoutFollowService.shouldPoll` / `isLoopbackUrl`): loopback targets (`127.0.0.0/8`, `::1`, `localhost`) always poll, even with no active network.

## Testing

TDD throughout. New/updated tests:

- `NightscoutFollowEmptyObjectEntriesTest` — reproduces the exact `Expected BEGIN_ARRAY but was BEGIN_OBJECT` failure, now tolerated.
- `NsServerCapabilitiesTest` — default/sticky/URL-reset + self-heal (re-probe after window, immediate clear on success).
- `NightscoutFollowDeviceStatusFallbackTest` — 400 → sticky skip.
- `JugglucoEmulatorContractTest` — executable documentation of the Juggluco contract (`{}` / array / 400), driven end-to-end through `NightscoutFollow.work()`.
- `NightscoutFollowServiceNetworkTest` — extended with loopback / `shouldPoll`.

Full `./gradlew test` passes.

Fixes #4592

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollow.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollow.java
@@ -20,6 +20,7 @@ import java.util.List;
 
 import okhttp3.ResponseBody;
 import retrofit2.Call;
+import retrofit2.Response;
 import retrofit2.http.GET;
 import retrofit2.http.Header;
 import retrofit2.http.Headers;
@@ -129,15 +130,11 @@ public class NightscoutFollow {
                     }
                 }
             }
-            if (JoH.ratelimit("nsfollow-devicestatus", 5 * 60)) {
+            if (NsServerCapabilities.supportsDeviceStatus(urlString)
+                    && JoH.ratelimit("nsfollow-devicestatus", 5 * 60)) {
                 try {
-                    getService().getDeviceStatus(session.url.getHashedSecret()).enqueue(
-                            new NightscoutCallback<>("NS devicestatus download", session,
-                                    statusList -> {
-                                        if (!statusList.isEmpty()) {
-                                            applyDeviceStatus(statusList.get(0));
-                                        }
-                                    }, null));
+                    getService().getDeviceStatus(session.url.getHashedSecret())
+                            .enqueue(new DeviceStatusCallback(session, urlString));
                 } catch (Exception e) {
                     UserError.Log.e(TAG, "Exception in devicestatus work() " + e);
                 }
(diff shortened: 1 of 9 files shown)


ahaarrestad:
@jamorham discussion in #4592 and fix has been tested and verified by reporter.

jamorham (approved):

Navid200:
Thanks to @ahaarrestad for sorting this out so fast.
Thanks to @jamorham for the rapid review.

Issue #1503, opened 2020-11-11
Notification Channel settings missing on new phone


dbnsacc:
I am replacing my older Pixel 2 XL with a Pixel 5 and I am having an issue with not being able to properly configure the glucose alerts notification channel on the new phone. I transferred my settings using the QR code method and have set my glucose alerts up from scratch the same as the old phone but I am unable to set the channel options as I was on the old phone so that the alarms can vibrate and break through Do Not Disturb mode at night reliably.

I have tried toggling various options inside the alerts section to see if I can get this to appear including changing the Alert Volume Profile to no avail.

I've attached screenshots showing the different options I have on the two phones (both are running Android 11). If I press on the text for the 'bgAlertChannel' notifications on the Pixel 5 it simply toggles on or off.

![xdrip_GANCP2XL_11112020_1302](https://user-images.githubusercontent.com/44261610/98815199-b210db00-241e-11eb-9f71-88a6a94154df.jpg)
![xdrip_GANCP5_11112020_1304](https://user-images.githubusercontent.com/44261610/98815203-b3420800-241e-11eb-985a-109bff32f63d.png)

tolot27:
@dbnsacc I'm a bit lost. Were are the screenshots from? From Android System Settings or xDrip settings? The skin looks completely new to me. Which version of xDrip do you have?

dbnsacc:
These are inside the system settings where you can control the notification channels for each app. On a Pixel you can access the settings quickly by half/short swiping a fired notification and tapping the cog. The skin is essentially as close to stock Android as you get with a Pixel nowadays.

I suspect that the reason I can't adjust the notification channel on the new phone is that I haven't yet found a way to 'fire' that notification type within xDrip+ on it - it always seems to fire the one in my previous posts screenshot with just the vibration symbol (whereas on the old phone at some point in time I must have managed to fire the alternate type with the music note) which exposed the individual settings allowing me to break through Do Not Disturb mode (notification settings aren't exposed until a notification fires in recent versions of Android).

I'm on the very latest version of xDrip+ available on both phones: a91b660-2020.10.01 and attached a screenshot of my system status and one of my configured notifications as an example of how I have them set. I've tried reconfiguring each of these options and test firing them to no avail. I also have the option to speak an alert enabled under Settings > Speak Readings but disabling this has no effect. My volume profile is currently on ascending also but again changing to medium, for example, does nothing.
[...]

Issue #206, opened 2017-10-22
Rewrite History after calibration is only done on master, should be done also on followers


dideldum:
When Rewrite History in the settings is enabled, this is only applied on the master, who adjusts the curve to the calibration value(s).
On the follower this does not happen. Followers thus have a sudden jump in the curve, also showing then completely wrong tendency arrows.
Followers should adjust the curve to calibration values at least, if they are configured in the settings to rewrite history like the master.
Maybe followers should be forced to rewrite history if the master is set to rewrite, regardless the setting in the follower.

TecMunky:
I like the idea of having followers rewrite history when a calibration is performed, and I would prefer that this depend on the setting on each follower device.  This seems a much simpler task than re-sending rewritten data from the sync master to sync followers.

Navid200:
@dideldum Would you please close this issue if you are satisfied?
If not, would you please respond to the suggestion you have received?
Thanks

dideldum:
@Navid200 I'm sorry, that I cannot confirm, that this issue is resolved. It has not been resolved for a very long time, and now I don't use xDrip+ anymore, so I don't know, if the followers handle a rewrite history with a calibration now as it is done on the master.
Nonetheless I close this issue.

TecMunky:
This would still be a useful feature

tolot27:
@TecMunky Do you experiencing the same problem?

Navid200:
@TecMunky Rewrite history is a graphics setting, not a calibration setting.
I don't see how we can (and why we should) have the graphics manipulations on master to be transferred to the follower or vice versa.

Please explain what feature exactly you are talking about.

TecMunky:
When a calibration is executed, the readings for up to the last 3 hours are
adjusted on the Master.  All followers do not see any adjustment of the
data, and thus show different data than the Master.  I would prefer that
followers would also see the same data immediately after a calibration.

This is not a critical issue for me.  Just a minor annoyance.

Navid200:
I don't think we will expand rewrite history or extend it to any other platforms.
https://github.com/NightscoutFoundation/xDrip/issues/366#issuecomment-894306546

PR #4601, opened 2026-07-03
Retiring Android 7

Navid200:
This is a cleaned up version of a previous PR that I have now closed: https://github.com/NightscoutFoundation/xDrip/pull/4525  
<br/>  
  
---  
  
**Why we need this**  
We need to use newer packages and tools.  So much changes with Android 8 and a lot of very important functions will be affected in xDrip.  That's why we have delayed this for so long.   But, I feel the time has come that we should go ahead and retire Android 7.  I am not very happy forcing users to go our and buy phones.  But, I feel we have delayed this long enough.  
<br/>  
  
---  
  
**Changes to the UI and functionality**  

| Before | After |  
| ------- | ------ |  
| Screenshot_20260703-092820 | Screenshot_20260703-105809 |  
| Screenshot_20260703-120032 | Screenshot_20260703-112654 |  
| Screenshot_20260703-093022 | Screenshot_20260703-113650 |  

  
<br/>  

**Why is the notification grouping option removed?**  
Verifications take a very long time.  Every additional option means we have to run the entire verification tests one more time.  
Your alert notification will not be grouped with the ongoing notification.  There may be an exception that I will cover in detail below.  
If your persistent high alert and high alert trigger together, they will be grouped together.  Forcing these two not to be grouped would make this PR considerably more complicated than it already is.  
  
**Why is the option to run the collector in the foreground removed?**  
I don't know why there may be a need to disable this.  Please tell me if you believe there is a legitimate case when you need to disable it.  
  
**Why is the large number icon and arrow in the ongoing notification option removed?**  
These options have not been operational for 6 years: https://github.com/NightscoutFoundation/xDrip/discussions/4590  
If we want to fix this feature and bring it back, we can.  It has to be done in a separate PR.  
  
<br/>  
  
---  
  
**Android 11 issue**  
https://github.com/NightscoutFoundation/xDrip/issues/1503  
This PR happens to fix the problem on my Android 11 phone.  
I will suggest updating to the corresponding release after merge in that issue as a possible solution.   
  
<br/>  
  
---  
  
**Tests**  
  
This has an impact on too many parameters.  It is possible that I have not seen all the impacts.  Therefore, I wish for this to be tested by others as well before we merge it.  
  
We cannot have a release that causes an alert to become silent or use a different custom sound file.  
  
To test the alerts, I have used the fake data source set to run fast.  This would have it go through a full cycle in about 1 hour and 15 minutes.  
A user could have notification channels enabled or disabled right now.  For that reason, I had to test both conditions.  

If a user decides to go back to an older release, their alerts should not go silent or use a different custom sound file.   That has been tested as well.  

I have used Pixel 6a for all of the tests.
I have also run tests with a Motorola Android 11.
I have also run some tests with a Motorola Android 8.

I am using this as my active xDrip on Pixel 10a running Android 17.  It is collecting directly from G7.  It is uploading to Nightscout and Tidepool and broadcasting to AAPS.  
<br/>  
  
---  
  
**Number icon in notification area and grouping**  
Android 16 groups notifications from an app if it decides that the app generates too many notifications.  

This PR avoids that grouping as long as you do not enable "Number icon in notification area".
But, if you enable that feature, Android 16 may still group the xDrip notifications if the frequency is higher than a certain value.  
I can cause this in my tests when I set the fake data source to fast and enable enough notifications to cause about one notification every 10 minutes.
But, the grouping ends if the frequency drops below one alert per 30 minutes.  
<br/>  
  
---  
  
**Preparing the users**  
I will make an announcement on facebook as well as in our discussions.
Users should be aware that after this update, notification channels will be enabled for them and they will have no option to disable them thanks (NOT) to Google.    
This may require some understanding.  I don't want anyone to get the impression that xDrip is an unstable app shifting left and right for no reason.  
So, the announcement will explain to the users what to expect from Notification channels.  
<br/>  
  
---  
  
**Existing problems**  
Google's Notification channels have major issues.  
We may be able to possibly make changes to xDrip to make the experience better.  But, this is something that exists already in xDrip if you enable notification channels.  This PR enables notification channels whether you want it or not.  The resulting problems have been addressed if they had been caused by this PR.  But, if the problems existed already, we can plan and address them in future PRs.  

For example if you change the sound file in any of your other alerts, a new notification channel will be created for you.  But, the previous notification channel will still remain (zombie) and you will have no way of deleting it other than reinstalling xDrip.  This is not something that this PR is causing.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -89,10 +89,10 @@ android {
 
     defaultConfig {
         applicationId "com.eveningoutpost.dexdrip"
-        minSdkVersion 24
+        minSdkVersion 26
         // increasing target SDK version can cause compatibility issues with Android 7+
         //noinspection ExpiredTargetSdkVersion
-        targetSdkVersion 24
+        targetSdkVersion 26
         // change versionCode only when downgrade should be prevented
         // eg, when data structures are incompatible
         versionCode 1603091400
@@ -225,7 +225,7 @@ android {
             // set minSdkVersion to 21 or higher. When using Android Studio 2.3 or higher,
             // the build automatically avoids legacy multidex when deploying to a device running
             // API level 21 or higher—regardless of what you set as your minSdkVersion.
-            minSdkVersion 24
+            minSdkVersion 26
             versionNameSuffix "-dev"
             buildConfigField "int", "buildVersion", "2021010100"
             buildConfigField "String", "buildUUID", "\"0f79a60a-5616-99be-8eb1-a430edcfd9fd\""
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -110,7 +110,6 @@
     <application
         android:name=".xdrip"
         android:allowBackup="false"
-        android:appComponentFactory="android.support.v4.app.CoreComponentFactory"
         android:icon="@drawable/ic_launcher"
         android:label="@string/app_name"
         android:largeHeap="true"
@@ -118,8 +117,7 @@
         android:networkSecurityConfig="@xml/network_security_config"
         android:requestLegacyExternalStorage="true"
         android:theme="@style/AppTheme"
-        android:usesCleartextTraffic="true"
-        tools:replace="android:appComponentFactory">
+        android:usesCleartextTraffic="true">
         <activity
             android:name=".cgm.glupro.GluProActivity"
             android:label="@string/glupro_activity_title"
@@ -332,13 +330,6 @@
             <intent-filter>
                 <action android:name="android.bluetooth.headset.profile.action.CONNECTION_STATE_CHANGED" />
                 <action android:name="android.bluetooth.headset.profile.action.AUDIO_STATE_CHANGED" />
-            </intent-filter>
-        </receiver>
-        <receiver
-            android:name=".wearintegration.ExternalStatusBroadcastReceiver"
-            android:exported="true">
-            <intent-filter>
-                <action android:name="com.eveningoutpost.dexdrip.ExternalStatusline" />
             </intent-filter>
         </receiver>
         <receiver android:name=".utils.DisconnectReceiver">
@@ -640,14 +631,6 @@
         <receiver android:name=".LibreAlarmReceiver">
             <intent-filter>
                 <action android:name="com.eveningoutpost.dexdrip.FROM_LIBRE_ALARM" />
-            </intent-filter>
-        </receiver>
-        <receiver android:name=".NSEmulatorReceiver">
-            <intent-filter>
-                <action android:name="com.eveningoutpost.dexdrip.NS_EMULATOR" />
-                <action android:name="com.eveningoutpost.dexdrip.OOP2_DECODE_FARM_RESULT" />
-                <action android:name="com.eveningoutpost.dexdrip.OOP2_DECODE_BLE_RESULT" />
-                <action android:name="com.eveningoutpost.dexdrip.OOP2_BLUETOOTH_ENABLE_RESULT" />
             </intent-filter>
         </receiver>
         <receiver android:name=".receivers.aidex.AidexReceiver">
--- a/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/Home.java
@@ -3392,11 +3392,10 @@ public class Home extends ActivityWithMenu implements ActivityCompat.OnRequestPe
         new AsyncTask<Void, Void, String>() {
             @Override
             protected String doInBackground(Void... params) {
-                int permissionCheck = ContextCompat.checkSelfPermission(Home.this,
-                        Manifest.permission.READ_EXTERNAL_STORAGE);
-                if (permissionCheck != PackageManager.PERMISSION_GRANTED) {
+                if (ContextCompat.checkSelfPermission(Home.this, Manifest.permission.READ_EXTERNAL_STORAGE) != PackageManager.PERMISSION_GRANTED ||
+                        ContextCompat.checkSelfPermission(Home.this, Manifest.permission.WRITE_EXTERNAL_STORAGE) != PackageManager.PERMISSION_GRANTED) {
                     ActivityCompat.requestPermissions(Home.this,
-                            new String[]{Manifest.permission.READ_EXTERNAL_STORAGE},
+                            new String[]{Manifest.permission.READ_EXTERNAL_STORAGE, Manifest.permission.WRITE_EXTERNAL_STORAGE},
                             0);
                     return null;
                 } else {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/NSEmulatorReceiverWrapper.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/NSEmulatorReceiverWrapper.java
@@ -1,0 +1,32 @@
+package com.eveningoutpost.dexdrip;
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+
+/**
+ * Lightweight manifest-registered receiver.
+ *
+ * This keeps the registered receiver tiny until a broadcast is actually received,
+ * then lazily creates and reuses the real receiver.
+ */
+public final class NSEmulatorReceiverWrapper extends BroadcastReceiver {
+
+    private static volatile NSEmulatorReceiver receiverInstance;
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        getReceiverInstance().onReceive(context, intent);
+    }
+
+    private static NSEmulatorReceiver getReceiverInstance() {
+        if (receiverInstance == null) {
+            synchronized (NSEmulatorReceiverWrapper.class) {
+                if (receiverInstance == null) {
+                    receiverInstance = new NSEmulatorReceiver();
+                }
+            }
+        }
+        return receiverInstance;
+    }
+}
--- a/app/src/main/java/com/eveningoutpost/dexdrip/models/JoH.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/models/JoH.java
@@ -1258,15 +1258,9 @@ public class JoH {
 
     public static boolean setMediaDataSource(final Context context, final MediaPlayer mp, final Uri uri) {
         try {
-            if (uri.toString().startsWith("/")) {
-                UserError.Log.d(TAG, "Setting old style uri: " + uri);
-                mp.setDataSource(context, uri);
-            } else {
-                UserError.Log.d(TAG, "Setting new style uri: " + uri);
-                val pfd = context.getContentResolver().openFileDescriptor(uri, "r");
-                mp.setDataSource(pfd.getFileDescriptor());
-                pfd.close();
-            }
+            if (uri == null || mp == null) return false;
+            // Use the context-aware method for all URIs to satisfy Android 11+ requirements
+            mp.setDataSource(context, uri);
             return true;
         } catch (IOException | NullPointerException | IllegalArgumentException | SecurityException ex) {
             UserError.Log.e(TAG, "setMediaDataSource from uri failed: uri = " + uri.toString(), ex);
--- a/app/src/main/java/com/eveningoutpost/dexdrip/ui/NumberGraphic.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/ui/NumberGraphic.java
@@ -9,12 +9,12 @@ import android.graphics.Paint;
 import android.graphics.Rect;
 import android.graphics.Typeface;
 import android.graphics.drawable.Icon;
-import android.os.Build;
 
 import com.eveningoutpost.dexdrip.models.JoH;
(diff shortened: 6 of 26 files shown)


Navid200:
**Android Notification channels**

If you enable notification channels in xDrip before this PR, the xDrip notification channels appear at the top.
Screenshot_20260704-223940  
  
After we merge this PR, if you have been using xDrip already with notification channels enabled and update, some notification channels will remain at the top and some will appear under "Other".  
Screenshot_20260704-183514  


After we merge this PR, if you install xDrip for the first time or after uninstall, notification channels will appear under "Other".  
Screenshot_20260704-210017


Regardless, the alert behavior will be retained going through an upgrade or going to a release before this PR.

ahaarrestad (commented):
Thanks again for all the work here, Navid — really solid cleanup overall. I left a few inline questions below, all meant as things to sanity-check together rather than blockers. 🙂

ahaarrestad:
I built and traced this branch. One thing in `ExternalStatusService` looks like a real regression (not style):

Off `IntentService`, `onStartCommand` runs on the **main thread** and calls `update()` synchronously. `update()` hits the DB on that thread — `APStatus.last()` (a `Select…executeSingle()`), `Profile.getBasalRate()`, and `APStatus.save()` when the value changed — plus the regex parsing. The old `onHandleIntent` did all this on `IntentService`'s worker thread. So it's now synchronous SQLite I/O on the UI thread on every AAPS status broadcast: StrictMode disk/DB-on-main-thread, UI jank, and ANR risk if the DB is contended.

One subtlety: `IntentService` processed intents **sequentially** on a single worker thread, so the fix should stay single-threaded too — otherwise overlapping broadcasts could race `createEfficientRecord()`'s read-then-write. This keeps it off the main thread while preserving that ordering (compiles clean here, `compileFastDebugJavaWithJavac`):

```java
private static final ExecutorService executor = Executors.newSingleThreadExecutor();

@Override
public int onStartCommand(Intent intent, int flags, int startId) {
    if (intent != null) {
        final String action = intent.getAction();
        if (ACTION_NEW_EXTERNAL_STATUSLINE.equals(action)) {
            final String statusLine = intent.getStringExtra(EXTRA_STATUSLINE);
            // onStartCommand runs on the main thread; keep the DB write + parsing off it.
            // Single-thread executor preserves the old IntentService's sequential processing.
            executor.execute(() -> {
                update(JoH.tsl(), statusLine, true);
                stopSelf(startId);
            });
            return START_NOT_STICKY;
        }
    }
    stopSelf(startId);
    return START_NOT_STICKY;
}
```
(adds `import java.util.concurrent.ExecutorService; import java.util.concurrent.Executors;`)

If useful I can open a small PR against your branch with this — otherwise feel free to grab the snippet.

Navid200:
@ahaarrestad Fantastic!  
Thanks a lot for catching that.  
And thanks for the fix.

Navid200:
@ahaarrestad I would like to understand your offer.
Would you be open to testing this?
Do you mean you take this, correct it and open a full PR to contain the correction you have suggested?
That would be wonderful because then you will be able to test it too and even can make any other corrections it may need.

Considering your experience, that would make this a robust solution.
I am very worried about this PR breaking something.
So, if that is what you suggest and if you have the time to test, that would be wonderful.

But, it is possible I have misunderstood.  Please don't feel obligated that you now have to spend time on this.

I worked on this because it needs to get done.  But, this is an extremely complicated PR and it would feel really good if it can benefit from your verification and improvements.

ahaarrestad:
Yes — I'll open a PR against your branch with the executor fix and test it on an emulator (StrictMode on) to confirm the DB write is off the main thread before you merge. I'll keep it to just this fix. Give me a day or two.

Navid200:
@ahaarrestad Thanks

With respect to the lazy thing, if you attempt to post a notification, unless there is a notification channel for it, Android will just ignore it unlike with SDK 24.  So, I had to make changes.

Android 16 groups notifications from an app if it is more than so many per hour.  This is fine for many apps.  But for xDrip, we cannot have the ongoing notification grouped with an alert.  When the user pulls down from the top to see which alert is triggered and to snooze it, he will not see either.  That's why I have had to make even more changes.

Testing alerts, unfortunately, takes a very long time.  You have to customize the alert sounds and let them all trigger and create notification channels, then update to this PR and wait for every alert to trigger again and make sure the sound still reflects the chosen sound.

I cannot test OOP.  I don't know how.  Neither can I test WearOS.  

You also made a comment about a follower may not work right away.  I didn't notice an issue with followers that I tested, Nightscout and xDrip Sync.  But, that would be a show-stopper.  So, it would be great if you could advise about how to fix that as well.

Thanks a low

ahaarrestad:
Two things:

**The main-thread fix** — opened as a PR against your branch: Navid200/xDrip#206. It moves the status update onto `BackgroundQueue` (xDrip's own sequential background processor). Verified on an API 34 emulator that the `APStatus` insert now runs on a background thread. Please build/test it and adjust as you see fit.

**On the follower question** — I dug into it and couldn't reproduce a breakage. In normal follower operation the ongoing channel is present (it's even pinned by the active foreground service, so the system won't let you delete it). And when I forced the channel to be genuinely missing on API 34, `startForeground` doesn't crash — the system just silently drops that one foreground notification, and the service keeps running and records the status. So it's not a show-stopper. If you want the belt-and-suspenders version, a one-line `NotificationChannels.getChan(builder)` in `ExternalStatusService.onCreate` guarantees the notification always shows, but it's cosmetic, not a bug fix.

ahaarrestad:
On WearOS: I paired a phone + Wear OS 3 emulator (via Android Studio's pairing assistant) and ran your branch with the #206 fix on top, end-to-end. Your notification-channel changes do execute in this path — I can see your `NotificationChannels` creating the ongoing, icon-test and hashed alert channels, and `XdripNotification` posting on `ongoingChannel` — while a Nightscout follower feeds real BG all the way through to the watch (confirmed a reading in the watch's DB, source "Nightscout Follow"). No crashes. Your PR doesn't touch the wear module itself, so this is really a regression check that your app-side channel/foreground changes don't break the phone→watch pipeline — and they don't.

ahaarrestad:
I've been putting your branch through its paces on phone + Wear OS emulators — with a Nightscout follower feeding real BG so I could actually exercise the alert and notification paths. Summary: it holds up well.

- **Notification channels**: alert channels come out `IMPORTANCE_HIGH` and ungrouped, and the ongoing channel is `LOW` + silent + no badge — exactly as intended. Both sound models work: the main High/Low alerts play through `AlertPlayer` (channel stays silent), while the "other" alerts (persistent-high etc.) and calibration set the sound on the channel, and changing the sound still maps to a distinct channel (your hash logic).
- **Removed notification prefs** (`use_notification_channels`, `run_service_in_foreground`, grouping, number-icon-large): no leftover references that would crash the settings screens; `legacySettingsFix` pins them to the right values.
- **ForegroundServiceStarter**: the collector always moves to foreground now — verified with a follower, which was the old edge case — with no crash on API 34/36.
- **IdempotentMigrations**: your additions follow the existing "force legacy settings" pattern.
- **Phone → watch**: full NS-follow → phone → Wear data-layer sync works end to end (a reading landed on the watch, sourced from "Nightscout Follow").

Everything I threw at it behaved. This is a big change and it's holding together well.

ahaarrestad:
On the OOP side — the one thing your PR touches there is moving `NSEmulatorReceiver` (which handles the `OOP2_DECODE_*` result broadcasts) from the manifest to a dynamic registration in `xdrip.onCreate`. I tested that decode-result delivery still works: I sent a synthetic `OOP2_DECODE_BLE_RESULT` broadcast with a crafted `json` payload, and the dynamically-registered receiver picks it up and hands the full payload to the BLE handler (`extractParams` parses it, then it reaches the normal `ROW_ID` correlation check). So the manifest→dynamic move doesn't break OOP2 reception while xDrip is running — and the receiver's own decode code is untouched, so nothing else there changes.

Navid200:
I have run my notification tests and they work fine.
I rebooted my Pixel 10a that is running this as a collector from G7 and after reboot, unlocked the sim card and unlocked the screen.  I did not bring up xDrip.
I monitored Nightscout in a browser.  Nightscout received an upload from xDrip on this phone on schedule meaning the collector woke up automatically on the scheduled wake.

jamorham:
So this looks like it should work.

I'm not sure you needed to move the receivers from the manifest because if they're directed towards the package explicitly then those still work.

The advantage of having them in the manifest is no memory is consumed when not in use and the app can be woken up by receiving a broadcast. I'm not against them being registered at runtime but I wonder if its worth adding a lightweight frontend class which instantiates the bigger class if it actually receives a broadcast.

One of my main concerns with android 8 is how foreground services can be blocked from starting from the background. This is actually quite tricky to test. One test for example is to reboot the phone and don't open xDrip at all but then observe in logcat whether the collector service starts correctly and doesn't get shut down by the system. For collection services I think this should be covered correctly by the PR but for anything else using ForegroundServiceStarter I am not 100% sure if it will work.

Navid200:
@jamorham Thanks for the review.  
I am right now looking into the manifest issue and will get back.

But, I have tested the reboot already.  My xDrip collects directly from a G7 and uploads to Nightscout.  So, I don't need to look at logcat.  I rebooted the phone.  After the reboot, I only unlocked the phone.  But, I never opened xDrip.
I then opened Nightscout in a browser on my computer.  And I could see that a new reading arrived exactly 5 minutes after the previous one while I still had not opened xDrip.  
This tells me that xDrip without being opened uploaded a new reading to Nightscout.  The only way my xDrip could have a new reading was to have successfully brought up the collector.

Navid200:
@jamorham 
Are all broadcasts directed specifically to xDrip, AAPS, BYODA, Tasker, ...?  I don't know.  Are we sure about all?  For all I know there are some that we may not even know about!

@ahaarrestad fixed a problem already: https://github.com/NightscoutFoundation/xDrip/pull/4601/changes/9be621e5077a35f744f8399d2c7a01b1ba4a1220 
Isn't that now effectively a lightweight front end as you have suggested?

Navid200:
This is what I have been running now consistently as my main active xDrip for almost a week.  I see no change in the phone battery drain.

jamorham:
You don't need to make any changes regarding the manifest. The most important is to see if this works in terms of all features still being okay.

Where you do `new NSEmulatorReceiver()` you're allocating the memory for that object which stays resident all the time whether the user ever uses that feature or not. When its in the manifest it only gets created when an intent is directed towards it. So if the user doesn't use the NSEmulatorReceiver then no memory is used.

What I'm saying is with dynamic registration which you're doing here, you can create a new very short class which just wraps it, then only that short class is memory resident when the receiver is not in use:

```
import android.content.BroadcastReceiver;
import android.content.Context;
import android.content.Intent;

/**
 * Lightweight manifest-registered receiver.
 *
 * This keeps the registered receiver tiny until a broadcast is actually received,
 * then lazily creates and reuses the real receiver.
 */
public final class NSEmulatorReceiverWrapper extends BroadcastReceiver {

    private static volatile NSEmulatorReceiver receiverInstance;

    @Override
    public void onReceive(Context context, Intent intent) {
        getReceiverInstance().onReceive(context, intent);
    }

    private static NSEmulatorReceiver getReceiverInstance() {
        if (receiverInstance == null) {
            synchronized (NSEmulatorReceiverWrapper.class) {
                if (receiverInstance == null) {
                    receiverInstance = new NSEmulatorReceiver();
                }
            }
        }
        return receiverInstance;
    }
}
```

Navid200:
I hope I did not misunderstand your review.
I have now added two wrapper classes for the two receivers I had moved out of the manifest.
And am callng the wrappers in xDrip.

I repeated the reboot test:
After rebooting the phone, unlocked the sim card and unlocked the phone.  After seeing the xDrip icon appear in the notification area, without ever opening xDrip, locked the screen.
Nightscout in a browser on my computer showed that xDrip was uploading new readings to Nightscout.

Since there has been no change to anything related to notifications/alerts, I have not repeated my notification channel tests.  But, am running a set anyway.  If I find anything wrong, will notify.

Navid200:
I have tested everything again (including going back to a previous release) after the latest changes.  Everything is as good as it gets.  

I have to help train users to get used to notification channels.
And we may need to remove the overrise silent mode setting from the "Other" alerts in a future PR, which I will be happy to open later.

Navid200:
My thanks go to @ahaarrestad for having helped with this PR.  
He has fixed an oversight I had originally, and he also has tested cases I could not have.

ahaarrestad:
> for anything else using ForegroundServiceStarter I am not 100% sure if it will work

Tested the reboot / background-start case for the follower path. Emulator on API 34, xDrip set up as a Nightscout follower, rebooted, and I did not open the app.

After boot, `AutoStart` starts the process from the `BOOT_COMPLETED` broadcast, `NightscoutFollowService` comes up and reaches foreground, and it re-arms its poll alarm — all from the background. logcat shows `Background started FGS: Allowed` and `NightscoutFollowService is running in foreground`, with no exception and no crash. topResumedActivity stayed on the launcher the whole time, so nothing brought the UI up.

So the follower foreground services start fine from the background after a reboot on this PR.

jamorham (approved):
Okay lets try this!

PR #4604, opened 2026-07-04
OB1/G7 collector: unrecoverable signal-loss paths (stuck scans, stale GATT cache, watchdog gaps)

psonnera:
Detailed problem analysis to follow.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
@@ -30,6 +30,7 @@ import static com.eveningoutpost.dexdrip.models.JoH.tsl;
 import static com.eveningoutpost.dexdrip.models.JoH.upForAtLeastMins;
 import static com.eveningoutpost.dexdrip.plugin.Dialog.txIdMatch;
 import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.STATE.BOND;
+import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.STATE.CHECK_AUTH;
 import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.STATE.CLOSE;
 import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.STATE.CLOSED;
 import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.STATE.CONNECT;
@@ -38,7 +39,9 @@ import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.STATE.D
 import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.STATE.GET_DATA;
 import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.STATE.INIT;
 import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.STATE.PREBOND;
+import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.STATE.RESET;
 import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.STATE.SCAN;
+import static com.eveningoutpost.dexdrip.services.Ob1G5CollectionService.STATE.UNBOND;
 import static com.eveningoutpost.dexdrip.utilitymodels.Constants.DAY_IN_MS;
 import static com.eveningoutpost.dexdrip.utilitymodels.Constants.G5_CALIBRATION_REQUEST;
 import static com.eveningoutpost.dexdrip.utilitymodels.Constants.G5_SENSOR_FAILED;
@@ -1214,8 +1217,12 @@ public class Ob1G5CollectionService extends G5BaseService {
 
 
             scheduleWakeUp(MINUTE_IN_MS * 6, "fail-over");
-            if ((state == BOND) || (state == PREBOND) || (state == DISCOVER) || (state == CONNECT))
+            if ((state == BOND) || (state == PREBOND) || (state == DISCOVER) || (state == CONNECT)
+                    || (state == CONNECT_NOW) || (state == CHECK_AUTH) || (state == GET_DATA)
+                    || (state == RESET) || (state == UNBOND)) {
+                UserError.Log.e(TAG, "Fail-over wakeup: stuck in " + state + " after " + niceTimeScalar(msSince(last_connect_started)) + ", resetting to SCAN");
                 state = SCAN;
+            }
 
             checkAndEnableBT();
 
@@ -1608,6 +1615,17 @@ public class Ob1G5CollectionService extends G5BaseService {
             case DISCONNECTED:
                 connection_state = "Disconnected";
                 JoH.releaseWakeLock(floatingWakeLock);
+                // Fallback if BLE state dropped but connect flow did not fail-cleanly.
+                final long since_connect = msSince(last_connect_started);
+                if ((state == CONNECT || state == CONNECT_NOW || state == DISCOVER
+                        || state == CHECK_AUTH || state == PREBOND || state == BOND
+                        || state == GET_DATA || state == RESET || state == UNBOND)
+                        && since_connect > Constants.SECOND_IN_MS * 10
+                        && JoH.ratelimit("ob1-disconnect-fallback-retry", 5)) {
+                    UserError.Log.e(TAG, "Fallback retry: disconnected in " + state + " after " + niceTimeScalar(since_connect) + ", forcing scan");
+                    stopConnect();
+                    changeState(SCAN);
+                }
                 break;
         }
         static_connection_state = connection_state;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
@@ -489,6 +489,10 @@ public class Ob1G5CollectionService extends G5BaseService {
                 UserError.Log.d(TAG, "Not avoiding scanning due to connect failure level: " + connectNowFailures);
                 connectNowFailures++;
             }
+            if (alwaysMinimize && preScanFailureMarker) {
+                alwaysMinimize = false;
+                UserError.Log.e(TAG, "Not avoiding scanning due to mid-session connection failure");
+            }
         }
         if (transmitterMAC == null) {
             UserError.Log.d(TAG, "Do not know transmitter mac inside minimize scanning!!");
@@ -1535,6 +1539,7 @@ public class Ob1G5CollectionService extends G5BaseService {
         if (state == CHECK_AUTH || state == GET_DATA || state == PREBOND
                 || state == BOND || state == RESET || state == UNBOND) {
             UserError.Log.e(TAG, "Connection failure in " + state + ": " + throwable.getMessage() + ", forcing scan");
+            preScanFailureMarker = true;
             changeState(SCAN);
         }
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
@@ -1478,9 +1478,15 @@ public class Ob1G5CollectionService extends G5BaseService {
 
         if (state == DISCOVER) {
             // possible encryption failure
-            if (!resetBondIfAllowed(false) && android_wear) {
-                UserError.Log.d(TAG, "Trying alternate reconnection strategy");
-                changeState(CONNECT_NOW);
+            final boolean bondReset = resetBondIfAllowed(false);
+            if (!bondReset) {
+                if (android_wear) {
+                    UserError.Log.d(TAG, "Trying alternate reconnection strategy");
+                    changeState(CONNECT_NOW);
+                } else {
+                    UserError.Log.e(TAG, "Connection failure in DISCOVER, bond reset not allowed, forcing scan");
+                    changeState(SCAN);
+                }
             }
             return;
         }
@@ -1523,6 +1529,13 @@ public class Ob1G5CollectionService extends G5BaseService {
                 tryGattRefresh();
                 changeState(SCAN);
             }
+            return;
+        }
+
+        if (state == CHECK_AUTH || state == GET_DATA || state == PREBOND
+                || state == BOND || state == RESET || state == UNBOND) {
+            UserError.Log.e(TAG, "Connection failure in " + state + ": " + throwable.getMessage() + ", forcing scan");
+            changeState(SCAN);
         }
 
     }
@@ -1611,6 +1624,11 @@ public class Ob1G5CollectionService extends G5BaseService {
                 break;
             case DISCONNECTING:
                 connection_state = "Disconnecting";
+                if (state == CONNECT || state == CONNECT_NOW || state == DISCOVER
+                        || state == CHECK_AUTH || state == PREBOND || state == BOND
+                        || state == GET_DATA || state == RESET || state == UNBOND) {
+                    UserError.Log.e(TAG, "Disconnecting in active state: " + state + " after " + niceTimeScalar(msSince(last_connect_started)));
+                }
                 break;
             case DISCONNECTED:
                 connection_state = "Disconnected";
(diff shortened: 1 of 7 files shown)


psonnera:
Hello, this is not a PR but a suggestion to have a look at some weaknesses that have impacted my use of xDrip with a small chinese phone NX1. I had the same issue with the Cubot Pocket and understand it cannot be easily reproduced on higher quality phones like Pixels. Basically xDrip can sometimes lose connection to the sensor without being able to recover. I fed a lot of data into LLM, logs, adb dumps, and tried to find out why Juggluco would be able to recover when xDrip wasn't. The result is functional: I've been testing it for some months now and I never saw any occasion where xDrip would not recover the signal in more than 20 minutes. Bottom line, this is just a suggestion to look at what could be improved to make recovery more efficient. Not my work, but hopefully it will give you some ideas. Touching the OB1 collector has serious impact on many other things, so I'll easily understand it shouldn't be touched at all. Below the LLM analysis.

psonnera:
Problem

G7 users can hit prolonged signal loss that xDrip never recovers from on its own — readings stop until the user manually toggles Bluetooth or restarts the app, sometimes hours later (typically noticed on waking up). Tracing these incidents through Ob1G5CollectionService shows they are not one bug but several independent weaknesses that compound:

1. A stuck BLE scan is a black hole the state machine cannot escape

When the Android BLE stack wedges, the scan subscription stays active but delivers no callback at all — no result, no error. Every recovery mechanism in the collector is callback-driven or runs inside automata(), so none of them can fire:

- onScanFailure() is never called (it only fires on exceptions, not on silence);
- the automata never re-enters, so even the existing last-resort collector restart (automata() case SCAN, after 30 min without connection) is unreachable — it can only run when the state machine runs;
- the 320s stop-scan watchdog (STOP_SCAN_TASK_ID) was only scheduled when minimize_scanning was enabled, so the majority of users had no time-based escape path whatsoever.

The result is indefinite silent scanning: the phone looks like it's collecting, the transmitter is broadcasting every 5 minutes, and nothing connects.

2. Repeated scan timeouts never escalate to the one thing that fixes a wedged stack

When the stack is wedged, stopping and restarting the scan often isn't enough — only an adapter restart clears it. The collector has a Bluetooth watchdog for connect failures (connectNowFailures % 12 == 7), but consecutive scan timeouts never triggered it, so a wedged-stack signal loss persisted even once timeouts were detected.

Conversely, any escalation added here must be bounded: an unconditional "restart BT every Nth timeout" keeps cycling the adapter every ~16 minutes forever when the transmitter is merely absent (expired sensor, phone left at home) — which disrupts every other BLE device on the phone. For looping users with Bluetooth pumps this is worse than the original problem. Distinguishing "recent signal loss" from "transmitter gone" (e.g. by time since last successful connection) is essential.

3. After a mid-session failure, the recovery reconnects into the same broken state

Two related issues:

- Stale GATT cache: after a failure in CHECK_AUTH / GET_DATA / BOND etc., forcing a rescan reconnects using Android's cached GATT service table. If that cache is the corrupt thing that caused the failure, the failure just repeats on every retry.
- Minimize-scanning shortcut: with ob1_avoid_scanning (or WholeHouse) active, the recovery path skips the real scan entirely and jumps to CONNECT_NOW on the remembered MAC — repeating the exact sequence that just failed instead of re-establishing ground truth.

4. Smaller defects in the recovery utilities themselves

- tryGattRefresh() consumes its once-per-60s ratelimit token before checking whether a refresh is even possible; its if (connection != null) guards only a log line while connection.queue(...) runs unconditionally. A refresh attempted on a dead/null connection is a no-op that still blocks the next genuine refresh opportunity for 60 seconds.
- Once the 320s scan timeout applies to all users, users who deliberately run continuous scanning get silently duty-cycled through prepareToWakeup() (up to ~5-minute blind gaps per cycle) — a behavior change they never opted into.

Reference implementation

This branch contains a minimal working fix for the above (a few lines per item) that compiles and is running on my device: always-scheduled 320s scan timeout; adapter restart after 3 consecutive scan timeouts gated on bluetooth_watchdog and a connection within the last 3 hours; GATT refresh + forced real scan (preScanFailureMarker) on mid-session failure; null/ratelimit guard in tryGattRefresh(); immediate scan restart on timeout when minimize-scanning is off.

Treat it as one possible implementation — the issue descriptions above are the substance, and I'm happy to rework the code if the maintainers prefer a different approach. Log lines to watch when reproducing (walk out of range mid-session): Scan timeout watchdog, Gatt refresh skipped - no connection, Restarting scan immediately as minimize scanning is off.

🤖 Generated with Claude Code

psonnera:
Please note commit cc28a842717530490be06b21883d9f74072c1d0b is irrelevant, that one fixed the random new data chime (sorry Navid, I couldn't test your version as it would have removed the OB1 tricks that made my life easier)

Navid200:
Thanks

Are you certain that you have seen improvements with this so you have gone back and forth and it is obvious to you that this effectively enables a phone that currently is effectively unusable?

If yes, as much as I hate new settings, we cannot just change the Dexcom protocol and jeopardize the behavior of the phones that are fine now.  It is also not possible to test all phones with it.  
Can you add a new setting under Dex Debug settings to be disabled by default?  Can you make the code such that it would be clear to the reviewers that if the new setting is disabled, there will be no change to the existing process?

Can you place this new setting behind engineering mode?

If we do all of that, we can ask users of that phone to try it out, which means they will need to enable engineering mode unfortunately.
If all the responses are positive, we can also ask these guys to try it out: https://github.com/NightscoutFoundation/xDrip/discussions/4527 

If there are no negative surprises, we then remove it from engineering mode.  We can then ask everyone to try it and provide feedback.  
If no one has a problem with it enabled, we can then retire the new setting and enable it for everyone.

If it turns out that some brands behave better with the setting disabled, we can leave the new setting just like that.  

I would like to be involved in choosing the title and summary and placement of the new setting.  

This all is my opinion.  We can/should (either way) wait for jamorham's review.

psonnera:
@Navid200, sure will do something safer like this, yes. And then I'll test for a month.
I found out from the adb logs that MTK processors (chinese phones) just don't work with Minimize scanning and that can't be fixed. I ended up in the code after trying to understand the impact of all xDrip settings, and trying the most logical combinations.
By the way, having an indication of the issue on the master phone looking at the follower main view message is fantastic (sensor error #, sensor # not within reach, ...). It really helps when you're a caregiver.

ahaarrestad:
Thanks so much for digging this deep — honestly the problem write-up is excellent, and that `tryGattRefresh()` ratelimit-before-null-check catch is a lovely find on its own. 🙌

Since you're already working with Claude, I'm less worried about *whether* we can get tests and more curious about *which* ones would actually help us here — what do you think? A couple of thoughts, very much open to your take:

- Would it make sense to start with one test that shows *the new setting disabled → same decision path as today's master* (watchdog not scheduled without minimize-scanning, `prepareToWakeup()` rather than `changeState(SCAN)`, etc.)? That feels like it would give reviewers the "setting off = no change" confidence @Navid200 is after — does that match how you're seeing it?
- After that, maybe the genuinely new bits: the `scanTimeouts % 3` + `bluetooth_watchdog` + `< 3h since last connect` gate (including the "transmitter gone >3h → no reset" case), the `preScanFailureMarker` path, and the `BtPermissionCache` fresh/stale/fallback logic? The last one looks almost pure — could be an easy first one?
- There's already a Robolectric + reflection pattern for testing this exact service if that's useful — want me to point you at it? To test the gates cleanly you might find it easier to pull each decision into a small `@VisibleForTesting` method — would that fit how you'd want to structure it?

No pressure on volume at all — honestly a small handful you'd feel comfortable standing behind is probably more useful than broad coverage. Does that sound reasonable to you? Thanks again for all this. 🙏

app/src/main/java/com/eveningoutpost/dexdrip/cloud/jamcm/Pusher.java
```diff
@@ -537,9 +537,11 @@ public static boolean sendMessage(Bundle input) {
                 return false;
             }
 
+            val id = JamCm.getId();
```
ahaarrestad: I think this might be the "random new data chime" change you mentioned (commit `cc28a842`) rather than part of the signal-loss work? Would it make sense to split it into its own PR — mostly so this one stays focused and both are easier to reason about on their own? Happy either way if there's a reason to keep them together.

app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
```diff
@@ -1214,9 +1239,15 @@ public int onStartCommand(Intent intent, int flags, int startId) {
 
 
             scheduleWakeUp(MINUTE_IN_MS * 6, "fail-over");
-            if ((state == BOND) || (state == PREBOND) || (state == DISCOVER) || (state == CONNECT))
+            if ((state == BOND) || (state == PREBOND) || (state == DISCOVER) || (state == CONNECT)
```
ahaarrestad, line 1254: This widens the 6-minute fail-over reset to also cover `CONNECT_NOW`, `CHECK_AUTH`, `GET_DATA`, `RESET`, `UNBOND`. `CHECK_AUTH`/`GET_DATA` stuck for 6 min does sound genuinely stuck to me — but I wasn't sure about `CONNECT_NOW`/`RESET`/`UNBOND`, since they interact with the minimize-scanning reconnect flow. Do you think existing users could notice a difference there? Might be another good candidate for living behind the new pref?

app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
```diff
@@ -1516,16 +1553,28 @@ private void onConnectionFailure(final Throwable throwable) {
...
+            UserError.Log.e(TAG, "Connection failure in " + state + ": " + throwable.getMessage() + ", forcing scan");
+            preScanFailureMarker = true;
+            tryGattRefresh();
+            changeState(SCAN);
         }
 
     }
 
     public void tryGattRefresh() {
+        if (connection == null) {
```
ahaarrestad: Really nice catch on this one — moving the `connection == null` check ahead of the ratelimit makes total sense. Before, a refresh on a null connection would still eat the 60s token and block the next real refresh, right? Lovely fix. 👍

ahaarrestad (commented):
This is such a useful investigation — all four problem areas ring true to me, and I'd love to keep the diagnosis in the PR description regardless of where we land on the implementation.

Building on @Navid200's thoughts: right now most of the changes seem to touch the default path for everyone (always-scheduled 320s watchdog, `changeState(SCAN)` instead of `prepareToWakeup()`, the wider fail-over reset, the new DISCONNECTED fallback). Would it feel workable to tuck all of that behind a single new pref (off by default, under Dex Debug / engineering mode)? My thinking is it'd let a reviewer confirm "pref off = no change" at a glance — and it pairs nicely with that first characterization test. Does that seem sensible to you, or would it get in the way of what you're trying to do?

A few smaller things inline — none of them are doubts about the *idea*, just things I was wondering about. 🙂

app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
```diff
@@ -1604,10 +1653,26 @@ private synchronized void onConnectionStateChange(RxBleConnection.RxBleConnectio
...
+                        || state == GET_DATA || state == RESET || state == UNBOND) {
+                    UserError.Log.e(TAG, "Disconnecting in active state: " + state + " after " + niceTimeScalar(msSince(last_connect_started)));
+                }
                 break;
             case DISCONNECTED:
                 connection_state = "Disconnected";
                 JoH.releaseWakeLock(floatingWakeLock);
+                // Fallback if BLE state dropped but connect flow did not fail-cleanly.
+                final long since_connect = msSince(last_connect_started);
+                if ((state == CONNECT || state == CONNECT_NOW || state == DISCOVER
```
ahaarrestad: One thing I was wondering about here: in rxandroidble, doesn't a dropped connection sometimes trigger *both* `onConnectionFailure()` and `onConnectionStateChange(DISCONNECTED)`? If so, would this new path and the existing failure handler both end up calling `changeState(SCAN)` in the normal case? I can see why you need it for the *silent* disconnect case, so I'd definitely keep it — just curious whether the double path is harmless, or whether it'd be worth gating so only one wins. What have you seen in your logs?

app/src/main/java/com/eveningoutpost/dexdrip/utils/BtPermissionCache.java
```diff
@@ -0,0 +1,90 @@
...
+import androidx.core.content.ContextCompat;
+
+import com.eveningoutpost.dexdrip.models.UserError;
+
+/**
+ * Caches BLE permission check results to avoid calling PackageManager during
+ * scan/connect operations. Refreshed once per service wake cycle (~5 min).
+ * Falls back to a live PackageManager check when the cache is cold or stale.
+ */
+public class BtPermissionCache {
```
ahaarrestad, line 20: A couple of tiny housekeeping things, nothing functional: could we add an `@author` tag (we try to keep that on new source files), and would you be open to `JoH.tsl()` instead of `System.currentTimeMillis()` for the freshness window, just for consistency with the rest of the codebase? Also — this class looks like the easiest thing in the whole PR to unit-test, if you fancy a quick win. 🙂

app/src/main/java/com/eveningoutpost/dexdrip/cloud/jamcm/Pusher.java
psonnera: Yes you're right, it landed there but is not relevant to the OB1 work... Sorry!

app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
psonnera: The synchronous state = SCAN write in changeState() means whichever callback fires first flips state, so the second one's guard fails and it doesn't double-fire.
Please note I need to provide logs for real time usecase.

psonnera:
@ahaarrestad thanks for reviewing this. I created this draft because I believe something is not working the way it should, but the analysis being LLM driven I just wanted real developers to use it as a base of reflection for eventual code changes. Not the code proposed.
Let me shield the new functions behind a developer only feature for now, ob1RecoveryEnabled.
On the logs side, I unfortunately found out how to better see the OB1 logs with the Ob1G5CollectionService:v,Ob1G5StateMachine:v tags only after I've been fighting with it.
My plan now is:
1 - Implement these changes in my kid's xDrip
2 - Monitor and gather logs for meaningful events, for a month
3 - Feedback and give it back to you, real developers
Thanks!

ahaarrestad:
Really nice turnaround — the engineering-mode gate makes the "off = no change" story easy to verify, and thanks for splitting out the chime fix and adding the `BtPermissionCache` test. Looks like a solid base to gather field logs on. 👍

jamorham:
@psonnera thanks for taking the time to look at this and being grounded in terms of just using this for pointers rather than solutions.

Unfortunately I don't think a current LLM can understand a system like this and the multi-threaded nature of how OB1 collector works is exactly the sort of non-linear system that LLMs struggle with. Although I keep an open mind and am interested in anything Claude provides, I think this discussion also a good example of AI confidently leading people down the wrong track.

OB1 was originally designed to give compatibility even down to Android 4 and has grown with many work-arounds for different stacks as well as having G7 retrofitted to it. Its got pretty messy as a result. I'm happy to take another deep dive and see if I can resolve any issues. 

I have an unmerged branch which attempts to update the libraries and behavior but in testing this was less reliable which is why I haven't merged it. 

What I would say goes as follows:

- There definitely are bugs. I've seen them myself but only around sensor changes.
- We may need to revisit the use of scanning - with GluPro collector I found that scanning was essential even on the most common and up-to-date handsets which was surprising. 
- My plan is to check the state flow going back to first principles with live testing rather than analysis of the code.

psonnera:
@jamorham thanks for this feedback, this is exactly what I needed. I do not want LLM code to add burden to maintaining xDrip: just wanted to see if pointers that proved efficient in my own version could be of use. My main issue is non-recoverable connection loss with the G7, not frequent, but when it occurs my son needs to reboot his phone. Sometimes forcing xDrip close will allows recovery, this is why I'm not 100% sure the phone is the culprit.
Bottom line, how would you like feedback to address potential issues? Only logs and adb dumps? Or adding LLM analysis can add value? Thank you!

psonnera:
Ooops ignore the last 2 commits I need to change upstream!

jamorham:
@psonnera enable the ob1 debugging (google for the QR code) and then next time you get stuck issue persist for a while - upload logs and contact me to let me know they've been submitted and I can see if there are any clues.

Navid200:
@psonnera 

> sorry Navid, I couldn't test your version as it would have removed the OB1 tricks that made my life easier  
  

Would you please remind me what you mean by my version?

psonnera:
@Navid200 regarding the random new data chime, filtering is embedded in my own version https://github.com/NightscoutFoundation/xDrip/commit/6ab9366e25c6b4f6d298e2e04f0ce190c7eb06d4. Cannot find the original chat in the too many discussions, ignore for now.

PR #4608, opened 2026-07-06
Tidepool: don't force interactive re-login on transient token-refresh network errors

ahaarrestad:
### Problem

On unreliable networks, xDrip repeatedly pops the Tidepool browser/login page
(reported in #4605). The reporter's log shows:

    TidePoolAuth: failed to fetch configuration
    TidePoolAuth: Failing to use access token - trying initial login again
    TidePoolAuth: Got exception token: AuthorizationException:
        {"type":0,"code":3,"errorDescription":"Network error"}

### Root cause

`AuthFlowIn.handleTokenLoginAndStartSession()` refreshes the Tidepool token via
AppAuth's `performActionWithFreshTokens`. When the access token needs refreshing
while the network is down, AppAuth's refresh call fails and hands the callback
`accessToken == null` with `GeneralErrors.NETWORK_ERROR` (`type:0, code:3`). The
callback only checked whether an access token came back, so on `null` it always
called `retryInitialLogin()` → `doTidePoolInitialLogin()` →
`performAuthorizationRequest()`, which launches the interactive login page.

So a transient connectivity failure was treated identically to a rejected
credential. The stored refresh token is still valid — the device just couldn't
reach the server that moment. A genuinely expired/revoked refresh token surfaces
differently (`TokenRequestErrors.INVALID_GRANT`, type 2), so it is unaffected by
this change and still triggers a real re-login.

This path has existed since 2023 (`e33663e55`); it is not a recent regression —
it surfaces only on genuinely flaky networks.

### Fix

Add `isTransientTokenError(AuthorizationException)`: a `TYPE_GENERAL_ERROR` with
code `NETWORK_ERROR` or `SERVER_ERROR` is transient. On a transient failure the
fresh-token callback now keeps the session and lets the next sync retry the
silent refresh; every other failure preserves the existing fall-back to
interactive login. The callback body was also lifted out of the lambda into a
small `onFreshTokenResult(...)` method so the flow reads top-to-bottom —
behavior and all existing log messages are unchanged.

### Testing

New `AuthFlowInTest` pins the classifier boundary: `NETWORK_ERROR` / `SERVER_ERROR`
→ transient; `INVALID_GRANT`, a non-network general error
(`JSON_DESERIALIZATION_ERROR`), and `null` → not transient (so re-login still
fires for genuine credential problems). Full `./gradlew test` green.

### Verifying this build

This is grounded in code analysis — the AppAuth token-refresh path plus the
reporter's own `{"type":0,"code":3}` log — rather than a staged reproduction
(the failure needs the access token to expire *during* a network outage). The
recovery path is verified in code: a network error never discards the stored
refresh token, and `TidepoolEntry.newData()` re-drives the silent refresh
(≤20 min), so Tidepool resyncs on reconnect with no login prompt. Field
confirmation from anyone hitting this on flaky networks is welcome.

Refs #4605

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/tidepool/AuthFlowIn.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/tidepool/AuthFlowIn.java
@@ -14,9 +14,9 @@ import com.eveningoutpost.dexdrip.utilitymodels.Inevitable;
 import com.eveningoutpost.dexdrip.utilitymodels.Pref;
 
 import net.openid.appauth.AppAuthConfiguration;
+import net.openid.appauth.AuthState;
 import net.openid.appauth.AuthorizationException;
 import net.openid.appauth.AuthorizationResponse;
-import net.openid.appauth.TokenResponse;
 
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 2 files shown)


Navid200:
Damn it's good to have you on our side.

jamorham:
Are you able to test this with forcing a network failure during the refresh process? I use tidepool and am regularly transiting between networks and I've never seen an authentication pop-up.

ahaarrestad:
I have looked i to trying this, but as far as i can ser it requires both some tidepool accounts which i dont have, and some lucky network hickups at the right time. I have however tried my best to test as much as possible on the change.

Navid200:
I just got this with the latest Nightly with a Tidepool login popup:
Screenshot_20260724-071341  
  
I now remember I have got this before.

ahaarrestad:
That's exactly the case this PR fixes. `{"type":0,"code":3}` is AppAuth's `NETWORK_ERROR` — a transient network hiccup during the silent token refresh, not an expired credential. On master that falls straight into `retryInitialLogin()`, which is the "trying initial login again" line and the popup. `isTransientTokenError()` classifies type 0 as transient, keeps the session, and lets the next sync retry the refresh silently, so no popup.

This is also the repro we couldn't stage — it needs the network to drop in the exact refresh window. If you can install the PR build and catch it again, that'd be the field confirmation @jamorham asked for.

Navid200:
I don't get this once a month.  I may get it once every 3 months.  So, I don't know if there is any point running it.  I keep switching my install for testing different things.  There is no way I can just be running this and nothing else for the next 3 months.

ahaarrestad:
@jamorham Managed to force it after all — turned out I didn't need a Tidepool account. The refresh path just needs an AuthState in prefs; AppAuth validates nothing on deserialize, it only reads `expires_at`. And with the network down the token POST fails before Tidepool is ever contacted, so a fake token takes the identical path. Seeded an expired state on an emulator, killed wifi and mobile data, triggered a sync.

master (703395e73):
```
E/TidePoolAuth: Got exception token: AuthorizationException: {"type":0,"code":3,"errorDescription":"Network error"}
E/TidePoolAuth: Failing to use access token - trying initial login again
E/TidePoolAuth: failed to fetch configuration
```
plus the "Connecting to Tidepool" toast on screen — same as Navid's screenshot above.

This branch, same steps:
```
E/TidePoolAuth: Got exception token: AuthorizationException: {"type":0,"code":3,"errorDescription":"Network error"}
D/TidePoolAuth: Token refresh failed transiently - keeping session, will retry on next sync: ...
```
No re-login, no toast — session kept, next sync retries silently.

On not seeing it yourself: it needs the network gone in the exact window where the token has expired and the refresh fires. Navid puts it at once every few months. Roaming between networks that stay reachable won't hit it.

Also added two tests over `onFreshTokenResult()` — the existing ones only pinned the classification, not what it's used for. `invalid_grant` still goes to interactive login.

jamorham (approved):

PR #4610, opened 2026-07-06
nsfollow: backfill 24h on first run for a new phone / fresh install

ahaarrestad:
## Problem

On a **new phone** or **fresh install**, Nightscout Follow only pulled the newest ~10 entries instead of a full day of history, so the graph never backfilled on a fresh install.

`NightscoutFollow.work()` chose the entry query based on whether a local `BgReading` already existed:

```java
if (lastTs > 0) {
    // 24h date-filtered backfill via find[date][$gt]
} else {
    getEntries(secret, 10, ...); // only 10 newest entries
}
```

A fresh install has no local readings, so `lastTs == 0` always took the `else` branch. This regressed in `9aa1aa999` ("replace MissedReadingsEstimator with server-side date filter"), which reduced the first-run fetch from ~288 entries (24h) to a hardcoded 10.

## Fix

The cutoff already floors to `now - 24h`:

```java
final long cutoff = Math.max(lastTs, JoH.tsl() - Constants.DAY_IN_MS);
```

That holds for `lastTs == 0` too (`max(0, now-24h) == now-24h`), so the first-run branch is redundant. Drop it and always fetch via `find[date][$gt]` with the existing `count=2880` safety cap. Existing users are unaffected (they always have `lastTs > 0`); only the first poll on a blank device changes.

Juggluco and real Nightscout both answer the date filter — covered by `JugglucoEmulatorContractTest`.

## Tests

- Rewrote `NightscoutFollowEntriesWorkTest.work_usesCountOnly_whenNoLastReading` → `work_backfills24Hours_whenNoLastReading`, asserting first run now uses `find[date][$gt]` with a ~now-24h cutoff and `count=2880` (TDD: watched it fail, then pass).
- Updated the stale "count-based query" comment in `JugglucoEmulatorContractTest`.
- Full `./gradlew test` green.


ahaarrestad:
@jamorham - discovered this as I've been testing a lot on "new emulators" lately - and it only backfills 10 samples.... I was expecting 24 hours.... so not ideal... It is an edge-case, but still worth fixing I belive.

ahaarrestad:
Merged current master in — only conflict was in the entries test, since #4657 refactored the helper it used. Tests green. Ready for merge.

ahaarrestad:
@jamorham it would be great if we could merge this one and fix the bug for new installs for nightscout follow :-)

I have merged main and verified the tests.

PR #4620, opened 2026-07-11
nsfollow: remove redundant absolute timestamps from the follower status page

ahaarrestad:
## Problem

The Nightscout Follow status page shows the same fact twice — as an age relative to now, and again as an absolute wall-clock time:

| Kept | Removed |
|---|---|
| `Latest BG` — "5 min ago" | `Last BG time` — "14:32" |
| `Next poll in` — "2 min" | `Next poll time` — "14:37" |

The relative row answers the question directly. The absolute one makes you do the subtraction yourself.

@Navid200 raised the clutter on #4476. His two comments name opposite items — [April 5](https://github.com/NightscoutFoundation/xDrip/pull/4476#issuecomment-4188804176) says drop the absolute rows, [May 29](https://github.com/NightscoutFoundation/xDrip/pull/4476#issuecomment-4575492863) says drop the relative ones. This follows April.

## After

Nightscout Follow status with the change — both absolute timestamps are gone, only the relative rows remain:

Nightscout Follow status after the change

## Fix

Drop the two absolute rows from `NightscoutFollowService.megaStatus()`:

```java
-if (lastBg != null) {
-    statuses.add(new StatusItem("Last BG time", JoH.dateTimeText(lastBg.timestamp)));
-}
-statuses.add(new StatusItem("Next poll time", JoH.dateTimeText(wakeup_time)));
```

Display only. No change to polling, backfill or data handling.

## Tests

- `NightscoutFollowServiceMegaStatusTest`: two cases asserting the absolute rows are gone and the relative ones remain. The BG case inserts a real reading first — that row was conditional on one existing, so without it the test would pass for the wrong reason.
- TDD: both failed before the change.
- Full `./gradlew test` green.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollowService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/nsfollow/NightscoutFollowService.java
@@ -270,7 +270,9 @@ public class NightscoutFollowService extends ForegroundService {
 
         statuses.add(new StatusItem("Last poll", lastPollText + (lastPoll > 0 ? " ago" : "")));
         statuses.add(new StatusItem("Next poll in", JoH.niceTimeScalar(wakeup_time - JoH.tsl())));
-        statuses.add(new StatusItem("Buggy handset", JoH.buggy_samsung ? gs(R.string.yes) : gs(R.string.no)));
+        if (JoH.buggy_samsung) {
+            statuses.add(new StatusItem("Buggy handset", gs(R.string.yes)));
+        }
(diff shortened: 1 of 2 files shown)


Navid200:
It looks like you are removing only one of the two parameters.  
The parameters that look redundant, at least to me, are two:
620475906-acad077c-45c0-473b-8645-2587e0cb1ea3
  
When I make a change to the UI, I always include a before/after image in the PR.  Sometimes, it has actually let me see that I have forgotten something.

Navid200:
Or may be you have removed it and I made a mistake looking at the changes made to the code.

ahaarrestad:
Thanks @Navid200 . I have added an "after" image in the PR.

Navid200:
This looks good to me.
Thanks

jamorham (approved):

PR #4510, opened 2026-04-28
Snooze glucose alerts with volume buttons, also from lock screen

petervanrijt:
At home, we are very happy with this (small) change, especially at night 😄. Both my girlfriend (T1D) and I (as her follower) are already using the `petervanrijt:petervanrijt-alarm-snooze-with-volume-buttons` branch. 

## Table of contents

* What this changes
* Setup after installation
* UI changes
* Security: lock screen behaviour
* Problems solved
* Implementation
* Changed files
* Safety review
* Testing
* Design consideration: separate wake-screen preference

## What this changes
With this change you can snooze a glucose alert by **pressing a volume button on the side of your phone twice**. This works in every situation:

- **Screen off or locked** -- the phone wakes up, shows the snooze screen on top of the lock screen, and waits for your button press. You do not need to unlock the phone.
- **Another app in the foreground** -- the snooze screen appears over the app you were using. After snoozing, the previous app is restored automatically.
- **xDrip already open** -- the snooze screen appears inside the app. The system volume does not change while the alert is active.

A single press is not enough -- you must press the **same** button **twice within 1.5 seconds** to prevent accidental snoozes. The first press shows a brief message asking you to confirm; the second press snoozes the alert.

After snoozing, the ongoing glucose notification (the persistent graph on your phone) shows a line like *"Alert low snoozed until 14:35"* so you can see at a glance when the alert will return.

A new setting (**Settings > Alarms and Alerts > Glucose Alerts Settings > Volume snooze from lock screen**) controls whether the snooze screen is allowed to wake your screen and appear over the lock screen. It is off by default; existing users see no change unless they enable it.

Alert testing via **Settings > Glucose Level Alerts List > Edit alert > Test alert** now also supports volume button snooze, so you can verify the full flow without waiting for a real alert.

## Setup after installation

Volume button snooze works out of the box inside the xDrip Home screen once the setting is enabled. For lock screen and overlay support a few extra steps are needed.

**Step 1 Enable volume button snooze in xDrip**

Open **Settings > Alarms and Alerts > Glucose Alerts Settings** and enable **Buttons silence alarms**. This is the parent setting; without it volume buttons have no snooze function.

**Step 2 Enable lock screen wake**

Under the same screen, enable **Volume snooze from lock screen** (only visible when *Buttons silence alarms* is on). This allows the snooze screen to wake the phone and appear over the lock screen when an alert fires. Leave it off if you only want volume button snooze while xDrip is already open.

[...]

PR #4623, opened 2026-07-12
Snooze glucose alerts from anywhere: over other apps, on the lock screen, and with volume buttons

petervanrijt:
# What this PR does

### What this PR does

Snoozing a glucose alert no longer requires opening xDrip first:

- **Snooze screen over the app in use** — when an alert fires while you are using another app, the snooze screen floats on top of it; after snoozing you are returned to that app automatically. Opt-in via *Settings → Alarms and Alerts → Glucose alerts settings*, requires the Android permission *Display over other apps* (xDrip asks for it at startup).
- **Snooze screen on the lock screen** — with *Override silent mode* (per alert) or *Wake Screen* enabled, an alert wakes the display and shows the snooze screen above the lock screen, without unlocking. Closing it returns to the lock screen.
- **Volume-button snooze, also from the lock screen** — press the **same volume button twice within 1.5 seconds** to snooze. While an alert is sounding, the volume buttons no longer change the phone volume. A held button does not count as a double press.
- **Snoozed-alert status** — the ongoing notification (and optionally the Extra Status Line) shows *"Alert &lt;name&gt; snoozed until HH:mm"* while an alert is snoozed, and hides it when glucose recovers.
- **Test alert with a 5-second delay** — time to switch apps or turn the screen off, so the new screens can be tested from *Edit Alert → Test alert*.
- **Robustness fixes** — crash when disabling alerts while the alert type is missing; REMOVE ALERT leaving an orphaned active alert (and its sound) behind; test alerts not being snoozable; full-screen intents silently ignored due to notification channel importance (fixed with a migration that only runs for users who need it).

The alert sound itself is never reduced or suppressed by any of this: every failure direction is "more conspicuous" or "convenience unavailable", never "quieter or missed".

### What existing users need to change

After updating, xDrip asks once per start to grant **Display over other apps**; without it everything keeps working except the pop-up over other apps. Note two behaviour changes: snoozing with a volume button now takes a **double press** (was a single press), and alerts with *Override silent mode* now wake up into the **snooze screen** instead of the main screen. Users of xDrip notification channels who use *Wake Screen* or override-silent alerts get a fresh high-importance alert channel; re-apply any custom channel settings there if desired.

### Documentation (included in this PR)

Design, decisions, safety review and per-change test plans:

- [CHG1-snooze-over-other-apps.md](https://github.com/user-attachments/files/30242951/CHG1-snooze-over-other-apps.md)
- [CHG2-fullscreen-snooze-on-lockscreen.md](https://github.com/user-attachments/files/30242959/CHG2-fullscreen-snooze-on-lockscreen.md)
- [CHG3-volume-key-snooze.md](https://github.com/user-attachments/files/30242961/CHG3-volume-key-snooze.md)
- [CHG4-volume-key-double-press.md](https://github.com/user-attachments/files/30242965/CHG4-volume-key-double-press.md)
- [CHG5-statusline-snoozed-alert.md](https://github.com/user-attachments/files/30242968/CHG5-statusline-snoozed-alert.md)
- [CHG6-review-chg1-chg5.md](https://github.com/user-attachments/files/30242972/CHG6-review-chg1-chg5.md)
- [CHG7-pr4510-adoptions.md](https://github.com/user-attachments/files/30242975/CHG7-pr4510-adoptions.md)
- [CHG8-notification-snooze-line.md](https://github.com/user-attachments/files/30242981/CHG8-notification-snooze-line.md)
- [CHG9-delayed-test-alert.md](https://github.com/user-attachments/files/30242983/CHG9-delayed-test-alert.md)
- [CHG11-buttons-silence-alarms-dependencies.md](https://github.com/user-attachments/files/30242984/CHG11-buttons-silence-alarms-dependencies.md)
- [CHG12-consistent-snooze-text-and-prompt-refresh.md](https://github.com/user-attachments/files/30242986/CHG12-consistent-snooze-text-and-prompt-refresh.md)
- [CHG13-final-review.md](https://github.com/user-attachments/files/30242990/CHG13-final-review.md)
- [CHG14-adopt-upstream-notificationchannels.md](https://github.com/user-attachments/files/30242997/CHG14-adopt-upstream-notificationchannels.md)
- [CHG15-remove-string-resources.md](https://github.com/user-attachments/files/30242999/CHG15-remove-string-resources.md)
- [file_overview.md](https://github.com/user-attachments/files/30243002/file_overview.md)

### Notes for reviewers

- Supersedes draft #4510: scope cleaned up per the feedback there (no extra wake switch — the existing *Wake Screen* setting is reused; single shared key handler; documented test plans per change).
- The robustness fixes (CHG7/A1) are preconditions for the new snooze paths (consistent `ActiveBgAlert` state, snoozable test alerts), not standalone refactoring.
- No new string resources: per review feedback the new user-facing texts are plain English literals for now; a dedicated follow-up PR will move them into translatable resources once the wording is agreed. Existing translated strings are untouched.
- Phone app only; the Wear module is untouched.

# Full disclosure
This pull request is accelerated using Claude Code (model Opus 5 with effort setting on high or maximum).


Navid200:
Thanks

It will take a while to make the strings perfect.  I suggest removing all the strings from this PR.

petervanrijt:
Ok. Will do tonight or tomorrow. Just synced with xDrip-master. I will
review it before sending it to dev team for the final review. When curious
you can already build it yourself :)



Op ma 13 jul 2026, 20:10 schreef Navid ***@***.***>:

Navid200:
If you prefer to keep the strings, we can work on them.  It is just that it will only delay the PR in going through.
The reason I suggest keeping them as text in English is that then they will not be show stoppers as long as they get the point across.
We will then be able to open another PR later and spend as much time on it to agree on the strings without delaying the functionality.  That's the only reason I have suggested that.  I will work with you to get this done any way you prefer.  
My suggestion is to expedite this PR by making it contain as little as possible that requires a review.  Strings are very critical and can take a long time to review and come to agreement on.

jamorham:
I don't really like the look of what looks like changes to the alerting logic. To snooze an alert you should just be able to call `  AlertPlayer.getPlayer().OpportunisticSnooze();` which is what is done for incoming alert snoozing from other sources. As far as I know this function works very well and doesn't need any changes to the alert code.

From what I understand this PR should be UI changes only to support the lock screen snoozing.

Navid200:
I would like us to work on this and make it do what it needs without doing anything else.
Your original PR changed 9 files.  This one changes 31.  
We should be able to accomplish our common intent with a simpler PR.

I could be wrong about all your intentions.  My hope is for the user to be able to snooze an alert without having to look at the screen.

The existing mechanism in xDrip requires xDrip to be in foreground.  This is why the existing mechanism is not adequate.  I always think of someone who gets an alert in an inconvenient situation for example while driving.  The intent is to be able to silence the alert without having to take your eyes off the road.

I understood from your original PR that you had found a way to force Android to bring up an xDrip page on screen when you pressed the volume button.  Now that xDrip is in foreground, pressing the button a second time should snooze it because we already have the mechanism.  
Am I misunderstanding your intent or your findings?
If not, we should be able to accomplish this still with changing only 9 or even less files.  

I am more than happy to work on this to assist you if you are OK with that.

jamorham:
The documentation looks like it probably should be attached to the PR rather than added to the source tree as it seems to relate to the changes only.

petervanrijt:
Thank you!

### Processed

- :white_check_mark: merged [NightscoutFoundation:master](https://github.com/NightscoutFoundation/xDrip/tree/master) into [alarm-snooze-v2](https://github.com/petervanrijt/xDrip/tree/alarm-snooze-v2) — local merge 13 July 2026 13:44, pushed 14 July 2026.
- :white_check_mark: Comment [4961115319](https://github.com/NightscoutFoundation/xDrip/pull/4623/#issuecomment-4961115319) @Navid200: "It will take a while to make the strings perfect. I suggest removing all the strings from this PR."
- :white_check_mark: Comment [5014902562](https://github.com/NightscoutFoundation/xDrip/pull/4623/#issuecomment-5014902562) @jamorham: "The documentation looks like it probably should be attached to the PR rather than added to the source tree as it seems to relate to the changes only." 

### Help

In response of comment [5008280148](https://github.com/NightscoutFoundation/xDrip/pull/4623/#issuecomment-5008280148) @Navid200: Ofcourse your help is more than welcome, please do. With a (third) newborn my time is somewhat limited :sweat_smile:.

### To answer the other questions

Snoozing itself uses the existing path: the snooze screen and the volume keys both end up in AlertPlayer.Snooze() — the same call OpportunisticSnooze() wraps. I deliberately didn't call OpportunisticSnooze() itself, because its internal 3-second rate limit would swallow the second press of the double-press confirmation, and it shows its own toast. 

Why UI alone doesn't get there: volume keys need a foreground activity to receive key events, and the lock-screen UI needs the notification's full-screen intent — which Android silently drops on upgraded installs whose alert channel importance is below HIGH. That is the only reason for the small channel migration; it enables the screen, nothing else. A MediaSession-based approach could intercept volume keys without any UI, but it competes with other media apps for the keys and behaviour varies per OEM — for a glucose alarm I preferred the predictable failure mode (if no screen shows, the alarm simply keeps sounding).

Beyond the UI there are, deliberately, a few small non-UI changes:

1. After a snooze from any source (snooze screen, notification, watch, broadcast), the pop-up snooze screens close and the ongoing notification refreshes right away — small hooks in Snooze()/stopAlert(). Reason: a snooze screen must never stay up, and the notification must never show a sounding alert, for an alert that has already been handled elsewhere; the UI has to reflect the true alert state at all times.
2. GuessDefaultSnoozeTime() falls back to the per-type default when an alert has no snooze time set. Reason: the volume-key and lock-screen paths snooze without showing a picker, so the default they apply must always be a sane value.
3. The orphaned-alert fixes — ActiveBgAlert.alertTypegetOnly() no longer deletes the active alert record as a side effect when the alert type is missing, and REMOVE ALERT now clears its active alert. Reason: the double-press path relies on ActiveBgAlert.getOnly() returning consistent state; without these fixes a double press can show success while the alert state is inconsistent — apparent success without effect, the worst failure mode for a glucose alarm — and test alerts, exactly how users will try the new snooze paths, cannot be snoozed at all.

Over the coming week I'll retest everything against the documentation. If all tests pass, I'll mark this PR as Ready for review.

PR #4628, opened 2026-07-14
Missing notification channels

Navid200:
After bumping the SDK level to 26, we have some notifications that don't use a notification channels.
They were all shown under Other previously.  Now, Android just ignores them and does not show a notification.
One of the consequences is that the sensor expiry notification currently is not shown.

This PR adds a new notification channel for general use.
It also creates a new dedicated channel for sensor expiry.
We need to enhance the sensor expiry alert.  So, we will need a dedicated channel for it.  

I will be testing sensor expiry shortly.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/alert/SensorExpiry.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/alert/SensorExpiry.java
@@ -6,6 +6,7 @@ import static com.eveningoutpost.dexdrip.models.JoH.niceTimeScalarNatural;
 import static com.eveningoutpost.dexdrip.models.JoH.showNotification;
 import static com.eveningoutpost.dexdrip.models.JoH.tsl;
 import static com.eveningoutpost.dexdrip.utilitymodels.Constants.SENSORY_EXPIRY_NOTIFICATION_ID;
+import static com.eveningoutpost.dexdrip.utilitymodels.NotificationChannels.SENSOR_EXPIRY_CHANNEL;
 
 import com.eveningoutpost.dexdrip.R;
 import com.eveningoutpost.dexdrip.g5model.SensorDays;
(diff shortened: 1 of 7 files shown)


Navid200:
I have verified that the sensor expiry alert now creates a notification and sound as as it did before.
Another advantage of this is that the user can choose a custom sound for the sensor expiry notification.  And they will be able to disable vibration as well if they so desire.

Navid200:
@ahaarrestad 
I don't know why I cannot find you under the list of reviewers.  
Would you please have a look at this and let me know if you find anything that needs to be adjusted?

Navid200:
There is a consequence to this whole SDK bump obviously.  We need to use a notification channel for all notifications.
We can create a dedicated notification channel for every single notification in xDrip.  That will result in too many notification channels.
Or, we can use General that I have added in this PR for most if not all the ones that never had a notification channel.  The consequence of the latter is that they will all sound the same way.  

We have to have a dedicated notification channel for every alert.  This is to allow users to customize their alerts.  
I hope that we can live with one notification channel for all the other notifications.
But, if I am overlooking something, please let me know.

Navid200:
I have to say that I may not have found every existing notification in xDrip.
I feel if there is a particular notification that no one is aware of, perhaps it is OK if it never makes a sound.  
Is this a mistake?  

I am not sure if there is a negative consequence (other than not having a notification) to having a notification that Android ignores because it has no notification channel assigned to it.

ahaarrestad:
I am looking into this and will supply some comments and possibly a suggestion for improvement. Found a possible hickup and a win.

ahaarrestad:
Built this and checked it on the device (Android 16). The fix works — sensor expiry gets a notification again, and the cause is what you said: no channel, so Android dropped it.

jamorham (approved):

jamorham:
Why don't you make `public static void showNotification(String title, String content, PendingIntent intent, int notificationId, String channelId, boolean sound, boolean vibrate, PendingIntent deleteIntent, Uri sound_uri, String bigmsg, boolean highPriority) ()` function in `JoH` set the general channel id if `channelId` is null? 

Wouldn't that catch anything using `showNotification()` that doesn't have a channel id set?

Navid200:
@jamorham That is a much better solution I didn't think of.  Thanks
I will make that change.  That should be a smaller PR.

Navid200:
Testing now

Navid200:
Testing shows that sensor expiry notification is generated.  It should be good to go.

jamorham:
let me know when you think ready to merge

Navid200:
I am OK with this to be merged right now so that sensor expiry notifications are shown again.

PR #4630, opened 2026-07-14
Don't set CATEGORY_ALARM unconditionally in XdripNotificationCompat.build()

ahaarrestad:
Fixes the regression reported in #4625: a Suunto watch stopped showing xDrip alerts after 2026.07.12, because its companion app ignores notifications carrying the alarm category.

`XdripNotificationCompat.build()` now sets `CATEGORY_ALARM` on every notification passing through it. Before, only `AlertPlayer` set it, and only when the alert had "Override Silent Mode" ticked and the volume profile was not vibrate-only or silent.

Most users see no change: the default alerts are created with Override Silent Mode on, so they already carried the category. It only affects users who turned it off — which is exactly what the reporter did, precisely because the watch drops alarm-category notifications.

Measured on the real app (Android 16, same database and alert config, only this line differing, number icon on so the ongoing channel is at `IMPORTANCE_DEFAULT`):

| build | alert category | alert grouped | ongoing grouped |
|---|---|---|---|
| current | `alarm` | yes | no (`overrideGroupKey=null`) |
| this PR | none | yes | no |

The category is not what keeps the ongoing notification out of the Android 16 alert cluster (#4611). Posting the same burst with and without it groups identically; what separates them is the channel importance split, and that is untouched here. This holds with the number icon on as well, where the ongoing channel is created at `IMPORTANCE_DEFAULT` rather than `IMPORTANCE_LOW`.

Tests: two Robolectric tests. Without the fix, `buildDoesNotAddACategoryOfItsOwn` fails with `expected: null but was: alarm`. The second one guards the behaviour that has to survive — a category set by the caller, as `AlertPlayer` does, is passed through untouched. Full `./gradlew test` green.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/XdripNotificationCompat.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/XdripNotificationCompat.java
@@ -27,8 +27,6 @@ public class XdripNotificationCompat extends NotificationCompat {
         builder.setGroup(null);
         builder.setGroupSummary(false);
 
-        builder.setCategory(NotificationCompat.CATEGORY_ALARM);
-
         final Notification n = builder.build();
 
         UserError.Log.d(TAG, "NotifCompat: chan=" + id +
--- a/app/src/test/java/com/eveningoutpost/dexdrip/utilitymodels/XdripNotificationCompatTest.java
+++ b/app/src/test/java/com/eveningoutpost/dexdrip/utilitymodels/XdripNotificationCompatTest.java
@@ -1,0 +1,61 @@
+package com.eveningoutpost.dexdrip.utilitymodels;
+
+import static com.eveningoutpost.dexdrip.utilitymodels.NotificationChannels.BG_ALERT_CHANNEL;
+import static com.google.common.truth.Truth.assertWithMessage;
+
+import androidx.core.app.NotificationCompat;
+
+import com.eveningoutpost.dexdrip.RobolectricTestWithConfig;
+
+import org.junit.Test;
+import org.robolectric.RuntimeEnvironment;
+
+import lombok.val;
+
+/**
+ * Tests for {@link XdripNotificationCompat}.
+ * <p>
+ * The alarm category is what tells a paired watch that a notification is an alarm, and some watches
+ * drop notifications carrying it. Deciding whether an alert is an alarm belongs to the caller:
+ * {@code AlertPlayer} sets the category only when the alert overrides silent mode. This helper must
+ * therefore carry a caller's category through untouched, and never add one of its own.
+ *
+ * @author Asbjørn Aarrestad - 2026.07
+ */
+public class XdripNotificationCompatTest extends RobolectricTestWithConfig {
+
+    @Test
+    public void buildDoesNotAddACategoryOfItsOwn() {
+        // :: Setup
+        val builder = alertBuilder();
+
+        // :: Act
+        val notification = XdripNotificationCompat.build(builder);
+
+        // :: Verify
+        assertWithMessage("category of a notification the caller left uncategorised")
+                .that(notification.category).isNull();
+    }
+
+    @Test
+    public void buildKeepsTheCategorySetByTheCaller() {
+        // :: Setup
+        val builder = alertBuilder();
+        builder.setCategory(NotificationCompat.CATEGORY_ALARM);
+
+        // :: Act
+        val notification = XdripNotificationCompat.build(builder);
+
+        // :: Verify
+        assertWithMessage("category set by the caller, as AlertPlayer does when overriding silent mode")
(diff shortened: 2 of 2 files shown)


Navid200:
Every xDrip notification is not marked as an alarm.  The title of this PR is misleading.  
If a watch does not trigger an alarm for an alarm, but does for any other notification, the problem seems to be in the watch and not in xDrip.  
I have provided more details in the linked discussion.

ahaarrestad:
Renamed to something descriptive. You're right that xDrip posts far more than six notifications — the six are just the ones routed through `XdripNotificationCompat.build()`, which is the only thing this PR touches. Everything that doesn't go through `build()` is unaffected either way.

The thing I'd flag is that `build()` didn't set a category at all before #4601 — `git blame` puts that line in with the targetSdk bump, alongside the grouping work, not as a decision to classify these as alarms. So this isn't removing a deliberate classification; it's restoring what `build()` did before 2026.07.12.

Of the six, only glucose alerts are alarms in the `CATEGORY_ALARM` sense, and `AlertPlayer` already sets that itself when the alert overrides silent mode (`AlertPlayer.java:574`). The other five — calibration reminder, "Other" alerts, compatible-apps prompt, parakeet-missing, and the generic `JoH.showNotification` path (sensor expiry etc.) — got the category only as a side effect of that one line. `CATEGORY_ALARM` isn't just a label: on the reporter's Suunto, the bridge forwards through a NotificationListenerService that drops alarm-category notifications, so those five plus any glucose alert with Override Silent Mode off stopped arriving.

On "the problem is in the watch" — the watch's handling didn't change; what changed is that xDrip started stamping the category on notifications that never carried it. Reverting that restores the prior behaviour for the Suunto with no change for default setups or xDrip-on-Wear (glucose alarms with Override Silent Mode on still carry the category, exactly as before). Worth confirming on the actual watch before merge, since I can only see what a notification listener sees.

Navid200:
I made the change because there were too many problems blocking progress.
Eventually, the problems were resolved.  It is possible that some changes are unnecessary and we can undo them.

All alerts, not just the glucose level alerts, are alerts that someone with diabetes has to deal with on a daily basis.  OK, not all.  You may hardly ever deal with the missed reading alert.  But, when the day comes that alert goes off, you had better hear it.  Waking up in the morning finding out that you have had no readings for the past 8 hours and for some reason, the alert made no sound is not a nice surprise.

When one of these alerts triggers, any of them, not just the glucose level alerts, they should not be grouped with the ongoing notification.
If the user enabled override silent mode for one of these alerts, the phone should respect that.  This is now controlled by the notification channel.  I am working on this now to make sure the user will never be unpleasantly surprised.

I know that the change has caused the watch to fail.  I have never disagreed with any of these statements.

I am simply saying we cannot just undo any of these simply because a watch is malfunctioning.  I am so sorry that this watch is malfunctioning.  And I want to fix it.  But, we cannot just undo this because it is the cause of the watch's malfunction.  

We wouldn't have to do any of this if Google hadn't forced it on us.  I am not happy that we have had to make these changes.  We need to look at all of this and decide how we can deal with all of the issues that will come up.

The day we need to worry about is when we release the next stable release.  When that day comes, everyone will switch to our final product.  What we have now is a Nightly.  I am not happy with it.  We have to work on it.
But, we cannot just go undo things because something has failed.

Navid200:
I can run my tests on this so that I can be happy that it will not cause problems.  
But, has it been tested to verify that it will solve the problem on the watch?

ahaarrestad:
Yes. The reporter in #4625 field-tested this build on the actual Suunto on 15 July — photos in that thread: the alert arrives on the watch, and both Dismiss and Snooze from the watch snooze correctly through `SnoozeOnNotificationDismissService`. He confirmed it again on 16 July with a log.

One correction from my side, since it bears on whether this causes problems. I re-measured notification behaviour on Android 16 / targetSdk 26, and removing the blanket `CATEGORY_ALARM` does cost something for other alerts: with the category they sound through DND priority, without it they are muted. It does not cost the ringer silent/vibrate override, because other alerts don't have that today either way — the channel is created with `USAGE_NOTIFICATION`, and the category alone doesn't override the ringer. I had that wrong in #4632 and I'm correcting it there.

So the tradeoff in this PR is: watch alerts work again, other alerts lose the DND-priority bypass. If you want alerts to stay categorized as alarms, that is the decision to make — and #4632 is where it should be made, since the two are the same mechanism.

ahaarrestad:
Closing this — #4680 removes the same line.

Its diff on `XdripNotificationCompat.build()` deletes `builder.setCategory(NotificationCompat.CATEGORY_ALARM)`, which is everything this PR does to production code. Once it lands there is nothing left here.

The two tests in this PR guard exactly that line, and going by the diff they pass against #4680's head unchanged. @Navid200 take them if you want them — otherwise I'll open them on their own once yours is in. There's no test on that behaviour today.

I'll reopen this if #4680 doesn't go through.

PR #4633, opened 2026-07-16
Harden notification-channel fallback and name the shared channels

ahaarrestad:
## What

Two small hardening fixes in the shared notification path, plus tests for the
whole chain.

1. `XdripNotificationCompat.build()` fell back to `BG_ALERT_CHANNEL` when the
   channel guesser threw. `getChan()` never creates that id (it posts to
   `id + hash`), so on API 26+ the notification is silently dropped. Fall back
   to `GENERAL_CHANNEL` instead and create it if missing, and log the degraded
   path.

2. `GENERAL_CHANNEL` and `SENSOR_EXPIRY_CHANNEL` were missing from the channel
   name map, so they showed under their raw id in Android's channel settings.
   Added the names already used elsewhere ("General", "Sensor expiry").

## Why

The fallback is a safety net, not a routing path — but if it ever fires it must
land on a channel that exists. General is also the channel the null-channel
source fix (#4628) now routes to, so it needs a readable name.

Glucose-alert behaviour is unchanged: those resolve through `getChan()` as
before and never touch the fallback.

Trade-off: the degraded fallback lands on `GENERAL` (IMPORTANCE_DEFAULT) rather
than the per-alert HIGH. That is strictly better than a silent drop.

## Tests

Robolectric, no device needed:
- `JoHShowNotificationTest` — drives `showNotification()` end to end and asserts
  it is delivered on a channel that exists (null-channel, sensor-expiry, and the
  shorter overload).
- `XdripNotificationCompatTest` — the fallback regression: an unresolvable
  builder still delivers on an existing General channel; `build()` never leaves
  a null channel id.
- `NotificationChannelsTest` — extended: channels are keyed by their
  customisation, setup methods create named channels, ongoing channel stays
  silent; kept the reflection canary.

Full `./gradlew test` green.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NotificationChannels.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NotificationChannels.java
@@ -75,6 +75,8 @@ public class NotificationChannels {
         map.put(BG_PERSISTENT_HIGH_CHANNEL, xdrip.getAppContext().getString(R.string.persistent_high_alert));
         map.put(CALIBRATION_CHANNEL, xdrip.getAppContext().getString(R.string.calibration_alerts));
         map.put(ONGOING_CHANNEL, "Ongoing Notification");
+        map.put(GENERAL_CHANNEL, "General");
+        map.put(SENSOR_EXPIRY_CHANNEL, "Sensor expiry");
     }
 
 
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/XdripNotificationCompat.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/XdripNotificationCompat.java
@@ -18,8 +18,11 @@ public class XdripNotificationCompat extends NotificationCompat {
         try {
             id = NotificationChannels.getChan(builder).getId();
         } catch (Exception e) {
-            // Fallback to generic alert channel if the guesser fails
-            id = NotificationChannels.BG_ALERT_CHANNEL;
+            // Guesser failed. Fall back to the general channel and make sure it exists,
+            // otherwise the notification is silently dropped on API 26+.
+            id = NotificationChannels.GENERAL_CHANNEL;
+            NotificationChannels.setupGeneralChannel();
+            UserError.Log.e(TAG, "getChan() failed, fell back to GENERAL_CHANNEL: " + e);
         }
         builder.setChannelId(id);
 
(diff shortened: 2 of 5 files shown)


Navid200:
Is there a malfunction that the user currently experiences that this PR is addressing?  If yes, can you please provide steps for recreating the malfunction?  

If not, I suggest we focus on addressing the issues that affect users right now first.  There will be some of this that will be addressed in the process anyway.

ahaarrestad:
No live repro — that's the nature of it. The fallback in `build()`'s catch points at `BG_ALERT_CHANNEL`, which `getChan()` never creates, so if that catch ever fires the notification is dropped silently on API 26+. Today the catch only fires if the reflection in `getNotificationFromInsideBuilder()` breaks, so it's latent, not something a user hits now.

No objection to prioritising the live issues first. This can wait — it's a safety net for the shared funnel, not a change to any alert behaviour. Happy to leave it open until the rest is cleared.

jamorham:
Have other changes superseded this or is this still required?

Navid200:
I have significant changes coming to the notification channels.  It is of course a PR and can be rejected.

In summary, Other alerts will use media player instead of relying on the notification channel for making sound or vibration etc.
Please hang on to all PRs that are related to notification channels.

Unfortunately, the facebook group  being hacked and other issues coming up like my NSEmulator PR needing more work are keeping me away from finishing up my notifications PR.

I respectfully suggest putting on hold all PRs related to notification channel for a few weeks.  Thanks

ahaarrestad:
Still required — nothing has superseded it. It builds on @Navid200's #4628 rather than competing with it: the fallback now lands on his `GENERAL_CHANNEL`, which is what his own comment on that constant says it is for.

`XdripNotificationCompat.build()` calls `NotificationChannels.getChan(builder).getId()` inside a try, and `getChan` returns null by design when the `mNotification` reflection isn't available — *"Note we return null to fallback old behavior if we can't get reflected access"*, `NotificationChannels.java:219`. That null is dereferenced immediately, so the catch is the intended path for that case rather than an exotic one. The catch then sets the bare `bgAlertChannel` id, and `getChan` only ever creates `bgAlertChannel<hash>`; the only bare channels the app creates are the icon test, general and sensor expiry ones. On API 26+ a notification posted to a channel that doesn't exist is dropped, so the "fallback to old behaviour" is silence.

It's still latent — the reflection works against the androidx version we ship, so nobody hits it today. It's a safety net for the shared funnel, not a change to any alert behaviour.

@Navid200 asked on 5 Aug to hold notification-channel PRs while he reworks Other alerts onto MediaPlayer. Happy to leave this open until that lands, then rebase.

ahaarrestad:
Closing this — #4680 supersedes it.

The reason this PR existed was the fallback in `build()`: the catch set `BG_ALERT_CHANNEL`, which `getChan()` never creates, so on API 26+ the notification was dropped silently. #4680 removes the try/catch entirely and lands a null channel id on `GENERAL_CHANNEL`, which it does create in `setupAllChannels()`. That covers it better than this did.

@jamorham — this is the real answer to your 5 Aug question. It wasn't superseded then. It is now.

Some of the tests here still apply, some go with the code they cover: `SENSOR_EXPIRY_CHANNEL`, `setupTestChannel` and `setupGeneralChannel` are all gone in #4680. I'll sort out which ones are worth carrying over once it lands.

I'll reopen this if #4680 doesn't go through.

PR #4636, opened 2026-07-17
Add scheduled transmitter ID switch for G7 soak & handoff

databoy2k:
## Summary
Discussed in #4635 — thanks @Navid200 for the discussion there.

Adds an optional **New Transmitter ID** field to the Dexcom (OB1) data source settings, for users who "soak" a new G7/One+ before the current one expires. Note that this assumes that users only use one live G7 at a time; this is not for soaking while reading two CGMs.

It automates the manual changeover: queue the next sensor's pairing code and a pre-selected delay, and xDrip performs the switch when the delay expires and the user taps a notification.

## What it does
- Enter the next sensor's pairing code in **Settings → Hardware Data Source → New Transmitter ID**, then a delay (default 30 min, remembers last used, capped at the end of the current sensor's life to ensure uninterrupted readings).
- At the scheduled time a **New Sensor Ready** notification appears. The notification opens System Status and then sets the new ID, clears the old session, and restarts the collector. The user must still approve the Android pairing request (not automatable).
- The pending switch is shown in the field summary and under System Status → Other Notes.
- Cancel by clearing the new field. Change new transmitter id or delay by going into the new field as well. Manually changing the active ID clears any pending switch.

## Scope / safety
- **Entirely opt-in** — with no next ID entered, no code path changes.
- **Reuses the existing manual transmitter-ID change path** — no changes to the BLE scan/connect/collection code, so it is no different on the Bluetooth side from a manual switch.
- ~230 net lines across 5 code/resource files, plus a documentation page.

## Known limitation
The notification rides the normal reading cycle rather than an exact alarm, so it appears within a few minutes of the target time while data is flowing (the intended case: switching a still-active sensor). If the app isn't running and no data is flowing at the scheduled time, it appears once data resumes. The schedule itself survives reboot/crash (stored in preferences). This is noted in the docs page.

## Testing
Built and exercised on an emulator (settings flow, delay dialog, summary, System Status notes — screenshots in the docs page), and running on my own device across multiple real changeovers.

## Docs
Adds `Documentation/G7_Soak_and_Handoff.md`, linked from the README.


ahaarrestad:
Had a proper look at this and built and ran it on an emulator. The overall approach looks sound — reusing the existing manual transmitter-ID path keeps the BLE side untouched, which seems like the right call.

Three things I think need addressing, plus a note on tests.

**1. The `createOngoingNotification` changes look unrelated to this feature**

The hunk in `Notifications.java` (`useOngoingChannel()`, `setGroup("xDrip ongoing")`, `setPriority(PRIORITY_HIGH)`, the channel-less builder branch) re-introduces code that was removed in #4601 (`10f57b82b`) and #4628 (`67c19479d`).

I checked whether it regresses anything, and it doesn't. `createOngoingNotification` ends with `XdripNotification.build(b)` (`Notifications.java:711`), which overwrites the group and category and sets the channel via `getChan()` — and `getChan()` always returns `ONGOING_CHANNEL` regardless of what the builder carries. On an emulator with default prefs (`use_notification_channels=true`, `ongoing_notification_channel=false`), `dumpsys notification` gives `channel=ongoingChannel ... sound=null category=status`, same as master. `setPriority` is ignored on API 26+ with minSdk 26, and the `setSound(null)` branch is unreachable with default prefs.

So no harm done, but it's dead code that conflicts with cleanup merged a few days ago. Suggest dropping the hunk to keep the PR to the feature.

**2. `dex_txid_next` shows on every collection type**

`dex_txid` is explicitly removed from `collection_category` for non-Dexcom collectors and re-added for DexcomG5 (`Preferences.java:2084` and `:2590`). `dex_txid_next` has no equivalent, so "New Transmitter ID" appears for Libre, NS Follow, Medtrum and the rest.

Confirmed on an emulator running `dex_collection_method=NSFollower` — the field sits between the collector selector and "Nightscout Follow URL", and the entry dialog opens normally:

New Transmitter ID showing on a Nightscout Follower setup

It isn't inert there either — if it gets filled in, `checkSoakTimer` will raise "New Sensor Ready" notifications and `SystemStatusFragment` will overwrite `dex_txid`.

**3. The delay dialog can lose a queued switch silently**

In `showDelayDialog`:

- OK with an empty or non-numeric delay skips the whole `if (isNumeric(val))` block, so the ID is never stored. The user has typed the pairing code and pressed OK, but nothing is queued and there's no feedback.
- The negative button is labelled "Cancel" but clears `dex_txid_next`, `dex_txid_next_time` and `dex_txid_delay`. With `setCancelable(false)` there's no way to open the dialog and leave an existing schedule intact.
- While a switch is pending, `setOnPreferenceClickListener` returns true and consumes the click, so the text dialog never opens and the ID can't be edited or blanked. The docs say "Cancel it: clear the New Transmitter ID field (enter a blank)", which isn't reachable in that state.

**Automated unit tests**

I'd like to see automated unit tests added for this before it goes in. The reason is that this feature schedules a changeover on a live sensor, and all three problems above are the quiet kind — nothing throws, nothing logs, the user just finds out later that the switch didn't happen or that their queued ID vanished. Tests are what catch that class of bug.

The test infrastructure is already there, so this shouldn't be much work: Robolectric 4.16 is configured in `app/build.gradle`, and `PrefTest` and `NotificationChannelsTest` under `app/src/test/` are good templates to copy from.

*One refactor first, to make it testable*

Right now the logic lives inside a `PreferenceFragment` and inside `SystemStatusFragment.onResume()`, so there's no way to exercise it without driving the UI. On top of that, the due check is duplicated verbatim in three places:

- `SystemStatusFragment.checkAndPerformAutoSwitch()`
- the block added to `SystemStatusFragment.setNotes()`
- `Notifications.checkSoakTimer()`

If those three ever drift apart, the settings summary, the notification and the actual switch start disagreeing with each other about when the changeover is due — which would be very confusing to debug from a user report.

Both problems have the same fix: pull the three prefs (`dex_txid_next`, `dex_txid_next_time`, `dex_txid_delay`) and the due predicate into one small plain class, with the current time passed in as a parameter rather than read from `JoH.ts()` inside. The three call sites then share one implementation, and everything below becomes an ordinary JUnit test with no Android UI involved.

*What to cover, and why each one matters*

- **The due predicate** — not due before `switchTime - 60000`, due exactly at it, due after, and not due when the ID is blank or the time is 0. This is the core timing rule and it's currently written out three times.
- **`setAndValidateDelay` clamping** — `< 1` → 1, above remaining sensor life → capped, remaining `<= 0` → no cap applied. There's a real edge case worth pinning here: if the remaining sensor life is under a minute, `maxMinutes` works out to 0, and the `< 1` clamp has already run by that point, so minutes ends up 0 and the switch fires immediately instead of being refused.
- **Empty or non-numeric delay** — this is the silent no-op from point 3. A test asserting the ID actually gets stored (or that the user gets told it didn't) is what stops it coming back.
- **`updateNextSwitchTime`** — `switchTime == now + delay * 60000`, and a blank ID resets it to 0. Cheap to test, and it's the arithmetic the whole schedule rests on.
- **The queue/cancel lifecycle** — queueing sets all three prefs consistently, blanking the ID clears both the time and the delay, a manual `dex_txid` change clears the pending switch, and Cancel leaves an existing schedule alone. This is where a half-cleared state would leave a switch armed that the UI says is gone.
- **Prefs visibility per collection type** — assert `dex_txid_next` is removed from `collection_category` for non-Dexcom collectors, parameterised over the collection types. That covers point 2. The new field is the requirement here; the existing `dex_txid` handling isn't yours and I'm not asking you to take it on. That said, once the parameterised test exists, adding `dex_txid` to the same cases is close to free and would put a first test around behaviour that's currently maintained by hand in two places — worth doing if it falls out easily, fine to skip if it doesn't.

databoy2k:
NGL, the second point is such a no-brainer that I'm quite embarassed that I didn't catch it. Sorry. 

The first point is because I started developing this long before those two PRs. We'll catch it up and make sure that we're on a modern branch. 

I'll look at those tests as well.

Sorry guys - I really didn't want this to waste a bunch of your time. Fixes incoming in the next few hours; can I tag you @ahaarrestad? Thanks for your feedback.

databoy2k:
Ok; took a lot of advice from Claude about how to best set this up. Here's what it came up with and we've implemented:
1. Removed stale createOngoingNotification hunk. createOngoingNotification and useOngoingChannel() are back to master; the only Notifications.java changes now are the soak additions (the checkSoakTimer call, soakTimerAlert, the notification id, and the OtherAlert content-intent overload). Like I said earlier - I branched this before #4601 and #4628 landed.

2. Fixed dex_txid_next on every collection type. It now follows the same DexcomG5 gate as the g5_settings_screen prefs: added to collection_category only for DexcomG5 and removed otherwise, in both the initial setup and the collection-method-change path. AFAIK this field no longer shows for Libre/NSFollow/Medtrum/etc.

3. Silent switch losses:
- Empty/non-numeric delay now falls back to the default and still queues, instead of the silent no-op.
- The Cancel button now dismisses and leaves any existing schedule untouched (and it's no longer setCancelable(false)).
- The click handler no longer consumes the tap, so the id editor always opens pre-filled with the pending id: editing re-queues, blanking cancels. Documented this.
- Re your setAndValidateDelay note: the "<1 min of sensor life left → maxMinutes 0 → fires immediately" edge is now caught. clampDelay returns a cannot-schedule sentinel and the dialog refuses with a message instead of switching on the spot.

We refactored and tested; the three prefs and the due predicate are now in a small plain class, SoakSchedule (cgm/dex/), with the current time passed in. All three call sites (checkAndPerformAutoSwitch, setNotes, Notifications.checkSoakTimer) go through SoakSchedule.isDue(now), so they can't drift. SoakScheduleTest (Robolectric, off your PrefTest template) covers the due-predicate boundaries, clampDelay (incl. the <1-min refusal regression), the switch-time arithmetic, and the queue/deactivate/clearAll lifecycle.

4. Notes from Gemini over in Antigravity (which... accidentally got pushed and had to get yanked back out)
We looked for other silent failures and found two; both now fixed:
- Threading in performAutoSwitch: it was calling LocationHelper.requestLocationForBluetooth from a background thread via safeGetContext(), which falls back to the app context after a fragment detach, so instanceof Activity was false and the permission prompt got silently skipped (and it was UI work off the main thread regardless). Now the Activity is captured on the UI thread and the prompt runs via runOnUiThread.
- Notification lifecycle: the "New Sensor Ready" notification had no auto-cancel, so it lingered after the switch, a manual ID change, or a cancel. Added Notifications.cancelNotification and wired it into all three.

Also documented the 60s grace window in the docs page (it can fire a switch up to a minute early so it isn't missed between reading cycles; a consequence is that a sub-2-minute delay is effectively immediate — the 30-minute default and normal soak times are unaffected).

Builds clean, tests green.

On the parameterised prefs-visibility test you asked for: I've held off, because driving the whole Preferences fragment through Robolectric looked disproportionately brittle for that one assertion, and the fix routes dex_txid_next through the identical DexcomG5 gate as the existing Dexcom prefs (verified across collection types on the emulator). Happy to add it if you'd still like it pinned in code; didn't want to haul the fragment into a unit test without checking.

ahaarrestad:
Built the branch (ee38e975a), ran the full test suite (green) and checked the settings behaviour on an emulator.

All three points are addressed:

- The ongoing-notification hunk is gone. `createOngoingNotification` is now byte-identical to master, and the `useOngoingChannel()` helper it introduced is removed. The only remaining Notifications.java changes are the soak ones.
- `dex_txid_next` now follows the same DexcomG5 gate as `dex_txid`. Confirmed on an emulator: not present under Nightscout Follower, present under Dex — and via both routes, switching collector in place and reopening Settings fresh. `dex_txid_delay` stays hidden in both, as intended.
- The delay dialog no longer loses a queued switch. `SoakScheduleTest` covers the substance here — blank input falling back to the default and still queueing, the lifecycle around deactivate/clearAll, and `clampDelay` returning CANNOT_SCHEDULE under a minute of sensor life. That last one is the right call; refusing beats switching on the spot.

SoakScheduleTest is 15/15 here and the full suite is green. Routing all three call sites through `SoakSchedule.isDue(now)` removes the drift risk I was worried about.

That's everything from me on the code. Whether the feature itself goes in isn't my call; that's for the maintainers. CONTRIBUTING says to feel free to ask about progress if a PR looks like it's waiting on action.

databoy2k:
Thanks for the CR. @Navid200 any thoughts following up on the discussion?

Navid200:
I tried to explain my ideas in the discussion.  I will not repeat it all.  But, my point was this would complicate xDrip for addressing something that can be done already (may be with a few more taps) using xDrip as is.   

You then asked "what if I opened a PR ..."?
This is not my repository.  Everyone can open a PR.  I cannot tell anyone they cannot open a PR.  So, I said yes, you can.  
It is not me who has to decide if the PR will be approved or not.  

I can answer any questions you may have here or in the discussion if you have any for me.
I hope I have not misled you.  If I have, I am sorry.

databoy2k:
No worries. I opened the pr so that we could continue with the code rather
than talking in theory. I understood the point of opening a discussion just
to be to root out completely useless ideas.

Point taken. Let's see if it gets merged. As i said, if it's stupid then
let it get nuked. If there's value here, I'd say ship it.

I for one will be cherrypicking it going forward. So much easier to have
xdrip handle my whole cgm rather than spreading it around other apps.

PR #4644, opened 2026-07-21
Fix crash loop when Wear OS integration is enabled

simelis:
With wear_sync enabled, the app is killed by Android roughly every 30 seconds:

  FATAL EXCEPTION: main
  android.app.RemoteServiceException$ForegroundServiceDidNotStartInTime
  Exception: Context.startForegroundService() did not then call
  Service.startForeground(): ServiceRecord{... .wearintegration.
  WatchUpdaterService ...}
      at CollectionServiceStarter.startServiceCompat(...:510)
      at CollectionServiceStarter.start(...)

startForegroundService() is a promise that the started service will call startForeground() shortly afterwards. WatchUpdaterService never does: it is a WearableListenerService, it holds no ongoing notification, and unlike the collectors (and PebbleWatchSync, which extends ForegroundService) it declares no android:foregroundServiceType in the manifest. It was never meant to be a foreground service - everywhere else in the app it is started with a plain startService(), see Home and NewDataObserver. The file's own TODO already states the invariant this violates: "everything started here must be capable of being foreground if it is started with compat".

The existing try/catch cannot help, because the exception is delivered asynchronously on the main looper seconds after the start call has returned successfully.

Start it plainly instead. The crash is most visible while readings are missing - a sensor change, a pairing problem - because MissedReadingService then restarts the collector every couple of minutes and each restart re-issues the doomed call, so the app crash loops exactly when it is most needed. Losing the app also loses its foreground collector, so this compounds rather than recovers.

Existing behaviour is otherwise unchanged: collectors and the pebble service still start with startForegroundService(), and a plain start is safe from the background while a collector foreground service is running, which is how the app already starts this service from NewDataObserver on every new reading.

Fixes #4672

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/CollectionServiceStarter.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/CollectionServiceStarter.java
@@ -251,7 +251,7 @@ public class CollectionServiceStarter {
             if (prefs.getBoolean("wear_sync", false)) {//KS
                 boolean enable_wearG5 = prefs.getBoolean("enable_wearG5", false);
                 boolean force_wearG5 = prefs.getBoolean("force_wearG5", false);
-                startServiceCompat(WatchUpdaterService.class);
+                startPlainServiceCompat(WatchUpdaterService.class);
                 if (!enable_wearG5 || (enable_wearG5 && !force_wearG5)) { //don't start if Wear G5 Collector Service is active
                     startBtWixelService();
                 }
@@ -276,7 +276,7 @@ public class CollectionServiceStarter {
             if (prefs.getBoolean("wear_sync", false)) {//KS
                 boolean enable_wearG5 = prefs.getBoolean("enable_wearG5", false);
                 boolean force_wearG5 = prefs.getBoolean("force_wearG5", false);
-                startServiceCompat(new Intent(context, WatchUpdaterService.class));
+                startPlainServiceCompat(new Intent(context, WatchUpdaterService.class));
                 if (!enable_wearG5 || (enable_wearG5 && !force_wearG5)) { //don't start if Wear G5 Collector Service is active
                     startBtShareService();
                 }
@@ -293,7 +293,7 @@ public class CollectionServiceStarter {
             if (prefs.getBoolean("wear_sync", false)) {//KS
                 boolean enable_wearG5 = prefs.getBoolean("enable_wearG5", false);
                 boolean force_wearG5 = prefs.getBoolean("force_wearG5", false);
-                startServiceCompat(new Intent(context, WatchUpdaterService.class));
+                startPlainServiceCompat(new Intent(context, WatchUpdaterService.class));
                 if (!enable_wearG5 || (enable_wearG5 && !force_wearG5)) { //don't start if Wear G5 Collector Service is active
                     startBtG5Service();
                 } else {
@@ -318,7 +318,7 @@ public class CollectionServiceStarter {
             if (prefs.getBoolean("wear_sync", false)) {//KS
                 boolean enable_wearG5 = prefs.getBoolean("enable_wearG5", false);
                 boolean force_wearG5 = prefs.getBoolean("force_wearG5", false);
-                startServiceCompat(new Intent(context, WatchUpdaterService.class));
+                startPlainServiceCompat(new Intent(context, WatchUpdaterService.class));
                 if (!enable_wearG5 || (enable_wearG5 && !force_wearG5)) { //don't start if Wear G5 Collector Service is active
                     startBtWixelService();
                 }
@@ -511,6 +511,21 @@ public class CollectionServiceStarter {
         } catch (Exception e) {
             // If foreground fails (e.g. app is already in foreground), fall back to standard start
             mContext.startService(intent);
+        }
+    }
+
+    private void startPlainServiceCompat(final Class service) {
+        startPlainServiceCompat(new Intent(xdrip.getAppContext(), service));
+    }
+
+    // Starts a service which does not run itself in the foreground.
+    @SuppressWarnings("ConstantConditions")
+    private void startPlainServiceCompat(final Intent intent) {
+        try {
+            Log.d(TAG, String.format("Starting service: %s", intent.getComponent().getClassName()));
+            mContext.startService(intent);
+        } catch (Exception e) {
+            Log.e(TAG, "Could not start service: " + e);
         }
     }
 


Navid200:
Can you identify a change in xDrip that caused this.
There are many who use WearOS.  I am surprised that we have not got any complaints related to a crash from anyone.  

@ahaarrestad  Do you mind having a look at this please?

ahaarrestad:
@Navid200 i will look into this as soon as possible

simelis:
Three things have to line up:
Wear integration enabled.
CollectionServiceStarter.start() actually running. In steady state that's rare — boot, a settings change, a collection restart, an NFC scan. A single crash looks like the OS killing the app, and it comes straight back; easy to blame Android rather than xDrip.
The app being permitted to start an FGS at all. On Android 12+, when it isn't, startForegroundService() throws synchronously — the existing try/catch catches it, falls back to startService(), and no startForeground() obligation is ever created. So a lot of background-initiated starts fail silently and safely. The catch is masking the bug in exactly the cases where it would otherwise fire.
It only becomes a visible loop when the collector is already unhealthy: MissedReadingService then restarts collection every couple of minutes and each restart re-arms the doomed obligation. Which is why it surfaced for me during a sensor change with pairing trouble — 55 identical fatal exceptions in one crash buffer — and why a stable install would see at most an occasional restart, which may be hard to connect to Wear OS.

ahaarrestad:
@Navid200 I reproduced it, and it is possibly worse than the PR description suggests.

Emulator, API 36, master at f64625443, `wear_sync` on, collection method DexcomG5. No watch connected. Launching the app is enough:

```
D CollectionServiceStarter: Starting foreground service: ...WatchUpdaterService
W ActivityManager: Bringing down service while still waiting for start foreground:
                   ServiceRecord{... .wearintegration.WatchUpdaterService ...}
E AndroidRuntime: FATAL EXCEPTION: main
   android.app.RemoteServiceException$ForegroundServiceDidNotStartInTimeException
```

Dead about 30 seconds after launch, then again on restart. No sensor change, no pairing trouble, no watch.

`WatchUpdaterService` also looks like the only service `CollectionServiceStarter` starts this way that never calls `startForeground()`. I went through the fan-out from `getCollectorServiceClass()`: the follow services and `PebbleWatchSync` extend `ForegroundService`, the G5 services extend `G5BaseService`, Medtrum extends `JamBaseBluetoothService`, and `DexCollection`/`DexShare`/`Wifi`/`DoNothing` use `ForegroundServiceStarter` directly. `UiBasedCollector` has no `startForeground()` either, but it is in neither `usesBluetooth` nor `newerCollector`, so this path never reaches it. So the fix looks scoped correctly to me.

@simelis a few things I couldn't settle from here — you have the devices, so they may be quicker for you.

1. There is one part of your explanation I'm unsure about. You wrote that on Android 12+ `startForegroundService()` throws synchronously when the app isn't permitted to start an FGS, and that the existing catch then falls back safely. As far as I can tell that restriction applies to apps *targeting* Android 12 (API 31), and xDrip targets 26 — so I don't think the catch is masking anything here. Have you seen it fire? If it doesn't, the bug may simply be less rare than your three conditions suggest, which is also what the repro above points to.

2. On @Navid200's question about what changed: #4601 raised targetSdk from 24 to 26 and landed on master on 12 July (bb5c5c766). You were on a roughly six-month-old alpha before switching to nightly — do you know whether the crash ever happened on that older build? If not, targetSdk looks like the trigger, and building current master twice with only `minSdkVersion`/`targetSdkVersion` changed would settle it. I haven't verified whether the `startForeground()` obligation is targetSdk-gated at all, so this is only a hypothesis.

3. One thing about the fix itself. `WatchUpdaterService` is started before the collector in all four branches, so on the boot path (`AutoStart` → `restartCollectionServiceBackground`) no foreground service is up yet, and a plain `startService()` from the background can be refused for an app targeting 26 — in which case the new catch would swallow it and wear sync would quietly not start. Since you have the patch running: does the log show `Could not start service:` after a reboot? Moving the wear start after `startBt*Service()` would cover it, though not the `force_wearG5` branch, where no collector is started at all.

On tests, in case it comes up: I don't think this can be covered by a unit test. Robolectric 4.16's `ShadowContextImpl.startForegroundService()` is `return startService(intent)`, so both calls land in the same recorded list with no way to tell them apart, and it doesn't model the ActivityManager timeout that actually kills the app. An instrumented test could, using the logcat signature above.

I have the emulator set up for this now, so happy to run any of these if that's easier.

simelis:
Thanks — this is a much sharper analysis than my writeup, and you've caught a real error in it. I also pulled fresh data off the master to settle 2 and 3.

1. The catch-masking claim is wrong — drop it. I conflated two different things: the background-FGS-start ban (ForegroundServiceStartNotAllowedException, target-31+, which doesn't apply since xDrip targets 26) with the did-not-start-in-time kill (ForegroundServiceDidNotStartInTimeException, which is what actually fires). Because the app targets 26, startForegroundService() doesn't throw synchronously and the catch never fires for that reason, so nothing is masked. My "three conditions" framing was wrong, and your launch-time repro is the honest picture: wear_sync on + any CollectionServiceStarter.start() = crash.

2. targetSdk looks like the trigger, and I was wrong to call it "not a regression." My earlier reasoning only covered the guard removal — I didn't look at the targetSdk 24→26 bump in the same PR, which is the better lead. Data point from here: the master ran an old alpha (~November 2025, so pre-bump, target 24) and I never saw this crash on it; it only began after moving to nightly. That's consistent with the kill being enforced at target-26 while the same code was dormant at target-24 — i.e. the mechanism predates the PR, but the enforcement was switched on by the targetSdk raise. I'd stop short of calling it proven, since an 8-month-old alpha differs in far more than targetSdk; your two-build experiment (current master, only min/targetSdk changed) is the clean isolator, and since you've got the emulator up that would be more authoritative than anything I can produce.

For what it's worth, the crash is well-corroborated on real hardware beyond the emulator: dumpsys dropbox on the master still holds a batch of ForegroundServiceDidNotStartInTimeException traces, all from the last pre-fix build (7e62326ec), and none from any build carrying the fix (running 1+ day since, including a reboot).

3. Rebooted the master with logcat capturing across the boot to check your point directly. The boot path runs exactly as you described — wear started before the collector FGS:

— but Could not start service: does not appear, there's no crash, and wear sync comes up normally (100+ WatchUpdaterService events since boot). So the BOOT_COMPLETED background-start allowance covers the plain start even in the wear-before-collector ordering. On this device the reorder isn't required for correctness, though I'm happy to add it as hardening if you'd prefer it in — it'd also cover non-boot background restarts that land outside any grace window, at the cost of not helping the force_wearG5 branch (where no phone-side collector is started, and the service comes up on watch traffic via its WearableListenerService filters anyway).

Tests: agree entirely. ShadowContextImpl.startForegroundService() just delegates to startService() and records both in the same list, and it doesn't model the AM timeout, so a unit test can neither tell the calls apart nor reproduce the kill. An instrumented test keyed on the logcat signature is the only thing that'd catch a regression. Of the experiments, the two-build targetSdk isolation is the one I'd most value from your side, if it's quick on the emulator you already have set up.

ahaarrestad:
Ran the two-build isolation on the emulator (`Medium_Phone_API_36.1`, API 36). Master `f64625443` built twice, only min+targetSdk changed, everything else identical (`wear_sync` on, `DexcomG5`, no watch, launch only):

- **target 26:** `ForegroundServiceDidNotStartInTimeException` on `WatchUpdaterService` at exactly +30s, then again on restart.
- **target 24:** no crash. Same path — `WatchUpdaterService` starts and lingers with `startForegroundCount:0` (the `startForeground()` obligation is still unmet), but the platform doesn't enforce the timeout for a target-24 app. Process stayed up well past the deadline.

So it's targetSdk-gated: the 24→26 bump in #4601 is what turned the missing `startForeground()` into a hard crash. Your fix is the right call.

(Had to drop minSdk to 24 in the app and three local modules to build target 24 at all, since min can't exceed target — doesn't affect the result.)

jamorham:
This looks to be the identical issue to reported: https://github.com/NightscoutFoundation/xDrip/discussions/4672

Question is, is this the correct fix or should the wear updater also be a foreground service? 

@Navid200 what would the behavior have been prior to the android 7 changes - assuming that user's left the default for running services in the foreground option enabled?

Navid200:
There was a complaint that there were too many logs related to automatic WearOS update failing.  So, we added a checkbox to disable automatic updates:
https://github.com/NightscoutFoundation/xDrip/pull/4384  

But, there was never any report of crashes related to Wear OS before retiring Android 7.

simelis:
> This looks to be the identical issue to reported: #4672
> 
> Question is, is this the correct fix or should the wear updater also be a foreground service?

Yes — #4672 is the same crash: same service, same exception, same trigger. ahaarrestad's finding there that it needs targetSdk 26+ matches what I saw on my devices. Happy to add "Fixes #4672" to the description so the issue closes with this.

On the question — I believe the plain start is the correct fix, and making the wear updater a foreground service would be the wrong direction:

1. **It isn't that kind of service.** WatchUpdaterService extends Google's `WearableListenerService`, whose lifecycle the system manages — Android starts it by itself whenever the watch sends something. It runs for a few seconds to shuttle data to or from the watch, then stops. Foreground services are for long-running work the user must be aware of, like the collectors holding a Bluetooth connection around the clock. This is short, event-driven work.

2. **A foreground service must show a permanent notification.** Everyone using wear sync would get a second ongoing xDrip notification for a service that is idle almost all the time. It would also need a `foregroundServiceType` declared once targetSdk reaches 34, and the type that fits (dataSync) is exactly the one Google has started putting runtime limits on in Android 15. It fixes nothing today and buys problems later.

3. **Plain start is already how the rest of the app starts this exact service.** Every new reading triggers one via NewDataObserver, and Home and the settings screens do the same — all plain `startService()`. CollectionServiceStarter was the one call site doing it differently. This change brings the odd one out in line with a pattern that has run dozens of times a day on every wear-sync device for years.

4. **The failure modes aren't symmetric.** A plain start can only be refused while the app is fully in the background with no foreground service running. I verified the tightest case on a real device: at boot, this service started successfully *before* the collector's foreground service was up (apps get a grace period after boot). If a start were ever refused anyway, the cost is one skipped watch update, repeated automatically on the next reading — against the current cost, which is Android killing the entire app, collector included. And promoting it wouldn't remove that corner case either: from targetSdk 31 up, starting a foreground service from the background is itself forbidden, so the foreground route ends up back at the same restriction with a notification on top.

jamorham (approved):
ok lets try this

PR #4645, opened 2026-07-21
Fix followers being left with a stale collector status

simelis:
A follower can keep showing an old collector status indefinitely - most visibly "Searching for <transmitter>" long after the master has found the sensor and is collecting normally again.

The master sends its status only when the text changes, and marks it as sent by storing it before the send is attempted. The send can then not happen at all: sendNanoStatusUpdate() drops the update when its 30 second rate limit is in effect, which is easily hit while a sensor start moves the status through several values, and the send is deferred by half a second so it is also lost if the process dies first. Either way the stored value already matches what was computed, so the update is never retried and the follower is stuck with whatever it last received.

Store the status once it has actually been handed to the sender, so a dropped update is simply picked up again on the next cycle.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/GcmActivity.java
@@ -516,11 +516,14 @@ public class GcmActivity extends FauxActivity {
         return false;
     }
 
-    public static void sendNanoStatusUpdate(final String prefix, final String json) {
+    // returns true when the update was actually queued for sending
+    public static boolean sendNanoStatusUpdate(final String prefix, final String json) {
         if (JoH.pratelimit("gcm-nscu" + prefix, 30)) {
             UserError.Log.d(TAG, "Sending nano status update: " + prefix + " " + json);
             sendMessage("nscu" + prefix, json);
+            return true;
         }
+        return false;
     }
 
     public static void sendMimeoGraphUpdate(final String json) {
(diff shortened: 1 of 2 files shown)


Navid200:
Can you please explain how I can recreate the problem this PR is supposed to address?

simelis:
Recipe A — the 30 second window (no crash needed)
Master + follower connected; note the status the follower shows.
On the master, cause a collector status change and confirm the follower picks it up. Easiest lever: set the transmitter ID to a bogus value → Searching for <id>.
Within 30 seconds, cause a second change — set the ID back.
The second update is swallowed by the gcm-nscu rate limit but is already recorded as sent. The follower stays on Searching for <id> — not for 30 seconds, but until some later, unrelated status change happens to get through.
Master-side confirmation (debug logging): Sending nano status update: appears for change #1 and not for #2, while the master's own UI shows the new status.
Recipe A also occurs naturally during a sensor stop/start, since StopSensor calls keepFollowerUpdated() directly and the status then moves through several values quickly.
Recipe B — how I actually hit it
The master was crash-looping for an unrelated reason (the WatchUpdaterService foreground-service bug in the companion PR — this is how the two are connected). Each crash landed inside the 500 ms deferral, so the update was stored and never sent. Result: my follower displayed Searching for <transmitter> for hours after the master had paired a new sensor and was collecting normally. Any repeated process death — battery optimiser kills included — reproduces it.

Navid200:
This is for xDrip Sync, right?  
If the follower does not get readings for a while, it will provide a missed reading alert.  Wouldn't the problem resolve if you tap on restart collector in the follower?

simelis:
> This is for xDrip Sync, right? If the follower does not get readings for a while, it will provide a missed reading alert. Wouldn't the problem resolve if you tap on restart collector in the follower?

It's not about missed readings, it is about a status line near the bottom like "Searching for XXXX". This message got stuck on my (follower) phone for hours, even though the readings were coming in and my child's master phone had dismissed this message long ago.

Restarting the collector on the follower cannot resolve it, because the follower isn't holding the problem — the master is. The "fix" without this patch could only be done on the master side: force its status text to change (e.g. restart its collector). The follower's owner — the person actually looking at the wrong message — can do nothing, and has no way to know the message is stale rather than current.

jamorham:
This makes sense, have you tested that this functions correctly with changes to the status between master and follower?

simelis:
> This makes sense, have you tested that this functions correctly with changes to the status between master and follower?

I've been running a version with this fix on both master and follower phones since I posted this PR, so about 3 weeks already, and everything looks to be working fine, including regular status changes. Didn't do the comprehensive before/after fix comparison though.

jamorham (approved):

PR #4646, opened 2026-07-21
Ns readonly treatment download

simelis:
# Let xDrip read AAPS-entered BG checks (and notes) from Nightscout, without turning on upload

**The problem**

A finger BG check entered in AAPS never appears in xDrip. This is by design on the AAPS side, not a bug you can see: AAPS broadcasts *carbs and insulin* to xDrip locally (its XdripPlugin sends them straight over), but it does **not** broadcast therapy events like BG checks — those it sends only to Nightscout. So the only way xDrip can obtain an AAPS BG check is to read it back from Nightscout. Today xDrip refuses to download treatments from Nightscout unless the *upload* master switch is also on — which would make xDrip a second uploader to a site AAPS already uploads to, duplicating everything.

**How to reproduce** (xDrip + AAPS on one phone, AAPS uploading to Nightscout):

- Enter **carbs** in AAPS → appears in xDrip within seconds (local broadcast).
- Enter a **BG check** in AAPS → never reaches xDrip; it goes only to Nightscout.

(Verified by watching the broadcasts: carbs go out as `info.nightscout.client.NEW_FOOD` to xDrip; the BG check produces no broadcast toward xDrip at all, only an NSClientV3 upload. xDrip's own AAPS receiver even has a commented-out `mgdl` read at `NSClientReceiver.java:303`, but it's moot — AAPS sends nothing to read.)

**What this changes**

- Download no longer depends on the upload switch. Upload remains its own switch, default off, and with it off xDrip **never uploads** — this is a read-only Nightscout treatment client, not "upload disabled."
- The Base URL and download screen are freed from the upload dependency, and download is skipped quietly when the URL is unset/placeholder/invalid instead of throwing.
- Downloaded blood tests are fanned out over xDrip Sync, so in a family the other phones receive them without each configuring Nightscout.
- A non-Finger entry (e.g. an AAPS BG check defaults to type *Sensor*) is logged once instead of every download pass.

**What this is not**

- **Not a second data source.** Glucose readings still come from the single configured source (xDrip Sync, Dexcom, etc.). Only *treatments* are pulled — the same treatment download that already exists, minus the upload gate.
- **Not the Nightscout-follower data source.** This is the *Cloud Upload → Nightscout* treatment download, which is independent of the data source. That's deliberate: it serves a device whose readings arrive some other way (in my case, Dexcom or xDrip Sync) while its treatments live in Nightscout.

**For the reviewer — the open question**

The confusion seen in comments is probably about the names of the settings and their placement: a read-only download option living under *Cloud Upload* menu reads oddly. The suggested alternative — make the *Nightscout-follower* data source download treatments — wouldn't serve this case, because the reading source here is Dex or xDrip Sync+ Follower or anything else, so switching to Nightscout follower is not an option - nightscout is not a source, but a supplement for the data which is not received over the other existing channels, like local broadcast from AAPS. If we could rename "Cloud Upload" to "Cloud Sync", that would probably remove the discrepancy.  Also, the Enabled switch in the root of Nightscout Sync would probably have to be renamed to "Enable Uploads". That is your call and I'm happy to make any required changes to the settings, which I tried to keep minimal with this PR.

**Why it's worth the complexity**

Any AAPS user can now see AAPS-entered BG checks in xDrip's graph — with xDrip's 10-minute delayed marker and a visual match against sensor values, which AAPS doesn't offer — and keep BG checks, carbs and insulin all in one place, entered once in AAPS.

---

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/DoNothingService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/DoNothingService.java
@@ -21,6 +21,7 @@ import com.eveningoutpost.dexdrip.utilitymodels.Constants;
 import com.eveningoutpost.dexdrip.utilitymodels.ForegroundServiceStarter;
 import com.eveningoutpost.dexdrip.utilitymodels.InstalledApps;
 import com.eveningoutpost.dexdrip.utilitymodels.NanoStatus;
+import com.eveningoutpost.dexdrip.utilitymodels.NightscoutUploader;
 import com.eveningoutpost.dexdrip.utilitymodels.Pref;
 import com.eveningoutpost.dexdrip.utilitymodels.StatusItem;
 import com.eveningoutpost.dexdrip.ui.helpers.Span;
@@ -139,6 +140,14 @@ public class DoNothingService extends Service {
                 if (minsago > 6) {
                     if (Home.get_follower()) GcmActivity.requestPing();
                     sleep_time = (minsago < 60) ? ((minsago / 6) * 1000) : 1000; // increase sleep time up to 10s for first hour or revert
+                }
+
+                if (Home.get_follower()) {
+                    // read-only Nightscout treatments download (BG checks / notes entered in
+                    // AAPS) also from the background, not only when the Home screen is opened
+                    // - no-op unless cloud_storage_api_download_enable is set, internally rate
+                    // limited and never uploads anything
+                    NightscoutUploader.launchDownloadRest();
                 }
 
                 try {
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutTreatments.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutTreatments.java
@@ -28,6 +28,7 @@ public class NightscoutTreatments {
 
     private static final HashSet<String> bad_uuids = new HashSet<>();
     private static final HashSet<String> bad_bloodtest_uuids = new HashSet<>();
+    private static final HashSet<String> logged_non_finger_ids = new HashSet<>();
 
     public static boolean processTreatmentResponse(final String response) throws Exception {
         boolean new_data = false;
@@ -90,7 +91,9 @@ public class NightscoutTreatments {
                                 UserError.Log.d(TAG, "Already a bloodtest with uuid: " + uuid);
                         }
                     } else {
-                        if (JoH.quietratelimit("blood-test-type-finger", 2)) {
+                        // the same entries reappear in every download pass - complain only once
+                        // per entry, but keep rechecking so a later edit to Finger gets imported
+                        if (logged_non_finger_ids.add(nightscout_id)) {
                             UserError.Log.e(TAG, "Cannot use bloodtest which is not type Finger: " + tr.getString("glucoseType"));
                         }
                     }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutTreatments.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutTreatments.java
@@ -1,5 +1,6 @@
 package com.eveningoutpost.dexdrip.utilitymodels;
 
+import com.eveningoutpost.dexdrip.GcmActivity;
 import com.eveningoutpost.dexdrip.Home;
 import com.eveningoutpost.dexdrip.models.BloodTest;
 import com.eveningoutpost.dexdrip.models.DateUtil;
@@ -30,6 +31,7 @@ public class NightscoutTreatments {
 
     public static boolean processTreatmentResponse(final String response) throws Exception {
         boolean new_data = false;
+        boolean new_bloodtest = false;
 
         final JSONArray jsonArray = new JSONArray(response);
         for (int i = 0; i < jsonArray.length(); i++) {
@@ -77,6 +79,7 @@ public class NightscoutTreatments {
                                 bt.uuid = uuid; // override random uuid with nightscout one
                                 bt.saveit();
                                 new_data = true;
+                                new_bloodtest = true;
                                 UserError.Log.ueh(TAG, "Received new Bloodtest data from Nightscout: " + BgGraphBuilder.unitized_string_with_units_static(mgdl) + " @ " + JoH.dateTimeText(timestamp));
                             } else {
                                 UserError.Log.d(TAG, "Error creating bloodtest record: " + mgdl + " mgdl " + tr.toString());
@@ -204,6 +207,11 @@ public class NightscoutTreatments {
                     }
                 }
             }
+        }
+        if (new_bloodtest) {
+            // fan newly downloaded blood tests out over xDrip sync so other family
+            // devices get the marker without downloading from Nightscout themselves
+            GcmActivity.syncBloodTests();
         }
         return new_data;
     }
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutUploader.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/NightscoutUploader.java
@@ -190,13 +190,34 @@ public class NightscoutUploader {
     }
 
     public static void launchDownloadRest() {
-        if (Pref.getBooleanDefaultFalse("cloud_storage_api_enable")
-                && Pref.getBooleanDefaultFalse("cloud_storage_api_download_enable")) {
+        // Download does not require upload (cloud_storage_api_enable) to be on: with only
+        // the download switch enabled this is a read-only Nightscout treatments client,
+        // e.g. so a device can receive BG checks / notes entered in AAPS via Nightscout
+        // while something else (AAPS) remains the sole uploader.
+        if (Pref.getBooleanDefaultFalse("cloud_storage_api_download_enable")
+                && haveValidBaseUrl()) {
             if (JoH.ratelimit("cloud_treatment_download", 60)) {
                 final NightscoutUploader uploader = new NightscoutUploader(xdrip.getAppContext());
                 uploader.downloadRest(500);
             }
         }
+    }
+
+    // quietly skip download when the base url is unset, template or unparseable
+    private static boolean haveValidBaseUrl() {
+        final String base = Pref.getString("cloud_storage_api_base", "").trim();
+        if (base.isEmpty() || base.contains("{") || base.contains("}")) return false;
+        for (String url : base.split(" ")) {
+            url = url.trim();
+            if (url.isEmpty()) continue;
+            try {
+                new URI(url);
+                return true;
+            } catch (URISyntaxException e) {
+                // try any further space separated entries
+            }
+        }
+        return false;
     }
 
     public static boolean isNightscoutCompatible(String url) {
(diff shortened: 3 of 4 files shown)


Navid200:
Thanks for your PRs.

If a particular issue is being addressed, please explain how we can recreate that issue.
I am looking at your description, and I see references to xDrip Sync and follower uploading and logs repeating ...
Would you please tell me how I can recreate each one of those issues?

It is not a good idea to provide your API_SECRET to someone you want to follow you.  You can provide a read token to them.  Then, they can only read and will never be able to upload anything:
https://navid200.github.io/xDrip/docs/Nightscout/Tokens.html

simelis:
On the read token — you're right, and it's compatible with this PR.
Agreed that sharing an API_SECRET is bad practice, and a read-only token is the correct way to let someone follow you. But that's orthogonal to what this PR fixes, and I think it actually reinforces it.
My case isn't following a third party — it's my own child's Nightscout site, where AAPS is the sole uploader and xDrip should only read. The problem is not which credential is used, it's that xDrip will not download unless the upload switch is on. cloud_storage_api_enable genuinely enables the uploader: UploaderQueue sets the NIGHTSCOUT_RESTAPI destination bit, UploaderTask and SyncService act on it, and Treatments queues records for it.
So if a user follows your advice and uses a read token, they must still enable upload to get downloads — and xDrip will then queue and attempt REST uploads that cannot ever succeed against a read-only token. The read-token approach really wants a read-only nightscout client in xDrip, which is exactly what this PR makes possible.

Navid200:
Can you explain exactly how you are setting up xDrip as a Nightscout follower please?

Navid200:
You shouldn't use Nightscout under Cloud Upload for setting up xDrip as a Nightscout follower.  
This is how you should set it up:
https://navid200.github.io/xDrip/docs/Follow/FollowNightscout.html

simelis:
> You shouldn't use Nightscout under Cloud Upload for setting up xDrip as a Nightscout follower. This is how you should set it up: https://navid200.github.io/xDrip/docs/Follow/FollowNightscout.html

Oh but why not? As it does include downloading data from nightscout? I am happy with my current setup - my nightscout on free northflank tier is not stable enough to keep it as a single source of data for followers. I have my AAPSClient bound to nightscout, and using xdrips as xdrip sync followers, which has helped me a lot already - when nightscout is down, I still get readings on xdrip, and when xdrip sync is down (it happened a few times in a couple of years), I still get readings over nightscout. I've seen reports that the commercial nightscout is not too stable and gets offline from time to time too, so keeping it either xdrip sync or nightscout wouldn't be a good idea. Also, this PR is about downloading treatments from nightscout, not BG data.

Navid200:
> Oh but why not? 

Because that is not a data source.  
Nightscout follower is.  Please read the guide I provided so that you can see how everyone uses xDrip as a Nightscout follower.

xDrip can be a Nightsccout follower.  xDrip can also upload to Nightscout.  You have to set up xDrip differently depending on if it is meant to be a follower or an uploader.  
Please set it up as a follower.  
If you do that and still see a problem, we can talk about fixing it.

simelis:
> > Oh but why not?
> 
> Because that is not a data source. Nightscout follower is. Please read the guide I provided so that you can see how everyone uses xDrip as a Nightscout follower.
> 
> xDrip can be a Nightsccout follower. xDrip can also upload to Nightscout. You have to set up xDrip differently depending on if it is meant to be a follower or an uploader. Please set it up as a follower. If you do that and still see a problem, we can talk about fixing it.

xDrip can also be xDrip Sync follower and receive missing treatments from Nightscout with this little change. I don't want to set it up as nightscout follower because of the reasons I stated above. What is the reason to have this limitation?

Navid200:
OK, I finally understand what you wan to do.  
You could have made this much easier by explaining that you want to set the hardware data source to xDrip Sync and enable upload to Nightscout, but not allow it to upload.

But, why not just switch the hardware data source to the one that works at any time.  If xDrip Sync works, leave it like that.  When it stops working, change the hardware data source to Nightscout follower.
Wouldn't that let you accomplish what you are after?

simelis:
> OK, I finally understand what you wan to do. You could have made this much easier by explaining that you want to set the hardware data source to xDrip Sync and enable upload to Nightscout, but not allow it to upload.
> 
> But, why not just switch the hardware data source to the one that works at any time. If xDrip Sync works, leave it like that. When it stops working, change the hardware data source to Nightscout follower. Wouldn't that let you accomplish what you are after?

Not really, as what I initially want is to get the finger BG treatment data entered in AAPS and broadcasted to xdrip master and followers. It doesn't work now - do you know why? The finger bg check data entered in AAPS doesn't sync over local broadcast to xdrip. If it did, I would be happy too, but I'm afraid AAPS local broadcast change would be required for that, so I took this direction - getting that data from nightscout. I could possibly achieve what I want if I had one of the followers to follow nightscout - others would probably get that data synced from that one, but it would still be a workaround where a solution is possible. The proposed solution would actually benefit a setup with no followers at all - one could enter a bg check in aaps and get it in xdrip while using local broadcast, which probably doesnt send this data now (not sure, needs testing).
Anyway, don't you agree that getting nightscout treatment data in addition to the BG data received from whatever datasource would be a nice feature?

Navid200:
xDrip is a CGM.  AAPS is not.  
Why would you want to enter calibrations in AAPS?
Why would AAPS need to see blood glucose measurements?
Can you provide a link to what AAPS would do to a blood glucose measurement?

We shouldn't add redundancies to xDrip.  There is enough in it already.
The reviewer will need to know why this is needed.  
He needs to decide if the benefit justifies the complexity.

simelis:
AAPS does nothing with it, its for information only. And its not necessarily for calibrations, its more to let the parents know whats an actual BG. Its more convenient to use a single app to enter everything - aaps is used to enter carbs, insulin and other stuff anyway, and has a nice way to enter finger bg too - but we cannot use it because it simply doesnt sync to xdrip. I had a hard time making my child to enter finger bg data into xdrip, maybe because he is not used to it or what, but now when he can do it in AAPS (with this fix), it gets entered more often.

Navid200:
If we set up xDrip as a Nightscout follower and it does not download blood glucose readings from Nightscout, it makes sense to add that ability.  I didn't know it did not do that already.  

Setting up xDrip to upload to Nightscout and add a checkbox to disable its upload is not a good idea in my opinion.  If someone doesn't want xDrip to upload, they should not go to the Cloud upload menu in xDrip.

simelis:
No, it's not that. Take a setup with no followers at all - xdrip is with dexcom (or any other direct cgm) source, broadcasting BG data to AAPS over local broadcast, and receiving AAPS related data (insulin, carbs etc.) directly from AAPS. No nightscout involved. This setup does not sync finger BG check values (and many other treatment values) from AAPS to xDrip. We can either fix this local sync to include whats missing - that would probably involve a change to AAPS as well, or use nightscout as a mediator - as it has all data uploaded from AAPS, we can just download it in xDrip from nightscout, with just the proposed fix.

Navid200:
So, if you set up xDrip as a Nightscout follower, does it download blood glucose entries?

Navid200:
If AAPS broadcast to xDrip does not include blood glucose, we can ask AAPS to add that.  If xDrip blocks it from AAPS, we need to fix that.
We cannot add a switch to xDrip to stop uploading when it is meant to upload.

simelis:
It probably does, I don't really know that, as its not the follower sync problem that is being adressed.

> So, if you set up xDrip as a Nightscout follower, does it download blood glucose entries?

Navid200:
A setup that has no followers at all must have AAPS installed then.  That user can see all the blood glucose readings in AAPS already.

I was talking about your case where you are following your child.  You can set up your xDrip to use xDrip Sync or Nightscout follower.  If either one of those is skipping in downloading something, we can talk about fixing those.

simelis:
> If AAPS broadcast to xDrip does not include blood glucose, we can ask AAPS to add that. If xDrip blocks it from AAPS, we need to fix that. We cannot add a switch to xDrip to stop uploading when it is meant to upload.

We are not adding a switch to stop uploading. In the existing nightscout sync (rest api) infrastructure, we only enable the nightscout url entry and the switches to enable treatment data download while the upload switch is still off. That does look strange, as it looks like master switch, but it now would become not a master, but just a switch for the Upload functionality, letting the download functionallity to be toggled independently.

simelis:
> A setup that has no followers at all must have AAPS installed then. That user can see all the blood glucose readings in AAPS already.
> 

But he cannot see the AAPS finger BG checks in xDrip, which would be a lot more convenient as it provides a 10 minutes delay and a nice visual match of sensor vs finger readings inside the graph. AAPS doesnt have that.

Navid200:
> But he cannot see the AAPS finger BG checks in xDrip, which would be a lot more convenient as it provides a 10 minutes delay and a nice visual match of sensor vs finger readings inside the graph. AAPS doesnt have that.

So, the problem is that your son's xDrip does not show blood glucose readings from AAPS.
That seems like the problem that should be fixed.  Why are those not transferred to his xDrip?  Is that because something missing in AAPS or xDrip?

simelis:
Yes, we can narrow down the problem to that. xDrip does have a line for reading finger BG check value from AAPS, which is commented out. I haven't checked the AAPS side - will do it now to see if it gets sent from AAPS. 

Yet the provided fix involves more than just a finger BG readings - xDrip now also receives additional events like "AAPS started", "Missed BG readings", any notes entered manually in AAPS and much more. Should all of this be included in the data transferred from AAPS to xDrip locally? Or should it have some boundaries and send only the critical data, as the remaining not that critical treatments could be received from nightscout (using the provided fix)?

Anyway, will come back soon with the information about what's being sent from AAPS.

Navid200:
You can wait for the lead developer review.  Hopefully, he will not have as much trouble as I did understanding what the problem is.  

xDrip receives data only from one data source.  There have been requests to allow one to get data from two at the same time.  It has never been approved.
If we are going to allow something along those lines, it will be a new feature and should be easy to find for someone who doesn't know it exists.  A switch on the Nightscout upload page may not be where someone who has no idea such a feature exists would look.

Isn't what you are looking for the AAPS client?  It lets you see what your son sees in AAPS.

simelis:
I do have AAPS client and I can see what my son sees in AAPS, that is not a problem - I particularly want to see BG check data in xDrip, as AAPS doesn't provide a convenient way to match it with sensor data. Also, it would be neat to have all the data around, and not have to look for some things in one app, and other things - in another one.

It's not a "several data sources" problem. The datasource is still single, we just enable the treatment download from nightscout (which doesn't include the sensor BG data). That is already available with the nightscout upload enabled - it allows downloading from nightscout (while already having configured another physical data source), it just requires uploads to be enabled too, which I try to become an option, not a requirement.

simelis:
OK I also tested what AAPS actually does when I enter things.
I entered two things in AAPS, one after the other:

1 g of carbs — showed up in xDrip within seconds, on its own.
A blood-glucose check (a finger-prick value) — was never sent to xDrip at all. It only went up to Nightscout.

It's AAPS's XdripPlugin (app.aaps.plugins.sync.xdrip, DataSyncSelectorXdripImpl). It selects which data types to broadcast to xDrip — processChangedCarbs fired for the carbs and sent NEW_FOOD. There is no equivalent for therapy events, so BG checks never leave AAPS toward xDrip locally; they go only to Nightscout. That's the whole asymmetry, and it's a deliberate limitation on the AAPS side, not something xDrip can fix by listening harder.

simelis:
> You can wait for the lead developer review. Hopefully, he will not have as much trouble as I did understanding what the problem is.

Fair enough. I also updated the initial PR description to reflect our conversation here, hope it is more understandable now.

PR #4648, opened 2026-07-21
fix: detect source unit from companion app package name in UiBasedCollector

ryceg:
Split out of #4529 so it can be reviewed on its own merits.

Many companion apps encode their glucose unit in the package name (e.g. `.mmol`, `.mmoll`, `.mgdl`). UiBasedCollector previously interpreted scraped values using xDrip's own display unit preference, so an mmol/L value from a companion app could be parsed as mg/dL (or vice versa) when the two disagreed. This detects the source unit from the companion package name and falls back to xDrip's preference when the name carries no unit hint.

Includes unit tests for the package-name detection.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/UiBasedCollector.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/UiBasedCollector.java
@@ -371,7 +371,7 @@ public class UiBasedCollector extends NotificationListenerService {
         int mgdl = -1;
         try {
             val ftext = filterString(text);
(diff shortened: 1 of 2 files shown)


jamorham:
The unit test appears to be failing:

```
UiBasedCollectorTest > sourcePackageUsesMgDlTest FAILED
    camaps mmoll ca
    value of: sourcePackageUsesMgDl()
    expected to be false
        at com.eveningoutpost.dexdrip.services.UiBasedCollectorTest.sourcePackageUsesMgDlTest(UiBasedCollectorTest.java:257)
```

PR #4650, opened 2026-07-22
Keep receivers in the manifest

Navid200:
I removed the receiver from the manifest in order to satisfy the SDK 26 background restrictions when we retired Android 7.
As a result, Esel broadcasting to xDrip (explicit class targeting) failed as reported here: https://github.com/NightscoutFoundation/xDrip/discussions/4647 

This PR places the receivers back in the manifest to ensure compatibility with explicit broadcasts.
But, the receivers are wrappers to keep them lightweight.  

The intent is to still satisfy SDK 26, but have everything that was functional still functional.

A test release from this PR was created and verified by the reporter.  

I have repeated my test with BYODA with this PR and it remains operational.

Is this acceptable?


Navid200:
@ahaarrestad 

Do you see any oversights in this?

Thanks

ahaarrestad:
Verified on an emulator (targetSdk 26, API 34) — it restores the explicit-broadcast path that Esel uses.

One FYI, not a blocker: the actions are now registered both in the manifest and at runtime (`xdrip.onCreate`), so both receivers fire for a package-targeted broadcast while the app is running. Measured on this branch for `NS_EMULATOR`:

- explicit component (what Esel sends) → 1 delivery (manifest only)
- plain implicit → 1 (runtime only; manifest blocked on O)
- `setPackage(...)` implicit → 2 (both fire)

I checked whether that double delivery actually duplicates a reading, in follower mode with a real `entries`/sgv payload: it doesn't. Both deliveries reach `bgReadingInsertFromData`, but `readingNearTimeStamp()` dedups the second, so exactly one row is stored — same as a single delivery. So it's only wasted work (an extra thread/wakelock, double OOP2 decode), not a data problem. No internal code sends these five actions, so it only touches external senders that use `setPackage` instead of an explicit component — and Esel uses explicit, so it's fine.

Minor: the new `.NSEmulatorReceiver` entry has no `android:exported`, while `ExternalStatusBroadcastReceiver` sets `exported="true"`. Works today (defaults to true with an intent-filter at targetSdk 26) but would become a manifest-merger error if targetSdk ever moves to 31+ — worth making explicit for consistency.

Navid200:
@ahaarrestad Thanks

I have added android:exported="true" to the NSEmulatorReceiver entry for consistency and future-proofing as you suggested.

Navid200:
After that last updated, I have tested it again and it is working fine with BYODA.
The Esel user has also tested it again after the last update and is happy with it.

jamorham:
I think its easy to misunderstand the restrictions on receivers. As far as I know, manifest receivers still work fine if the xDrip package is specified by the broadcasting app. As far as I know that is always the case because broadcasts became unreliable unless the package was specified some time ago even without changing the target sdk version.

I believe that means that it isn't necessary to register the receivers programmatically and manifest only is fine, however I can't control how other apps may or may not be using broadcasts so I can't say for 100% certain.

My concern here is that registering both in the manifest and in the application startup is going to cause double delivery of messages which we don't want. 

What do people think?

Navid200:
I don't understand why BYODA does not work for me any longer.  All releases before retiring Android 7 work as receivers for BYODA.
I thought I had tested it and it had worked.  But, all releases after retiring Android 7 fail to get any readings from BYODA.

My logs show the broadcasts are being 'skipped by policy':

reason: skipped by policy at enqueue: Background execution not allowed: receiving Intent { act=com.eveningoutpost.dexdrip.NS_EMULATOR flg=0x30 xflg=0x4 (has extras) } to com.eveningoutpost.dexdrip/.NSEmulatorReceiver
Historical Broadcast #22:

I just tested and this PR makes it work.

How about this?
I can add a switch that would only appear if the data source is set to 640G/Eversense.  
The switch will say something like "Standard" and enabled by default.
The summary would explain that disabling the switch would use legacy mechanism.

Internally, I will set things such that we either use the manifest receiver path (legacy) or we use the code receivers (standard).  So, we will never have data coming from both paths.  
<br/>  
  
---  
  
I would also like to rename the 640G/Eversense to a more general title that would not refer to any particular product.  Something like "Inter-app (receiver)".  But, in line with keeping the PRs as small as possible, let's leave that for follow-up PR.

Navid200:
We may even be able to make the switch an automatic switch.  

We could start a method when xDrip starts and 640G is the data source, and also when 640G is enabled.
The method will enable the manifest path of 640G to begin with.  The code path of 640G will always be active.  
The method will monitor the code path of 640G.  As soon as it sees data coming from that path, it will disable the manifest path.
This way we will only get one data point from both paths after a start.  After that, data will arrive only from one path.

This will remove the need for a new switch and having to train users and keep the guide updated.

Navid200:
@ahaarrestad When you have a chance, would you please provide a critical review?

When xDrip starts or when the user switches the hardware data source to 640G, both paths, code and manifest, become enabled.
Following that, as soon as xDrip receives data from the code path, it will automatically switch off the manifest path.  So, we will receive one redundant data following a reboot or a data source switch.  But, after that, data will arrive only from on path.

There is a shortcoming that we have to live with:
Let's say the broadcasting app broadcast explicitly.  Therefore, data arrives from both paths and xDrip shuts down the manifest path.  Now, lets say the user decides to force stop xDrip.  
When the broadcasting app broadcasts its next data, xDrip will not wake up automatically.  In this case, the user must start xDrip or else, it will never start on its own.

ahaarrestad:
You asked for a critical review, so I put the three "Intelligent switch" commits on an emulator (API 34, targetSdk 26). I counted how many times the receiver ran, and watched the manifest receiver go on and off with `dumpsys package`.

The manifest entry works. Nothing I saw argues against it.

One thing surprised me: when a sender targets the receiver class directly, like Esel does, the broadcast never reaches the code path — so nothing ever calls the disable. For the Esel case this PR started from, the switch never comes into play.

The "one extra reading, then never again" part didn't match what I measured. Two places write that on/off state and neither knows about the other. Every reading on the code path turns the manifest receiver off, and `CollectionServiceStarter.start()` turns it back on for 640G on a timer — I measured 120 seconds, hitting the same second every time. Over five minutes with a reading every minute I got 7 deliveries for 5 broadcasts. The extra reading keeps coming back.

The part I'd like you to look at is what happens when xDrip's process is gone. The manifest receiver is what lets an incoming reading start it again. I killed the app, checked it was really gone, and broadcast:

- manifest on → xDrip started, reading arrived
- manifest off → nothing, app stayed dead

I didn't measure how quickly the scheduled jobs would have brought it back on their own. During those five minutes the manifest receiver was off about 80 % of the time. You raise this for force-stop, and it's wider than that — any time the process dies you land in the same place. The off state also survives a reboot and an app update, and nothing tells the user.

There's also a mismatch in what the two paths cover. The manifest entry declares one action, `NS_EMULATOR`. The code-registered receiver takes four — `NS_EMULATOR` plus the three `OOP2_DECODE_*` ones. `onReceive` doesn't look at the action before switching the manifest receiver off, so an OOP2 result from Libre decoding turns off the 640G cold-start path. The manifest receiver could never have received that broadcast, so there was no duplicate to prevent.

One more, and this one isn't a regression — nobody has the manifest path on master today. The receiver accepts `NS_EMULATOR` data in three cases (`NSEmulatorReceiverCore:77`): follower mode, 640G, or `external_blukon_algorithm`. Line 243 only enables the manifest entry for 640G, so the other two stay where they are now, with no way for an incoming reading to start xDrip. Worth noting that `get_follower()` is the plain `Follower` source only — it doesn't match NSFollow. If the manifest entry is the fix for cold start, those two cases have the same problem.

Since the switch doesn't come into play for the direct-broadcast case, and `readingNearTimeStamp()` already throws away the duplicate reading, my suggestion is to drop the three "Intelligent switch" commits — `6710a46f9` is what the Esel user tested and confirmed.

If you'd rather not drop commits, it's four edits on top:

- `NSEmulatorReceiver.onReceive()` — remove the `if (isStandardInstance) setManifestReceiverEnabled(context, false)` block
- remove the `isStandardInstance` field and the `NSEmulatorReceiver(boolean)` constructor
- `xdrip.java:120` — back to `new NSEmulatorReceiver()`
- `xdrip.java:110-112` — the enable on startup can go too, since nothing disables it any more

That leaves `CollectionServiceStarter:243` as the only place writing the state, keyed off the data source the user picked. No race, and 640G users still get the manifest path. That doesn't fix the follower/blukon gap above — widening the condition on line 243 to match the receiver's own gate would, if you want those cases covered too.

If you want to keep the switch, the minimum is that line 243 stops overwriting the decision — right now that's what undoes it every 120 seconds — and that the disable checks the action first.

Happy to send either as a patch, just say which.

Navid200:
@ahaarrestad If I change the automated switch to a manual switch, will all the problems be resolved?

The user will enable the switch if their broadcaster does not send explicit broadcasts.  And they will disable the switch if their broadcaster does send explicit broadcasts.
Will that address the two paths fighting?

Can you tell me the relationship between the NSEmulator and OOP2?
Are there devices that use OOP2 that broadcast to 640G/NSEmulator?

Navid200:
This comment is from my PR that retired Android 7.  
Screenshot 2026-08-06 094633
  
I was blinded by tests passing.  I should have looked at that comment and stopped the whole process.  

So, it seems to me, I should place the receivers back in the manifest and remove the receivers and wrappers I added to the code.

Navid200:
I will close this PR and open another PR that puts the receivers back in the manifest.

PR #4654, opened 2026-07-25
[cleanup] Replace android.preference.PreferenceManager with androidx (Batch 2: 4 files)

ahaarrestad:
## Why

`android.preference.*` was deprecated wholesale in API 29 and is no longer maintained. xDrip
already ships `androidx.preference` and already uses it in parts of the app, so today the codebase
carries **two** preference frameworks and each file silently picks one. That is the thing being
removed here, one small batch at a time.

Three reasons this is worth doing rather than leaving alone:

- **It unblocks Category 3.** The real work — `Preferences.java` and the
  `PreferenceFragmentCompat` migration — cannot be reviewed on its own merits while ~56 unrelated
  files still import the old `PreferenceManager`. Getting the mechanical swaps out of the way first
  means the interesting PR is only about the interesting change.
- **It removes deprecation warnings** from the build, so the ones that remain still mean something.
- **Small batches, because this is medical software.** A 56-file sweep is unreviewable. Five files
  or fewer per PR is reviewable in a sitting, and a regression has a small blast radius.

**Why it is safe:** both `PreferenceManager` classes return the *same* default `SharedPreferences`
file. No user data moves, no key is renamed, nothing is migrated. That claim is not asked for on
trust — it is what the tests below assert, and what the emulator run demonstrates.

## What

Category 2 (Batch 2) of the deprecated-API cleanup started in #4479, continuing #4595. Four files,
**one import line each**, no logic changes:

- `utilitymodels/Pref`
- `LicenseAgreementActivity`
- `utilitymodels/pebble/PebbleDisplayAbstract`
- `stats/DBSearchUtil`

```diff
-import android.preference.PreferenceManager;
+import androidx.preference.PreferenceManager;
```

Production diff is 4 files × (1 insertion, 1 deletion). Everything else in the diff is tests.

## Scope correction to #4479

While selecting files for this batch, two of them turned out **not** to be simple import swaps, so
the claim in #4479 that the whole category is `getDefaultSharedPreferences()`-only needs narrowing:

- `ColorCache.java:43` and `app/.../xdrip.java:95-101` call `PreferenceManager.setDefaultValues()`.
  That call **inflates the preference XML**, so under androidx it needs the XML's classes to extend
  `androidx.preference.Preference`. It is Category 3 work, not a one-liner. Both files have been
  moved out of Category 2 (58 → 56) and into Category 3 (8+1 → 8+3, alongside `Preferences.java`).
- The two are not equally visible if broken. `xdrip.java`'s call is unguarded, so a failure is a
  crash. `ColorCache`'s sits inside `catch (Exception e) {}`, so a failure is **swallowed** — the
  colour defaults would simply never load, with nothing in the log. Worth knowing before anyone
  attempts it.
- `wear/src/main/res/xml/preferences.xml` contains only stock tags, so `wear/xdrip.java` is
  **unproven, not broken**. I am not claiming it fails; I have not shown that it does.

Unit tests that only exercise `getDefaultSharedPreferences()` stay green while `setDefaultValues()`
breaks, which is why this was worth catching by reading rather than by testing.

**Batch 1 is unaffected.** I re-checked all five files merged in #4595 — none of them calls
`setDefaultValues()`; they are pure `getDefaultSharedPreferences()` swaps as described.

## Tests

Three new test classes, 13 tests. Each was run against the **original** `android.preference`
import and passed, then against the swapped import and passed identically. That ordering is the
whole point: a test that only ever saw the new import proves nothing about the old behaviour.

- `LicenseAgreementActivityTest` (4) — the stored `I_understand` preference decides whether the
  agreement checkbox is ticked on open, and Save writes the checkbox state back. Both directions,
  through the view hierarchy and the preference store.
- `PebbleDisplayAbstractTest` (5) — `getBatteryString()`, `getDexCollectionType()` and
  `isDexBridgeWixel()` translate stored preferences into what is sent to the watch, including the
  default when nothing is stored. Context is injected through the real `initDisplay()` entry point
  rather than by touching the protected field.
- `DBSearchUtilTest` (4) — the range counts classify readings against `highValue`/`lowValue`,
  interpreted in whichever unit `units` names. The same seeded readings are asserted under mgdl and
  under equivalent mmol thresholds, which is what pins the unit rescaling rather than the numbers
  being used raw.

Full `./gradlew test` green.

### Coverage limitation, stated plainly

`Pref` has **no new test**. Its `SharedPreferences` handle is a private static cache with no reset
seam, so a cross-API "old API writes, new API reads the same file" assertion cannot be written —
the first test to touch it fixes the instance for the rest of the JVM. It rides on the existing
`PrefTest` round-trip, and store identity is proven by the other three files instead. I would
rather say that than add a `@VisibleForTesting` reset seam to production code to make a test
possible.

## Manual emulator test (per the #4479 review condition)

`API34_Phone` (Android 14), FastDebug build of this branch.

**Part 1 — preference search**, the inherited script:

| Step | Result |
|---|---|
| Settings renders, search icon present | ✅ |
| Search opens; `glucose` returns live results | ✅ single- and multi-level breadcrumbs |
| Tap a nested result | ✅ navigates into the correct sub-screen |
| `insulin` | ✅ four-level breadcrumbs resolve |
| `zzqqxx` | ✅ graceful "No results" empty state |

**Part 2 — Statistics**, which is specific to this batch. `DBSearchUtil` backs the Statistics
screens and reads `units` / `highValue` / `lowValue` through the swapped `PreferenceManager`, so
rendering that screen is direct evidence here rather than a ritual carried over from Batch 1.

144 synthetic readings were written into `BgReadings` (12 h, 5-minute spacing, sine 55–260 mg/dL).
Host-side SQL over those rows gives 27 below / 48 in range / 69 above.

The app rendered **`Absolute #s: 48/69/27`** and **`Range Settings: 70 - 170 mg/dl`** — matching
exactly, with the thresholds read back through the swapped code. Had the swap resolved to a
different preference store, the thresholds would have fallen back to the hard-coded defaults and
the counts would not have matched. All three Statistics reports render.

`adb logcat -b crash` empty; no `InflateException`, no `PreferenceParser` errors.

### What this does not cover

- **`PebbleDisplayAbstract`** — no Pebble hardware and no Pebble emulator. Unit tests only, not
  manually verified.
- **`LicenseAgreementActivity`** — first-install screen; the emulator's install was already past
  it, so it was not re-triggered by hand. Covered by its unit tests.
- **`Pref`** — exercised throughout Part 1, but only indirectly, per the limitation above.

## Screenshots

| Settings + search | Search opened | `glucose` |
|---|---|---|
| Settings | Search opened | glucose |

| Nested result | `insulin` (deep breadcrumbs) | No-match |
|---|---|---|
| nav | insulin | no results |

| Statistics — range counts | Distribution | Percentile |
|---|---|---|
| ranges | pie | percentile |

## After this

52 Category-2 files remain in the app module. `wear/src/main` holds a further 27 with the same
import that no category list tracks — separate track, not this one. Batch 3 next, same shape.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/LicenseAgreementActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/LicenseAgreementActivity.java
@@ -3,7 +3,7 @@ package com.eveningoutpost.dexdrip;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.os.Bundle;
-import android.preference.PreferenceManager;
+import androidx.preference.PreferenceManager;
 import android.view.View;
 import android.widget.Button;
 import android.widget.CheckBox;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/stats/DBSearchUtil.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/stats/DBSearchUtil.java
@@ -4,7 +4,7 @@ import android.content.Context;
 import android.content.SharedPreferences;
 import android.database.Cursor;
 import android.database.sqlite.SQLiteDatabase;
-import android.preference.PreferenceManager;
+import androidx.preference.PreferenceManager;
 
 import com.eveningoutpost.dexdrip.models.JoH;
 import com.eveningoutpost.dexdrip.models.UserError.Log;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Pref.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/Pref.java
@@ -1,7 +1,7 @@
 package com.eveningoutpost.dexdrip.utilitymodels;
 
 import android.content.SharedPreferences;
-import android.preference.PreferenceManager;
+import androidx.preference.PreferenceManager;
 
 import com.eveningoutpost.dexdrip.models.JoH;
 import com.eveningoutpost.dexdrip.models.UserError;
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/pebble/PebbleDisplayAbstract.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/pebble/PebbleDisplayAbstract.java
@@ -4,7 +4,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.os.BatteryManager;
-import android.preference.PreferenceManager;
+import androidx.preference.PreferenceManager;
 
 import com.eveningoutpost.dexdrip.BestGlucose;
 import com.eveningoutpost.dexdrip.models.BgReading;
--- a/app/src/test/java/com/eveningoutpost/dexdrip/LicenseAgreementActivityTest.java
+++ b/app/src/test/java/com/eveningoutpost/dexdrip/LicenseAgreementActivityTest.java
@@ -1,0 +1,140 @@
+package com.eveningoutpost.dexdrip;
+
+import android.widget.Button;
+import android.widget.CheckBox;
+
+import androidx.preference.PreferenceManager;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.robolectric.Robolectric;
+import org.robolectric.RuntimeEnvironment;
+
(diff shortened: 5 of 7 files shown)


jamorham (approved):
thanks for such a comprehensive PR for this housekeeping work.

PR #4657, opened 2026-07-26
tests: wait for conditions instead of sleeping (suite 86.8s → 55.3s)

ahaarrestad:
Test-only change. No file outside `app/src/test/` is touched.

Six test classes waited for async Retrofit callbacks by sleeping a fixed 300–400 ms and then draining Robolectric's paused main looper once. This replaces that with polling for the condition the test is about to assert.

## Result

`:app:testFastDebugUnitTest` 86.8s → 55.3s (three clean runs, 55.3s each). 549 → 552 tests, 0 failures.

| Class | Before | After |
|---|---:|---:|
| `NightscoutUploaderCallerTest` | 19.595s | 0.515s |
| `NightscoutFollowEntriesWorkTest` | 4.635s | 0.492s |
| `NightscoutFollowDeviceStatusWorkTest` | 2.835s | ~1.45s ¹ |
| `JugglucoEmulatorContractTest` | 1.074s | 0.369s |
| `NightscoutFollowDeviceStatusFallbackTest` | 0.984s | 0.202s |

¹ The raw XML says 6.320s, but the class now sorts first in the suite and absorbs Robolectric's one-time sandbox bootstrap: first test 5.085s, the other six 1.234s combined. The same test measured 0.215s when the class did not run first.

## The 19.6s class

`MockWebServer.getHostName()` returns `localhost`. `NightscoutUploader.TryResolveName()` treats a host without a dot as an mDNS candidate, appends `.local` and blocks for `Mdns.NORMAL_RESOLVE_TIMEOUT_MS` (5 s) before falling back to the original URL. Three tests, three lookups, 15 s. The rest was a drain loop that kept calling `takeRequest` after it had already found its match, timing out 2 s twice.

Fixed in the test by starting the server on `127.0.0.1` explicitly. The address is pinned rather than derived from the server because an IPv6 loopback renders as `[::1]` in a URI — also dotless, same slow path.

**This is a production bug, not a test artefact.** Any Nightscout URL with a dotless host blocks the uploader 5 s per attempt (`NightscoutUploader.java:377` and `:485`). Not addressed here — it belongs in its own issue.

## Three tests did not verify what they were named

Found while removing the sleeps; this is the part that is not about speed.

- `work_sendsDeviceStatusRequest_whenNotRateLimited` asserted only `assertThat(server.getRequestCount()).isGreaterThan(0)`. It would have passed with the devicestatus call deleted from production entirely. It now waits for, and asserts on, the devicestatus request itself.
- `work_doesNotSendDeviceStatusRequest_whenRateLimited` and its sibling in `NightscoutFollowDeviceStatusFallbackTest` looped `for (i < server.getRequestCount())` over zero requests and passed on an empty loop. Verified by removing the new anchor: `expected to be at least: 1, but was: 0`. The 300 ms sleep was the only thing hiding it. The fallback one also wrapped its assertion in `if (r != null && r.getPath() != null)`, which skips silently.

Each of the three was verified discriminating by breaking its precondition on purpose and confirming it fails.

## `Await`

New test helper, `app/src/test/java/com/eveningoutpost/dexdrip/Await.java`. Drains the main looper, re-evaluates the condition every 5 ms, gives up after 2 s.

It never throws. The condition it waits on is usually the same one the test is about to assert, so throwing would pre-empt the real assertion and replace `expected 88.0 but was 65.0` with an uninformative timeout. Waiting is an optimisation; the assertion stays the only oracle.

`NightscoutFollowEntriesWorkTest` uses a local helper that *does* throw, because those tests assert on the contents of a request — if it never arrived there is nothing to assert against, and a named failure is the useful outcome.

Timing uses `System.nanoTime()`, not `JoH.tsl()`: `PersistTest` moves `JoH.tsl()` forward 100 hours via `ShadowSystemClock`, and a cap measured against a shifted clock does not work.

## Deliberately not changed

- **`PersistTest`** (~2.9 s) — investigated, left alone. `@Config(instrumentedPackages = {"…models.JoH"})` cannot be removed: without it `ShadowSystemClock.advanceBy()` does not reach `JoH.tsl()` and both tests fail. Verified by removing it. The cost is the second classloader and it is not reducible. Added a comment saying so, so the next person does not repeat the experiment.
- **Timing margin in the two negative tests** — the anchor now fires ~5 ms after the entries request instead of the old fixed 300 ms. Narrower in principle. A probe run passed 60/60 including under 3× CPU oversubscription. The alternative considered — `takeRequest(250 ms)` returning null — costs 250 ms per test in the passing case, which eats most of the win.

## Two choices worth a second opinion

- `Await` sits in `com.eveningoutpost.dexdrip` to match the package of the test classes that use it. A `test`/`support` package would be tidier if you would rather keep helpers separate.
- The three strengthened tests make this more than a pure speedup. Keeping them together was deliberate — the sleeps are what hid the vacuous passes, so the two are the same finding — but they will split cleanly if you want the correctness fixes reviewed on their own.

## Verification

`./gradlew :app:testFastDebugUnitTest :wear:testFastDebugUnitTest` green three times on the final commit. Timings above are from three `cleanTest` runs.

### DIFF
--- a/app/src/test/java/com/eveningoutpost/dexdrip/Await.java
+++ b/app/src/test/java/com/eveningoutpost/dexdrip/Await.java
@@ -1,0 +1,63 @@
+package com.eveningoutpost.dexdrip;
+
+import static org.robolectric.Shadows.shadowOf;
+
+import android.os.Looper;
+
+import java.util.function.BooleanSupplier;
+
+/**
+ * Waits for a condition to hold instead of sleeping for a fixed duration.
+ * <p>
+ * Retrofit delivers {@code enqueue()} callbacks on a background OkHttp thread and posts the
+ * result to the Android main looper. Robolectric's PAUSED looper does not run on its own, so
+ * a test must drain it before the effect of a callback is visible. Sleeping a fixed 300 ms and
+ * draining once is a bet that 300 ms is enough — a bet that does not hold on a loaded machine.
+ * <p>
+ * <b>This class never throws and never asserts.</b> The condition it waits on is normally the
+ * same one the test is about to assert, so throwing here would pre-empt the real assertion and
+ * replace a precise Truth message ("expected 88.0 but was 65.0") with an uninformative timeout.
+ * Waiting is an optimisation; the assertion that follows remains the only oracle.
+ *
+ * @author Asbjørn Aarrestad - 2026.07
+ */
+public final class Await {
+
+    /**
+     * Generous on purpose. Nothing is paid when the condition holds quickly, so a high cap
+     * costs nothing in the normal case while tolerating a slow or heavily loaded machine.
+     */
+    private static final long DEFAULT_TIMEOUT_MS = 2000L;
+
+    private static final long POLL_INTERVAL_MS = 5L;
+
+    private Await() {
+    }
+
+    /** Waits up to {@link #DEFAULT_TIMEOUT_MS} for the condition to hold. */
+    public static void awaitAtMost(final BooleanSupplier condition) {
+        awaitAtMost(condition, DEFAULT_TIMEOUT_MS);
+    }
+
+    /**
+     * Drains the main looper and re-evaluates the condition every {@value #POLL_INTERVAL_MS} ms
+     * until it holds or the cap elapses. Returns normally in both cases.
+     */
+    public static void awaitAtMost(final BooleanSupplier condition, final long timeoutMs) {
+        // System.nanoTime(), not JoH.tsl(): this measures real wall clock. JoH.tsl() can be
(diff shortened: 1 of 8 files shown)


jamorham:
looks helpful, let me know when ready for merge

ahaarrestad:
Synced to master and re-ran: 565 tests, no failures. That is 13 more than when I measured, since #4654 and #4595 added three test classes while this sat open — they run fine against the new waiting code.

Numbers are unchanged from the description: 86.8s → around 55s on `:app:testFastDebugUnitTest`. Re-timing after the sync gave 51s and 60s on two runs, so 55s is the honest figure, not the lower one.

Tests only, no production code touched.

Ready for merge.

jamorham (approved):

PR #4663, opened 2026-08-01
Add Medtrum Follower source to get glucose values from Medtrum EasyView service

T-o-b-i-a-s:
Please see https://github.com/NightscoutFoundation/xDrip/discussions/3429#discussioncomment-17865504 for more details.

I implemented a medtrum-follower source with AI support that was proposed in the linked discussion above to get glucose values from the Medtrum EasyView service. 
Additional information about the API was fed to the AI from https://github.com/rodecker/medtrum-tray and https://github.com/pachi81/GlucoDataHandler, thanks to these authors for their work.

The glucose values are only available when you have an Internet connection, similar to a Nightscout follower setup.

The prerequisites to use the feature are as follows:

* Create an additional account with a different e-mail address as a "caregiver" on the [Medtrum Easy View](https://easyview.medtrum.eu/v3/#/login/login) system
* After loggin in as the caregiver, add a connection to the main patient account by entering the e-mail address of the primary patient account you use with your EasyPatch app
* In the EasyPatch App, go to Settings -> Account Security and add the newly created caregiver account with the secondary e-mail address as a remote follower

In xDrip, then select "Medtrum Follower" as the source and enter the mail address of your secondary caregiver account as "EasyView Username" and its password under "EasyView Password".

### DIFF
--- a/wear/src/main/java/com/eveningoutpost/dexdrip/utils/DexCollectionType.java
+++ b/wear/src/main/java/com/eveningoutpost/dexdrip/utils/DexCollectionType.java
@@ -40,6 +40,7 @@ public enum DexCollectionType {
     NSFollow("NSFollower"),
     SHFollow("SHFollower"),
     Medtrum("Medtrum"),
+    MedtrumFollow("MedtrumFollower"),
     Disabled("Disabled"),
     Mock("Mock"),
     Manual("Manual"),
@@ -186,6 +187,7 @@ public enum DexCollectionType {
             case Medtrum:
                 return MedtrumCollectionService.class;
             case Follower:
+            case MedtrumFollow:
             case LibreReceiver:
                 return DoNothingService.class;
             case NSFollow:
@@ -265,6 +267,8 @@ public enum DexCollectionType {
                 return "Network libre";
             case NSFollow:
                 return "Nightscout";
+            case MedtrumFollow:
+                return "Medtrum EasyView";
             case SHFollow:
                 return "Share";
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -884,6 +884,9 @@
             android:name=".cgm.nsfollow.NightscoutFollowService"
             android:foregroundServiceType="dataSync" />
         <service
+            android:name=".cgm.medtrumfollow.MedtrumFollowService"
+            android:foregroundServiceType="dataSync" />
+        <service
             android:name=".insulin.inpen.InPenService"
             android:foregroundServiceType="connectedDevice|location" />
         <service
--- a/app/src/main/java/com/eveningoutpost/dexdrip/MegaStatus.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/MegaStatus.java
@@ -217,7 +217,7 @@ public class MegaStatus extends FloatingLocaleActivityWithScreenshot {
                 addAsection(NIGHTSCOUT_FOLLOW, "Nightscout Follow Status");
             }
             if(dexCollectionType.equals(MedtrumFollow)) {
-                addAsection(MEDTRUM_FOLLOW, "Medtrum Follow Status");
+                addAsection(MEDTRUM_FOLLOW, "Medtrum follow status");
             }
             if(dexCollectionType.equals(SHFollow)) {
                 addAsection(SHARE_FOLLOW, "Dex Share Follow Status");
--- a/app/src/main/java/com/eveningoutpost/dexdrip/MegaStatus.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/MegaStatus.java
@@ -10,6 +10,7 @@ import static com.eveningoutpost.dexdrip.Home.startWatchUpdaterService;
 import static com.eveningoutpost.dexdrip.utils.DexCollectionType.DexcomG5;
 import static com.eveningoutpost.dexdrip.utils.DexCollectionType.GluPro;
 import static com.eveningoutpost.dexdrip.utils.DexCollectionType.Medtrum;
+import static com.eveningoutpost.dexdrip.utils.DexCollectionType.MedtrumFollow;
 import static com.eveningoutpost.dexdrip.utils.DexCollectionType.NSFollow;
 import static com.eveningoutpost.dexdrip.utils.DexCollectionType.SHFollow;
 import static com.eveningoutpost.dexdrip.utils.DexCollectionType.WebFollow;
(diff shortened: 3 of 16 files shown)


jamorham:
Please confirm that this has been tested

T-o-b-i-a-s:
Hello @jamorham , I run the branch on my phone for two days now and also used the Medtrum follower source temporarily for running my AndroidAPS closed loop. The following screenshots show a comparison of the blood glucose graph from the Medtrum Follower source and a parrel running G6 sensor, a screenshot from the Medrum EasyPatch app and a screenshot from xDrip using this data at approximately the same time and finally the settings of my xDrip (in German) with the new Medrum Follower source settings.
medrum vs  G6
signal-2026-08-03-08-43-42-522
signal-2026-08-03-08-43-42-522_002
signal-2026-08-03-08-42-54-538_003

While testing, I experienced teh following challenges, which are inherent to the follower approach:
* Gaps in data delivery when no Inetrnet connection is available
* "Jumpy" graph as the original source data is delivered in mmol/l with a one digit after the point precision and converted to mg/dL
* The mg/dL data send to Nightscout also were float numbers with digits after the point, I added a fix for that in my local repo, but the commit was not included in this PR so far, I need to check how to do that. For PRs in local repos, commits added later are usually added to teh PR automatically.

Navid200:
Thanks for this work.

I always suggest avoiding to introduce any strings in a PR that brings in new features.
Then, the review can focus on the functionality and the code  standards and ignore grammar for now.

The problem with strings is that if the PR goes through, the strings will be translated.  After that, changing the strings will not be easy because it will impose extra work on the translators.

I suggest removing all strings.  
<br/>  
  
---  
  
This PR introduces a new data source using arrays that looks like this:  
<item>Medtrum Follower</item>  
  
I proposed using sentence case here:  https://github.com/NightscoutFoundation/xDrip/discussions/4408  
I add that jamorham never made a comment.  So, I am not 100% if he is OK with my proposal.  

My suggestion would be to change that to:
<item>Medtrum follower</item>   

I know that will be different than the other items on the list like Nightscout Follower.  
I would like to change all of those.  But, I have never got to it.

wear/src/main/java/com/eveningoutpost/dexdrip/utils/DexCollectionType.java
```diff
@@ -186,6 +187,7 @@ public static Class<?> getCollectorServiceClass(final DexCollectionType type) {
             case Medtrum:
                 return MedtrumCollectionService.class;
             case Follower:
+            case MedtrumFollow:
```
jamorham, line 190: I'm thinking this is not correct as it should point to the service class?

T-o-b-i-a-s: The current implementation maps it to the DoNothingService.class; for all of the following sources: 
Follower, MedtrumFollow and LibreReceiver. The other sources like NSFollow or Medtrum point to specific classes that have the same "Do Nothing" empty implementation for the wear artifact. I do not see added value in adding another empty class here. What would be your recommendation?

T-o-b-i-a-s:
@Navid200 , I extracted all visible string resources and converted them to sentence case.

Navid200:
You are still adding strings in this PR.
After strings are added to xDrip, translators will translate them.  If we then decide to change anything in them, it will be very difficult because it will require to ask translators to translate them again.

It is my advice for any PR to avoid adding any strings.
Every now and then, users identify text they prefer to appear in their language.  We then take those texts and convert them to strings after spending time on them making sure they are the best in English.

We can do that in this PR.  It means more work will have to be done before the PR is ready to merge.  That's the reason for my advice to speed up the review process.

T-o-b-i-a-s:
Hello @Navid200 , sorry that was a misinterpretation of your request "avoiding to introduce any strings in a PR that brings in new features." - I understood that no strings must be introduced in the code/classes themselves and rather separate them in the strings.xml - That is why I extracted even more strings into the xml file to comply with what I thought was your requirement. Should I revert these changes and have everything in plain English strings directly in the classes?

Navid200:
I am so sorry that my request was not clear.
I am happy to work with you to make sure the strings are OK.

Please focus on what jamorham has asked  you to do.  When it gets to the point that we get close to merge, I may make suggestions to change a word or two.  Hopefully, none.  

But, my advice in general is not to change the .strings file in any PR that is about anything other than strings.

jamorham:
Unit tests are failing:

```
MedtrumFollowDataTest > selectsConfiguredPatientAndConvertsMmol FAILED
    java.lang.AssertionError: expected:<127.91068899999999> but was:<128.0>
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:555)
        at org.junit.Assert.assertEquals(Assert.java:685)
        at com.eveningoutpost.dexdrip.cgm.medtrumfollow.MedtrumFollowDataTest.selectsConfiguredPatientAndConvertsMmol(MedtrumFollowDataTest.java:29)

MedtrumFollowDataTest > parsesValidGraphRowsAndSkipsMalformedOrOldRows STARTED

MedtrumFollowDataTest > parsesValidGraphRowsAndSkipsMalformedOrOldRows FAILED
    java.lang.AssertionError: expected:<90.07795> but was:<90.0>
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:555)
        at org.junit.Assert.assertEquals(Assert.java:685)
        at com.eveningoutpost.dexdrip.cgm.medtrumfollow.MedtrumFollowDataTest.parsesValidGraphRowsAndSkipsMalformedOrOldRows(MedtrumFollowDataTest.java:87)
```

Suggestion: Check conversion constant in use matches throughout and increase floating point tolerance in checks where it tolerance may be set at 0 - Confirm operation of unit tests is correct.

T-o-b-i-a-s:
Hello @jamorham , thank you for pointing me to the failing tests. When I ran the original implementation without [Commit "Round conversion to mg/dL"](https://github.com/T-o-b-i-a-s/xDrip/commit/e4eac2799c70f15eb53c2092bbc4f99c1942a0d4), I had a problem with mg/dL float numbers being displayed in Nightscout. In my setup, xDrip broadcasts the blood glucose values to AAPS, which then uploads blood glucose + Insulin + meal data to Nightscout. After adding the commit above, this problem was fixed. However, I did not check the tests again - shame on me. Do you want me to revert the commit above and add a Math.round() to java/com/eveningoutpost/dexdrip/utilitymodels/BroadcastGlucose.java when mg/dL data is broadcasted?

As a fix, I just increased the tolerance in the test to accept rounded values.

Navid200:
I don't think it is a good idea to relax the test accuracy requirements.

Is there no way to receive data in mg/dL?

T-o-b-i-a-s:
The json data I can download from the service only contain the values as mmol/l, if the patient sets the unit to mg/dL, the conversion is done in the web app. It's weird, but I could not find a way to get the raw mg/dL values. The service does not really provide an API, the current implementation is a "hack" as it simulates the Easy View web application to get the glucose values that the web app displays.

Navid200:
I am sorry I overlooked that limitation.  We have a similar issue with the companion app feature.  It can only work when the user sets the units in xDrip to match the units the companion app shows.  So, in this case, it will work fine if the user sets the xDrip units to mmol/L, right?

I don't know how to deal with this or the companion app issue.  
You can ignore my comment and wait for jamorham to review again.

T-o-b-i-a-s:
Hello @Navid200 , I reverted the addition of more strings and just left the strings that are used in Android view elements (e.g. app/src/main/res/xml/pref_data_source.xml).
Unfortunately, I have not tested the mmol/l display in xDrip and now my test sensor is no longer active. Maybe someone else with medtrum sensor can check or we need to keep the PR paused until we have another test sensor available.

PR #4675, opened 2026-08-05
okhttp4 Phase C: swap deprecated RequestBody/ResponseBody.create() argument order

ahaarrestad:
okhttp 4 deprecated `RequestBody.create(MediaType, X)` and `ResponseBody.create(MediaType, X)` in favour of `create(X, MediaType)`. The old order is gone in okhttp 5, so this clears the way for that upgrade. Argument order only, no behaviour change.

With #4593 merged, this finishes the job: **no use of the deprecated overloads remains in the repo.**

### What changed

44 call sites across 18 files, plus one new test.

- **Production — 17 sites, 10 files:** `NightscoutUploader` (8), `TidepoolUploader`, `tidepool/BaseMessage`, `cgm/nsfollow/messages/BaseMessage`, `cgm/webfollow/Cmd`, `carelinkfollow/client/CareLinkClient`, `carelinkfollow/auth/CareLinkAuthenticator`, `deposit/WebDeposit`, `cloud/nightlite/NightLiteClient`, `sharemodels/ShareRest`
- **Tests — 27 sites, 8 classes:** `NightscoutServiceIntegrationTest` (8), `AuthenticatingCallbackTest` (6), `GzipRequestInterceptorTest` (4), `TidepoolIntegrationTest` (3), `NightscoutUploaderGzipInterceptorTest` (3), `InfoInterceptorTest`, `ShareRestTest`, `NightscoutCallbackTest`

Three commits: the new test first, then production, then tests.

### Why it is behaviour-preserving

The deprecated overloads are one-line delegations to the new ones. From `javap -c` on okhttp 4.12.0, `ResponseBody$Companion.create(MediaType, String)` is `aload_2; aload_1; invokevirtual create(String, MediaType)` — it swaps the arguments and calls the same method. `RequestBody$Companion.create(MediaType, byte[])` reaches `create$default` with the same target as the new-order overload. Same bytes on the wire either way, including when the media type is null.

Three sites take a null media type (`CareLinkAuthenticator`, `ShareRest`, and `AuthenticatingCallbackTest`). They carry an explicit `(MediaType)` cast so the reader can see which overload is selected; the compiler does not require it, and there is no null-specific path on either side.

### Not in this PR

`MediaType.parse` and `HttpUrl.parse` are **not** deprecated in 4.12.0 (checked with `javap`) and are left alone. `get` throws on invalid input where `parse` returns null, so converting them would be a behaviour change rather than a cleanup.

### Verification

`./gradlew :app:testFastDebugUnitTest :wear:testFastDebugUnitTest` — all green.

A javac deprecation lint over `:app` production and unit-test sources reports **zero** remaining uses of either method, down from 44.

`ShareRestNullContentTypeTest` is new. It drives the real `ShareRest` interceptor through a `MockWebServer` and asserts the rebuilt response body: null content type and payload preserved when the server sends no `Content-Type`, type and subtype preserved when it sends `application/json`. The interceptor consumes the original body in order to log it, so reading the returned body at all proves the rebuild ran. It is committed before the swap and was green against unmodified production code, so it pins existing behaviour rather than the change.

The gzip interceptor tests assert on the encoded request body, so those 7 sites are covered behaviourally too. Elsewhere the compiler is the real check — a wrong argument order does not type-check — and the integration tests that were touched build their own request bodies rather than asserting on these call sites. `CareLinkAuthenticator#refreshBrowserToken()` builds its client inline and has no unit-test seam.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/carelinkfollow/auth/CareLinkAuthenticator.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/carelinkfollow/auth/CareLinkAuthenticator.java
@@ -65,6 +65,7 @@ import okhttp3.ConnectionPool;
 import okhttp3.Cookie;
 import okhttp3.FormBody;
 import okhttp3.HttpUrl;
+import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.RequestBody;
@@ -440,7 +441,7 @@ public class CareLinkAuthenticator {
                 .build();
         requestBuilder = new Request.Builder()
                 .url(url)
-                .post(RequestBody.create(null, new byte[0]))
+                .post(RequestBody.create(new byte[0], (MediaType) null))
                 .addHeader("Accept", "application/json, text/plain, */*")
                 .addHeader("Accept-Language", "en;q=0.9, *;q=0.8")
                 .addHeader("Connection", "keep-alive")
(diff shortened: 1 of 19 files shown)


jamorham (approved):
thanks

Issue #4652, opened 2026-07-24
#4650 manifest receivers double-deliver OOP2 result broadcasts on the hot path


ahaarrestad:
Discussion thread: https://github.com/NightscoutFoundation/xDrip/discussions/4651 (see also #4650).

Opening this at @Navid200's request to track it as an issue.

### Problem
#4650 re-adds manifest `<receiver>` entries for the OOP2 result actions that `xdrip.onCreate` also registers at runtime: `OOP2_DECODE_FARM_RESULT`, `OOP2_DECODE_BLE_RESULT`, `OOP2_BLUETOOTH_ENABLE_RESULT`.

While the app is running, a `setPackage`-scoped broadcast reaches **both** the manifest and the runtime receiver → 2 deliveries. These results fire on every Libre reading, and each delivery spawns a thread + a 60s wakelock + a full OOP decode. If the farm sends them with `setPackage` rather than an explicit component, re-adding the manifest receiver doubles that work on the hot path for every OOP2 user. The `Process.myPid()` filter in the receiver looks like the farm broadcasts wide, which would point to `setPackage`.

(The reading/follower actions such as `NS_EMULATOR` are not the concern — their duplicate delivery is collapsed downstream by `readingNearTimeStamp()`, so only one row is stored.)

### Open question
How does the OOP2 farm target these results — explicit component or `setPackage`? With an explicit component there is no second delivery and nothing to do here.

### If it's `setPackage`
Two candidate fixes:

1. **Scope the manifest `<receiver>` to only the cold-start action(s) Esel needs**, leaving the OOP2 results runtime-only — keeps #4650's fix without doubling the decode.
2. **A central dedup manager** (@Navid200's proposal): record each incoming broadcast by path (manifest vs runtime) and drop the cross-path duplicate. Two constraints for it to help — it must drop the duplicate at the **front** of the receiver, before the thread/wakelock/decode; and it should key on the **broadcast's identity**, not a wall-clock timestamp (two real readings can share a second → risk of dropping a legitimate one).

### Goal
Keep #4650's Eversense/Esel fix without doubling the OOP2 decode on the hot path.

Navid200:
Is this resolved now?

ahaarrestad:
I don't know yet, so I measured what I could. Android 14 emulator, current master, a log line in the receiver printing which copy arrived.

[...]

PR #4676, opened 2026-08-07
Keep receivers in the manifest (second attempt)

Navid200:
Edit: What I meant was that BYODA does not work with our current Nightly.  This fixes it.  
<br/>  

I put the receivers back in the manifest with no other change.  BYODA does not work.

This one also has receivers in the code path again.  But, they are blocked if they are not needed.

@ahaarrestad Please have a look at this when you can.
Thanks


jamorham:
when you say `BYODA does not work` do you mean it should work here but still doesn't ?

Navid200:
After we bumped the SDK level, BYODA does not work based on my current tests.   I don't understand why no one has complained.
It works after this PR.

However, ahaarrestad is still testing this now.  I would like to wait for his review.  I cannot find him under the the list of developers I can ask for a review in the right pane.  Would you please add him?

ahaarrestad:
I tested this on an emulator, API 34, your head `a7da7975c` with master as the control. Both target
SDK 26.

**The direction is right.** On master I killed xDrip, checked it was really gone, and sent the
reading three ways: aimed straight at the receiver, plain, and with `setPackage`. Nothing arrived,
not one of the three, and the app stayed dead. Putting the receivers back in the manifest is the only
thing that can wake it, so that part has to happen.

**But the PR doesn't get there yet.** Same test on your branch. Android starts the process for the
broadcast, and it dies before the receiver runs:

```
Start proc 5338:com.eveningoutpost.dexdrip for broadcast {.../.NSEmulatorReceiver}
FATAL EXCEPTION: main
  BackgroundServiceStartNotAllowedException: Not allowed to start service
  Intent { cmp=.../.services.PlusSyncService }: app is in background
    at PlusSyncService.startSyncService(PlusSyncService.java:57)
    at xdrip.onCreate(xdrip.java:142)
Process com.eveningoutpost.dexdrip (pid 5338) has died
```

`xdrip.onCreate` calls `PlusSyncService.startSyncService`, which ends in a plain `startService`. A
process started for a broadcast counts as being in the background at that moment, so the call throws
and takes the process with it. The job scheduler brings it back and it crashes again.

This is the other half of the same break. #4601 (`10f57b82b`) raised `targetSdkVersion` from 24 to
26, and 26 is where Android stops a background app from calling `startService`. I built the commit
before it and ran the same test — same emulator, same settings, app killed and confirmed gone:

```
Start proc 7254:com.eveningoutpost.dexdrip for broadcast {.../.NSEmulatorReceiver}
D jamorham nsemulator: NSEmulator onReceiver: com.eveningoutpost.dexdrip.NS_EMULATOR
```

It starts, the receiver runs, nothing crashes. That build has the same two receivers in the manifest
and the same `startSyncService` call in `onCreate`, so the SDK level is the only difference. July 8th
cost us two things, and this PR fixes one of them.

The answer to the other one is already in the tree. `CollectionServiceStarter.startPlainServiceCompat`,
from #4644, is `startService` in a try/catch that logs. `PlusSyncService` never calls
`startForeground`, so that is its shape. I built it that way and measured it, then found it was not
enough on its own and added the same guard to both `JoH.startService` overloads (`JoH.java:1363` and
`:1381`). Cold NSEmulator broadcast, process confirmed dead first, each time:

| build | receiver ran | crashes | process after | reading stored |
|---|---|---|---|---|
| your branch as it stands | no | 2 | dead | no |
| guard in `PlusSyncService` only | yes | 1 | dead | no |
| guard in `PlusSyncService` + `JoH.startService` | yes | 0 | alive | `151.0 NSEmulator Follow` |

The middle row is the one worth knowing about: the receiver gets to run, then the process dies a
second later on `MissedReadingService.delayedLaunch()`, which is the first line of `onCreate` and has
no setting in front of it. Guarding `PlusSyncService` alone moves the crash rather than removing it.

The cost of the guard is that sync and missed-reading checks don't start on a process that was woken
by a broadcast. That seems right — it came up to take one reading, and both start as usual next time
the app or a collector runs.

**Who this actually reaches.** If a collector is running, xDrip holds a foreground service and the
process is never dead — I could not kill it, and the receiver registered in code already serves those
users. The ones that can be dead are the ones whose data source is NSEmulator, which starts no
foreground service at all. That is rudysroost's case, and the only case the manifest entry changes.

**On the status line receiver:** it behaves the same way — dead without the guards, delivered with
them. But AAPS sends that broadcast plain (`XdripPlugin.kt`, `sendStatusLine()`, no `setPackage`, no
component), and at SDK 26 plain only reaches the receiver registered in code, so the manifest entry
adds no delivery for the one sender we have. I'd leave it out and keep this PR to the NSEmulator one.

To your earlier question about NSEmulator and OOP2: they share the receiver class, not a data source.
The three `OOP2_DECODE_*` actions only work if `ROW_ID` matches our own pid, which
`LibreOOPAlgorithm.sendIntent` stamps when the request goes out. A process just started for a
broadcast has a different pid, so it can never finish an OOP2 result — the manifest entry gives those
three exposure but no capability. `NS_EMULATOR` on its own is enough there, and that also settles
#4652.

So: the manifest entry is right and should go in, but it has to land together with the two guards.
On its own it turns a silent drop into a crash loop for exactly the users #4647 is about.

Navid200:
@ahaarrestad 

Thanks a lot for the detailed review.

1- I believe OOP2 is a broadcaster we can change if needed.  I have been told there is a checkbox in OOP2 that enables intents.  it is off by default.

2- I can make changes to fix this if you tell me what to change.  

Or if you prefer, would you like to fix it?

Navid200:
The third row in the table represents the solution, right?
Please open a PR to change my fork to fix it or open a PR altogether.  Whichever is easier for you.

Navid200:
@ahaarrestad You have already spent a lot of time reviewing this.
I am testing it right now so that I can see the crash.  And I will fix it.
Thanks so much.

ahaarrestad:
Working on a fix. Just need to check it. Should be ready later today

Navid200:
OK, I will wait.  Thanks so much.

ahaarrestad:
Yes, the third row. It's up as Navid200/xDrip#210, against `Navid_2026_08_01`.

Both `JoH.startService` overloads and `PlusSyncService.startSyncService` get the try/catch, same
shape as `startPlainServiceCompat`. I added two more while I was in there:
`ActivityRecognizedService.startActivityRecogniser` and `BluetoothGlucoseMeter.start_service`. Both
are called from `xdrip.onCreate` behind a setting, so with motion tracking or a Bluetooth meter
enabled they break the same way. I didn't want the fix to depend on which settings the user has.

Refusals are logged once an hour, not once per call.

The guard has to sit in `JoH.startService` and not only in `onCreate`: `Inevitable.poll()` runs the
deferred work on a plain thread with no catch, so a start refused there kills the process off the
`onCreate` stack. That is the crash still left in row 2.

Checked on the emulator again, API 34, process confirmed dead first, same broadcast both ways:

- your head as it stands: `FATAL EXCEPTION ... BackgroundServiceStartNotAllowedException ... at
  xdrip.onCreate(xdrip.java:142)`, process dies, comes back for the next job and dies again
- with the PR: `E jamorham plussync: Could not start sync service: ...`, no crash, process alive

Unit tests all green, `:app:testFastDebugUnitTest` and `:wear:testFastDebugUnitTest`.

Navid200:
@ahaarrestad Thanks for the fix.

I compiled it after your fix.
I have BYODA getting readings and broadcasting to xDrip and xDrip receives them.
I then killed xDrip (force stop).   It still remained dead and was not woken up by the next broadcast.

Am I doing something wrong in my test?

jamorham:
Force stopped app can't be woken up by a broadcast afaik.

ahaarrestad:
Your test is fine, it just doesn't reach the code this fix changes.

Force-stop does two things at once: it kills the process, and it puts the package in the stopped
state. A broadcast from another app is not delivered to a stopped package unless the sender sets
`FLAG_INCLUDE_STOPPED_PACKAGES`, so BYODA's broadcast never reaches the receiver at all. Nothing is
started, so the guards never run. That is what jamorham is describing, and this fix doesn't change it
either way.

The crash is on the other path: process not running, but the package not stopped — after a reboot, a
system kill, or the previous crash in the loop. There the broadcast does start the process,
`onCreate` runs, and the service start is refused.

adb gets past the stopped state, so you can drive that path on the phone:

    adb shell am force-stop com.eveningoutpost.dexdrip
    adb shell am broadcast -a info.nightscout.client.NS_BRIDGE -n com.eveningoutpost.dexdrip/.NSClientReceiver

Measured just now on API 34 with your fix compiled in: `Start proc 26681:com.eveningoutpost.dexdrip
for broadcast {com.eveningoutpost.dexdrip/.NSClientReceiver}`, then `E jamorham plussync: Could not
start sync service: BackgroundServiceStartNotAllowedException`, process alive, 0 FATAL. Unguarded,
from the A/B I posted earlier: `FATAL EXCEPTION: main ... at xdrip.onCreate(xdrip.java:142)`.

That was a plain `adb shell` on a `user` build, uid 2000, `ro.debuggable=0`, no root — the same thing
you get on a phone with USB debugging on, so it should behave the same on your device.

Two things worth knowing if you retest. The shell's intent carries flag `0x00400000` on its own, so
adb reaches a force-stopped app with or without `-f 0x00000020` — use adb to test the fix, not to
check the force-stop behaviour you saw. And by hand on the phone there is no good way to reach
"dead but not stopped": force-stop from Settings always sets the stopped flag, and opening xDrip to
clear it starts the process. adb or a reboot are the practical routes.

Navid200:
Thanks

I am OK with this if the two of you are OK with it.

I cannot thank @ahaarrestad enough not just for this but for all his contributions in this repository.

jamorham (approved):

PR #4679, opened 2026-08-08
Delete treatments which have been deleted in AndroidAPS

simelis:
Deleting a treatment in AndroidAPS leaves it on the xDrip graph.

AndroidAPS does not send a removal message: it resends the same treatment with `isValid`
false. That flag was never looked at in xDrip, so the resent copy was handled as an addition and
then discarded as a duplicate of the original. Nightscout keeps deleted treatments in the
same way, marked invalid, so upon deleting a treatment via AAPSClient, it did not get deleted in xDrip, it actually even recreated previously deleted treatments.

From a phone log, deleting a 1 g carb entry:

```
Processing treatment from NS: {"eventType":"Carb Correction","carbs":1,
  "isValid":false,"date":1786220236000,"_id":"eff7a672-a9d5-..."}
converting treatment from json: {"timestamp":1786220236000,
  "eventType":"Carb Correction","carbs":1,"notes":"","created_at":...}
Duplicate treatment for: 1786220236000
```

This change:

- deletes a treatment marked invalid rather than adding it, on both paths
- carries the sender's id through so the treatment can be identified — `pushTreatmentFromJson`
  already adopts it as the uuid, only the AndroidAPS path was dropping it
- does not upload these deletions back, since the invalidated record is what tells other
  apps the treatment was deleted

Note: Treatments received before this change hold a locally generated uuid, so deleting those
elsewhere will still leave them on xDrip graph - the change will only work for the treatments created using the version of xDrip with this change.

## How to see it

1. AndroidAPS and xDrip on one phone, with broadcasts to other apps enabled: add a small carb
treatment, wait for it to appear in xDrip, then delete it in AAPS. It now disappears from xDrip; without this change it would stay indefinitely.
2. Using follower phone and AAPSClient - add 1 carb treatment, wait for it to appear in xDrip, then delete it in AAPSClient. It now gets deleted in xDrip, without this change, it doesn't get deleted and even restores previously deleted entries.

## Testing

Used on AndroidAPS connected to xDrip over local broadcast, with two follower phones in xDrip follower sync mode and AAPSClient apps, sharing a Nightscout site.
Treatments deleted in AndroidAPS, or from another device via AAPSClient, are now removed in
xDrip master and all followers.


jamorham:
This makes sense, I haven't reviewed the operation but can you describe whether this has been properly tested?

simelis:
I have tested it on my setup, works pretty well for carbs, insulin and other treatments. Looks like it doesn't work well together with #4646  - I can add a BG Check, notes or announcements in AAPS, and it goes to xDrip, but if I then remove it in AAPS - it doesn't get deleted in xDrip. As #4646 is not yet merged into master, I'll update that PR if I manage to find a fix for that - here it doesn't have what to work with.

What exactly is proper testing? Is this enough for such PR, or should I follow some document, or it should be tested by a team?

simelis:
I committed the BG related fixes here after all, as it's all related to deletion. No more issues found since using this version.

jamorham:
Unit tests are failing due to the changes. Can you either wrap the original behavior so the existing tests pass or justify the change and update the unit tests.

Issue #4642, opened 2026-07-20
Audio focus manager


Navid200:
xDrip offers audio focus for its alerts.
However, it does not for the speak reading feature.

We need an audio focus manager to handle requests from different methods and to negotiate with the OS so that we can offer audio focus for speak readings as well.

Issue #215, opened 2017-11-14
Guardian Connect


Schoggiweggli:
Since Medtronic is reaaaaally slow at releasing an Android App or even update the iOS one to be considered a competition for Dexcom, is it possible to get it to work with xDrip?

I i am stuck with an iPhone because of it and i would love to change the device.

hartlijn:
For now only in combination with android uploader phone with the BG Meyer attached via OTG cable.

Schoggiweggli:
@hartlijn that's for the Pump combination. Guardian Connect is a standalone bluetooth transmitter.

shtefanilie:
@Schoggiweggli any news? Did you try with xdip? Does it work?

Schoggiweggli:
> @Schoggiweggli any news? Did you try with xdip? Does it work?

I don't know, sorry. Changed to Dexcom G6 a year ago and gave up on Medtronic entirely.

tolot27:
Duplicate of #131

PR #4680, opened 2026-08-09
New mechanism for the Other alerts

### LINKED ISSUE #4678: Negative consequences of having bumped the SDK level
1- Wear OS crash
  
2- Some apps that broadcast to xDrip fail now.   This is one example: https://github.com/NightscoutFoundation/xDrip/discussions/4647  
  
3-  Suunto watch failure: https://github.com/NightscoutFoundation/xDrip/discussions/4625  
  
4- Override silent mode setting in Other alerts does not work.  
  
---  
  
1- This PR should have fixed it: https://github.com/NightscoutFoundation/xDrip/pull/4644  
  
2- This PR hopefully will fix it: https://github.com/NightscoutFoundation/xDrip/pull/4676  
  
I will have a PR in the next few days that is meant to fix 3 and 4.

Navid200:
I have opened this so that you can see what it looks like.
If you see something that is completely out of the question, please let me know.
But, please don't do a detail review yet.
I am still testing and some more tweaks may still be needed.  
  
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/4678


Navid200:
The latest Nightly has the following related issues:
A- The existing "Override silent mode" in any of the "Other alerts" has no effect.
B- The existing sound file for any of the "Other alerts" will result in a new notification channel being created leaving the previous one as a zombie.  

I wanted to address these by guiding the users to learn to rely on notification channels and tolerate the zombie notification channels.  
It turned out it was not that simple and more and more issues resulted as I made progress.  It got to a point that it made no sense to invest so much effort for a band aid solution that I decided to address them differently.
We already have glucose level alerts that have none of these problems and they have been tested for years.

This PR makes Other alerts use media player the same as glucose level alerts make sound instead of depending on notification channels.  This PR addresses both existing issues.

There are currently two remaining issues with this PR I still need to resolve:
1- After an update to this PR, the user will need to restart the phone.  Otherwise, Other alerts will make sound but will not create as notification that can be dismissed!  This makes no sense and I should be able to close this today or tomorrow.
2- This PR still relies on notification channels for vibration.  This is not good.  We have had a request for years to offer custom vibration patterns.  We need to remove the dependence on notification channels for vibrations as well.  

The second item is not easy to address.  It will make this PR larger.  I have a proposal.  But, to explain it, I first need to talk about zombie channels again.
It is very easy to address zombie channels.  I will have a new page in the guide focused on addressing zombie notification channels.  It will start by recommending to the user to have a Nightscout site, or at least use Tidepool.  It then explains how the user can backup their settings as well as their data base.  Then, after highlighting that backup is a requirement before an uninstall, it explains that the user needs to uninstall xDrip and install a recent version of xDrip.  
This will eliminate all zombie notification channels.

The idea of uninstalling and reinstalling xDrip is not very good.  But, thanks to Google we have lots of zombie channels and there is no other way for cleaning them out.

Assuming the idea of cleaning up the zombie notification channels is acceptable, we can go ahead with this PR as is with dependence on notification channels for vibrations.
When the day comes that we are ready for merging a PR that offers custom vibration patterns, the user will have to follow the page in the guide for addressing zombie notification channels.  However, they will not be doing it to remove zombie channels.  They will be doing it to replace their existing notification channels that control vibration to notification channels that have vibration disabled.  Only then, they can benefit from the new custom vibration patterns.

To address item 2, we can either wait for this PR to be changed enough so that it can also offer custom vibration patterns, or we go ahead and some time in the future, we will have to ask users to reinstall again.

Navid200:
I will continue working on this to improve it so that the Other alerts directly control vibration as well.  This way, if we decide to offer custom vibration patterns in the future, it will not require asking users to re-install.  

I would also like to add here that the sensor expiry notification still relies on notification channels even after this PR.

Navid200:
This bypasses notification channels for sound and vibration for everything.
All alerts, reminders, sensor expiry, number wall icon have been tested.

The calibration alerts or parakeet alerts will not work.  I can fix those if we need them.
There may be some notifications that are now silent.  If that is the case, we can fix it easily later.

Vibration cannot be disabled using notification channels anymore.  That's why I added switches to the alerts that trigger often, Other alerts, persistent high and forecasted low.  
We can add more of these switches if we want easily later.  For example, we can add it to sensor expiry.

What this PR addresses is that after a fresh install, the override silent mode of any alert works even for the very first alert.
The sound file can be changed resulting in no zombie notification channels.
It opens the door for so many improvements we can bring like custom vibration pattern or volume profile for missed reading alert.  But, all of these are left for future to minimize the size of this PR.  

It should fix the watch that was broken.  
There should be no missing notification channel for a catch.  

This is not meant to make everything perfect.  It is meant to address some serious issues so that we can stabilize the Nightlies and be able to soon introduce a new beta.  

There will be a page in the guide walking the user through a reinstall process to address zombie notifications.  
Alternatively, the user can change the existing notification channels and disable vibration and sound in them.  

I have been using this for a while as my active xDrip as it has evolved.  

@ahaarrestad Please let me know what you think of this.   Please let me know if you see anything broken?

Navid200:
Screenshot_20260813-121208  Screenshot_20260813-121219  Screenshot_20260813-121245

Navid200:
I'm sorry I find that an update from our current alpha (SDK 24) results in silent alerts.  This may have to do with file access permissions.  
I have to find out what causes this and fix it before we can consider this.

ahaarrestad:
Direction looks right — `getChan()`'s hash-suffixed channels are the zombie generator, so removing
that is the real fix. Four things from reading the diff against master:

**1. No vibration in vibrate mode.** `VibrateNotifyMakeNoise` gates vibration on
`alert.override_silent_mode || isLoudPhone(context)`, and `isLoudPhone()` is false for
`RINGER_MODE_VIBRATE`. Correct for sound, inverted for vibration. With `BG_ALERT_CHANNEL` now created
with `enableVibration(false)` there is no fallback: phone in vibrate mode, `override_silent_mode` off,
and a low alert is completely silent. Master vibrates.

**2. Other alerts and BG alerts now share one MediaPlayer.** `startGenericAlert` → `playFile` →
`stopAndReleasePlayer(mediaPlayer)`. A missed-reading alert cuts off a playing low alert. Same in
`SnoozeOnNotificationDismissService`, which now calls `stopAlert` for every type except `bg_alerts` —
dismissing a missed-reading notification silences an active low alert, and with `ClearData=false` it
stays active but mute.

**3. `JoH.showNotification` is now `if (sound || vibrate)` and plays a file.** Vibrate-only callers
like `CheckBridgeBattery` start making noise. And `MIN_VOLUME = 0.5f` is a floor, so media volume at
10% plays at 50%.

**4. Zombie channels can be deleted in code.** Enumerate `getNotifManager().getNotificationChannels()`
and delete everything that isn't one of the four new IDs — the hash suffixes rule out deleting by
constant, but exclusion covers all of them. `deleteAllChannels()` already does this for groups. That
removes the reinstall requirement.

Minor: the four `TODO Navid` channel names are hardcoded English where they used to come from
`getString(...)`.

Navid200:
@ahaarrestad Thanks for the detail review.

I believe I have addressed 1 and 3 already.  

From the 4 notification channels we will have after this PR, 3 of them are brand new since the SDK bump.  I would like them to have titles that can be translated.   
The only notification channel that existed before is the one for the glucose level alerts.  It looked like this:
Screenshot_20260817-095718
  
I am creating 4 new strings.  This PR leaves them in English so that we can focus on everything else and leave the grammar and style review for a follow-up PR.  

I had never thought about alert collision.  Thanks for bringing it up.
I believe a missed reading alert should override any other alert.  I have opened an issue so that I remember to fix this: https://github.com/NightscoutFoundation/xDrip/issues/4688  

But, it is possible for a high alert and a persistent high alert to coexist.   It is also possible for a low alert and a forecasted low alert to coexist.
You have a point that this PR changes the behavior in such a way that snoozing the persistent high alert will also silence a coexisting high alert.  But, it will not deactivate it.  

Can we take a step back and think about the concept.  A CGM alert is meant to get your attention.  If you have picked up the phone and are snoozing them, it has done its job.  Can you tell me under what condition you would snooze one and leave the other one making sound?  I myself cannot think of one.  if I am up and am snoozing, I know what I need to do.  I don't need either of the alerts to continue to annoy me or anyone else around me.

We can spend a lot of time trying to come up with a scheme to snooze silence alerts individually instead of silencing them all.  But, is there really a reason to do that/

ahaarrestad:
> Can you tell me under what condition you would snooze one and leave the other one making sound?

No, I can't construct one either. High + persistent high is one event, low + forecasted low
likewise. Agreed — silencing them together is right, and an individual-snooze scheme isn't worth
building. That half of #2 is withdrawn.

Strings in English now, grammar in a follow-up: fine.

The rest of #2 isn't about snoozing. It fires with no user interaction.

The glucose path has only ever had one producer, which is why one `mediaPlayer`, one `streamType`,
one `volumeBeforeAlert` are safe. This PR makes it multi-producer. Master had one `playFile`
caller, head has two.

**1. A notification cuts off a sounding BG alert.** `playFile` kills whatever is playing
(`AlertPlayer.java:338`). `default_alert.mp3` is 32 s at 128 kbps, no looping, so that's the
window. The follower chime (`GcmListenerSvc.java:469`) fires on a new reading — the same event
that raises the alert. Also `Reminders`, `SENSOR FAILED`, `Battery Low`, `SensorExpiry`. Next
sound is `ClockTick` at the following `minutes_between`.

**2. Alarm volume is left raised.** The kill at `:338` doesn't `revertCurrentVolume`. The
notification sets `streamType = STREAM_MUSIC` (`:384`), overwrites `volumeBeforeAlert` (`:409`)
and reverts music on completion. `STREAM_ALARM` stays where the alert put it.
`onCompletionListener` reads `streamType` as a field at completion time (`:357`), not the value it
started with.

**3. `SnoozeOnNotificationDismissService` silences without snoozing.** `bg_alerts` with
`time_showing <= MINIMUM_CANCEL_DELAY` now falls through to `stopAlert(..., false, false, false)` —
the accidental swipe the 2 s guard is for. On master that branch only logged. The alert goes
quiet, stays active, unsnoozed.

Point 4 hasn't landed: `cleanAllNotificationChannels()` has no callers on either side and still
says `// TODO this isn't right yet`; `setupAllChannels()` (`xdrip.java:147`) only creates. Zombies
survive and the reinstall page is still needed. Wiring that loop up is worth more to users than
anything above.

None of this blocks the PR for me — it fixes two things broken for everyone all the time. #4688
covers 1 and 2 if you build precedence there; today it's "last caller wins".

Minor: `startGenericAlert` never nulls `mediaPlayer`, so `"ERROR, playFile sound already playing"`
logs on every later play whether or not anything was playing.

Navid200:
These are all excellent points.

We can take advantage of deleting zombie notification channels in code in a future PR.  Thanks.

I will have to address volume (and may be even audio focus) when an alert arrives while another is in progress.  

I will have to figure out how to handle a ping coming from anything while an alert is in progress.  Excellent point!  Thanks.  

With respect to the snooze confirmation, again excellent point.  I would like to get rid of it!
We have another PR that lets a user double press a volume button to snooze: https://github.com/NightscoutFoundation/xDrip/pull/4623 
I wonder if it could interfere with that too.  
If someone snoozes an alert and realizes they weren't sure what it was, they can always look at the logs to get more detail about it.  I understand the logs are not translated.  So, I need to come up with a scheme to address that.  But, is there really a point in keeping the sound going and asking for confirmation?  I prefer to snooze as soon as the user asks us to.  If they made a mistake, we have to provide a way for them to see what was going on (logs).

ahaarrestad:
I tested the confirmation screen on Android 16. It never appears.

Android does not let an app open a screen while it is in the background. xDrip is in the background when you swipe the alert away, so the screen is refused:

```
Background activity launch blocked!
START ... com.eveningoutpost.dexdrip/.Home (BAL_BLOCK)
```

To make it easy to hit I changed the 2 second limit to 10 minutes in a local build, so every swipe took that path. Nothing else was changed. I only tested Android 16. This restriction came in with Android 10, so older phones may behave differently.

So the 2 second rule does one thing today: the sound keeps playing. That is what tells the user the swipe was an accident. The confirmation screen behind it is already dead.

That helps your case and hurts it at once. The screen you want to remove is gone anyway. But the sound is the whole safeguard now, and this PR stops it, so a fast accidental swipe goes silent with the alert still active and nothing to notice it by.

On #4623: yes, it changes this. That PR asks for the Android permission "Display over other apps", which it needs for its own snooze screen. That permission also lifts the block above, for everything xDrip does. I granted it and ran the same test again, and the confirmation screen came back:

```
START ... com.eveningoutpost.dexdrip/.Home (BAL_ALLOW_SAW_PERMISSION)
```

and "Cancel Alarm? Please confirm to cancel the alert?" appeared on screen.

So the two PRs pull in opposite directions on the same screen: #4623 brings it back to life for users who grant that permission, this one removes it. Worth deciding once rather than twice.

ahaarrestad:
Correction to my last review: I said #4688 covers points 1 and 2. It doesn't — #4688 is about not
raising an alert when there are no readings, which is a different problem.

Your own #4642 is the closer fit. "Requests from different methods" is this case: two parts of xDrip
want to make a sound at the same time and nothing decides which one wins. That is the volume and
audio focus work you said you would have to do.

Navid200:
@ahaarrestad The last commit is supposed to stop a low-priority notification from interfering with a high priority alert.
When you get a chance, please have a look.

I will now work on fixing the volume so that we return the volume to exactly what it was at before the first alert triggered.

Navid200:
The volume should be managed correctly now even if there is a collision.  

Is there anything else that is broken due to all alerts using the media player?

ahaarrestad:
Most of it is fine. I went through all six places that build a notification this way, and four of
them are unaffected — they either play their own sound through the alert player, or never had a
sound to begin with.

Two are not.

**Calibration and Parakeet status notifications.** Those two are the only ones that hand the sound to
Android as part of the notification itself instead of playing it through the alert player. Android
only honours that if the channel it lands on has a sound of its own, and the channels in this PR are
all created without one. So on a phone that installs this build fresh, those two make no sound and
no vibration, and the calibration sound setting stops having any effect.

**The upgrade case goes the other way.** Android only applies a channel's sound the first time the
channel is created — after that the setting is ignored. Every phone already running xDrip already
has the general channel, and today it is created without an explicit sound, which means Android gave
it the default notification sound and vibration. That does not change when this PR lands. So on an
upgrade the general channel keeps its sound, and everything moved onto it inherits that sound —
including notifications that ask for no sound, like the Nightscout "upload failed" one.

That half is inherited rather than introduced here, but it does mean a fresh install and an upgrade
will not behave the same way.

I checked this on an emulator against your latest commit, reading back the channel settings Android
had actually stored, once from a fresh install and once after upgrading from master.

Routing those two notifications through the alert player like the rest would settle both halves.

Navid200:
Thanks for the review.

Parakeet is gone.  We need to remove everything related to Parakeet from xDrip.  There will be no Parakeet alert.

Calibration alerts have no reason to exist.  We needed them when we used G5.  G5 needed to be calibrated twice a day regardless of accuracy.
G5 and its likes do not exist any longer.  Today, the only reason to calibrate is if the user chooses to.  There is no requirement to calibrate twice a day and there is no reason for an alert.  There is no sensor today that xDrip supports internally that will ever nudge the user to calibrate.

Those who are using our latest Nightly and update to this PR only need to edit the General notification channel and set it to silent.  
A user may one day decide to assign a sound to their general notification channel (or any of the other three) as an experiment.  They will notice that when a notification triggers, they may hear an unexpected sound since the alert player and the notification channel both will play a sound.  The solution is the same.  Just edit the notification channel and set it to silent.

Navid200:
I take that back!  Please don't set the General notification channel to silent.
If you do, when you run the number icon test, instead of seeing 123 in the notification bar, you will see a square (grouped notifications).

1- Edit the General notification channel.
2- Disable vibration.
3- Set the sound to none.

ahaarrestad:
This is going in the right direction. Fifteen channels down to four, and the reflection into the
builder's private field gone — that workaround has a "please report" line logged inside it because
nobody trusted it to keep working. Removing things is the harder half, and this project does not do
much of it.

That is also why I would rather see it go further, though not here. If calibration alerts and
Parakeet are dead, the settings should follow them out: `calibration_notification_sound` and the ones
under `calibration_notifications` stay in the UI, editable, doing nothing. That is a cleanup of its
own and no reason to grow this PR. And withdraw my earlier suggestion about routing those two through
the alert player — if they are dead, that is the wrong direction. You know the user base here and I
will take your word on it.

The one I would still flag is the general channel. The Nightscout upload-failure notification asks
for no sound and no vibration, and on an existing install it will now make one, since Android keeps
the sound a channel already has. That one is a live feature, and the people who hear it are the ones
who never read this PR.

On your last comment, I want to make sure I follow: it opens with not setting the general channel to
silent, and then step 3 is to set the sound to none. Which of the two did you mean? Good catch on the
icon test either way — that would have been easy to miss.

Navid200:
There is a setting in the notification channel called silent, which removes the vibration checkbox.  If you enable this setting, you effectively change the importance level of the notification channel.  We don't want the user to do that, which will lead to the grouping of the general channel and the ongoing channel.  

But, we also don't want the notification channel to make a sound or to vibrate.  The checkbox for vibration should be disabled.
Then, the user should tap on sound as if they were going to choose a sound file.  But, they should choose none after they choose the my sounds folder.  If the user chooses none, the notification channel will be set to make a sound.   But, the sound it is set to will be nothing.  This is the best setting.  And that is what we want.
Please remember a failure here will result in an alert making more sound than it should!  So, someone experiencing this will post and ask and I will be guiding them to do exactly what I explained here.

A failure leading to no sound instead of a sound would be unacceptable.  So, I am happy this is the worst failure of this.  I admit that I could have handled all of this better.  The problem is at first, I was going for relying on notification channels for Other alerts.  We have only exposed those who use the Nightly channel to this.  The real release is when the next beta release comes out.
That is why I would like all of these open issues to be closed as soon as possible before then.

Navid200:
I always make a PR as minimal as it can be to help with the review.
This one is not a minimal PR.  It was not possible to make it minimal.

But, on top of that, I have to confess that there are two changes that were not necessary and I can remove them if it would help:

1- If you lower the volume, currently your Other alerts will become extremely quiet.
This PR puts a minimum on that.  0.6 is the minimum imposed on the missed reading alert and the persistent high alert.  Medium is 0.7.  So, this imposed minimum is lower than medium.  
For the remaining Other alerts, the imposed limit is 0.4.  For general notifications, the behavior remains as before meaning if you lower the volume to 0.1, the notification will play at 0.1.  

2- If you set your volume profile to silent, currently you will still hear all your Other alerts.  That doesn't make sense.  This PR silences Other alerts as well in that case.  

Neither of these two changes address something that is broken due to using the media player for Other alerts.  So, I can remove them if it helps with the review.  

Something else I should mention is the table of priority.  This PR introduces the table based on which, in case of a collision, it is decided which sound will be heard.  If you think I have chosen an incorrect priority order, we can always adjust the order in the table even in a follow-up PR if not now.

ahaarrestad:
Keep it as one PR. I don't think it needs trimming.

The two changes you offered to take out aren't really two changes. The volume floor and the silent-profile handling are the same condition:

```java
if (volumeFrac < minVolume && (overrideSilent || volumeFrac > 0)) volumeFrac = minVolume;
...
if (volumeFrac > 0) playFile(context, soundUri, volumeFrac, overrideSilent, overrideSilent);
```

`volumeFrac > 0` is doing both jobs. Removing either one means rewriting that expression rather than deleting a hunk, which is more risk than leaving it alone. And with override-silent on and the volume at zero, it's the floor that makes the alert audible at all — that is the #4632 fix.

Both are the right behaviour anyway. An alert nobody can hear isn't an alert, and Other alerts ignoring a silent profile never made sense. Good that you took both while you were in there.

The priority table is a real improvement. Deciding the order explicitly beats whatever it did before, and missed readings above low glucose is the right call — no data hides everything else. Nothing in the order looks wrong to me. If something shows up in use it's an easy follow-up, like you say.

Navid200:
One shortcoming of this PR is related to the light colors.
This PR lets us control sound and vibration independent of notification channels.
Unfortunately, I cannot find a way to control light colors independent of notification channels.
Therefore, this PR reduces the variety of colors we can have.

jamorham:
How is this looking now - testing?

Navid200:
I cannot test anything else.  Everything I can test with respect to the sensor and phones that I use has been tested.
It needs more testing through the Nightly channel unless you see something that makes no sense or you test yourself and see an unexpected behavior.

app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/AlertPlayer.java
```diff
@@ -666,4 +702,87 @@ private boolean isLoudPhone(Context ctx) {
...
+     */
+    private int getPriority(String tag) {
+        if (tag == null || tag.isEmpty()) return 0;
+        String t = tag.toLowerCase();
+        if (t.contains("bg_missed_alerts")) return 95;
+        if (t.contains("low_glucose_level")) return 90;
+        if (t.contains("persistent_high_alert")) return 87;
+        if (t.contains("high_glucose_level")) return 85;
+        if (t.contains("bg_predict_alert")) return 80;
+        if (t.contains("BlueReader Alarm")) return 75;
```
jamorham: wont this fail because its a lowercase match only?

Navid200: Yes, it will.
Thanks for catching that.

app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/AlertPlayer.java
```diff
@@ -115,6 +116,7 @@ public class AlertPlayer {
     private final static String TAG = AlertPlayer.class.getSimpleName();
     private volatile MediaPlayer mediaPlayer = null;
     private final AudioManager manager = (AudioManager)xdrip.getAppContext().getSystemService(Context.AUDIO_SERVICE);
+    private static String activeTag = ""; // The alert that is currently playing or vibrating
```
jamorham: Thinking this needs volatile at a minimum

Navid200: Thanks

Will fix, test and submit.  It will take a few hours.  I don't believe it will be done (your) today.
Thanks a lot for the review.

app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/AlertPlayer.java
```diff
@@ -352,9 +355,10 @@ protected synchronized void playFile(final Context ctx, final String fileName, f
 
         mediaPlayer.setOnCompletionListener(mp -> {
             Log.i(TAG, "playFile: onCompletion called (finished playing) ");
+            activeTag = "";
```
jamorham, line 369: does this also need to be set to empty in setOnErrorListener() ?

app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/AlertPlayer.java
Navid200: Done

app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/AlertPlayer.java
Navid200: Done

app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/AlertPlayer.java
Navid200: Done

Navid200:
@jamorham Please don't forget this.

Navid200:
Ah great.  Now there is a conflict.

Navid200:
@ahaarrestad Do you want to take over this?

I have spend so much time on this.  And it now has conflicts.  
I now have to make changes to this to address the conflicts caused by your PR.  But, the problem is I have no idea what will happen after I address the conflcts.  For all I know, another PR will be merged and it will have even more conflicts.

I trust you have the knowledge and expertise to address this.
I am getting tired.  Would you please take over this?

ahaarrestad:
I have opened Navid200/xDrip#215 against this branch. It is the master merge, nothing else.

The conflict was one hunk, in CheckBridgeBattery.java. #4718 deleted checkParakeetBattery(), and this branch changes one line inside that method. I kept the deletion. Your same change in checkBridgeBattery() is untouched, and the other 89 files merged without conflict.

There was a second break that git did not report. NotificationChannelsTest comes in with the merge and lists eleven channels; this branch keeps four, so it stopped compiling. That test is mine, from #4718, so that one was on my side. I narrowed it to the four you keep, without changing what it asserts. Tests are green on the result.

I would rather not take the PR itself. It is your design and you know what each change is there for. The merges I can do, and I will keep doing them whenever master moves under you - just say so and I will open another one.

Navid200:
@ahaarrestad This PR was ready to be merged.
There was no communication that it needed more work after I had addressed the last requests.  
But, now it has a conflict.

You are asking me to spend even more time on it.  But, you cannot tell me that it will be merged if I do.  Do you see the predicament?

ahaarrestad:
Yes, I see it. And no, I can't tell you it will be merged - I don't decide that, and I'd rather say so than pretend otherwise.

For what it's worth, I think it is ready. I thought I had said so in my comments; if that didn't come across, I'm saying it plainly here.

On #215: it's a merge button on your branch, nothing more. The conflict is already resolved and the tests are green, so there is no work left in it for you - you don't have to open CheckBridgeBattery.java at all. I opened it so the conflict wouldn't land on you. If you'd rather close it, that's fine.

Nothing I did was meant to get in front of this PR. The order things get merged in isn't mine to set.

Navid200:
I never meant anyone did anything intentionally to do this.
The PR you have opened into my branch changes 90 files!
It is not pressing the merge button that takes time.   It is repeating the tests that had already completed that I now have to run again.  
And for all I know, something else may be merged again after that instead of this.  There has to be a better way for doing all of this.  We need to prioritize PRs and go in the right order.  This PR is critical.  It has been holding up progress with releasing the beta channel.
Although, it is possible that I am wrong about all of that and this PR is not really as high priority as I had thought.

ahaarrestad:
The 90 files are master moving, not something I added. They arrive on your branch either way round - the only thing my PR changes is that you see them before this lands instead of after.

I checked them against your work. You changed 21 files. The merge touches 4 of them, and none of those are the alert ones - AlertPlayer, NotificationChannels, Notifications, ActiveBgAlert and the notification prefs are exactly as you left them. Nothing in the 90 goes near the alert path.

So I don't think the alert testing needs redoing. What you tested is what is still there.

I ran the unit tests and a full prod release build on the merged branch too. Both green.

You are right that this one matters, and you are not wrong about the ordering either - that part just isn't mine to decide. If master moves under you again I will redo the merge. It costs me very little and it should cost you nothing.

Navid200:
For all I know, jamorham is still testing this.  This is a massive change and more testing will only help us.

There was no conflict.  The file shown as having conflict was a file that this PR was not making any change to!
I resolved the conflict.  But, it now fails tests.  This sometimes happens and it requires the tests to be triggered again.  jamorham can do that.

Navid200:
Let me close and reopen.  That may trigger the tests again.

Navid200:
I don't know why this is failing tests.  I can compile it.

ahaarrestad:
Nothing is wrong with your code - the app compiles. What fails is a test file that came in with the master merge.

`NotificationChannelsTest` lists eleven channel names. This branch keeps four, so the nine it no longer defines don't compile:

```
> Task :app:compileProdReleaseUnitTestJavaWithJavac FAILED
NotificationChannelsTest.java:40-49: error: cannot find symbol
```

That test is mine, from #4718, so this one is on me - I listed eleven channels while this PR was open to remove nine of them.

The fix is one file: in that list, the eleven names become the four this branch defines. Navid200/xDrip#215 already has exactly that, and it is still clean against your branch after your merge, so it is one button. Editing the file yourself works the same.

Navid200:
No, we are not going to edit a good file in order to please a test file.
This is an example of when automated tests hurt instead of help.

ahaarrestad:
To be clear: nothing in your PR changes. Your code stays exactly as it is.

The file that needs the edit is the test file, not `NotificationChannels.java`. It holds a list of channel names, and that list has to match whatever channels the branch defines - eleven on master, four here. That is all it is.

Navid200/xDrip#215 already has that edit.

Navid200:
I have to verify that the sensor expiry notification behaves as expected.  It is possible that it plays the default xDrip alert sound, which is not the existing behavior.  I need to test and if this is true, I need to fix it.  I am sorry about this late transition back to draft.

Navid200:
Before using the media player, we asked the OS to play the sound that the user had chosen as the notification sound for the phone and the OS did the work.
We cannot do that any longer now that we want to use the media player directly.

We already have a default alarm sound inside the repository and we can play that with no trouble.  We have a few other sounds in the repository.

The problem is for simple and general notifications, we cannot play the alarm sound.  We need a shorter less disruptive sound.
Why not also add a notification sound to the repository?
If we do that, we can play that sound without having to rely on the phone and the OS doing the right thing.  

So, I am looking at the Android Open Source Project Sounds to avoid any copyright infringement possibility.
I hope to have this ready for a review in a day or two.

Navid200:
I'm sorry about my impatience.  This makes major changes and should be checked thoroughly.  Thanks for all the reviews and feedbacks.

Sensor expiry notification was sounding the default alarm sound.
I have added a new notification sound.  
The sensor expiry notification uses that now.
The reminder uses the reminder sound as it did before bumping the SDK.

I have created the sound file.  I did not download it from anywhere.
I created it for xDrip.  Considering xDrip is free and is for everyone's use who may have diabetes, I have created it for xDrip with no strings attached.

If you don't like the sound as the default notification, please tell me and tell me what you like tweaked and I will edit it.

I continue to use this and if I find anything else broken, I will fix.  If you see anything that needs attention, please let me know.

Thanks

ahaarrestad:
Built your branch and ran it on an emulator, next to a build of 2026.03.01 — the current stable release, so that is what people would notice a change from. Tests and the release build are green:

    ./gradlew :app:testFastDebugUnitTest :wear:testFastDebugUnitTest :app:assembleProdRelease

**Works**

- High and low alerts sound as before, and the volume is raised for the alert and put back afterwards.
- Reminders and Other alerts play their own sound.
- Override silent for Other alerts works both ways: phone silent and the setting on, the missed reading alert sounds; setting off, it stays quiet.
- The new softer sound never replaces the high or low alarm sound.
- Once an alert has made a sound, a reminder a few minutes later still gets through while the alert is still running.

**Does not work**

**1. An alert that has not made a sound yet blocks every quieter alert.** The default profile only vibrates for the first three minutes, and a reminder due in that window is dropped. Same with Vibrate only, or during a call: nothing quieter gets through until the alert is snoozed or BG is back in range. On 2026.03.01 the reminder always sounds. Your Other alerts code already handles this, the high/low path does not.

**2. The BG alert profile now silences reminders and Other alerts too.** Profile set to Silent, phone not silent, no BG alert running: the reminder makes no sound. On 2026.03.01 it sounds. That setting used to affect only high and low alerts.

**3. Upgrading from 2026.03.01 leaves the old notification channels behind.** Android's notification settings for xDrip then list the old "Glucose Alerts Settings" and "Reminders" next to the new "Glucose level alerts", "Other alerts" and "General". The old ones do nothing, and whatever the user had changed on them no longer applies.

Navid200:
@ahaarrestad Thanks for the review.

1- The default profile is High.  And the alert will not only vibrate for the first 3 minutes.  The only time the alert only vibrates for the first 3 minutes is if you choose the ascending volume profile and do not disable the sound delay.  Please let me know if I am misunderstanding.  Because there is a lot more I can say about this.  But, it would be pointless if I am misunderstanding you.  

2- This is a change I have made and it is what makes sense to me.  If a user chooses the silent profile, why should xDrip not be silent?  
If the user only wants the glucose level alerts to be silent but not the rest, the user can choose a silent sound file for the alerts he/she wants to be silent.  But, choosing a silent profile for xDrip to me means the user wants silence.  A user who is deaf would choose that (silent) profile.  Why would we (xDrip) still make noise?  

3- That's the expected behavior.  As soon as this PR is merged, I will open another PR that will use what you suggested to delete all notification channels other than the expected ones.   I am already testing this to save time when that day comes.

As soon as this PR is merged, I will be posting on facebook as well as in the discussions here explaining to users that they should stop editing notification channels because they will have no impact on the sound and vibration that xDrip makes; that they have still as much control and even more on xDrip's behavior using xDrip's settings, which can be easily transferred to a new phone unlike notification channel settings.  This is the expected behavior.  
The same post on facebook and discussions here will explain that the user must undo any edits they may have made to the General notification channel.   If they have enabled vibration in it or if a sound file is assigned to it, they need to undo those.  

Again, it is possible I am misunderstanding you.  Please let me know.

ahaarrestad:
You are right about the default, and I got that wrong. High is the default, and the quiet first
three minutes only happen with the ascending volume profile and the sound delay left on. I had set
ascending myself before measuring, then described it as the default. Sorry for the confusion.

I went back and measured the same cases on 2026.03.01, and one of my three was wrong as well:
during a call, 2026.03.01 keeps the reminder quiet too, so that one is not a change. Two are, both
with a low alert running and a reminder due:

- Vibrate only — no sound on your build, sounds on 2026.03.01
- Ascending inside the sound delay — no sound on your build, sounds on 2026.03.01

From the outside it looks like this: while an alert is actually making a sound, quieter alerts are
held back, which seems to be the intention. But when the alert makes no sound at all, they stay held
back for as long as it keeps running — I saw a reminder still blocked twelve minutes in.

Navid200:
Thanks.  
That (an active alert that is currently not using the media player blocking a reminder) is not what I expect.  Thanks for reporting this.  

Let me test and fix.

Navid200:
I had the collision management at the alert level, which was wrong.  Thanks for catching that.
It is now at the hardware resource level (media player and vibrator).  

So, if a glucose level alert is active and the ascending volume profile is selected, a reminder will still go through and make sound unless it arrives exactly when the glucose level alert is making sound.  But, if it arrives in between the 1-minute intervals, it will go through and make sound.  

Unfortunately, there is no onComplete listener for the vibrator.  This PR deals with vibrator collision in a very simple manner.  In the future when we offer custom vibration patterns, we will have to add a feature to address it.  I did not want to do that in the PR, which would make it even more complicated.  
The result is that a lower priority vibration could interrupt a higher priority vibration for now.   I don't think this is a catastrophe at the moment.  
We can also deal with it in a follow-up PR even if we never add custom vibration patterns.  

Please check this when you get a chance.  

Thanks

ahaarrestad:
Ran your branch on an emulator next to 2026.03.01 — same data, ringer on normal, mg/dL.

**The blocking is gone.** With the ascending profile and a low alert at 50, the low alert stays quiet as intended and the reminder nine seconds later now makes its sound. That was the main thing, and it works.

One question about a side effect, in case it wasn't intended: with the alert profile set to **Silent** or **Vibrate only**, other alerts are now silent too, and turning on their own "Override silent mode" doesn't bring the sound back.

I tested it with Missed Reading, "Override silent mode" on for it, ringer on normal. On 2026.03.01 it sounded; on the branch the notification appears but stays quiet. Same for reminders.

Is that the behaviour you want — the glucose alert profile also deciding whether reminders and other alerts can make a sound?

Navid200:
Thanks for the review.

I have a schedule on my phone that enables DND late at night and disables it early in the morning because I sleep at nights.  
When my phone is in DND, I still need an alert that I can hear if my glucose goes too low.  That is the purpose of override silent mode in xDrip.  
So, the ability to silence somethings and not the others depending on their importance for me already exists.  

Now, someone may ask "what is the purpose of the silent volume profile"?
In my opinion, if someone is deaf, they cannot hear sound.  But, others around them can.  So, it would be inappropriate for their xDrip to make noise when their blood glucose goes low.  They need a different way for being alerted.  They can choose the silent or vibrate only for their volume profile.  They are deaf.  This is not something that only happens at nights.  

That's why in my opinion, setting the volume profile to silent is very different from enabling DND on your phone.  Therefore, overriding silent mode should not override the chosen silent volume profile.

But, I admit this is a fundamental question.  What is the purpose of the silent volume profile?  
It is not a technical question.  

I cannot claim that I added the silent volume profile to xDrip.  It had always bothered me until I started considering deaf people.  But, no early xDrip developer told me that.  So, I could be wrong.   

I am happy to add a feature to xDrip such that when a user chooses the silent or the vibrate only volume profile for a dialog to come on screen explaining that override silent mode will not override the silent or the vibrate only profile.  And the user will have to tap OK for the dialog to go away.

Navid200:
I'm now thinking of just adding zombie cleanup and strings to this same PR and be done with everything in a single PR.  

Screenshot 2026-09-06 082847

Navid200:
I can undo either or both of these if it helps with the review process.
I will run my tests now again.

Navid200:
This is how many different notification channels I had on my main phone having gone back and forth for testing through so many different versions:
Screenshot_20260906-083257
  
And this is what I have now after the last commits:
Screenshot_20260906-083611

Navid200:
It looks OK to me.
The test worked as expected.
I am using this now.

ahaarrestad:
Ran the merged tree — your branch merged onto current master, not the branch on its own.

Upgrading over an existing install: the old channels, including the hashed `bgAlertChannel  📳` and `reminderChannel  🎵📳💡`, are gone, and exactly the four are left. Stable across repeated restarts. Matches your screenshots.

Sound still works on all three paths: a low glucose alert, a reminder, and Missed Reading all play, and the volume xDrip raises for an alert is put back afterwards.

Tests and the build are green on the merged tree:

    ./gradlew :app:testFastDebugUnitTest :wear:testFastDebugUnitTest :app:assembleFastDebug

Keep both the zombie cleanup and the strings — the cleanup is what makes an upgraded phone end up with four channels instead of four more.

From my side this is **ready to merge**.

Separately, and explicitly not a blocker for this PR — it belongs in its own PR later, if anyone wants it at all: if the media player fails to start an alert sound, xDrip can stay in an "alert is active" state until the next time an alert stops, and while it is in that state a lower-priority alert will not vibrate. Same shape as the `setOnErrorListener` case — there are four more exits in `playFile` after `activeTag` is set that do not clear it. Nothing here should hold up the merge.

Navid200:
@ahaarrestad Many thanks for the tests and catching the 4 exists.
I can fix them now.  But, I agree with you that we can go ahead and fix these in a follow-up PR.  

@jamorham We can fix the 4 remaining exits now by adding activeTag = ""; to each.  
Or we can do that in a follow-up PR.

Navid200:
@jamorham Would you please tell me what this needs before it can be merged?

jamorham (approved):
Okay, we can merge this now but I think it probably deserves a follow up PR to tidy any loose ends.

Navid200:
Absolutely.

PR #4682, opened 2026-08-11
Fixed crash in Gabbro/Round2 and some other minor display issues.

jstevensog:
Discovered a code error that caused the watch face to crash, only on Gabbro.
Added some minor display fixes, so this is now almost identical to the Chalk face.


jstevensog:
Closing this request, as I discovered I introduced a crash bug in Aplite face whilst fixing Gabbro crash.

Issue #4688, opened 2026-08-14
Missed readings should block alerts


Navid200:
I am worried about the following scenario:

1- Your glucose is high and you get a High glucose level alert.  You snooze it.
2- For some reason, xDrip stops receiving readings.
3- The snooze time for your high alert ends and there is still no new readings.  
4- Will xDrip issue a High alert reraise?

If the answer is yes, that could lead to a catastrophe as for all we know your glucose by then may have come in range or even has gone low.  
xDrip should never issue an alert if there are no readings.  

I will need to verify this safety measure is in place.  If not, I will need to open a PR to add one.

PR #4693, opened 2026-08-18
Unify alerts so override silent has the same meaning for all alerts

00george:
Currently other alerts such as missed readings alert do not sound if the phone volume is at 0 even if the override silent is enabled. This PR fixes that by unifying alerts to follow the same system as high and low alerts by the sound via the media player with a silent notification channel. 

This is important as currently people are unaware of missed readings when their phone volume is off. 

I am creating this as a draft to get some early feedback and may have over simplified this. I have removed isSoundBlockedBySystem and it doesn't actually seem to be used in any actual logic and is just used for logs. It doesn't seem useful but perhaps it has been used for debugging the past?


Navid200:
There is ongoing work on this already: https://github.com/NightscoutFoundation/xDrip/issues/4688

PR #4699, opened 2026-08-22
Behavioural test coverage for the HTTP paths (no production changes)

ahaarrestad:
Behavioural test coverage for the HTTP paths, written ahead of an okhttp 4 → 5 upgrade so that upgrade has something to be measured against.

**No production code changes. No version changes.** Five new test classes, one existing test class extended, and one `testImplementation` dependency (`okhttp-tls`, matching the okhttp version already declared). Nothing here ships in the APK.

### Why separately

`master` has no round-trip coverage of these paths today. It is not that they are untested in every sense — `DesertCommsTest` and `CareLinkClientTest` exist and assert structure, e.g. that a factory returns non-null. What is missing is any test that sends a request and looks at what arrived:

- `OkHttpWrapperTlsTest` — the shared client completes a TLS handshake and a cleartext round trip.
- `OkHttpWrapperSharedClientTest` — extended: what `newBuilder()` derivations inherit, and what they can override. Twenty-two call sites across `:app:` and `:wear:` derive from this one client.
- `DesertTrustManagerTest` — the naive trust manager reaches a self-signed peer; the pinning hostname verifier refuses an unpinned one.
- `EditableCookieJarTest` — CareLink issues its session cookie on one host and is called on another under the same registered domain. That decision goes through okhttp's public suffix database, which okhttp 5 relocates into the Android artifact's assets.
- `WireMessageRoundTripTest` — the phone → watch message format, including unknown-field preservation. okio crosses a major version in the upgrade.
- `NocturneApiClientTest` — the Nocturne SDK driven over the shared client.

### What this does not cover

Every one of these is findable by anyone reading the diff, so they are stated up front rather than left to be discovered.

- **The certificate pin's *accept* path is not asserted.** Not for want of the certificate: `res/raw/localhost_cert.bks` holds it and its key, and its SHA-256 is `TrustManager.CERTIFICATE_PIN` verbatim. Serving it from `MockWebServer` still would not reach the accept branch — see the next point.
- **The pin's hash comparison never runs in these tests at all.** `TrustManager.getXdripHostVerifier()` reads the peer chain through the long-deprecated `SSLSession.getPeerCertificateChain()`, and on a modern JDK that chain cannot be read: the JDK no longer ships the implementation behind `javax.security.cert` (`ClassNotFoundException: com/sun/security/cert/internal/x509/X509V1CertImpl`), and Conscrypt masks it as `IllegalArgumentException: Self-causation not permitted`. Production catches `Exception` and returns false, so the rejection test proves the client refuses to transmit — it does not prove a hash was compared. `peerCertificateChain_isUnreadableSoTheHashIsNeverCompared` says so in the file and goes red the day that changes. This is unrelated to okhttp and unchanged by the upgrade. The pin is presumed live on Android, which still ships those classes; this does not measure that. Related, and also not addressed here: on JDK 21 the same call resolves to `NoClassDefFoundError`, an `Error` that `catch (Exception)` does not catch.
- **The trust tests pin the mechanism, not the wiring.** `DesertComms.getHttpInstance()` is private and statically cached, so the tests reassemble its builder chain by hand. If `.hostnameVerifier(...)` were dropped from production tomorrow, they would still pass.
- **`NocturneApiClientTest` covers client registration over cleartext, not the SGV upload.** The SDK is shown to work on the shared client; the upload path, the device-flow endpoints and the SDK's progress-reporting bodies carry no runtime coverage here.
- **`WireMessageRoundTripTest` encodes and decodes inside `:app:` only.** The real contract is app-encode → wear-decode across two okio majors, and that crossing is not tested.
- **`ShareRest` is not an uncovered custom client** — `getOkHttpClient()` derives from `OkHttpWrapper.getClient().newBuilder()` like the rest. What is untestable is its own wiring: the method is private, the Retrofit base URLs are hardcoded to Dexcom's hosts, and the injectable-client constructor bypasses it. Its hostname verifier returns `true` unconditionally, so unlike Desert Sync there is no pin to test. Adding a production seam for it is out of scope.

### Verification

All green, and the release gate passes:

```
./gradlew assembleProdRelease testProdReleaseUnitTest
bash ./etc/CheckBuild/check-release.sh test
```

https://claude.ai/code/session_01FPhzyz5ToRGSBC9pECwqZu

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -376,6 +376,7 @@ dependencies {
     testImplementation "org.robolectric:robolectric:4.16"
     testImplementation "com.google.truth:truth:1.4.5"
     testImplementation 'com.squareup.okhttp3:mockwebserver:4.12.0'
+    testImplementation 'com.squareup.okhttp3:okhttp-tls:4.12.0'
 
     testImplementation 'org.mockito:mockito-inline:2.13.0'
     testImplementation 'org.mockito:mockito-core:4.11.0'
--- a/app/src/test/java/com/eveningoutpost/dexdrip/cgm/carelinkfollow/auth/EditableCookieJarTest.java
+++ b/app/src/test/java/com/eveningoutpost/dexdrip/cgm/carelinkfollow/auth/EditableCookieJarTest.java
@@ -58,7 +58,7 @@ public class EditableCookieJarTest extends RobolectricTestWithConfig {
                 .build();
     }
 
-    // ===== session handoff =======================================================================
+    // ===== Session handoff =======================================================================
 
     /** A cookie set by the server is stored by the jar and replayed on the next request. */
     @Test
@@ -89,7 +89,7 @@ public class EditableCookieJarTest extends RobolectricTestWithConfig {
      * okhttp's bridge interceptor during {@code execute()} — not in this test's setup.
      */
     @Test
-    public void jar_replaysParentDomainCookieAcrossHosts() throws Exception {
+    public void cookiePolicy_replaysParentDomainCookieAcrossHosts() throws Exception {
         // :: Setup
         server.enqueue(new MockResponse()
                 .addHeader("Set-Cookie", AUTH_COOKIE + "=abc123; Domain=minimed.eu; Path=/")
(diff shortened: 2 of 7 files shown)


jamorham (approved):
Thanks for the additional tests

PR #4701, opened 2026-08-22
Send the Nocturne access token as a Bearer header

### LINKED ISSUE #4700: Nocturne upload never runs: NocturneUploader constructor always throws
Nocturne upload never runs. Once you have authenticated, every upload attempt throws inside the `NocturneUploader` constructor and the exception is swallowed. Nothing is uploaded and nothing is shown in the UI.

Found on `master` at `f30226459`, from #4529.

### What happens

`NocturneUploader`'s constructor ends with:

```java
// NocturneUploader.java:110
apiClient.setAccessToken(accessToken);
```

In `nocturne-java` 0.2.4 that method has no body other than a throw:

```
public void setAccessToken(java.lang.String);
   0: new    // class java/lang/RuntimeException
   4: ldc    // String No OAuth2 authentication configured!
   9: athrow
```

No condition, no early return. `ApiClient.init()` leaves `authentications` an empty map, so there is no auth scheme to set a token on. `HttpBearerAuth` ships in the jar but is never registered.

`UploaderTask.java:180` constructs the uploader inside a `try` whose `catch (Exception e)` logs and returns, so the throw never surfaces:

```java
final NocturneUploader nocturneUploader = new NocturneUploader(xdrip.getAppContext());
```

### Who this affects

Only users who have actually completed the device flow. Without a token, `getValidAccessToken()` returns null and the constructor takes the early return at `NocturneUploader.java:96` with a log line. With a valid token it goes on to line 110 and throws. So the failure is invisible in exactly the case where upload is supposed to work.

`NocturneOAuthService` is unaffected — it never calls `setAccessToken`, which is why registration and the device flow work.

### Notes

The `ApiClient` surface that does work is `addDefaultHeader`, already used two lines above for `Origin`.

Unrelated to okhttp — measured at 4.12.0. Turned up while writing test coverage for the HTTP paths (#4699): the planned test could not construct the uploader at all.

ahaarrestad:
Fixes #4700. Nocturne upload has never worked for anyone who completed the device flow.

`NocturneUploader`'s constructor ended with `apiClient.setAccessToken(accessToken)`. In nocturne-java 0.2.4 that method has no body other than a throw:

```
public void setAccessToken(java.lang.String);
   0: new    // class java/lang/RuntimeException
   4: ldc    // String No OAuth2 authentication configured!
   9: athrow
```

`UploaderTask` constructs the uploader inside a `try` whose `catch (Exception)` swallows it, so nothing uploaded and nothing was reported. Users without a token were unaffected — `getValidAccessToken()` returns null and the constructor returns early — so it failed invisibly in exactly the case where upload is supposed to work.

### The fix

Drop `setAccessToken` and send the token as a default header on the builder that already sets `Origin`.

This is not a workaround. The SDK registers no authentication scheme at all: `getAuthentications()` returns a map that both `ApiClient` constructors wrap in `Collections.unmodifiableMap`, so `HttpBearerAuth` cannot be registered at runtime even though it ships in the jar. It also never needs to be — every generated call site passes a zero-length `authNames` array to `buildCall` (checked all 93 of them across the twelve API classes this uploader uses), so `updateParamsForAuth` iterates nothing and its `Authentication undefined:` branch is unreachable. `processHeaderParams` applies `defaultHeaderMap` to every request, and `Authorization` appears in no generated API class, so nothing overwrites it. All streams share the one `ApiClient`, so all of them carry the header.

The header is fixed at construction, which is safe because `UploaderTask` builds a fresh uploader per run and `getValidAccessToken()` refreshes before returning.

### Tests

New `NocturneUploaderAuthTest` drives the real `(Context)` constructor against `MockWebServer` and reads the headers off the wire.

The existing `NocturneUploaderTest` is thorough but goes through the package-private `NocturneUploader(boolean)` seam, which leaves `apiClient` null. Nothing in the suite had ever executed the `(Context)` constructor, which is why this shipped.

One thing worth knowing if you touch this file: `NocturneUploaderTest` switches four `nocturne_upload_*` flags on and never switches them back, and `Pref` caches its `SharedPreferences` for the life of the JVM. `uploadDeviceStatus()` has no data gate at all and three more streams read the database directly, so a leaked flag puts an extra request on the wire. The new test forces all seven off in `@Before` rather than trusting the defaults.

### Not fixed here

- **`UploaderTask`'s swallowing `catch`.** It is shared by Nightscout, InfluxDB, Mongo and wear; widening it is a behaviour change well outside this bug. It no longer hides anything here once the constructor stops throwing.
- **No recovery when the server revokes a token before its local expiry.** `getValidAccessToken()` only refreshes inside the 60 s window, so a revoked grant means every request 401s and the 6-minute retry runs forever without triggering a refresh. Pre-existing, and unreachable until now because no request was ever made.
- **An empty access token passes the constructor's guard.** Line 88 checks `== null` but not `isEmpty()`, so a 200 refresh response missing `access_token` would send `Bearer ` and 401 permanently. One word closes it, but it changes constructor behaviour in an edge case, so it is not in this PR. The branch one line up does guard `isEmpty()`, so the asymmetry looks unintentional.
- **The basal-injection delete gap** (deleted long-acting entries stay on the server — nocturne-java has no `basalInjectionDeleteBySyncIdentifier`). Separate defect.

### Verification

All green:

```
```

https://claude.ai/code/session_01FPhzyz5ToRGSBC9pECwqZu

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneUploader.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/nocturne/NocturneUploader.java
@@ -106,8 +106,11 @@ public class NocturneUploader {
                 .build())
                 .setBasePath(basePath)
                 // Cloudflare-fronted instances reject POSTs whose Origin looks cross-site
-                .addDefaultHeader("Origin", basePath);
-        apiClient.setAccessToken(accessToken);
+                .addDefaultHeader("Origin", basePath)
+                // nocturne-java 0.2.4 registers no authentication scheme: setAccessToken() is an
+                // unconditional throw and getAuthentications() is unmodifiable. The generated APIs
+                // pass an empty authNames array, so nothing consults or overwrites this header.
+                .addDefaultHeader("Authorization", "Bearer " + accessToken);
         ready = true;
     }
 
--- a/app/src/test/java/com/eveningoutpost/dexdrip/nocturne/NocturneUploaderAuthTest.java
+++ b/app/src/test/java/com/eveningoutpost/dexdrip/nocturne/NocturneUploaderAuthTest.java
@@ -1,0 +1,184 @@
+package com.eveningoutpost.dexdrip.nocturne;
+
+import com.eveningoutpost.dexdrip.RobolectricTestWithConfig;
+import com.eveningoutpost.dexdrip.models.BgReading;
+import com.eveningoutpost.dexdrip.models.JoH;
+import com.eveningoutpost.dexdrip.utilitymodels.Constants;
+import com.eveningoutpost.dexdrip.utilitymodels.PersistentStore;
+import com.eveningoutpost.dexdrip.utilitymodels.Pref;
+import com.eveningoutpost.dexdrip.xdrip;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.concurrent.TimeUnit;
+
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import okhttp3.mockwebserver.RecordedRequest;
+
+import static com.google.common.truth.Truth.assertThat;
+
+/**
+ * How {@link NocturneUploader} authenticates against a Nocturne instance.
+ * <p>
+ * These tests go through the real {@code (Context)} constructor rather than the package-private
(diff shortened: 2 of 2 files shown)


jamorham:
Thanks for the work on this, PR is marked as draft, whats the status?

ahaarrestad:
Need to check the code once more before I can send it to review.

ahaarrestad:
@jamorham ready for review

jamorham (approved):
Thanks for finding and fixing

PR #4702, opened 2026-08-22
Make snooze alert texts translatable

olorinmaia:
Some texts on the Snooze screen were hardcoded in English and could not be translated:

- The snooze time picker values ("30 minutes", "2.0 hours")
- The status line "Active alert exists named ... (X minutes left)"
- The "you re-enable" part of "All/Low/High alerts disabled until ..."

This PR moves them to string resources:

- The picker reuses the existing `unit_minutes` / `unit_hours` strings, so it is translated right away in all languages.
- The active alert status line becomes two format strings with placeholders (snoozed / will rerise).
- "you re-enable" becomes a new string, and the three duplicated `MessageFormat` blocks are replaced by one small helper method.

No visible change in English. Time formatting is unchanged (`DateFormat.getTimeInstance()` gives the same format as `{n,time}` did).

**Tested:** built and ran the app with temporary Norwegian translations of the new strings. All three places show the translated texts as expected. 

Note: `wear/.../SnoozeActivity.java` has the same hardcoded texts, but this PR stays phone-only on purpose. The wear module currently ships no translations and is not part of `crowdin.yml`, so moving its texts to string resources would not change anything for users. If the wear module is added to Crowdin at some point (a maintainer decision), I can do the same fix for wear in a follow-up PR. See discussion below.

**Some screenshots when xDrip is using Norwegian display language**

| Before | After |
|--------|--------|
| image | image |
| image | image |
| image | image |

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/SnoozeActivity.java
@@ -1,6 +1,6 @@
 package com.eveningoutpost.dexdrip;
 
-import java.text.MessageFormat;
+import java.text.DateFormat;
 import java.util.Date;
 
 import android.app.Dialog;
@@ -118,9 +118,9 @@ public class SnoozeActivity extends ActivityWithMenu {
 
     static String getNameFromTime(int time) {
         if (time < 120) {
-            return time + " minutes";
+            return time + " " + gs(R.string.unit_minutes);
         }
-        return (time / 60.0) + " hours";
+        return (time / 60.0) + " " + gs(R.string.unit_hours);
     }
 
     static int getTimeFromSnoozeValue(int pickedNumber) {
@@ -374,8 +374,10 @@ public class SnoozeActivity extends ActivityWithMenu {
         } else {
             sendRemoteSnooze.setVisibility(View.GONE);
             if(!aba.ready_to_alarm()) {
-                status = MessageFormat.format("Active alert exists named \"{0}\" {1,choice,0#Alert will rerise at|1#Alert snoozed until} {2,time} ({3} minutes left)",
-                        activeBgAlert.name, aba.is_snoozed ? 1 : 0 , new Date(aba.next_alert_at),(aba.next_alert_at - now) / 60000);
+                status = getString(aba.is_snoozed ? R.string.active_alert_snoozed_until : R.string.active_alert_rerise_at,
+                        activeBgAlert.name,
+                        DateFormat.getTimeInstance().format(new Date(aba.next_alert_at)),
+                        String.valueOf((aba.next_alert_at - now) / 60000));
             } else {
                 status = getString(R.string.active_alert_exists_named)+" \"" + activeBgAlert.name + "\" "+getString(R.string.bracket_not_snoozed);
             }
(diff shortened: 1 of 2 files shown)


parapenT1sta:
Please do also for wear devices.

olorinmaia:
> Please do also for wear devices.

I looked into it, but I will keep this PR phone-only, and here is why.

The wear module ships no translations at all — it only has an English `values/strings.xml`, and `crowdin.yml` only registers the phone app's strings file.

So even with the texts moved to resources, the watch would still show English in every language. Fixing the wear texts only makes sense together with adding the wear module to the Crowdin project, and that is a maintainer decision — it would open ~250 wear strings for translators, and there may be reasons it was left out.

If the wear module gets added to Crowdin, I am happy to do the same fix for the wear app in a follow-up PR.

jamorham (approved):
This looks good and improves things. I'm thinking we need to expand the translation unit tests so that they can catch if any of the localized strings are missing the format string placeholders.

jamorham:
@ahaarrestad do you want to tackle adding unit tests which can detect wherever format string place holders have been broken by translators in the localized strings?

Navid200:
It would be great if we could also somehow catch if any translation was going to show up in logs.

ahaarrestad:
@jamorham Yes, I can take that.

Comparing the placeholder set per string key against the English `values/strings.xml`: 14 mismatches across 5 locales today.

Three are Bulgarian dropping the placeholder — `label_show_steps`, `label_show_steps_km`, `label_show_steps_mi`. The first is formatted at `BGHistory.java:216`, so in Bulgarian the label shows without the step count.

The other eleven are the opposite, a `%` the English does not have, mostly literal percent signs typed by translators such as Turkish `% düşük`. They break nothing today only because those strings are read with plain `getString()` or from a preference XML title — the day someone passes an argument to one of them, that locale throws. Bulgarian `number_error_` has an added `%s`.

The test: read every `values-*/strings-*.xml`, compare the placeholder multiset per key against the English, fail on any difference. Runs in the existing unit suite.

One thing to settle first: where do the fixes go? If I correct the XML directly, the next Crowdin sync overwrites it and the test goes red on your l10n PR. Fix them in Crowdin, or should the test carry an exception list?

Nothing to do for wear — no translations there at all, as olorinmaia says.

@Navid200 catching translated text that reaches the log is a separate test — it walks the source for `getString()` results going into `Log.*` rather than comparing resources. Doable, noisier, and I would do it after this one.

ahaarrestad:
Picking this up, and flagging the order so nothing here is a surprise.

The sweep is written and green: every key where the English string takes at least one readable
format argument, compared against all locales. The scope is deliberate — a `%` in a translation of
a string we never format is a percent sign, not a broken placeholder — so there is no exception
list to keep alive.

Three Bulgarian strings are the only ones failing today. `label_show_steps`, `_km` and `_mi` were
given description text instead of the label, so `%1$d`/`%1$s` are gone and Bulgarian users see no
step count or distance. #4751 deletes them; the English fallback with the number beats a sentence
without it.

They come from Crowdin, so a later sync brings them back unless the source is fixed, and that needs
your access.

The test itself follows as its own PR once #4751 is in, so it lands green rather than red.

PR #4707, opened 2026-08-23
Retire calibration alerts

Navid200:
I don't believe we need calibration alerts any longer.
We did when we supported G5.  A G5 needed to be calibrated twice a day.  
We no longer support any sensor that requires frequent calibrations and calibrated locally by xDrip.

This PR removes the calibration alerts from the menu.
Screenshot_20260823-181916
  
Installing an older version after updating to this will not cause a crash.
Associated settings in Preferences have been removed as well.  

Compiled and installed.  Looks OK.


jamorham:
Will this have a negative effect on people calibrating libre2 sensors locally? 

I'm also wondering about whether we should be forcing `play_sound_for_initial_calibration` to true when that is in the initial status helper, maybe drop that line?

Navid200:
@psonnera @PhilTU 

Does someone using xDrip with a Libre sensor today get any calibration alerts?  
I thought these alerts have no use any longer and opened this PR, which removes them from xDrip.  But, you can see jamorham's question above.  I have never used a Libre sensor.  So, I cannot answer his question.  

What is your view on this?

PhilTU:
I select "No calibration" in the OOP2 settings of xDrip. If I wish to manipulate the reported value I seek out the settings. Any alert or pop up about calibration is a nuisance to me.

Some people use calibration and I don't know exactly if the alerts are of any value to them.

Navid200:
Thanks

Let me post on facebook and ask this from everyone.

Navid200:
I asked on facebook.  Several Dex and Libre users stated there was no need for calibration alerts.  

There was one who stated that occasionally when they forget to enter the 4-digit calibration code for their G6, they would like to continue to receive the notification reminding them to calibrate once a day.   I told them that those notifications came from the G6 transmitter and they will continue even if we remove calibration alerts.  

One individual stated they used calibration alerts to remind them to calibrate their G6 once a day.  I told them they should use xDrip reminders instead.  

This is my concern:
An inexperienced user may see the alert and think that xDrip developers calibrate once a day and for that reason decide to calibrate once a day.  

I'm not suggesting we remove calibration from xDrip.  Only the calibration alerts.  
12 years (may be even more) ago when the Dex G4 intelligence was 90% inside the receiver, and xDrip wanted to replace the receiver, I can only assume that the developers had to include the calibration alert because that is what the receiver did.  The G4 manufacturer required calibration on a regular basis.
Today, there is no manufacturer that requires calibration on a daily basis.  

If a day comes that such a sensor comes to market again and we add support for it as a collector, I will be the one opening a PR bringing the calibration alerts back.  

I'm still waiting for more responses from users on facebook.  We can give it a few more days.

PR #4709, opened 2026-08-24
Invalid treatment time

### LINKED ISSUE #4640: Treatment menu time tab lock
If you start entering time in the time tab of the treatment menu, you can go to another tab before finishing.
If you do that, you will never be able to submit the form.
As long as you do not enter a valid time on the time tab or delete it completely, the green checkmark will not appear and you will not be able to submit.

It makes sense to expect a valid time entry.  But, users post on facebook asking why they cannot enter treatments.  It turns out they have accidentally entered a value on the time tab.

We need to come up with some sort of sign or symbol or note to highlight this to help the user.

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/4640  

Recreate the problem:
Go to the treatment menu.  Enter an insulin value.  You will see a green checkmark for submission.  
Now, go to the time tab and enter just 1.  This is an invalid time value.  Go back to the insulin tab.  You will see that the green checkmark has disappeared.  

This PR shows the green checkmark as long as there is something (insulin, carbs or glucose) to submit even if the time value is invalid.
Tapping on it with an invalid time will bring up a dialog:

Screenshot_20260824-164835

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/PhoneKeypadInputActivity.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/PhoneKeypadInputActivity.java
@@ -2,6 +2,7 @@ package com.eveningoutpost.dexdrip;
 
 import static com.eveningoutpost.dexdrip.Home.startHomeWithExtra;
 
+import android.app.AlertDialog;
 import android.graphics.Color;
 import android.os.Bundle;
 import android.util.DisplayMetrics;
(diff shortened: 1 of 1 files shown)


jamorham (approved):
Looks reasonable enough, do you want to follow up with more validation on whether the time entered is valid and to extract the strings once this new behavior is confirmed good?

Navid200:
I'm sorry I don't follow.

Navid200:
Once the user enters a valid time, and taps on the checkmark again, it will be accepted.

This worked already.  What was missing was a way to inform the user why the checkmark had disappeared.
But, if we just bring the checkmark back, it will not be good if the value is invalid.  The dialog tells the user what they need to do.  

The combination of the two changes this PR makes tells the user what the support community has been telling users for years.  

Please explain what additional changes you think we should make.

jamorham:
for example does it allow a time like 9999 ?

Navid200:
I see.
OK, we should fix that: https://github.com/NightscoutFoundation/xDrip/issues/4714

PR #4710, opened 2026-08-25
[cleanup] Replace android.preference.PreferenceManager with androidx (Batch 3: 9 files)

ahaarrestad:
## Why

Continues #4479 → #4595 (Batch 1) → #4654 (Batch 2). Same mechanical swap, same reasoning as the
two merged batches; nothing new to argue. Nine more files off the deprecated class.

Both `PreferenceManager` classes return the same default `SharedPreferences` file, so no user data
moves and no key is renamed. That is asserted below, not assumed.

## What

Nine files, **one import line each**, no logic changes:

- `AlertList`
- `MissedReadingActivity`
- `NavDrawerBuilder`
- `influxdb/InfluxDBUploader`
- `localeTasker/receiver/FireReceiver`
- `services/PlusSyncService`
- `sharemodels/ShareRest`
- `utilitymodels/IdempotentMigrations`
- `utilitymodels/NightscoutUploader`

```diff
-import android.preference.PreferenceManager;
+import androidx.preference.PreferenceManager;
```

Production diff is 9 files × (1 insertion, 1 deletion). Everything else in the diff is tests.

## Correction to the file count

#4479 has Category 2 at 56 files after Batch 2's scope correction. Four of those were already clean
in `master` when this batch measured them — `LicenseAgreementActivity`, `stats/DBSearchUtil`,
`utilitymodels/pebble/PebbleDisplayAbstract` and `utilitymodels/Pref`. Those are Batch 2's own
files; the list was never decremented when #4654 merged. The real figure going in was 52.

After this batch, **43 remain**. Measured, not counted from the list:

```
$ grep -rl 'android\.preference\.PreferenceManager' app/src/main | wc -l
46
```

46 = 43 Category 2 + the three Category 3 files (`utils/Preferences`, `utilitymodels/ColorCache`,
`xdrip`).

## Why these ten, and what was left out

Selection criterion is reachable public behaviour, not file size — a file entered the batch only if
its preference read can be observed through a public API.

Left out on purpose:

- **`ErrorsActivity`, `NSClientReceiver`** — both reach preferences through a
  `private static SharedPreferences` that nothing resets. The first test to touch it freezes the
  in-memory value cache for the rest of the JVM, so a characterization test cannot be written. Same
  exclusion class as `Pref` in Batch 2.
- **`services/MissedReadingService:224`, `FollowerManagementActivity:56`** — dead reads.
  `getOtherAlertSnoozeMinutes(SharedPreferences prefs, …)` never touches its `prefs` parameter, and
  `FollowerManagementActivity` assigns a local it never uses. Removing dead code is a scope
  decision, not a mechanical swap, so it is left alone.
- **`ParakeetHelper`** — dropped from this batch on Navid's request. The Parakeet alerts were
  retired in #4063: the settings are gone from `pref_notifications.xml` and `legacySettingsFix()`
  forces `parakeet_status_alerts` and `parakeet_charge_silent` to false, so the notification path
  is unreachable. Migrating an import on code that is heading for deletion only makes more to
  delete later. Left on `android.preference` until its fate is decided.
- **`AlertPlayer`, `SnoozeOnNotificationDismissService`, `NSEmulatorReceiver`** — Navid is
  rewriting that code for #4632 and the #4650 replacement. Not a technical exclusion; just staying
  out of the way.
- **`utilitymodels/ColorCache`, `xdrip`** — Category 3, per the scope correction in #4654.

## Tests

Per file: characterize against the **unmodified** `android.preference` import, prove green, swap the
import, prove green again. Test and swap are separate commits so the ordering stays visible in the
history. A test that only ever saw the new import proves nothing about the old behaviour.

New behavioural tests drive each file through its public entry points — the units preference
formatting alert thresholds, the missed-reading form round trip, the two drawer feature flags, the sync kill switch, the Share host region, the Tasker snooze
actions, and the legacy REST URI migration.

`./gradlew :app:testFastDebugUnitTest :wear:testFastDebugUnitTest` — all green.

### Two test files still import `android.preference`, deliberately

`NightscoutUploaderCallerTest` and `InfluxDBUploaderCallerTest` seed preferences through the **old**
framework class while the production code they exercise now reads through **androidx**. Green there
is a live cross-API assertion that the two classes address one store — the strongest evidence in
this PR, and the only kind that survives into the merged state.

Please don't tidy those two imports away. It looks like leftover cleanup and it is the proof.

### Coverage, stated plainly

- **`InfluxDBUploader` and `NightscoutUploader` have no new test.** They are covered by the two
  pre-existing `*CallerTest` classes above — which is also why those two are the cross-API pair.
- **`PlusSyncService:88` (`onCreate`) is not covered.** Robolectric never instantiates the service;
  the tests go in through `startSyncService`.
- Every other migrated read has at least one test that goes red when that read is removed. Checked
  by mutation, not assumed.

## Manual emulator test (the #4479 review condition)

`API34_Phone` (Android 14), FastDebug build of this branch. Preference search walks the preference
XML tree at runtime, which unit tests cannot exercise.

| # | Step | Result |
|---|------|--------|
| 1 | Settings opens, tree renders | ✅ |
| 2 | Search `glucose` | ✅ 3 hits, each with its breadcrumb path |
| 3 | Tap a nested result | ✅ "Glucose Level Alerts List" lands on **Alarms and Alerts** |
| 4 | Search `insulin` | ✅ 3 hits; "Insulin Sensitivity" lands on **Predictive simulations** |
| 5 | Search `zzzznomatch` | ✅ "No results", no crash |
| 6 | Navigation drawer | ✅ licence-accepted set, both feature flags off |
| 7 | Missed readings round trip | ✅ enabled the alert, 30 → 42 min, left, reopened: both held |
| 8 | Show Data Tables toggle | ✅ on → both table screens appear; off → both gone |
| 9 | `adb logcat -b crash -d` | ✅ no entries from `com.eveningoutpost.dexdrip` |

Step 7 was checked against the store, not just the form. After `MissedReadingActivity` — now reading
and writing through `androidx.preference.PreferenceManager` — closed:

```
$ adb shell run-as com.eveningoutpost.dexdrip \
    cat /data/data/com.eveningoutpost.dexdrip/shared_prefs/com.eveningoutpost.dexdrip_preferences.xml
    <boolean name="bg_missed_alerts" value="true" />
    <string name="bg_missed_minutes">42</string>
```

`<package>_preferences.xml` is the same default file `android.preference` writes. The swap addresses
the store the rest of the app already uses — on a device, not only under Robolectric.

## Screenshots

| Settings | `glucose` | Nested result |
|---|---|---|
| Settings | glucose | nav |

| `insulin` | Insulin result | No match |
|---|---|---|
| insulin | insulin nav | no results |

| Missed readings, before | Edited to 42 | Reopened |
|---|---|---|
| before | edited | reopened |

| Data Tables on | Drawer with tables | Drawer, flag off |
|---|---|---|
| on | drawer | drawer off |

## After this

43 Category-2 files remain in the app module. `wear/src/main` holds a further 27 with the same
import, still a separate track. Batch 4 next, same shape.


jamorham (approved):
Thanks

Navid200:
The changes you make to xDrip are extremely important.
But, they are not visible to many users.  That does not mean they are not crucial.  They are extremely valuable.  
Thanks so much for your work.  
It is wonderful to see your dedication and enthusiasm.  

Thanks

PR #4711, opened 2026-08-25
fix: carelink not working for medtronic go users

m0rt4l1n:
This PR updates the advertised app version from `3.6.0` to `3.8.0` for the Medtronic CareLink Follower, which fixes Medtronic Go users not being able to retrieve their data (see #4547)

This has been tested with a Medtronic 780G + Guardian 4 user and Medtronic InPen + Simplera user.

Note: This PR does not add the `os` field to the request. Manual testing showed no improvement while changing both the version and adding the field but after remove the field again (and keeping the new version) bg data was successfully retrieved.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/cgm/carelinkfollow/auth/CareLinkAuthenticator.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/cgm/carelinkfollow/auth/CareLinkAuthenticator.java
@@ -75,7 +75,7 @@ public class CareLinkAuthenticator {
 
     private static final String TAG = "CareLinkAuthenticator";
 
-    protected static final String CAREPARTNER_APP_DISCO_URL = "https://clcloud.minimed.eu/connect/carepartner/v13/discover/android/3.6";
+    protected static final String CAREPARTNER_APP_DISCO_URL = "https://clcloud.minimed.eu/connect/carepartner/v13/discover/android/3.8";
     protected static final String CARELINK_CONNECT_SERVER_EU = "carelink.minimed.eu";
     protected static final String CARELINK_CONNECT_SERVER_US = "carelink.minimed.com";
     protected static final String CARELINK_LANGUAGE_EN = "en";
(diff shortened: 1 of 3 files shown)


CarlosJPV:
Tested against a live **MiniMed Go + Instinct** account (EU, care partner role) by varying one field at a time. **Your PR's values work** — but I need to correct my earlier comment, which overstated the `os` part. Sorry about that.

**`os` is not required. The actual blocker is a minimum-version gate.**

Varying `appVersion` on `POST {baseUrlCumulus}/display/message`:

| `appVersion` | result |
| --- | --- |
| `3.5.0` | 206 `Upgrade application` |
| `3.6` | 206 `Upgrade application` |
| `3.6.1` | 206 `Upgrade application` |
| `3.6.9` | 206 `Upgrade application` |
| `3.7` | **200 OK** |
| `3.7.0` | **200 OK** |
| `3.8` | **200 OK** |
| `3.8.0` | **200 OK** |
| `4.0.0` | **200 OK** |
| `10.0.0` | **200 OK** |

Control: `appVersion: "3.8.0"` **without** any `os` field → **200 OK** with full data.

So the boundary is **≥ 3.7 accepted, ≤ 3.6.9 rejected**. Both `X.Y` and `X.Y.Z` formats are accepted. That explains the symptom exactly: xDrip was sending `3.6.0`, which Medtronic now refuses. Adding `os` is harmless (the real app sends it), it just wasn't the fix.

**Your PR is correct as written** — `3.8.0` clears the gate. Nothing to change.

One thing worth knowing for the future, because the two version fields behave differently:

- **Discovery path** `/connect/carepartner/v13/discover/android/<ver>` is constrained: `3.5` → Cumulus **v11**, `3.6`–`3.8` → **v13**, and **`3.9` and above fall back to a legacy config** with `baseUrlCumulus = /connect/carepartner/v2` and no Auth0 entry. So `3.8` in the URL is the right ceiling — do not bump it past that.
- **Body `appVersion`** has no upper bound (`10.0.0` works fine).

If Medtronic raises the minimum again, only the body value needs bumping; the discovery URL must stay at `3.8`.

Happy to test an APK against my Instinct sensor whenever you have one — same situation as @spindler-ctrl, I'm not set up to build Android. EU account, MiniMed Go, `appModelType: SIMPLERACC`, currently pulling ~500 readings at 1-minute resolution.

jamorham:
@benceszasz does this look okay to you?

benceszasz:
@jamorham yes, everything seems to be fine!
@m0rt4l1n, @CarlosJPV nice work!

jamorham (approved):
thanks

PR #4712, opened 2026-08-25
OB1: fix false "lost bonding" skip with Special Pairing Workaround (fixes #4706)

### LINKED ISSUE #4706: Regression: false "Appears to have lost bonding, skipping this connect!" with Special Pairing Workaround since 2026.07.08 — lost readings
## Summary

Since updating to a nightly containing commit 7213241f4 ("OB1: add mac caching fix", July 8), users with **Special Pairing Workaround** enabled (Samsung setting) get bursts of:

    Appears to have lost bonding, skipping this connect! @ 1 ... @ 10

The bonding is NOT actually lost. The check misfires, connects are skipped, and readings are lost until the safety valve at 10 skips forces a connect, or the user re-pairs or restarts the app. Several users are reporting this (e.g. on Facebook groups), and reports will likely grow as more people update.

I located the source of the problem myself and used Claude to help analyze the code history, gather the background details, and propose a possible fix.

## Root cause

Commit 7213241f4 added to `onStartCommand` in `Ob1G5CollectionService.java`:

```java
transmitterMAC = null; // clear on every service start
automata(); // sequence logic
```

The chain that turns this into a false positive:

1. `isDeviceLocallyBonded()` starts with `if (transmitterMAC == null) return false;`  — an unknown MAC is reported as "not bonded".
2. `onStartCommand` runs on every alarm wake-up and now clears the MAC every time,  right before calling `automata()`.
3. The state reset above it only maps `BOND/PREBOND/DISCOVER/CONNECT` back to `SCAN`  — **`CONNECT_NOW` is kept**. So `automata()` can enter `CONNECT_NOW` with a freshly cleared MAC and no `tryLoadingSavedMAC()` call on the way.
4. In `CONNECT_NOW`, with Special Pairing Workaround on, the guard fires:  `wasBonded.equals(getTransmitterID()) && !isDeviceLocallyBonded()` → skip connect  → missed reading. Repeats until `skippedConnects` reaches 10.

Users without the workaround setting never run this check, which matches reports that only Samsung users with the setting enabled are affected.

## Testing from my sons live-rig (Samsung SM-F711B / Z Flip3, Android 15, Dexcom G7)

I tested versions on the same afternoon (Aug 20), same sensor, same pairing:

| Build | Result |
|---|---|
| 049ce18-2026.03.27 | clean (4 days, zero occurrences) |
| c2a0ba1-2026.08.08 | skip bursts @1–@10 within hours |
| f49d55f-2026.06.17 | clean |
| f8b047f-2026.04.22 | clean |
| 805ba04-2026.03.01 | clean |
| back to c2a0ba1-2026.08.08 | skips return within 4 minutes |

All clean builds predate the July 8 commit; the only commit in the regression window touching this code path is 7213241f4. Burst pattern in the log matches the code exactly: counts to @10 (the `skippedConnects < 10` valve), ~5 s between attempts, ~5 min between clusters (G7 connect cycle). 

Recovery that worked: forget BT pairing + re-pair (or app restart — see below). On G7 every sensor is a new bonded transmitter, so each sensor change re-arms the condition.

## Suggested fix

Do not treat "MAC not known yet" as "bonding lost". Options:

- Call `tryLoadingSavedMAC()` at the top of the `CONNECT_NOW` case (or inside `isDeviceLocallyBonded()`) so the cached MAC is restored before the check, or
- Require `transmitterMAC != null` in the skip condition:

```java
if (wasBonded.equals(getTransmitterID()) && transmitterMAC != null
        && !locallyBonded && skippedConnects < 10) {
```

Either keeps the original purpose of both the July 8 MAC-clearing fix and the lost-bonding guard.

See proposed fix in 712eac9b uploaded to my repo

## Workaround for affected users until a fix ships

- Try to disable "Special pairing workaround". Forget BT device and re-pair. OR
- Rollback to recommended version for G7 - https://github.com/NightscoutFoundation/xDrip/releases/tag/2026.06.17

Logs:
[1787486871-xDrip-log.txt](https://github.com/user-attachments/files/31348328/1787486871-xDrip-log.txt)

Screenshot of issue:
| image | image |
|--------|--------|

Threads on facebook from users experiencing this error:
https://www.facebook.com/groups/xdrip/permalink/1998405371546783/
https://www.facebook.com/groups/xdrip/permalink/1988820212505299/
https://www.facebook.com/groups/xdrip/permalink/1998636931523627/

_Originally posted by @olorinmaia in https://github.com/NightscoutFoundation/xDrip/discussions/4705_

Tagging @jamorham after discussing issue with @Navid200 - let me know if there is anything else I can contribute with here.

olorinmaia:
Fixes #4706

## Problem

Since commit 7213241f4 ("OB1: add mac caching fix"), users with **Special Pairing Workaround** enabled get bursts of false `Appears to have lost bonding, skipping this connect! @ 1 ... @ 10` and lose readings, even though the bonding is fine. Full root cause analysis is in #4706. 

Short version: `transmitterMAC` is now cleared on every service start, and `isDeviceLocallyBonded()` treats an unknown MAC as "not bonded". When the service restarts while the state machine is in `CONNECT_NOW`, the lost-bonding guard fires with no MAC to check against, and connects are skipped.

## Fix

Two small changes inside the `specialPairingWorkaround()` branch of `CONNECT_NOW`:

1. Call `tryLoadingSavedMAC()` before the bonding check, so the cached MAC is restored first. This is the same call the normal `SCAN` path already makes every cycle (via `useMinimizeScanningStrategy()`), and it is cheap — it returns early when the MAC is already set.
2. Add `transmitterMAC != null` to the skip condition. If no saved MAC exists, the  bonding state is unknown, and connecting (the behavior before 7213241f4) is the safe default instead of refusing to connect.

The guard keeps its real purpose: with a known MAC and the device missing from the bonded device list, it still skips as designed. The purpose of 7213241f4 is also kept: `tryLoadingSavedMAC()` only loads the MAC stored for the *current* transmitter ID, so a stale MAC from a previous transmitter cannot leak back in.

Users without the Special Pairing Workaround setting never run this code path and are not affected by this change.

## Testing (and its limits — please read)

Tested on a Samsung Galaxy Z Flip5 (Android 16, Dexcom G7, Special Pairing Workaround enabled):

- **Without the fix (master):** false skip bursts are reproducible by restarting the collector while it is scanning/connecting, and sometimes appear on their own after a missed reading. Log in #4706.
- **With the fix:** the same procedures did not produce any false skips during a few hours of testing.
- **The guard still works:** after manually removing the BT pairing ("Unbonded mac" in the log), the fix build correctly detected the genuinely lost bonding and skipped connects. So the check still protects against the real Samsung lost-pairing case it was written for.

I have to be honest about the limits: I could only run the fix for a few hours, because the sensor had to move to my son's live rig (now on 2026.06.17 where the bug does not exist), and I have no other test equipment. The change is two lines, scoped to the workaround path, and restores pre-regression behavior — but longer real-world testing by other affected Samsung users, or a review by @jamorham against the intent of 7213241f4, would be very welcome before merging.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/services/Ob1G5CollectionService.java
@@ -365,9 +365,10 @@ public class Ob1G5CollectionService extends G5BaseService {
                         break;
                     case CONNECT_NOW:
                         if (specialPairingWorkaround()) {
+                            tryLoadingSavedMAC(); // transmitterMAC is cleared on every service start and an unknown MAC would read as not bonded
                             val locallyBonded = isDeviceLocallyBonded();
                             UserError.Log.d(TAG, "wasbonded = " + wasBonded + " local: " + locallyBonded);
-                            if (wasBonded.equals(getTransmitterID()) && !locallyBonded && skippedConnects < 10) {
+                            if (wasBonded.equals(getTransmitterID()) && transmitterMAC != null && !locallyBonded && skippedConnects < 10) {
                                 skippedConnects++;
                                 UserError.Log.wtf(TAG, "Appears to have lost bonding, skipping this connect! @ " + skippedConnects);
                                 changeState(CLOSE);


jamorham (approved):
Looks like a good catch, thanks

PR #4713, opened 2026-08-26
build: upgrade okhttp 4.12.0 to 5.3.2

ahaarrestad:
Second of three. #4699 was the preparation; this is the bump.

Four version lines in `app/build.gradle` (`okhttp`, `logging-interceptor`, `mockwebserver`, `okhttp-tls`), 4.12.0 to 5.3.2. No production source changes. Two test files needed work.

It stops at 5.3.2 because `okhttp-android:5.4.0` declares `minCompileSdk=36` and xDrip is on `compileSdk 34`. Raising that is the third PR.

## Fast fallback changes which exception surfaces

okhttp 5 defaults `fastFallback` to true. Every client derives from the singleton in `OkHttpWrapper` without overriding it, so dual-stack hosts now have their IPv4 and IPv6 routes raced instead of tried in order.

`FastFallbackExchangeFinder.find()` keeps the first failure to *complete*, attaches the rest with `addSuppressed`, and throws the first. Which route completes first is a race.

This broke the Desert Sync pinning test. `localhost` is dual-stack; MockWebServer listens on one route, and the verifier rejects the peer there only after a full handshake. The other route is refused outright, so `ConnectException` completes first and the real refusal ends up suppressed. The test now searches the chain (self, cause, suppressed) for `SSLPeerUnverifiedException`, and still asserts nothing was transmitted. Set the verifier to accept everything and it goes red.

Every `catch (IOException)` behaves as before. Both exceptions are `IOException`, and no production code catches either by name or matches on a message. Two places test the concrete type inside such a catch:

- `UpdateActivity.java:469` checks `instanceof SSLHandshakeException` to fall back to the system browser for the APK download. If the other route's connect failure completes first, that fallback stops firing and the download just fails. Update-check UI only.
- `ResponseGetterImpl.java:89` checks `instanceof SocketTimeoutException` to retry in the web follower. Read timeouts are thrown directly off an established connection, so the common case is unaffected. Only a connect-phase timeout could be masked.

Neither is touched here. Re-shaping them is a separate change.

## Public suffix list needs a Context

okhttp 5 reads its public suffix list from the AAR's assets, through a `Context` that `androidx.startup`'s `InitializationProvider` hands to `PlatformInitializer` at process start. That provider is in the merged manifest after the bump, so devices need no change.

Robolectric does not run it. Any test parsing a cookie whose `Domain` crosses a registered-domain boundary failed with `IllegalStateException: Unable to load PublicSuffixDatabase.list resource`. The failure is cached per sandbox, one per test class, so `OkHttp.INSTANCE.initialize(...)` runs at the top of the shared `@Before` in `RobolectricTestWithConfig`.

## Nocturne resolves above its declared okhttp

`nocturne-java:0.2.4`'s POM declares `okhttp:4.12.0`; Gradle resolves it up to 5.3.2, so the SDK runs against a major it was not built for. `NocturneApiClientTest` and `NocturneUploaderAuthTest` cover that path, both driving the real `(Context)` constructor over MockWebServer. The comment above the dependency claimed the SDK matched xDrip's pin; corrected.

## What else the bump pulls in

okhttp 5 publishes an Android artifact, so `okhttp:5.3.2` resolves to the `okhttp-android` AAR rather than a jar. With it:

- `androidx.startup:startup-runtime:1.2.0`, already pulled in by `androidx.work:work-runtime:2.9.1`. The `InitializationProvider` was in the merged manifest before this PR; okhttp adds a `meta-data` entry to it.
- `assets/PublicSuffixDatabase.list`, a baseline profile, and consumer ProGuard rules that are only `-dontwarn`. `app/proguard-rules.pro` already has broader `-dontwarn okhttp3.**`.
- okio 3.6.0 to 3.16.4.

No new permissions. One okhttp and one okio on the resolved classpath: `nocturne-java`'s 4.12.0 and `logging-interceptor`'s transitive both bump up cleanly. The `wear` module has no okhttp references.

`compileSdk` stays at 34 and must not go lower, since `startup-runtime:1.2.0` declares `minCompileSdk=34`. `minSdk` and `targetSdk` untouched.

## Verification

Unit suites all green:

```
```

On-device smoke test against a local Nightscout endpoint, Android 14 emulator, cleartext and TLS: HTTP 200 on both, TLS 1.3 on the https side, no `PublicSuffixDatabase.list` error.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -303,11 +303,13 @@ dependencies {
     implementation "com.google.android.gms:play-services-oss-licenses:15.0.0"
     implementation "com.google.protobuf:protobuf-java:4.34.0"
     implementation 'com.squareup.wire:wire-runtime:2.2.0'
-    implementation 'com.squareup.okhttp3:okhttp:4.12.0'
-    implementation 'com.squareup.okhttp3:logging-interceptor:4.12.0'
-    // Nocturne SDK: targets OkHttp 4.12, matching xDrip's pinned version. We exclude
-    // its gson transitive to keep xDrip's own pinned gson. Jackson is only a
-    // transitive of jackson-databind-nullable and unused at runtime on the gson path.
+    implementation 'com.squareup.okhttp3:okhttp:5.3.2'
+    implementation 'com.squareup.okhttp3:logging-interceptor:5.3.2'
+    // Nocturne SDK: declares OkHttp 4.12, so it resolves up to the 5.x pinned above rather than
+    // matching it. Covered by NocturneApiClientTest / NocturneUploaderAuthTest, which drive the
+    // real (Context) constructor over MockWebServer. We exclude its gson transitive to keep
+    // xDrip's own pinned gson. Jackson is only a transitive of jackson-databind-nullable and
+    // unused at runtime on the gson path.
     implementation('org.nightscoutfoundation:nocturne-java:0.2.4') {
         exclude group: 'com.google.code.gson'
         exclude group: 'com.fasterxml.jackson.core'
@@ -375,8 +377,8 @@ dependencies {
     testImplementation 'org.hamcrest:hamcrest-library:1.3'
     testImplementation "org.robolectric:robolectric:4.16"
     testImplementation "com.google.truth:truth:1.4.5"
-    testImplementation 'com.squareup.okhttp3:mockwebserver:4.12.0'
-    testImplementation 'com.squareup.okhttp3:okhttp-tls:4.12.0'
+    testImplementation 'com.squareup.okhttp3:mockwebserver:5.3.2'
+    testImplementation 'com.squareup.okhttp3:okhttp-tls:5.3.2'
 
     testImplementation 'org.mockito:mockito-inline:2.13.0'
     testImplementation 'org.mockito:mockito-core:4.11.0'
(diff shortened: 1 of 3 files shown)


jamorham:
Can you do an integration test using the release compiled version because there are proguard optimizations which can strip things and this needs to be tested as the behavior can differ vs unit tests or debug versions.

ahaarrestad:
Ran it on the prod release APK on an emulator. Four paths, each driven until a real server answered:

- **Update check** (plain okhttp) — parsed 200 back.
- **Dex Share follower**, junk credentials — TLS to share2.dexcom.com, then a 500 with the error text parsed out of the response body. That covers the Retrofit proxy, the gson type adapter and both interceptors.
- **Nightscout REST** against a live host — POST and GET treatments, both with a real status code back.
- **Desert Sync over HTTPS** — follower pointed at 127.0.0.1 so it talks to its own web service. TLS 1.3, three 200s, no verifier failures.

No `NoClassDefFoundError`, `NoSuchMethodError` or `VerifyError` anywhere in logcat, and no crash.

R8's own output agrees: nothing from okhttp3 or okio removed in `usage.txt`, publicsuffix kept in `seeds.txt`, and `configuration.txt` identical to the debug build.

Worth flagging: `DesertTrustManagerTest.peerCertificateChain_isUnreadableSoTheHashIsNeverCompared` records that the certificate pin is never actually compared under Robolectric. This run is the first time it was.

Tests and the release build are green:

    ./gradlew :app:testFastDebugUnitTest :wear:testFastDebugUnitTest :app:assembleProdRelease

Do you want a check in `etc/CheckBuild/check-release.sh` asserting okhttp and okio survive R8, so a future proguard change can't strip them silently?

jamorham (approved):

PR #4717, opened 2026-08-27
Treatment time validation improvement

### LINKED ISSUE #4714: Treatment time value validation
We need to reject values of hours and minutes outside the 0-23 and 0-59 ranges respectively.

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/4714 

It looks OK to me.
But, I don't use predictive simulations myself.  So, I may not be the best person for testing this.


jamorham (approved):
Looks like should work

PR #4723, opened 2026-08-28
Correct the incoming glucose broadcast page

ahaarrestad:
Follows up discussion #4716.

I have cut this back to a correction of the existing page. Earlier revisions of this PR
rewrote and expanded it. Navid's point was that the page needed correcting, not
expanding, and that tying it to xDrip's own SDK level makes it a liability the next time
we bump minSdk. Both fair.

The diff against master is now +14 -4. Everything below is verified against the source:

- `BgReading.slopeFromName()` -> `slopefromName()`. The name was wrong, and it is the
  only pointer the page gives for trend values (`NSEmulatorReceiver.java:128`).
- `Hardware Data Source` value `640G / Eversense` -> `640G / EverSense`, the actual
  label in `arrays.xml`.
- Intent Package now says `setPackage` is required rather than advisable, and why an
  implicit broadcast appears to work and then stops: xDrip also registers this receiver
  at runtime, so it keeps working until the process is not running.
- Added the `<queries>` entry a sender on targetSdk 30 or higher needs. Without it the
  broadcast is never delivered, and I think that is the most likely reason a current
  integration silently never arrives.
- One sentence that the array must hold exactly one reading. More than one is routed to
  the OOP parser and nothing is inserted (`NSEmulatorReceiver.java:107`).
- "xDrip+" -> "xDrip" in the title.

There is no reference to xDrip's own target or minimum SDK. The Android 8 restriction is
stated as the platform fact it is, which stays true through any future bump.

### DIFF
--- a/Documentation/technical/Incoming_Glucose_Broadcast.md
+++ b/Documentation/technical/Incoming_Glucose_Broadcast.md
@@ -1,57 +1,234 @@
-
 # Integration with xDrip+ via Broadcast Intent
 
 ## Overview
-This documentation provides guidance on how to send a broadcast intent from a third-party Android application to insert a glucose sensor record into the xDrip application. There are a few different broadcast receivers but this documents the Nightscout Emulation receiver just for glucose records.
+
+This page describes how a third-party Android application sends a glucose sensor reading to xDrip+
+using a broadcast intent, and what xDrip+ does with what it receives. It is written for app
+developers and sensor manufacturers.
+
+xDrip+ has several broadcast receivers. This page documents the Nightscout Emulation receiver, and
+only the insertion of glucose records.
+
+## Before anything else: xDrip+ must be set up to accept the broadcast
+
+The receiver rejects everything unless xDrip+ is configured to expect it. This is the xDrip+ user's
+setup rather than something your app controls, but nothing works until it is done, so it belongs at
+the start rather than at the end.
+
+The broadcast is accepted when any one of these holds:
+
+- `Hardware Data Source` is set to **`640G / EverSense`** — the normal choice for a broadcasting app,
+  or
+- `Hardware Data Source` is set to **`xDrip+ Sync Follower`**, or
+- the **`Out of process Libre algorithm.`** option is enabled (Advanced Settings).
+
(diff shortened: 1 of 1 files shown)


ahaarrestad:
The page as it renders after merge, in four parts.

**1. Overview, and the xDrip+ setup that gates everything**

Overview and setup

**2. Package visibility, the extras, and the payload tables**

Intent and payload

**3. Complete example, and how often you can send**

Example and rate

**4. How failures surface, and debugging**

Failures and debugging

Navid200:
Please don't use xDrip+ anymore.  
Please use xDrip instead.

ahaarrestad:
Renamed to xDrip throughout, and cut the page back: 1679 → 1019 words.

The two remaining `+` spellings are the UI labels themselves, `640G / EverSense` and
`640G/Eversense 1-minute`. Changing those would stop the text matching what the user sees.

Gone: the xDrip-side setup section — which followers qualify, that Hardware Data Source is
single-choice, the reboot. That is the user's concern, and what a developer needs from it is three
lines in the overview. Backfilling folded into the rate section, the worked 3-minute example cut to
a sentence, and the two failure tables merged into one.

I also dropped the note about a broadcast appearing twice in logcat. It was one of the two things I
left for you to decide, and it overlaps #4652, so it does not belong here while that is open.

The explicit broadcast is now the stated point of the page.

**1. Overview and sending the broadcast**

Sending

**2. The reading, and the complete example**

Payload

**3. How often to send, and when nothing arrives**

Rate and failures

ahaarrestad:
The page as it will look after merge. It is short enough now that it fits in one image.



The screenshots earlier in this thread are of the previous, longer version and no longer match the branch.

jamorham:
I think the last part will be wrong after https://github.com/NightscoutFoundation/xDrip/pull/4724 merges

ahaarrestad:
Good catch. I've updated that line to name `Inter-app broadcast`, and kept the old label in a second sentence since anyone on the current release still sees `640G / EverSense` in the menu. That way the page is right whichever order these two merge in.

PR #4724, opened 2026-08-29
Inter-app data source

Navid200:
This PR changes the title of one of the hardware data sources.
It renames 640G/Eversense to Inter-app.

The main reason I think we should change this title is that the existing title is too specific to only 2 of the possibilities.  
But, there are many other sources hat broadcast to xDrip and there will even be more in the future.
I believe we should change the title to something more general that would be valid for all of them.

Further, we should avoid using titles that may be trademarks.  

I know that this could cause confusion a little bit.
But, xDrip is alive, young and growing.  There will be changes.  We support users and help them adapt.  

I am not insisting this is the best title.
If you can think of alternatives, please let me know.

I did not like Broadcast receiver because receiver may bring to mind a receiving piece of hardware.
I didn't like receiver for the same reason.
I didn't like Listener because it was too vague.

I considered many others until I considered the existing "Inter-app settings" title.  Inter-app seems to be the best unless you can think of something better.

| Before | After|  
| ------- | ----- |  
| Screenshot_20260828-194351 | Screenshot_20260828-194508 |

### DIFF
--- a/app/src/main/res/values/arrays.xml
+++ b/app/src/main/res/values/arrays.xml
@@ -13,7 +13,7 @@
         <item>xDrip+ Sync Follower</item>
         <item>LibreAlarm</item>
         <item>Libre (patched App)</item>
-        <item>Inter-app</item>
+        <item>Inter-app broadcast</item>
         <item>Medtrum A6</item>
         <item>Nightscout Follower</item>
         <item>Dex Share Follower</item>
--- a/app/src/main/res/values/arrays.xml
+++ b/app/src/main/res/values/arrays.xml
@@ -13,7 +13,7 @@
         <item>xDrip+ Sync Follower</item>
         <item>LibreAlarm</item>
         <item>Libre (patched App)</item>
-        <item>640G / EverSense</item>
+        <item>Inter-app</item>
         <item>Medtrum A6</item>
         <item>Nightscout Follower</item>
         <item>Dex Share Follower</item>


Navid200:
There are existing references to the original title in documentation.
If this has merit, please let me know and I will add to this PR changes that will replace them accordingly throughout the entire repository documentation.

jamorham:
I would think "Inter-app broadcast" would make the most sense and other apps may label the emission as "broadcast to xDrip" - Good to ensure that its distinguished from companion app data source.

Navid200:
I can do that.

I have considered several phrases containing the word broadcast.
I dropped them all. My main concern was that xDrip can also broadcast.  xDrip broadcasts to AAPS.  That is set up somewhere else on the "Inter-app" settings.  It is called "Boradcast locally".  
But, if that setting was titled "Broadcast locally", it would make perfect sense?  Wouldn't it?

I thought if the the phrase ends with "broadcast", it is possible someone may think this means xDrip is the broadcaster!
Broadcast is a known.  But, it also can be a verb.  I know this is nitpicking.  I am just explaining why I had dropped it.  

It has taken me months to open this PR even though it may seem it should have taken 5 seconds.  I am still not 100% sure it is the best.  

And yes, I completely avoided companion because that title is used for something else and it is guaranteed would cause confusion if we now rename both.

jamorham:
Well this is in the data source section so hopefully it is clear that this is xDrip receiving a broadcast from another app.

jamorham (approved):

PR #4738, opened 2026-09-08
Retire ongoing notification channel setting

Navid200:
The ongoing notification setting has no purpose any longer.
I cannot see any references to it in the code.
It has to be active whether we want it or not.  

This PR removes the setting.  The following shows the setting page after this PR.  
Screenshot_20260911-114512


jamorham (approved):
thanks

PR #4740, opened 2026-09-09
Media player flag cleanup

Navid200:
@ahaarrestad Does this make sense for addressing the 4 exits that currently leave the activeTag flag occupied?

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/AlertPlayer.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/AlertPlayer.java
@@ -357,6 +357,7 @@ public class AlertPlayer {
         mediaPlayer = new MediaPlayerCreaterHelper().createMediaPlayer(ctx);
         if (mediaPlayer == null) {
             Log.wtf(TAG, "MediaPlayerCreaterHelper().createMediaPlayer failed !!");
+            activeTag = ""; // Clear the lock
             return;
         }
 
@@ -401,6 +402,10 @@ public class AlertPlayer {
         }
         if (!setDataSourceSucceeded) {
             Log.wtf(TAG, "FATAL: Default_alert failed to load!");
+            activeTag = ""; // Clear the lock
+            if (mediaPlayer != null) {
+                stopAndReleasePlayer(mediaPlayer);
+            }
             return;
         }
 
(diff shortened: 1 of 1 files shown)


ahaarrestad:
Yes — those are the right four exits, and clearing the flag on each is correct. Releasing the player
at the two exits that had created one is a real improvement too; before this they leaked.

One thing I measured on an API 34 emulator:

```
isPlaying() after release() -> threw java.lang.IllegalStateException
isPlaying() on a fresh player -> returned false
```

After the new release calls, `mediaPlayer` still points at the player that was just released. The
next alert that would play a sound starts by asking that same player whether it is playing, and that
is the one such call in the app not wrapped in a try/catch. So the alert *after* a failed one gets no
sound — and no notification either, because this happens before the notification is posted.

`stopAlert` already deals with this: it releases the player and then clears the reference. The new
exits release without clearing.

Both unit suites are green on your branch.

Navid200:
You are reporting something related to API 34.  Is that something we have to worry about when we change targetSDK to 34 or higher?  Or, is that something we have to worry about now?

In other words, do we have to fix anything in this PR or are you OK with this as is being merged?

ahaarrestad:
Now, not later. This is not an API 34 behaviour — 34 was just the phone I had. The `isPlaying()`
documentation is word for word the same from API 24 through API 36: it throws if the player has been
released. It does not depend on targetSDK, and xDrip's minimum is 26, so it applies to every device
the app runs on today.

One thing needs fixing before merge. After the two exits that release the player, the field still
points at the released one. Clearing it there — the way `stopAlert` already does right after it
releases — removes the crash.

As it stands the PR trades a leaked player for an alert that goes missing, and the missing alert is
the worse of the two.

Navid200:
Thanks a lot for catching this.
I have set the player to null after stop to address this.
Hopefully, it is OK now.

Navid200:
@ahaarrestad If/when you have a chance, would you please see if what I have done addresses your concerns?

Thanks

ahaarrestad:
Yes, this addresses it. All four exits clear the flag now, and the two
that release the player also clear the reference — that was the crash.

Both unit suites green on 5eb1b977c.

Navid200:
Thanks a lot

jamorham (approved):
thanks

PR #4741, opened 2026-09-09
Remove Parakeet support, part 2: the map, the helper and the extra-features gate

ahaarrestad:
Second of three. #4718 was the dead-code slice. This one removes the Parakeet features that are
still wired up, and renames what survives. Strings are left to the third PR so Crowdin sees one
removal pass.

### Removed

- **The map.** `MapsActivity`, its layout, the marker drawable, the manifest entry, and the two
  Home menu items (`showmap`, `parakeetsetup`). `play-services-maps` has been commented out in
  `app/build.gradle` for years, so the activity could not draw a map. With it go the
  `google_maps_key` manifest placeholder, `app/gradle.properties` — whose only content was that key
  — and `getLocalProperty()`, added for it and now unreferenced. That file held a real Maps API
  key rather than a placeholder; it stays in the repo's history either way, so it is worth
  revoking independently of this PR.
- **The `plu` follower action.** Its only job was feeding map locations to a follower. The sender
  side went in #4718.
- **`ParakeetHelper`** and its callers: the check-in notifications from `WixelReader`, the status
  string on the system status page, `notifyOnNextCheckin` from `Home.onCreate` and a Preferences
  listener. `Notifications.parakeetMissingId` (app and wear) goes with them.
- **The `plus_extra_features` and `plus_follow_geolocation` checkboxes.** The first gated only the
  two map menu items; its other consumer, `refresh_extra_items()`, had a fully commented-out body.
  The second was read only from a commented-out line in `MapsActivity`.
- **The Parakeet notification channel.** `Notifications.parakeetMissingId` goes with it.

### Renamed, not removed

The home screen `Parakeet Battery`, the `Uploader Battery` status item, `NightscoutBatteryDevice`,
`GcmActivity.sendParakeetBattery` and the `Wifi Wixel / Parakeet` collection method all become the
wifi uploader. Any WifiWixel reports that battery; only the label said Parakeet.

**The `parakeet_battery` preference key stays.** Renaming it would drop stored values on upgrade,
and it is read on six paths.

### Not touched: `IdempotentMigrations`

An earlier revision of this PR dropped the two lines writing `parakeet_status_alerts` and
`parakeet_charge_silent`. **They are back, and the file is at zero diff.**

Navid's point on review, and it is not visible from the diff: that file is what lets someone install
a very old release over a current one, or import a settings file exported years ago, without landing
on a setting in an undefined state. Removing an entry is not safe by inspection, and a cleanup of the
old ones is a decision for its own PR. So the two keys keep being written on every start — read by
nobody once `ParakeetHelper` is gone, and harmless.

### Screenshots

Both renames on one screen, and the collection method in the list:

| Wifi Wixel Status | Hardware Data Source |
|---|---|
| Wifi Wixel Status | Hardware Data Source |

### Three behaviour changes worth calling out

- **System status, WifiWixel row.** It showed the Parakeet check-in string when one was checked in,
  otherwise `No data`. It now always shows `No data` — what a WifiWixel user without a Parakeet
  already saw. The arm has to stay: without it the row falls through to the bluetooth path and
  reads `Not connected` for a collector that was never on bluetooth.
- **Nightscout devicestatus.** The uploader posts under `device` as `Wifi Uploader` where it said
  `Parakeet`. An existing site keeps the old pill, which stops updating and greys out; anything keyed
  on the name `Parakeet` stops matching. The battery value itself is unchanged.
- **Pebble bridge battery.** The uploader slot was reachable only by a user with a checked-in real
  Parakeet, since `isRealParakeetDevice()` required one. They now get the phone battery there, like
  every other passive collector. A plain WifiWixel is unaffected.

### Tests

New: `SystemStatusFragmentTest`, `PebbleDisplayAbstractTest`, `NightscoutBatteryDeviceTest`,
`WifiCollectionServiceTest`, `WixelReaderHttpTest`, `HomeLayoutAnchorTest`,
`CollectionMethodArraysTest`. Extended: `DexCollectionTypeTest` (app and wear).

`app/src/test/AndroidManifest.xml` gains one line, `SystemStatusFragmentTest$HostActivity`. That
is the unit-test manifest, so nothing new ships in the APK - Robolectric just will not launch an
activity that is not declared, and the fragment needs a host to attach to. The same edit merges
that file's two `<application>` elements into one; a child added to the first would otherwise
have landed in a second element that is ignored.

All green on current master:

```
./gradlew :app:assembleProdRelease
```


jamorham (approved):
thanks for the cleanup

PR #4745, opened 2026-09-11
Silent alert logs

### LINKED ISSUE #4744: Missing log for silent alert due to DND
Version: 2026.09.09
Phone set to DND
Override silent mode disabled for alert.

The log shows alert triggered.  It should also state that the alert makes no sound due to DND.  But, it doesn't.

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/4744 

If you wake up in the morning and notice your glucose has been low all night and you were never woken up and look at the logs and see that xDrip triggered alerts, you have no way of knowing if those alerts made any sound or not!  With our current Nightly release, the logs look exactly the same if the alerts were silent or made a sound.  

This PR changes that.  If an alert that triggers is silent, we now create a log explaining why it was silent.  After this, if you see no such logs, it means that xDrip did make a sound.  

Additional changes this PR makes:
1- There are AlertPlayer log tags in xDrip in classes other than AlertPlayer.  After this PR, a log tagged with AlertPlayer will only be a log really created by the AlertPlayer class.  
2- Rearranged a few items in the alert priority list.
3- The silent alert log is not added for basic notifications (priority less than 70)  
4- If your volume profile is the ascending volume and silent logs are required, they will be limited to one log per 20 minutes to avoid cluttering the logs.  

Please let me know if I have overlooked anything.

### DIFF
--- a/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/AlertPlayer.java
+++ b/app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/AlertPlayer.java
@@ -2,6 +2,7 @@ package com.eveningoutpost.dexdrip.utilitymodels;
 
 import static com.eveningoutpost.dexdrip.Home.startWatchUpdaterService;
 import static com.eveningoutpost.dexdrip.models.JoH.delayedMediaPlayerRelease;
+import static com.eveningoutpost.dexdrip.models.JoH.pratelimit;
 import static com.eveningoutpost.dexdrip.models.JoH.setMediaDataSource;
 import static com.eveningoutpost.dexdrip.models.JoH.stopAndReleasePlayer;
 import static com.eveningoutpost.dexdrip.receiver.InfoContentProvider.ping;
@@ -56,7 +57,7 @@ import lombok.Getter;
 // This is needed in order for the callbackst to work.
 class MediaPlayerCreaterHelper {
 
-    private final static String TAG = AlertPlayer.class.getSimpleName();
+    private final static String TAG = MediaPlayerCreaterHelper.class.getSimpleName();
     private final Object creationThreadLock = new Object();
     private volatile boolean mplayerCreated_ = false;
     private volatile MediaPlayer mediaPlayer_ = null;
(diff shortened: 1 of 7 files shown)


Navid200:
This screenshot shows silent logs after this PR.  

Screenshot_20260911-113931

app/src/main/java/com/eveningoutpost/dexdrip/utilitymodels/AlertPlayer.java
```diff
@@ -629,12 +633,14 @@ protected void VibrateNotifyMakeNoise(Context context, AlertType alert, String b
...
-                Log.i(TAG, "Vibration silenced due to ongoing call");
+                Log.i(TAG, "No vibration due to ongoing call");
             }
         }
+        if (profile == ALERT_PROFILE_SILENT && pratelimit("silent-alert-log", 1200)) {
+            UserError.Log.uel(TAG, "No " + tag + " with silent volume profile");
+        }
         // Let's keep this dummy pattern so the notification still mirrors to watches
         builder.setVibrate(new long[]{1, 0});
-        Log.ueh(TAG, contentLog);
```
jamorham, line 637: Can you explain this change, is it correct? Doesn't this mean we miss some logs?

Navid200: The log hasn't been removed.  It has been moved to an earlier spot, new line 619.
It now has a comment after it "Log.ueh(TAG, contentLog); // Glucose level alert log".  

The reason for this move is so that it appears before the new log that says the alert is silent.

Navid200:
The following screenshot shows the logs after this PR:
Screenshot 2026-09-18 080323


When I had originally made my PR and tested it, I had not rearranged any of the logs.  In that case, Log B apered before log A.
I didn't think that was right.  So, I moved the original log so that new (silent) log appears after.  
I did this both for the glucose level alert log as well as for the Other alert log.  

So now, we see a log that says we have an alert.  Then, we may see another log that says this is silent and it explains why.  Either because we have chosen the silent profile or because we have not enabled override silent mode in DND

PR #4747, opened 2026-09-13
Tests for the alert notification category

ahaarrestad:
Unit tests only - no production code.

`XdripNotificationCompat.build()` adds no category of its own and carries through one the caller set, on both of its paths (a named channel, and the general-channel fallback), and strips grouping.

`Notifications.OtherAlert` gives an alert the alarm category and a high-priority full screen intent exactly when it is set to override silent mode: per alert, inheriting the global setting for the other alerts, and off when neither is set. That is the behaviour #4680 introduced; these tests pin it so it cannot drift back unnoticed.

Deliberately not covered: the reraise window, and the audio stream picked in `triggerSoundAndVibration`.

All green: `./gradlew :app:testFastDebugUnitTest :wear:testFastDebugUnitTest`


jamorham (approved):
thanks

PR #4751, opened 2026-09-15
Remove three mistranslated Bulgarian step labels

ahaarrestad:
On Bulgarian the history screen shows no step count and no distance at all.

`values-bg` has description sentences where English has labels:

| key | English | Bulgarian, translated back |
|---|---|---|
| `label_show_steps` | `Steps:%1$d` | "Counting steps. They reset at midnight" |
| `label_show_steps_km` | `km:%1$s` | "show the distance travelled in km" |
| `label_show_steps_mi` | `mi:%1$s` | "distance travelled in miles" |

All three are formatted with an argument (`BGHistory.java:216`, `:220`, `:221`), so the value is
dropped silently — `String.format` ignores extra arguments. No crash, just missing numbers. The
neighbouring keys in the same file show the shape that was meant: `label_show_insulin` is `Е:%1$sе`.

Deleting them restores the English fallback, number included. I have not invented a Bulgarian
string I cannot read.

@jamorham — the source is Crowdin (`crowdin.yml` maps `bg-BG: bg`), so a later sync brings them
back unless they are corrected there, and that needs your access. Precedent for deleting in the
repo: `5ee1a23b4` removed `expires_days` from `values-he`, and it is still gone 67 Crowdin commits
later.

Why now: these are the only three keys in the repo that would fail the placeholder test you asked
about on #4702. That test comes as a separate PR once this one is in.

`assembleProdRelease` green.


jamorham:
I think this needs fixing at crowdin rather than here but thanks for the heads up

Navid200:
@ahaarrestad I can help with Crowdin.
I am looking at this right now to see what I can do.

Navid200:
I believe I have deleted those three translations from Crowdin now.

ahaarrestad:
Crowdin sync 0bb963d0a dropped the three strings, so this is done upstream. Closing — thanks both.

Navid200:
Thanks you for identifying it.

PR #4763, opened 2026-09-21
Muted alert warning

### LINKED ISSUE #4734: Show a dialog when silent volume profile is chosen
We should present a dialog that would explain the seriousness of the silent and the  vibrate only volume profiles when either is chosen by the user.

The dialog should highlight that the "Override silent mode" feature will not override the silent or vibrate only volume profiles.

Navid200:
Fixes: https://github.com/NightscoutFoundation/xDrip/issues/4734 

I wanted a simple solution.
This seems too complicated.
Is there a simpler way to accomplish this?

Please be critical.
I don't want to complicate the code unless absolutely necessary.   I believe the objective is important.  But, is this the simplest way to accomplish it?


Navid200:
Screenshot_20260920-222946

PR #4766, opened 2026-09-21
Play the soft notification sound for a system default sound URI

### LINKED ISSUE #4765: Ensure default sound for low-priority items is default notification
It is possible that the default alarm sound may be used  in some cases for items like the reminders.
Change this to the default notification sound.

ahaarrestad:
Fixes #4765. Reported by jamorham in chat: reminders play the glucose-alarm default.

**Cause.** A ringtone picker hands back `content://settings/system/...` when the user picks "Default", and the notification preferences persisted the same URIs as their defaults before #4680. `AlertPlayer.playFile` treats every such value as unset and never opens it — as @Navid200 noted, the row behind it is not reachable under all conditions — but substituted `R.raw.default_alert` at every priority. A reminder with that value sounded like a low-glucose alert.

**Fix.** Below priority 80 the substitute is now `R.raw.default_notification`, exactly as for the plain `"default"` marker. The URI is still never opened. At 80 and above nothing changes.

| Source | Priority | Before | After |
|---|---|---|---|
| Reminder with "Default" picked in the picker | 50 | `default_alert` | `default_notification` |
| `other_alerts_sound` on every install that predates #4680 and never changed it (unclear, rise, fall, BlueReader, session restart) | 20–75 | `default_alert` | `default_notification` |
| `persistent_high_alert_sound`, `bg_predict_alert_sound`, `bg_missed_alerts_sound` persisted before #4680 | 80–95 | `default_alert` | unchanged |
| High/low AlertType rows created before #4680 | 85 / 90 | `default_alert` | unchanged |

The second row is the largest group: for those installs the rise/fall/unclear alerts move from the glucose alarm to the notification sound, which is what a fresh install already plays for them. Calibration notifications are not affected either way: they set the sound on the builder and never reach `playFile`.

No migration: the stored values are fine, only their interpretation changed.

**Verified.** `AlertPlayerSystemSoundTest` pins the boundary from both sides (50, 80, 85) and the plain `"default"` marker. All green: `./gradlew :app:testFastDebugUnitTest :wear:testFastDebugUnitTest`. On an emulator, a reminder with `content://settings/system/notification_sound` went from 34 s of playback to 0.5 s; a High alert row with the same URI still plays the 32 s alarm.


Navid200:
This will fix the problem.
But, I am not sure it is a good idea to create a new variable titled systemDefaultSound.
I hope that a developer looking at this in the future will not decide to use that variable to assign a sound file.

xDrip had a default alarm sound.  It now also has a default notification sound.
Developers should use those two so that xDrip is in control and going from phone to phone will not cause a change to the alert sounds.  We should avoid using the system default sounds.  

Would it not be better to just replace the following existing line:
if ("default_notification".equals(fileName) || (priority < 80 && "default".equals(fileName))) {

with this:
if ("default_notification".equals(fileName) || (priority < 80 && ("default".equals(fileName) || fileName.startsWith("content://settings/system/")))) {  

Alternatively, if we have to create a new variable, I wish we would give it an obviously negative name clarifying that it should be avoided.

ahaarrestad:
Done — the variable is gone and the check sits in the condition, as you wrote it. I kept `fileName != null &&` in front of it: a missing sound name would otherwise crash the player instead of falling back to the bundled alarm. No current alert passes one at that priority, so it changes nothing today; there is a test for it.

Navid200:
That is fantastic.
Thanks a lot.

ahaarrestad:
Thank you — and for the pointer on the variable. The line reads better without it.

